PR #3, opened 2015-05-10
Remove iOS reference in podspec

orta:
Hello there, wanted to use this in a Mac app. Opted to remove the iOS only reference. The other option is to state a version for OS X and iOS. I have no major preference.


orta:
It also included your `.tt` and some module files that CocoaPods generates, so I've scoped it to only include swift files.

kzaher:
Hi,

this PR is really helpful, I've totally forgot about OSX target. I've only concentrated on iOS.

I'll definitely add OSX support in 1.3, should be really easy.

orta:
:+1: done

PR #9, opened 2015-05-14
Reducing the complexity of the examples with singleton and pipelines in new lines

carlosypunto:
It’s not technically better, but understanding has improved.


kzaher:
Hi, yep, I think this will improve readability and remove di noise.

PR #11, opened 2015-05-24
Develop

dtartaglia:
Created UIControl extension with rx_controlEvents method.
I want to be able to use the system for types of controls other than just UIButtons. Thanks for all the effort you put into this.


dtartaglia:
hmm... I'm new at pull requests. I think I did something wrong.

PR #15, opened 2015-05-25
playground minor corrections

carlosypunto:


kzaher:
I like your suggestion, it adds to code readability.

Could you just add a function like

func example(description: String, action: () -> ()) {
    println("--- \(number) ---")
    action()
}

to SupportCode.swift, and then use it instead of println like so

example("oneObservable") {
 		 
let oneObservable = just(32)		 let oneObservable = just(32)
let oneObservableSubscriber = oneObservable >- subscribe { event in
....
 }

carlosypunto:
sorry for all inconvenience, I close it, i clean my fork and I will make a PR in develop branch

PR #17, opened 2015-05-25
Allow remove table view cell

carlosypunto:
Now we can delete cells if we want
You can see a simple example in [this project](http://dl.dropbox.com/u/2878917/RemoveTableViewCell.zip)
perhaps, if this you like, I could add other optional callback for move, too
If you have any tip would do another PR


carlosypunto:
sorry, this is a dirty PR, I close it

PR #18, opened 2015-05-25
Allow remove table view cell

carlosypunto:
Now we can delete cells if we want
You can see a simple example in [this project](http://dl.dropbox.com/u/2878917/RemoveTableViewCell.zip)
perhaps, if this you like, I could add other optional callback for move, too
If you have any tip would do another PR

### DIFF
--- a/RxCocoa/RxCocoa/iOS/UITableView+Rx.swift
+++ b/RxCocoa/RxCocoa/iOS/UITableView+Rx.swift
@@ -10,10 +10,11 @@ import Foundation
 import RxSwift
 import UIKit
 
-// This cannot be a generic class because of table view objc runtime that checks for 
+// This cannot be a generic class because of table view objc runtime that checks for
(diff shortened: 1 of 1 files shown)


kzaher:
Hi Carlos,
I think that we'll need to implement this little differently.

The best solution I see right now is adding something like this to table view interface.

rx_rowDeleted() -> Observable<(UITableView, NSIndexPath)>
rx_rowMoved() -> Observable<(tableView: UITableView, from: NSIndexPath, to: NSIndexPath)>

That could be implemented by adding 

typealias RowDeletedObserver = ObserverOf<(tableView: NSIndexPath, row: NSIndexPath)>
typealias RowMovedObserver = ObserverOf<(tableView: NSIndexPath, from: NSIndexPath, to: NSIndexPath)>

var Bag<RowDeletedObserver> 
var Bag<RowMovedObserver>

to RxTableViewDataSource

PR #28, opened 2015-06-01
from creation function added

carlosypunto:
http://reactivex.io/documentation/operators/from.html

### DIFF
--- a/RxSwift/RxSwift/Observables/Observable+Creation.swift
+++ b/RxSwift/RxSwift/Observables/Observable+Creation.swift
@@ -56,6 +56,17 @@ public func returnElements<E>(values: E ...) -> Observable<E> {
     }
 }
 
+public func from<E, S where S: SequenceType, S.Generator.Element == E>(secuence: S) -> Observable<E> {
+    return AnonymousObservable { observer in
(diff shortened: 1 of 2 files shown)


kzaher:
Hi Carlos,
I'm usually extremely careful when pulling code into RxSwift :), but your change looks small and valuable.

There are just some minor tweaks I would want to ask you.

Would it be possible to merge all of the commits into one commit? I don't want to spam RxSwift git commit history :)

7 commits for 30 lines seem little too much :) So I would really appreciate if you would just merge all of this into one commit named something like "Adds from conversion function".

There is also a typo in source code. Other then that :+1:

RxSwift/RxSwift/Observables/Observable+Creation.swift
```diff
@@ -56,6 +56,17 @@ public func returnElements<E>(values: E ...) -> Observable<E> {
     }
 }
 
+public func from<E, S where S: SequenceType, S.Generator.Element == E>(secuence: S) -> Observable<E> {
```
kzaher: I think it should be "sequence" instead of "secuence". Not a big problem, but we should fix it.

PR #39, opened 2015-06-25
use ReplaySubject in Playgrounds

carlosypunto:
- use ReplaySubject in Playgrounds
- Some Observables+Connectable example


kzaher:
It's just spelled `third` instead of `thirth` :), but we can fix that in upcoming commits.

:+1:

PR #54, opened 2015-07-09
Swift 2.0 iOS 9.4 Fix for observeValueForKeyPath Change

garethknowles:
With this fix RxSwift and RxCocoa work with Swift 2.0 in the Xcode 7.0 beta 3. 
Due to Apples changes to NSKeyValueObserving found [here](https://developer.apple.com/library/prerelease/ios/releasenotes/General/iOS90APIDiffs/frameworks/Foundation.html)
All unit tests are currently passing on iOS and OS X


kzaher:
Awesome, thnx :+1:

PR #61, opened 2015-07-18
Make data source proxies own their data source

nathankot:
I was having trouble using `rx_subscribeItemsToWithCellIdentifier` and
kept getting retain unowned errors. Traced it down to the `dataSource`
of `RxTableViewDataSourceProxy` not being retained. Please correct me if
I'm wrong but I imagine that ownership should look like this:

```
ViewController -> TableViewDataSourceProxy -> DataSource
```

### DIFF
--- a/RxCocoa/RxCocoa/iOS/Proxies/RxCollectionViewDataSourceProxy.swift
+++ b/RxCocoa/RxCocoa/iOS/Proxies/RxCollectionViewDataSourceProxy.swift
@@ -32,7 +32,7 @@ class RxCollectionViewDataSourceProxy : DelegateProxy
     
     unowned let collectionView: UICollectionView
     
-    unowned var dataSource: UICollectionViewDataSource = collectionViewDataSourceNotSet
+    var dataSource: UICollectionViewDataSource = collectionViewDataSourceNotSet
     
     required init(parentObject: AnyObject) {
         self.collectionView = parentObject as! UICollectionView
--- a/RxCocoa/RxCocoa/iOS/Proxies/RxTableViewDataSourceProxy.swift
+++ b/RxCocoa/RxCocoa/iOS/Proxies/RxTableViewDataSourceProxy.swift
@@ -34,7 +34,7 @@ class RxTableViewDataSourceProxy : DelegateProxy
     
     unowned let tableView: UITableView
     
-    unowned var dataSource: UITableViewDataSource = tableViewDataSourceNotSet
+    var dataSource: UITableViewDataSource = tableViewDataSourceNotSet
     
     required init(parentObject: AnyObject) {
         self.tableView = parentObject as! UITableView


kzaher:
Hi,

We are working on documenting how everything works, but still not there yet :)

Without using Rx, the model tableview data source is unsafe_unretained. We really want to maintain that model because that's native behavior to UITableView, and there are reasons why that is the case :)

This is the ownership chain.

ViewController -> View -> TableView -> TableViewDataSourceProxy --- unsafe_unretained ---> DataSource

So now you are probably wondering what is retaining data source :)?

The observable subscription should, because it will funnel elements to reactive data source while it is producing elements.

One other thing will also retain it is subscription disposable added to disposeBag.

The idiomatic code usually looks like this. 

 ```
     observable
             >- tableView. rx_subscribeItemsToWithCellIdentifier(...
             >- disposeBag.addDisposable
```

I've found one place that has wrong code.

https://github.com/kzaher/RxSwift/blob/master/RxCocoa/RxCocoa/Common/DelegateProxyType.swift#L161

I need to make sure that original observable can never complete or error out and will retain data source until subscription disposable isn't disposed. (I'll probably add `>- variable` there, or something similar and make sure it crashes with error "You've tried to bind error to view, please use >- catch operator ....").

Can you please explain your situation in more detail so I can help you with it and find all the places that I need to fix?

* did you use observable that completed or errored? (you can use `>- debug` operator to debug that easily) -> assuming yes?
* `>- disposeBag.addDisposable` at the end -> assuming no?

**Comment by nathankot** [comment]
Hey @kzaher thanks for the detailed response :) I've been trying to re-create the issue using rxswift on `master` but I can't seem to anymore :( it's definitely not cocoapods cache because i've done a complete re-install after clearing the pods cache folder, I don't think it's ARC not cleaning up properly but it might be a peculiar state the app was in that caused it, I'm going to keep rxswift on `master` and see if the issue pops up again.

For your reference anyway (although take it with a grain of salt as I may have done something stupid) this was the code that didn't work before I removed `unowned`:

```swift
      stocks
      >- tableView.rx_subscribeItemsToWithCellIdentifier(
           "StockCell", configureCell: {
             (indexPath, stock, cell) in
             if let c = cell as? StockCell { c.stock = stock }
           })
      >- disposable // This is just an alias for `disposeBag.addDisposable`
```

Before I upgraded from the deprecated method it worked fine, so I assume somewhere in the new method nothing is retaining the datasource:

```swift
      // Use to look like this when it worked:
      stocks
      >- tableView.rx_subscribeRowsToCellWithIdentifier(
           "StockCell", configureCell: {
             (tableView, indexPath, stock, cell) in
             if let c = cell as? StockCell { c.stock = stock }
           })
      >- disposable
```

So to answer your questions:

* No, the `stocks` observable is continuous? it would never throw a completed or errored event
* Yes, I was using `addDisposable`

I can't remember the exact error but I tried removing `disposable` just to see if it would look familiar but nothing errored out either...

Thanks!

kzaher:
@nathankot I think I've fixed the problem with data source.

It was a bug. This should solve your problem.

I've made a patch release 1.8.1

nathankot:
@kzaher thanks so much! I haven't had this issue since to closing :)

PR #62, opened 2015-07-18
Add UITextView extension

tokorom:
Add UITextView+Rx that has same functions as those of UITextField+Rx.

### DIFF
--- a/RxCocoa/RxCocoa/iOS/UITextView+Rx.swift
+++ b/RxCocoa/RxCocoa/iOS/UITextView+Rx.swift
@@ -1,0 +1,20 @@
+//
+//  UITextView+Rx.swift
+//  RxCocoa
+//
+//  Created by Yuta ToKoRo on 7/18/15.
+//  Copyright (c) 2015 Krunoslav Zaher. All rights reserved.
+//
+
+import Foundation
+import RxSwift
+import UIKit
+
+extension UITextView {
+    public var rx_text: Observable<String> {
+        return NSNotificationCenter.defaultCenter().rx_notification(UITextViewTextDidChangeNotification, object: self)
+            >- map { notification -> String in
+                return (notification.object as? UITextView)?.text ?? ""
+            }
+    }
+}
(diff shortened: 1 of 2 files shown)


kzaher:
Hi,

I see you have a need for `UITextView` extensions.

I would love to merge this but can you please make these changes first.

Could you make a PR towards develop branch. We use gitflow model. Master branch contains latest released code.

RxCocoa/RxCocoa/iOS/UITextView+Rx.swift
```diff
@@ -0,0 +1,20 @@
...
+import Foundation
+import RxSwift
+import UIKit
+
+extension UITextView {
+    public var rx_text: Observable<String> {
+        return NSNotificationCenter.defaultCenter().rx_notification(UITextViewTextDidChangeNotification, object: self)
+            >- map { notification -> String in
+                return (notification.object as? UITextView)?.text ?? ""
+            }
```
kzaher: We would need to model this this way
```
return defer { [weak self] in
            let text = self?.text ?? ""
   NSNotificationCenter.defaultCenter().rx_notification(UITextViewTextDidChangeNotification, object: self)
            return self?.rx_delegate.observe("searchBar:textDidChange:") ?? empty()
                    >- map { a in
                        return a[1] as? String ?? ""
                    }
                    >- startWith(text)
        }
```

RxCocoa/RxCocoa/iOS/UITextView+Rx.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+//  Created by Yuta ToKoRo on 7/18/15.
+//  Copyright (c) 2015 Krunoslav Zaher. All rights reserved.
+//
+
+import Foundation
+import RxSwift
+import UIKit
+
+extension UITextView {
```
kzaher: You can also add 
```
    public var rx_delegate: DelegateProxy {
        return proxyForObject(self) as RxTextViewDelegateProxy
    }
```

... like other classes that have delegate have. But I can also add this myself if you don't want to investigate how this works right now.

tokorom:
@kzaher Thank you for your replying!

I'll retry to make a PR that was applied your teaching.

PR #70, opened 2015-07-27
Feature/swift2.0 - Added WatchOS 2 support in Cocoapods Spec File

mwermuth:
In order to use RxSwift also on WatchOS 2 we need to declare the support in the Spec File like so:

```
  if s.respond_to?(:watchos)
    s.watchos.deployment_target = '2.0'
  end

### DIFF
--- a/RxCocoa.podspec
+++ b/RxCocoa.podspec
@@ -16,6 +16,9 @@ Pod::Spec.new do |s|
 
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.9'
+  if s.respond_to?(:watchos)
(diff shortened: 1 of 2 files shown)


mwermuth:
Sorry I did notice, that we also need to change code in RxCocoa. Because there are availability checks for iOS and OSX but not for watchOS. I will try to work on that this week.

kzaher:
Oh, ok, let me know when you think it's ready for review.

kzaher:
Hi, I'm going to close this for now, because target branch is stale, we are commit all improvements to `rxswit-2.0` branch.

Issue #69, opened 2015-07-27
can I use this library on ios7?


shuoli84:
Any known issue? Thanks. 
I can't find an answer in readme.md and google.

kzaher:
Hi,
you actually can. I'll document this. But you'll need to manually copy the files to project and if you are using RxCocoa find replace all `import RxSwift` to `` because there is no RxSwift module if you copy the files.

I'm in the process of massive update to documentation, but yeah, some of the people have done that and they say it works for them.

The problem is that iOS 7 doesn't support dynamic frameworks

shuoli84:
Good to know. Thanks.

shuoli84:
I met a crash on ios7.1 in simulator. Not sure why. Any idea? 
![screen shot 2015-07-27 at 3 47 36 pm](https://cloud.githubusercontent.com/assets/3402035/8900905/e3787da8-3476-11e5-993b-fec23ff3b6cd.png)

shuoli84:
```
(lldb) po self.throttleState.cancellable
0x00007fb0605ce770
 {
  CaiShuo.DisposeBase = {}
  lock = (_lock = 0)
  state = {
    current = nil
    disposed = false
  }
}

po d
CaiShuo.SingleAssignmentDisposable
```

kzaher:
Hm, never seen this before.

I'm hoping this isn't swift compiler striking again :(

Can you maybe share the code snippet that caused this? What version of Xcode are you using?

shuoli84:
Xcode 6.4
```swift
        let chartTypeObservable = Variable(chartType as Observable<ChartType>)
            >- switchLatest
            >- throttle(0.3, MainScheduler.sharedInstance)
            >- distinctUntilChanged
        
        combineLatest(chartTypeObservable, stock) {(type:$0, stock:$1)}
            >- map { [unowned self] item -> Observable<ChartDrawType> in
                if item.type == .Minute || item.type == .FiveDays {
                    return self.getMinutesChartData(item.stock!.id, type: item.type)
                }
                else {
                    return self.getCandleChartData(item.stock!.id, type: item.type)
                }
            }
            >- switchLatest
            >- subscribeNext {[unowned self] item in
                switch item {
                case .Minute(let type, let data):
                    if type == .Minute {
                        self.minutesData.next(data)
                    }
                    else {
                        self.fiveDaysData.next(data)
                    }
                case .Candle(let type, let data):
                    switch type {
                    case .DailyCandle:
[...]

PR #72, opened 2015-07-28
Use RxMutableBox to workaround compiler bug

shuoli84:
#69 Changes on code template.


kzaher:
Hi @shuoli84,

Awesome job today. 

I've been trying to confirm your solution but unfortunately it still crashes for me in the same configuration iPhone 5S/iOS 7.1 and the codebase that I've sent to you.

The preprocessor code is in the project. If you want to run it, just choose Preprocessor scheme and set first argument to project root and second to some temp directory that preprocessor will use to generate temporary files.

I know this is little ugly for now, but it works.

Maybe you forgot to include some tiny detail?

screen shot 2015-07-28 at 22 35 34

shuoli84:
I also did some other fixes locally, which I thought not related with this. I'll filter out them.

shuoli84:
I just confirmed that Throttle has the same problem in
```swift
    typealias ThrottleState = (
        value: Element?,
        cancellable: SerialDisposable,
        id: UInt64
    )
```
Then I just wrap it with a RxMutableBox, it is fixed. Seems it is not only occurred when mixed. Not sure whether there are some other places needs the fix.

kzaher:
I've just confirmed the fix, awesome work!

I would just have one small request, could you make a PR to develop branch so I can merge this immediately. 

We are using gitflow, so master branch always contains latest released and tested version.

I'll merge this into master in a couple of days when releasing 1.9 version.

:bow:

shuoli84:
Sure. I created a pull request to develop branch. :palm_tree:

PR #75, opened 2015-07-29
Pull Request: Fix crash on iphone 5s ios7.1 simulater

shuoli84:


kzaher:
@shuoli84 I am little concerned about iOS 7 and Swift compiler (in general) because it looks like compiler is just not working properly, so it hard to know what else is out there, but I can merge this and see where this will get us.

I think rest of the changes (except workarounds around RxBox) are improvements.

If we figure out that helping compiler a little is not enough, and it just keeps generating invalid code, then we'll need to wait for compiler to start generating correct code.

Merging this :+1:

PR #76, opened 2015-07-29
Subjects playground

carlosypunto:


kzaher:
Hi Carlos,

nice work, just a small remarks.

I've added this header on top of every playground (it's still in feature/documentation branch), but it will be merge soon to develop.

Could you add this header to Subject.playground, so when we merge documentation branch, all playgrounds will have these instructions on top.

/*:
# To use playgrounds please open `Rx.xcworkspace`, build `RxSwift-OSX` scheme and then open playgrounds in `Rx.xcworkspace` tree view.
*/

Playgrounds/Subjects.playground/Contents.swift
```diff
@@ -0,0 +1,64 @@
+import Cocoa
+import RxSwift
+
+
+func writeSequenceToConsole(name: String, sequence: Observable<String>) {
+    sequence
+        >- subscribeNext {
+            println("Suscription: \(name), value: \($0)")
```
kzaher: subscription ;)

kzaher:
If you could squash those changes into that commit, that would be awesome.

PR #83, opened 2015-08-06
Fixed Typo in Subjects Playground

iandundas:
Note - BehaviorSubject still also needs a description (seems to be copy pasted from ReplaySubject)

### DIFF
--- a/Playgrounds/Subjects.playground/Contents.swift
+++ b/Playgrounds/Subjects.playground/Contents.swift
@@ -55,10 +55,10 @@ example("ReplaySubject") {
 
 ## BehaviorSubject a.k.a. Variable
 
-ReplaySubject emits to any observer all of the items, in the buffer, that were emitted by the source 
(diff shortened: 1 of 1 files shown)


kzaher:
Hi, thnx, you are right, we should fix that :)

I am confused is this an PR or an issue because it is partial fix and it has `[Edit me too]`.

We reuse a lot of documentation from reactivex.io, because it's all valid for our case.

I would suggest 

* using this description from [here](http://reactivex.io/documentation/subject.html)
```
PublishSubject emits to an observer only those items that are emitted by the source Observable(s) subsequent to the time of the subscription.
```
*  target this PR towards **develop** branch
* squashing the changes into **single** commit

so I can merge it, or reporting an issue if you want us to do it.

I'm fine with both options :)

iandundas:
I changed the description to one I found on reactivex.io - now to figure out how to target it to develop branch.. 😅

iandundas:
See pull request #90

PR #87, opened 2015-08-08
RxSwift 2.0

kzaher:
This is `develop` branch for RxSwift 2.0 version


kzaher:
Transition from `>-` to protocol extensions is practically done. There is still some polishing to do.

Unit tests looks like are passing.
Example apps work.
Opens a lot of questions regarding naming.

kzaher:
I think this one is going to be a looker :)

PR #97, opened 2015-08-17
NSURLSession extensions caused SourceKit crashes

lfaoro:
Checking for `nil` seemed to cause sourceKit crashes as of Xcode Beta 5.
Changing that to `if let` stopped the sourceKit crashes - can’t explain
why.


kzaher:
I think we would need to figure out what is exactly happening :)

I know that there are formatting tools that are changing whitespaces, but when using them in PRs they are causing a lot of noise and making reviews of real changes harder :)

If we do figure out what is exactly wrong, could we just commit that change alone :)

lfaoro:
True :)

PR #107, opened 2015-08-25
Fix renamed and deprecated zero rect and join

ChrisBuchholz:
Changes:

- Use `SequenceType.joinWithSeperator` instead of `Array.join`. I cannot seem to find verification for why `Array.join` doesn't work (it's like it doesn't exist (●__●)), but it doesn't work in any case, and `SequenceType.joinWithSeparator` does.
- `CGRect.zeroRect` has been renamed to `CGRect.zero`


kzaher:
:+1:

PR #111, opened 2015-08-25
Inline documentation proposal

lfaoro:
- style proposal for inline documentation
- example compiles in a playground
- todo: Improve operator's description @kzaher


lfaoro:
Sorry about this, was trying to send a pull from the github client instead of the command line, didn't go well :-1:

PR #112, opened 2015-08-25
Inline documentation proposal

lfaoro:
- style proposal for inline documentation
- example compiles in a playground
- todo: Improve operator's description @kzaher

### DIFF
--- a/Rx.playground/Pages/Introduction.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Introduction.xcplaygroundpage/Contents.swift
@@ -16,20 +16,22 @@ The key to understanding RxSwift is in understanding the notion of Observables. 
 ## Creating and Subscribing to Observables
 The first step in understanding this library is in understanding how to create Observables. There are a number of functions available to make Observables.
 Creating an Observable is one thing, but if nothing subscribes to the observable, then nothing will come of it so both are explained simultaneously.
+
+*NOTE: Activate the Debug Area*
 */
 
 /*:
 ### empty
-`empty` creates an observable that contains no objects. The only message it sends is the `.Completed` message.
+`empty` create an Observable that emits no values but terminates normally with a completed event.
 */
 
 example("Empty observable") {
     let emptyObservable: Observable<Int> = empty()
-
-    let emptySubscriber = emptyObservable .subscribe { event in
+    
+    let emptySubscriber = emptyObservable.subscribe { event in
         switch event {
         case .Next(let box):
-            print("\(box.value)")
+            print("\(box)")
         case .Completed:
             print("completed")
         case .Error(let error):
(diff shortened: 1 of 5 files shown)


kzaher:
Hi @lfaoro , thnx for your help.

I think the suggestion is really nice, but I would like to make it little more efficient.

IMHO, the purpose of inline documentation is to provide quick description and reference where to find additional information.

I would ideally like to use something more similar to what other rx distrubutions are using.

This is source for `map` (they call it `select` in Rx.NET, it's `map` also in RxJS)

[Inline documentation for map from Rx.NET](https://github.com/Reactive-Extensions/Rx.NET/blob/master/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable.StandardSequenceOperators.cs#L715)

[Inline documentation for map from RxJS](https://github.com/Reactive-Extensions/RxJS/blob/master/src/core/linq/observable/select.js#L2)

This is how I see ideal documentation for `map` operator should look like.

```Swift
    /**
        Projects each element of an observable sequence into a new form.
        [ReactiveX.io/Map](http://reactivex.io/documentation/operators/map.html)
    
        - Parameter selector: A transform function to apply to each source element.
        
        - Returns: An observable sequence whose elements are the result of invoking the transform function on each element of source.
    */
    public func map<R>(selector: E throws -> R)
            -> Observable<R> {
            return Map(source: self.asObservable(), selector: selector)
    }
```

The reason why I feel that example is unnecessary is because usage examples don't demonstrate IMHO anything useful. What you would ideally want is a marble diagram, but that is already displayed in that link to ReactiveX.io.

This is probably just a test PR, but we should be careful to not change anything except for documentation if the PR and commits says so :)

There are multiple reasons for that :) :

* reason 1
```
-            case .Error(let e):		+            case .Error(let element):
-                error = e		+                error = element
                 condition.lock()		                 condition.lock()
                 ended = true		                 ended = true
                 condition.signal()		                 condition.signal()
```

`e` was actually `error` and not element

* reason 2

Git blame hides real reason and date for that change, so it's harder to find what exactly happened there.

* reason 3

Every change, no matter how trivial (changing `var` to `let`) can change behavior of compiler and cause change in behavior of result code.

:)

**We should also be careful not to accidentally change whitespace all across unrelated areas**. They add extremely high amount of noise during CR, and again can change important blame lines ;), thus making my life a lot harder.

As far as for changing descriptions in playgrounds, I think we could help @carlosypunto with some of the formulations like the ones here.


We should fix the ones that contain either mistypes `Obserbables`

https://github.com/ReactiveX/RxSwift/blob/rxswift-2.0/Rx.playground/Pages/Combining_Observables.xcplaygroundpage/Contents.swift#L36
 
or the ones that could be formulated better

https://github.com/ReactiveX/RxSwift/blob/rxswift-2.0/Rx.playground/Pages/Combining_Observables.xcplaygroundpage/Contents.swift#L74

There is a lot of example texts in those playgrounds that are taken directly from ReactiveX, that we don't need to change.

You are correct, we should change this

```
Also note that you can have multiple subscribers following to the same observable (as `I` did in the example above.) <- and dot
```

I suggest we shouldn't use `we` either, but a passive form

```
Please also note there can be multiple subscribers of the same sequence as was demonstrated in the example above.
```

**I'm also not sure what is the proper way to formulate this. Maybe some native speaker could help us** :)

But I definitely prefer passive form then `we` form :), because it's impersonal. 
I try to write documentation in passive form, but sometimes fail miserably :)

I would suggest maybe first just compiling in one md documentation for all operators, after we review it, we can then easily make a PR.

lfaoro:
Noted, thanks for the generous comment :)

PR #114, opened 2015-08-27
Use static property for sharedInstance in MainScheduler

ikesyo:


kzaher:
Ah, haven't heard they've enabled static properties, thnx :+1:

PR #115, opened 2015-08-28
Reduction of alerts in examples project

carlosypunto:
- WikipediaSearchViewController now uses UISearchController
- Auto-layout in PartialUpdates Controller view fixes


carlosypunto:
since we need support iOS 7 this change does not make much sense

PR #116, opened 2015-08-28
Remove Box references in enums

ikesyo:
Those are not used/necessary any more in Swift 2.

### DIFF
--- a/RxCocoa/iOS/UIImageView+Rx.swift
+++ b/RxCocoa/iOS/UIImageView+Rx.swift
@@ -22,8 +22,7 @@ extension ObservableType where E == UIImage? {
             MainScheduler.ensureExecutingOnScheduler()
             
             switch event {
-            case .Next(let boxedValue):
-                let value = boxedValue
+            case .Next(let value):
                 if animated && value != nil {
                     let transition = CATransition()
                     transition.duration = 0.25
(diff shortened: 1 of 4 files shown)


kzaher:
Hi @ikesyo ,

I agree, the `RxBox` and `RxMutatingBox` will probably be removed from public API in upcoming versions :)

The reason why I'm still leaving it there because there are some compiler issues on iOS7/iPhone 5S that forced us to use it to solve internal compiler bugs (https://github.com/ReactiveX/RxSwift/pull/72), and I will maybe make it internal for some optimization purposes.

There is still a lot of investigations that I need to do there so I would like to leave it for now, together with RxMutatingBox :)

If there is something in Swift standard library I could use instead, I would be happy to remove it, but couldn't find anything :(

If you could make a PR with one commit that contains changes from `77fc2b422d6fc215d52c23e0691df70a48585171` and 
```
-// Because ... Swift		
-// Because ... Crash		
-// Because ... compiler bugs		
```

from the other commit, I would be happy to merge it :)

ikesyo:
@kzaher Thank you for the clarification! I've pushed a new commit that restores the `RxBox`.

> If there is something in Swift standard library I could use instead, I would be happy to remove it, but couldn't find anything :(

Unfortunately, me neither. :sob:

kzaher:
:+1:

PR #131, opened 2015-09-10
Using wrong cache

czeluff:
Previously, the 1st-level `imageCache` was never utilized. Items would be place into the cache, but the code would only check the 2nd-level cache for both an `Image` or `NSData`


kzaher:
Ah, thnx.

Could you just please point this PR to develop branch so I can merge it :)
We would ideally want to use gitflow and have master pointed to latest released version :)

kzaher:
@czeluff could you just direct PR to `develop` branch please :)

kzaher:
Dupe

PR #134, opened 2015-09-12
Locks using protocol extensions

dapenggao:
- Implements locks using protocol extensions (and replaces `var`s with `let`s)
- Simplifies the implementation of some `Disposable`s
- Adds `private` to some class members

### DIFF
--- a/RxSwift/Concurrency/AsyncLock.swift
+++ b/RxSwift/Concurrency/AsyncLock.swift
@@ -21,7 +21,7 @@ That means that enqueued work could possibly be executed later on a different th
 class AsyncLock : Disposable {
     typealias Action = () -> Void
     
-    private var lock = NSRecursiveLock()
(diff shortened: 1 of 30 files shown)


dapenggao:
I should do these in separate pull requests

kzaher:
Hi @dapenggao ,

We use `struct SpinLock` because we don't want to create separate objects on heap, and this helps reducing memory footprint.

Unfortunately `var` does describe what we want to do well and locks are mutating.

I do see a duplication of logic for spin and recursive lock but that duplication IMHO could be solved better by creating an internal protocol and extending it internally.

If you have questions about the code and why things were done the way they were, fell free to contact me on Slack.

dapenggao:
Hi @kzaher,
Thank you for your helpful explanation! I will try another pull request with the internal protocol instead.

PR #140, opened 2015-09-15
On event convienence methods

kousun12:


kzaher:
Hi @dtartaglia ,

I think that @kousun12 was proposing this for readability of dispatching events sake :) We've been exchanging messages for the past couple of days on Slack. Some other people were also curious about this.

Looks like some people do find it more readable to use it that way, I can't argue against that :)

I think we will be always using `.on(.Event)` in code and examples because it is simpler to explain to new users what is happening there and for our internal operators, we are trying to reduce that stack trace and increase performance, so every call counts :)

My concern with adding this was that adding this would confuse people because it would look like another way of doing the same thing.

I've agreed with @kousun12 that we could add those convenience methods if the comments explicitly describe that those are only convenience methods :)

I think he is right when he says that a lot of people would probably add those methods to their code anyway :)

Basically @dtartaglia , I think we all in agreement, and I do share your concerns :)

I think this is kind of a reasonable tradeoff :)

On the other hand, in observer implementation side the reason why we have `.on(.Event)` is because in that way you can reuse code inside a single method for multiple events, it does make code more concise IMHO and semantically more correct :)

I guess there aren't any perfect solutions :)

lfaoro:
What do you guys think about writing a concise "Best Practices" for creating Observables and subscribing to them. So far we know there are multiple ways of doing the same thing, awesome, whoever can choose how they prefer to work **but...** we have a preferred way elegantly explained to the user.

Same principle as for example: [Swift Style Guide](https://github.com/raywenderlich/swift-style-guide)

dtartaglia:
I'm not a big fan having multiple ways of doing the same thing. It adds complexity and I haven't heard any convincing argument as to why having the extra complexity is needed. There should be one obvious and public way of doing the job, not multiple ways.

It may be true that some programmers would add the extra methods into their code, and thus make their code base more complex, but that isn't an argument to make them part of the library.

And frankly, the fact that @lfaoro has now asked for new documentation in the form of a style guide helps  my case. Having two ways to do this adds no value to the library, creates complexity where none is necessary, and increases the documentation burden.

Lastly, I want to make clear that I have no problem @kousun12's commit. I have no problem with having three methods instead of one method that takes an enum with three states. I just don't like the fact that both both ways of dealing with the library are public. If `on` is an implementation detail then make it private. If we are going to recommend in a style guide that tells users to prefer one construct over the other, then the preferred construct should be the only pubic interface. That way we don't need the documentation.

I'll stop my ranting now... :smile:

lfaoro:
I don't know how you convinced me @dtartaglia, but in my hearth I know you're right. :+1:

kzaher:
Hi,

haha I'm now imagining @lfaoro as a salmon swimming near hungry Bear Grylls looking for proteins.

@dtartaglia you are correct and I have missed your "rants" :) Everything you've said is perfectly sane and true.

I think you also realize this is kind of a gray zone :) I'll try to convince you that this is not [Leeroy](https://www.youtube.com/watch?v=mLyOj_QD4a4) reasoning :)

We do have similar pattern for subscription. If we didn't have convenience methods, we would have to write

```
observable.subscribe(AnonymousObserver { e in
...
}
```
every time we would want to subscribe to fetch next element or subscribe on error.

It's workable, but it's less then optimal :)

One could also argue that `flatMap` is `map.merge`. I'll stop enumeration here because otherwise people will start to tweet like mad :)))

If I let people use this, the worst I can see is that people will write some ugly code like

```
observer.on(.Next(e))
observer.onComplete()
```

or 

```
observer.onNext(e)
observer.on(.Complete)
```

which is ugly, but it's on them :)

Or do PRs into RxSwift codebase with `observer.onCompleted()` which I do expect, and will probably haunt me forever :)

Since this is still alpha version, I'm fine with letting this in. If people find some mean ways to abuse this, I'll remove this from codebase before release, otherwise I'm with with merging this for now :)

@lfaoro I'll answer your concerns in the opened issue :)

PR #142, opened 2015-09-17
Use multiple case syntax instead of fallthrough.

yury:
What do you think about this syntax?


kzaher:
Love it :chocolate_bar: 

Looks much more readable and less code. Merging this.

PR #145, opened 2015-09-20
Add data to Error Response

zhigang1992:


kzaher:
Hi @zhigang1992  :),

I think your first commit eb16831 is fine to merge.

I wouldn't want to merge second commit adf2028 because:
* it's too specific for this library
* if we add that data to userInfo you should write extension method that does that
* I think it would be better practice in your case to actually hook up to lower `rx_response(request)` API in  case you need to parse error code response (and properly parse `content-type` etc.) :)

The reason why we don't parse content type for 200-300 responses is:
* because of simplicity
* APIs usually don't return different content types for single endpoint, so if doc says it returns JSON, it better returns JSON otherwise we don't care :)
* I know that it parsing success path could be improved, but that would add additional complexity

If you want to make a pr to `develop` branch with `eb16831`, I would be happy to merge it :)

zhigang1992:
@kzaher you're right, the second commit wasn't intended for this PR... I'll revert it

PR #146, opened 2015-09-20
Feature/rxunits ideas

kzaher:
**This is created for easier tracking of work done on units project**

This is kind of prototype of units. There is still a lot of work to be done, but I think this is a good start.

Units are essentially just simple struct wrappers for observable sequences that help to prove sequence have certain properties.

You can look at them as kind of observable sequence builders because each unit can be converted to observable sequence using `asObservable` method.

It will be later explained why that naming was chosen.

They are optional, nothing depends on them, and they really don't depend on anything.

The reason why I'll probably include driver unit in RxCocoa project and not create separate project called RxUnits is because:
* people will probably use them all the time when using cocoa extensions
* currently I don't see any reason why making them standalone project would increase value of the project since `Driver` is adapted for use in client applications
* We'll soon have 4+ platforms to support, so many, many targets as it is right now.

Most elaborate unit so far is `Driver`

The purpose of `Driver` unit is to ensure the underlying observable sequence has the following properties.

* can't fail, all failures are being handled properly
* elements are delivered on main thread
* sequence computation resources are shared

... still a lot, lot work to be done


bontoJR:
This can be a game changer, awesome! :+1: 

Totally agree to keep them in RxCocoa.

kzaher:
Time to take `Driver` for a spin ...

bontoJR:
Back from my duty service after a long week :)
I am definitely in for that.

PR #149, opened 2015-09-21
Add watchOS2 target

cfraz89:
We've recently successfully used RxSwift in a watchOS 2 app. We just had to do two things:
1. Add targets for watchOS
2. Turn 'embedded content contains swift code' to NO, otherwise the built app ends up with a second copy of the Swift runtimes. This should probably be disabled for the other targets too, however have left them as is for this PR.

Hopefully we can leave this target in the mainline RxSwift, and soon, tvOS :)

PR #154, opened 2015-09-22
[WIP] adds support for tvOS in RxCocoa

orta:
I don't know if I'll find the time to actually do whatever tests might be needed for this, and whatever is needed for carthage, or for anyone not using [CocoaPods-ExpertDifficulty](https://github.com/orta/cocoapods-expert-difficulty). So when someone decides to do this properly, this is the code you need.

### DIFF
--- a/RxCocoa/Common/CLLocationManager+Rx.swift
+++ b/RxCocoa/Common/CLLocationManager+Rx.swift
@@ -13,18 +13,18 @@ import RxSwift
 
 
 extension CLLocationManager {
-    
+
     /**
     Reactive wrapper for `delegate`.
-    
+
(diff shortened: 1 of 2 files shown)


orta:
Looks like there's a lot of whitespace changes in here too, would recommend ticking "automatically trim trainmen whitespace" in Xcode

screen shot 2015-09-22 at 10 45 40

orta:
This is a dupe! #149

orta:
This is not a dupe!

kzaher:
Oh damn it :/,

I think I'll need to turn this on because if somebody has that option it will mess up the whitespace, argh...

PR #162, opened 2015-09-24
CoreMotion+Rx

carlosypunto:


kzaher:
Hi Carlos,

great idea about wrapping CoreMotion!!!

I can help you polish that code before we merge it to develop, so let's meet on Slack when you'll have time so I can explain you what we need to do :)

I'll also need some time to investigate how those APIs exactly work regarding retries and error handling, but yeah, good idea, love it :+1:

I think we could also make a nice example in RxExample project to demonstrate this :)

bontoJR:
Has this still a real chance to be merged?

kzaher:
We'll try to do something similar in future, but probably not for this repo. Has some dependencies on other tasks.

PR #163, opened 2015-09-25
Add `rx_enabled` property to UIBarButtonItem.

yoshinorisano:
This pull request adds a new feature that enables us to use `rx_enabled` property on UIBarButtonItem.

For example, say we have code something like this:

```swift
func bindBoolToUI(source: Observable<Bool>, button: UIBarButtonItem) {
    source
        .subscribeNext { valid in
            button.enabled = valid
        }
        .addDisposableTo(disposeBag)
}
```

That code can be simpler by using `bindTo` and `button.rx_enabled` combo:

```swift
func bindBoolToUI(source: Observable<Bool>, button: UIBarButtonItem) {
    source.bindTo(button.rx_enabled).addDisposableTo(disposeBag)
}
```

Could you review this?

### DIFF
--- a/RxCocoa/iOS/UIBarButtonItem+Rx.swift
+++ b/RxCocoa/iOS/UIBarButtonItem+Rx.swift
@@ -46,6 +46,24 @@ extension UIBarButtonItem {
         return ControlEvent(source: source)
     }
     
+    /**
+    Bindable sink for `enabled` property.
+    */
+    public var rx_enabled: ObserverOf<Bool> {
+        return ObserverOf { [weak self] event in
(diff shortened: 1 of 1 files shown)


yoshinorisano:
Please ignore this pull request ...

I wrote this code based on master, which does not have that feature, and apply patch to develop tree, and I skipped to check the whole resulting code...
The latest develop tree has already the same code.

kzaher:
No problem :)

The good thing is that it's exactly the same like the code we have in develop, except for whitespace :)

I'll release this soon to avoid confusion :)

yoshinorisano:
@kzaher Thank you!

PR #165, opened 2015-09-26
Refactor CombineLatestSink.

yoshinorisano:
This pull request refactors `CombineLatestSink` by using `reduce`.
Could you review this?

### DIFF
--- a/RxSwift/Observables/Implementations/CombineLatest.swift
+++ b/RxSwift/Observables/Implementations/CombineLatest.swift
@@ -18,13 +18,15 @@ class CombineLatestSink<O: ObserverType> : Sink<O>, CombineLatestProtocol {
     typealias Element = O.E
    
     let lock = NSRecursiveLock()
-    
-    var hasValueAll: Bool
+
+    let arity: Int
+    var numberOfValues = 0
+    var numberOfDone = 0
     var hasValue: [Bool]
     var isDone: [Bool]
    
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @yoshinorisano ,

I think we can actually do something better then what was in CombineLatestSink :), you are correct.

Since performance is paramount for this layer, I think we can actually write this code without any loops or reducing. We basically want to squeeze as much performance as possible here :)

I've performed those optimizations to `CombineLatestCollectionTypeSink` already, but didn't have time to apply it to `CombineLatestSink`.

They use `numberOfValues` and `numberOfDone` counters instead of booleans.

I think it might be more performant to apply those same optimizations to `CombineLatestSink`.

yoshinorisano:
@kzaher 
Thank you for your review!
I optimized the code according to your advice.
Could you check it out?

kzaher:
Hi @yoshinorisano ,

Yep, that's what I've meant.

There is still one for loop left in that sink, but I guess I can also remove it :)

```
            for var i = 0; i < arity; ++i {
                if i != index && !isDone[i] {
                    allOthersDone = false
                    break
                }
            }
```

yoshinorisano:
@kzaher Thanks for merging!

> There is still one for loop left in that sink, but I guess I can also remove it :)

Is it something like this?

```swift
let allOthersDone = numberOfDone == arity - 1 && isDone[index] == false
```

PR #166, opened 2015-09-26
Use array literal for option sets

yury:


kzaher:
Ah ... now everything makes sense :), thnx :)

PR #168, opened 2015-09-27
Feature `average` operator

tottakai:
Needed the average operator for my project, so i though i would contribute it.


kzaher:
Hi @tottakai ,

I've reviewed your PR. First of all, I would like to say that it is a nice PR :) I really appreciate the effort with added unit tests.

I've actually intentionally added only `reduce` operator :) The reason why I've added only reduce is because all other operators are just a couple of lines of code.

I don't think it would be good idea right now to include other operators like `average`, `min`, `sum`, `count` ... explicitly in the library because:

* if somebody needs those operations, they are easily derived from `reduce`
* if somebody has a list of `int`s, it's not obvious that the result should be also `int`, maybe float or double is needed
* also, if somebody has a list of floats, maybe he wants to use doubles because of number instability
* there are edge cases that would be handled in a various of different ways
       * on zero elements one could fail the calculation or return default value
* every added operator creates maintenance debt 
* it's hard to find important things in large codebases, and I'm afraid including all of these operators and specializations could bloat our codebase a lot, and thus increasing library size a lot.

I think that including these operators right now would potentially make my life little harder :), so I would like to leave them out for now :) Hope that's ok with you.

My 2p, I would suggest that you `throw` error when there are 0 elements or adding default value as additional parameter because that is an undefined case otherwise.

tottakai:
Sounds logical, its also quite easy to include the code in the client side. thx

PR #175, opened 2015-09-30
Adding data for failed NSURLRequest with status code > 300

mr16839072:
Currently we just return "Server returned failure" and a `NSHTTPURLResponse` where we can only get the status code.  

Now we also always return `NSData` in `NSError` userInfo with the key `RxCocoaErrorHTTPResponseDataKey`.  The data will be an empty there if is no data. (`NSData()`)

This allows us to have more detail on failed requests with the data being retuned in raw data for the receiver to handle.

### DIFF
--- a/RxCocoa/Common/Observables/NSURLSession+Rx.swift
+++ b/RxCocoa/Common/Observables/NSURLSession+Rx.swift
@@ -136,7 +136,7 @@ extension NSURLSession {
                 return data ?? NSData()
             }
             else {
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @mr16839072 ,

There was already a PR with those changes to `develop` branch, and it was merged couple of days ago :)

Sorry for confusion, `master` always points to latest released version.

We always merge PRs to `develop` branch :)

We'll release bunch of changes this week in Alpha.4.

PR #176, opened 2015-10-01
Fix minor bugs of auto loading example

yoshinorisano:
I found some minor bugs when I was looking into the auto loading example :)
Could you review this?


kzaher:
Looks good to me :+1:

yoshinorisano:
@kzaher Thanks!

PR #178, opened 2015-10-02
Add @warn_unused_result attribute to subscribe* -> Disposable methods

yury:
Found useful swift attribute `@warn_unused_result`.

It emits a warning if result of function call is unused. I think it is very good use case for `subscribe*` methods family.

@kzaher  what do you think?

### DIFF
--- a/RxSwift/Observable+Extensions.swift
+++ b/RxSwift/Observable+Extensions.swift
@@ -15,6 +15,7 @@ extension ObservableType {
     - parameter on: Action to invoke for each event in the observable sequence.
     - returns: Subscription object used to unsubscribe from the observable sequence.
     */
+    @warn_unused_result
     public func subscribe(on: (event: Event<E>) -> Void)
         -> Disposable {
         let observer = AnonymousObserver { e in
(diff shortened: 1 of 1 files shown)


yury:
Also it helps to find some spots in Example app. Don't fix them, in order to show attribute work in action :)

kzaher:
Looks cool to me :)

I'll try to think of reasonable message to display to user :)

shahan312:
Hey, what's the reasoning for providing this compiler warning on subscribe/subscribeNext? On my current project, am now getting a ton of warnings. The code looks like this in most cases:

```swift
let observable: Observable<String> = …

observable.subscribeNext  { _ in

}
```

Is there a better way of doing this?

ikesyo:
Yes, it should be used for operators that returns `Observable`, I think.

kzaher:
I think the warnings would be appropriate in both places :)

* operators that return `Observable` (WIP)
* and subscribe methods

I've tried to explain basics of resource management [here](https://github.com/ReactiveX/RxSwift/blob/master/Documentation/GettingStarted.md#basics).

TLDR

In some cases it's ok to ignore the result, but it's always a safe bet to use `disposeBag` or `takeUntil` operator, otherwise you may risk leaking resources.

PR #182, opened 2015-10-07
Reorganized the GitHub Signup Example

samgreen:
Fixed a typo, used the type alias everywhere, and reorganized a few minor things for readability.

PR #184, opened 2015-10-08
Cleanedup GitHub signup example

samgreen:
Made the changes mentioned in #182. Opened this PR so it would merge to develop.

Thanks!

### DIFF
--- a/RxExample/RxExample/Examples/GitHubSignup/Views/GitHubSignupViewController.swift
+++ b/RxExample/RxExample/Examples/GitHubSignup/Views/GitHubSignupViewController.swift
@@ -13,9 +13,6 @@ import RxSwift
 import RxCocoa
 #endif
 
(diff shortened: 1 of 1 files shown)


RxExample/RxExample/Examples/GitHubSignup/Views/GitHubSignupViewController.swift
```diff
@@ -108,9 +108,8 @@ class GitHubSignupViewController : ViewController {
                 let validationColor: UIColor
                 
                 if let valid = v.valid {
-                    validationColor = valid ? okColor : errorColor
-                }
-                else {
+                    validationColor = valid ? self.okColor : self.errorColor
+                } else {
```
kzaher: I think this would create a memory leak because of referencing strong `self`.

I'll merge this and try to play with this and see how to make this nicer but without leaks :)

kzaher: We dispose all subscriptions on `willMoveToParentViewController`, so it's not a leak so far, my bad :)

PR #190, opened 2015-10-13
TakeWhile predicates can throw

yury:


yury:
Need to port Error tests

PR #197, opened 2015-10-15
Adds rx_ extensions for UICollectionView focus for tvOS

gregpardo:


kzaher:
Hi @gregpardo ,

I was looking at PR. I think it would make sense to have these methods in public interface to support tvOS

```
public func rx_didUpdateFocusInContextWithAnimationCoordinator -> Observable<(context: UIFocusUpdateContext, animationCoordinator: UIFocusAnimationCoordinator)>

public func rx_modelAtIndexPath<T>(indexPath: NSIndexPath)
```

Didn't work with tvos but I'm hoping the proposed interface makes sense :) It looks to me that this is more general interface that covers all of the scenarios and reduces duplication of code.

It seems to me that 
```
rx_nextFocusedUpdated
rx_previousFocusedUpdated
rx_modelForNextFocusedUpdated
rx_modelForPreviousFocusedUpdated
```

could be derived easily from these two methods I'm proposing, and they would also expose animationCoordinator if needed.

This method

```
func rx_modelAtIndexPath<T>(indexPath: NSIndexPath)
```

would be more generic then tvOS and we would refactor 

```
 public func rx_modelSelected<T>() -> ControlEvent<T> {
        let source: Observable<T> = rx_itemSelected .map { indexPath in
            let dataSource: RxCollectionViewReactiveArrayDataSource<T> = castOrFatalError(self.rx_dataSource.forwardToDelegate(), message: "This method only works in case one of the `rx_itemsWith*` methods was used.")
            
            return dataSource.modelAtIndex(indexPath.item)!
        }
        
        return ControlEvent(source: source)
    }
```

to be something like 
```
 public func rx_modelSelected<T>() -> ControlEvent<T> {
        let source: Observable<T> = rx_itemSelected .map { indexPath in
            return self.rx_modelAtIndexPath(indexPath)
        }
        
        return ControlEvent(source: source)
    }
```

Let me know does that make sense :)

gregpardo:
Let me know what you think about the changes.

kzaher:
Hey @gregpardo ,

It looks like that case was little more complex then expected, I think I understand the confusion.

I think that
```
public func rx_modelAtIndexPath<T>(indexPath: NSIndexPath) throws -> T
```
should be just a normal synchronous method :)

```
public func rx_modelAtIndexPath<T>(indexPath: NSIndexPath) throws -> T {
         let dataSource: RxCollectionViewReactiveArrayDataSource<T> = castOrFatalError(self.rx_dataSource.forwardToDelegate(), message: "This method only works in case one of the `rx_itemsWith*` methods was used.")
                
         guard let element = dataSource.modelAtIndex(indexPath.item) else {
                throw rxError(.InvalidState, "Items not set yet.")
         }

         return element
    }
}
```

This one is ok.
```
public var rx_didUpdateFocusInContextWithAnimationCoordinator: ControlEvent<ContextAndAnimationCoordinator> {
```

Then this could become
```
public func rx_modelSelected<T>() -> ControlEvent<T> {
        let source: Observable<T> = rx_itemSelected.flatMap { indexPath in
            guard let element = try? self.rx_modelAtIndexPath(indexPath) else {
#if DEBUG
                  rxFatalError("Items don't exist in model, but something is shown in the UI")
#else
                  return empty()
            } 
            return just(element)
        }
        
        return ControlEvent(source: source)
    }
```

I am also confused what should be the correct behavior when something is displayed in tableview, but model isn't set yet.

I'm concerned about including those ones for now
```
rx_nextFocusedUpdated
rx_previousFocusedUpdated
rx_modelForNextFocusedUpdated
rx_modelForPreviouslyFocusedUpdated
```

it looks to me that once you have 2+ elements as parameters, there is a lot of combinations that we would need to cover, so it's probably best to not include those for now in the main repo :)

In the case of `rx_itemSelected`, that was a pretty straightforward thing to include since it's only one combination and it's extremely useful :)

The good thing is that if we only support `rx_modelAtIndexPath` and `rx_didUpdateFocusInContextWithAnimationCoordinator` it's really trivial for people to code up the exact combination they need

```
self.collectionView.rx_didUpdateFocusInContextWithAnimationCoordinator.map { context, _ in
       .....
        let nextItem = try self.rx_modelAtIndexPath(nextFocusedIndexPath)
        let previousItem = try self.rx_modelAtIndexPath(previousFocusedIndexPath)
      ....
}
```

Hope this makes sense :)

gregpardo:
Yes it makes sense a bit. I'm kind of a noob with swift and rx but I'm starting to get it and appreciate the help. 

So I am having a problem with your example for this method

```Swift
public func rx_modelSelected<T>() -> ControlEvent<T> 
```

I had to change 'guard let ement = ' to 'guard let element:T = ' or the compiler was throwing an error. Also I don't think I'm supposed to using just and empty here as the compiler is also complaining it's looking for type T not Observable. So I'm left with something like this....

```Swift
 public func rx_modelSelected<T>() throws -> ControlEvent<T> {
        let source: Observable<T> = rx_itemSelected.map { (indexPath: NSIndexPath) in
            guard let element:T = try? self.rx_modelAtIndexPath(indexPath) else {
#if DEBUG
                    rxFatalError("Items don't exist in model, but something is shown in the UI")
#endif
                return ???
            }
            return element
        }
        
        return ControlEvent(source: source)
    }
```

This is still not working because i'm not sure what to return for empty here. After this is in I should be good to go

gregpardo:
Okay so for now I left rx_modelSelected as it was before and added the same 'rx_didUpdateFocusInContextWithAnimationCoordinator' to UITableView. If you have suggestions on updating the rx_modelSelected please let me know. Think were getting there though.

kzaher:
Hi @gregpardo ,

I'm fine pulling in these changes :+1:

Issue #187, opened 2015-10-09
Auto loading help


sergdort:
Hey, guys I'm just playing with the library, trying to create my own Autoloading tableView.
So I just did like in the example, but when error occurs it just dispose loading.

So I was able to do kind of like this, by recursively call this function again. 
But I feel that there is more elegant way do do this

```swift
   private func startLoad() {
      let nextPageTriger =  self.tableView
         .rx_contentOffset
         .flatMap { (offset) in
            Pagination.shouldLoadNextBatch(offset, table: self.tableView)
               ? just()
               : empty()
      }
      
      Marvel.heroList(self.heroViewModels.value.count,loadNextBatch: nextPageTriger)
         .map { (heroes) -> [HeroListViewModel] in
            return heroes.map(HeroListViewModel.init)
         }
         .subscribe(next:
            { (heroes) -> Void in
               self.heroViewModels.value.appendContentsOf(heroes)
            }, error: { (err) -> Void in
               print(err)
               self.startLoad()
            }, disposed: { _ in
               print("disposed")
         })
         .addDisposableTo(self.disp)
   }
```

kzaher:
Hi @sergdort ,

there is `catchError*` and `retry*` families of operators. Was that what you were looking for?

sergdort:
Hi @kzaher ,

So I was able to refactor this way

```swift

   private func recursiveLoad() -> Observable<[HeroListViewModel]> {
      let nextPageTriger = tableView
         .rx_contentOffset
         .flatMap { (offset) in
            Pagination.shouldLoadNextBatch(offset, table: self.tableView)
               ? just()
               : empty()
      }
      return Marvel.heroList(heroViewModels.value.count,loadNextBatch: nextPageTriger)
         .map(HeroListViewModel.transform)
         .catchError { err in
            print(err)
            return self.recursiveLoad()
         }
   }

// And using it like this

      recursiveLoad()
         .subscribeNext({ (heroes) -> Void in
            self.heroViewModels.value.appendContentsOf(heroes)
         })
         .addDisposableTo(disp)

```

Sure this is bad practice, because it's leads to infinite number of calls if error constantly occurs (and user trying to scroll to bottom)
But when I try to use retry* operators its start loading from the beginning of first recursive call again
For example if I have asking for 20, 40 .. got error it will ask again with offset of 20
I guess I'm expecting wrong from the normal behavior.
[...]

PR #207, opened 2015-10-24
Handling network errors with reachability api in rx example

vodovozovge:
simple handling network errors in RxExample -> AutoLoading and RxExample -> WikipediaImageSearch.
Using Reachability (https://github.com/ashleymills/Reachability.swift) with reactive wrapper
Reference: https://github.com/ReactiveX/RxSwift/issues/187


vodovozovge:
some updates - implemented ReachabilityService and now used in AutoLoading and Wikipedia image search examples

I have several concerns:
1. Not sure, that extension of `ObservableConvertibleType` with `retryOnBecomesReachable` is in the right place (in ReachabilityService file now). If not, where should it be?
2. I use different placeholders for loading (empty) and for retrying (the one with "loading" label) `WikipediaSearchCell`. For me if connection is ok, images are loaded quite fast, so this "loading" label doesn't make sense. But when connection is dropped with showing "loading" placeholder we can tell user, that something happens and he can wait. Is it ok?

kzaher:
Hi @vodovozovge ,

I think that `retryOnBecomesReachable` is in ok place.

I think it would be best to have Image enum
```
enum Image {
     case Content(UIImage)
     case OfflinePlaceholder
}
```
and return `OfflinePlaceholder` when there is error fetching image.

We can create convenience observer on UIImage to bind `Image` (maybe `rxea_image`). In it we can use some centered [placeholder image similar to](http://www.outsystems.com/forgeCore/_image.aspx/A28atdGCIn2i_ZW11S_0KWgJm3iqQCfV_w==/offline-apps) in case OfflinePlaceholder or AspectFill in case `Content` is being bound.

Using some image seems more performant and closer to real world case then generating that image :)

Let me know what you think.

vodovozovge:
hey @kzaher 
I agree that it's not good idea to generate images all the way.
I used this code generated image because didn't want to add image to assets somehow. It's not a real issue for sure, so I will search for some open source image which represent lack of connectivity (seems your example is from some commercial system) and add it to assets.

I need some time to become familiar with binding (I found some info here https://github.com/ReactiveX/RxSwift/tree/master/Documentation#bindings , but not so detailed, so I missed it. Will check how it's done in examples)

Another issue is that we already have `Image` type alias

```
#if os(iOS)
    typealias Image = UIImage
#elseif os(OSX)
    typealias Image = NSImage
#endif
```

and actually not every image is downloadable. So for not downloadable images `OfflinePlaceholder` doesn't make sense. Probably we can introduce `DownloadableImage` enum, which will be used only for such case of fallible downloads?

Btw, what `rxea` prefix means? Rx example application?

kzaher:
Hi @vodovozovge ,

I think `DownloadableImage` is a fine name for enum.

Take a look at `rx_image` extension for `UIImage`. You can easily create similar `rxex_image` extension that even uses `rx_image` internally or does something similar after converting `DownloadableImage` to `UIImage`.

For picture, I think we can just use image of (🔴Offline) text in large font (emojis should be in vector form) or something similar so we don't have copyright issues. Maybe there are some nicer emojis.

I thought `rxex` or something similar for prefix to identify rx example app.

Take your time to study the existing extensions. No rush :)

RxExample/RxExample/Services/UIImageView+DownloadableImage.swift
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    func rxex_downloadableImageAnimated(transitionType:String?) -> AnyObserver<DownloadableImage> {
+
+        return AnyObserver { [weak self] event in
+            MainScheduler.ensureExecutingOnScheduler()
+
+            switch event{
+            case .Next(let value):
+                for subview in self?.subviews ?? [] {
```
vodovozovge: don't like this part, but should be ok for demo

vodovozovge:
Hi @kzaher 
I updated PR with some improvements: 
1. added `DownloadableImage` enum
2. extended `ImageService` to support downloading image with retry
3. extended `UIImageView` with `rxex_downloadableImage` and `rxex_downloadableImageAnimated`. I use `UILabel` with "⚠️" symbol for image placeholder and add it to image view when necessary. 
4. WikipediaImageSearch now uses `DownloadableImage` for handling images in cells

kzaher:
Cool :), I think this is much better.

There are some details I would like to fine tune, but I'm fine merging this :)

vodovozovge:
@kzaher can you share your thoughts when you have time? I want to make it better :)

kzaher:
khju

kzaher:
(Cleaning keyboard :) Nothing drastic :) Just some details :) I'll show you the commit ...

vodovozovge:
thanks)

PR #214, opened 2015-10-26
TakeLast operator

tottakai:


kzaher:
Hi @tottakai ,

I think we are missing this operator, and the code looks pretty much inline with our current style so if we can fix the issues I would gladly merge this :)

I'll note the issues in the code or the PR ...

RxSwift/Observables/Implementations/TakeLast.swift
```diff
@@ -0,0 +1,61 @@
...
+import Foundation
+
+
+class TakeLastSink<ElementType, O: ObserverType where O.E == ElementType> : Sink<O>, ObserverType {
+    typealias Parent = TakeLast<ElementType>
+    typealias E = ElementType
+    
+    private let _parent: Parent
+    
+    private var _elements = [ElementType]()
```
kzaher: I would use `Queue` for `_elements`. `removeFirst` for array has `O(n)` complexity where removing from `Queue` has O(1).

RxSwift/Observables/Implementations/TakeLast.swift
```diff
@@ -0,0 +1,61 @@
...
+            _elements.append(value)
+            if _elements.count > self._parent._count {
+                _elements.removeFirst()
+            }
+        case .Error:
+            observer?.on(event)
+            dispose()
+        case .Completed:
+            if self._elements.count > 0 {
+            self._elements.forEach { element in
```
kzaher: Probably use `for e in _elements` when using `Queue`, we can optimize `Queue` for that case better.

RxSwift/Observables/Observable+StandardSequenceOperators.swift
```diff
@@ -77,7 +77,30 @@ extension ObservableType {
...
+// MARK: takeLast
+
+extension ObservableType {
     
+    /**
+    Returns a specified number of the last elements before the observable sequence completed.
+     
+     - parameter count: The number of elements to return.
+     - returns: An observable sequence that contains the specified number of elements from the end of the input sequence.
+     */
```
kzaher: Could you please copy the documentation from

https://github.com/Reactive-Extensions/RxJS/blob/8f12f812d497acf639588e90f74d504a9fc801ec/src/core/linq/observable/takelast.js#L27

... to be consistent :)

RxSwift/Observables/Observable+StandardSequenceOperators.swift
```diff
@@ -77,7 +77,30 @@ extension ObservableType {
...
+    @warn_unused_result(message="http://git.io/rxs.uo")
+    public func takeLast(count: Int)
+        -> Observable<E> {
+        if count == 0 {
+            return empty()
+        }
+        else {
+            return TakeLast(source: self.asObservable(), count: count)
+        }
```
kzaher: I think that optimization makes sense in the case of taking first elements, but when taking 0 elements from end, that's not equivalent to `empty` because sequence can error out before it completes, and that would diverge in this case.

TLDR, just remove
```
if count == 0 {
+            return empty()
+        }
+        else {
```

:)

RxTests/RxSwiftTests/Tests/Observable+StandardSequenceOperatorsTest.swift
```diff
@@ -2816,6 +2816,254 @@ extension ObservableStandardSequenceOperators {
...
+        
+        let res = scheduler.start {
+            xs.takeLast(0)
+        }
+        
+        XCTAssertEqual(res.messages, [
+            completed(200)
+            ])
+        
+        XCTAssertEqual(xs.subscriptions, [
```
kzaher: XCTAssertEqual(res.messages, [

])

XCTAssertEqual(xs.subscriptions, [200, 1000])

RxSwift/Observables/Implementations/TakeLast.swift
```diff
@@ -0,0 +1,61 @@
...
+            dispose()
+        }
+    }
+}
+
+class TakeLast<Element>: Producer<Element> {
+    private let _source: Observable<Element>
+    private let _count: Int
+    
+    init(source: Observable<Element>, count: Int) {
```
kzaher: Could you please add a check here
```
if count < 0 {
            rxFatalError("count can't be negative")
        }
```

I think we are missing the same check in `TakeCount`, so if you can add it there also, that would be great :)

kzaher:
... the unit tests are failing because you need to add `TakeLast.swift` to `RxExample` project.

We test integration without modules using `RxExample` app, so that's why you need to add that file to `RxExample-iOS-no-module` target,  group `NoModule/RxSwift/Observables/Implementations` :)

tottakai:
Thanks for the comments, fixed all of them and also added the `TakeCount` `count < 0 check`

kzaher:
Cool, I think it should be `_elements = Queue<ElementType>(capacity: parent._count + 1)`, but I can also fix that :)

PR #222, opened 2015-10-29
Feature/fix build warnings

Tableau-David-Potter:
### PLEASE IGNORE THIS PULL REQUEST
*I intended to make a pull request into develop, not master.  I've submitted a separate pull request.*

Fix warnings due to use in app extensions and to Xcode updates.

Xcode gives two warnings when including the Rx project and building in an app.

1. The first is due to use in a component that will be used in an extension. Xcode displays the message “linking against dylib not safe for use in application extensions”. To configure an app extension target to use an embedded framework, set the target’s “Require Only App-Extension-Safe API” build setting to Yes.

    https://developer.apple.com/library/prerelease/ios/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html

2. The second is due to changes in Xcode that want to update the project file to comply with the latest guidelines. Clicking the warning brings up a prompt to modify the project.

### DIFF
--- a/Documentation/API.md
+++ b/Documentation/API.md
@@ -21,6 +21,8 @@ Operators are stateless by default.
  * [`never`](http://reactivex.io/documentation/operators/empty-never-throw.html)
  * [`returnElement` / `just`](http://reactivex.io/documentation/operators/just.html)
  * [`returnElements`](http://reactivex.io/documentation/operators/from.html)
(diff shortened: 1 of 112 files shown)


kzaher:
Dupe of https://github.com/ReactiveX/RxSwift/pull/223

PR #225, opened 2015-10-30
Adds `window` operator

bontoJR:
This PR adds the operator version of [`window`](http://reactivex.io/documentation/operators/window.html) with time or count.

Other implementations added:
* `AddRef` operator
* `RefCountDisposable`
* `BooleanDisposable`


kzaher:
Hi @bontoJR ,

I would like to include this in beta.3 together with `retryWhen` and `flatMapFirst`. My mind is now focused on optimizations and wrapping beta.2 so I would like to do a detail CR this in a few days :)

I've glanced at it, thnx for including `test` prefix, yikes, I've overlooked it somehow :)

There is one unit tests commented in there, should we remove that code?

Looks good overall, but I would like to take a more relaxed in a few days :)

bontoJR:
> There is one unit tests commented in there, should we remove that code?

That test is for the simple version of window (the one with just count, that has a very limited use case), I forgot to remove it, so yes, it can be removed.

I have to focus on `retryWhen` with @yury and deliver before the next beta.

Take your time, with the `window` operator other implementations are coming like `addRef` and `RefCountDisposable` that are not very simple to understand at a first look. If you need everything else, just ask. :)

PR #227, opened 2015-10-30
Update os x framework target version

### LINKED ISSUE #224: Differing platform target versions
I've noticed that the target version for the os x podspec is 10.9 while the framework targets in the project have their version set is 10.10. Is there a reason for this or is it just a mistake?

petester42:
Sets the deployment target version to 10.9 for OS X just like it is for the podspec. This closes #224.


kzaher:
Cool :+1:

PR #237, opened 2015-11-07
Remove Unneeded dispose bag

zhigang1992:
Not sure if this line is here on purposely.

Seems like It's already initialized. Is it here for memory management?


kzaher:
Indeed, it looks like that is the case :)

PR #242, opened 2015-11-08
Added tvOS specific extension for UIButton

larslockefeer:
On tvOS, rather than the `.TouchUpInside` event, the `. PrimaryActionTriggered` event is sent when a User interacts with a button. As a result of this, `rx_tap` will not work as expected on tvOS.

This pull request adds a tvOS group which currently contains just one file, `UIButton+Rx.swift` to resolve this issue. I could have branched within the current file as well using `os(tvOS)`, but I expect that more differences will occur going forward. This seemed to be the cleaner solution.

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		9D71C4D21BF08191006E8F59 /* UIButton+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = C88254061B8A752B00B02D69 /* UIButton+Rx.swift */; };
 		B1B7C3BD1BDD39DB0076934E /* TakeLast.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1B7C3BC1BDD39DB0076934E /* TakeLast.swift */; };
 		B1B7C3BE1BDD39DB0076934E /* TakeLast.swift in Sources */ = {isa = PBXBuildFile; fileRef = B1B7C3BC1BDD39DB0076934E /* TakeLast.swift */; };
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @larslockefeer ,

I think this is a bug in tvOS. I don't have much experience there we'll probably need more help there :)

If you could rearrange original file to something like the following and point this PR to `develop` branch, I would be happy to merge this :)

```
//
//  UIButton+Rx.swift
//  RxCocoa
//
//  Created by Krunoslav Zaher on 3/28/15.
//  Copyright (c) 2015 Krunoslav Zaher. All rights reserved.
//

#if os(iOS)

import Foundation
#if !RX_NO_MODULE
import RxSwift
#endif
import UIKit

extension UIButton {
    
    /**
    Reactive wrapper for `TouchUpInside` control event.
    */
    public var rx_tap: ControlEvent<Void> {
		return rx_controlEvents(.TouchUpInside)
    }
    
}

#endif

#if os(tvOS)

import Foundation
#if !RX_NO_MODULE
import RxSwift
#endif
import UIKit

+extension UIButton {
+
+    /**
+     Reactive wrapper for `PrimaryActionTriggered` control event.
+     */
+    public var rx_primaryAction: ControlEvent<Void> {
+        return rx_controlEvents(.PrimaryActionTriggered)
+    }
+}

#endif
```

The reason why I would like to have this in a single file for now is because 
* we support CP and Carthage, and this currently would kill current version of CP integration (this can be done for sure, just didn't have time)
* having two separate files would make it harder for me to maintain our integrations and see per platform differences for same set of extensions (Cocoa and UIKit are vastly more different then UIKit specializations)
* tvOS is technically still UIKit, and yes, I agree, the current name of folder (iOS) doesn't describe it perfect, but maybe we can rearrange project structure in some other PR :)

larslockefeer:
Hi @kzaher,

I don't think it is a bug; for reference, see the following response in [this thread](https://forums.developer.apple.com/thread/17925):

screen shot 2015-11-09 at 08 16 36

I see your point in branching within the same file for now, I have updated my PR accordingly and pointed it at the `develop` branch.

larslockefeer:
See #243 for this PR pointed at `develop`. Thanks!

PR #243, opened 2015-11-09
Added tvOS specific extension for UIButton

larslockefeer:
This is a clone or PR #242, now pointed at the `develop` branch.

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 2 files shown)


kzaher:
Hey @larslockefeer ,

Merging this...

I think I'll rename `rx_tap` to `rx_primaryAction` after merge and add it to unit tests :) I think `rx_primaryAction` probably describes this better :)

Hope that makes sense.

larslockefeer:
@kzaher I have also been thinking about the name. For me as an iOS developer, I think `rx_tap` makes sense since this is the way you think about interacting with a `UIButton`. But for people doing `tvOS` only (which eventually will exist), `rx_primaryAction` will probably be better. Just like we don't talk about clicks on iOS like we do on OS X.

Trade offs :smile:

kzaher:
For OSX I have no idea what to use :)

I tap on everything on my laptop, some people click on everything :) I think we'll probably go with `rx_action` on OSX. What a mess :)

yury:
`rx_action`!

Issue #235, opened 2015-11-05
Is there a single result ensure operator?


frogcjn:
that operator can ensure the sequence have and only have one result before completed (maybe disposed?)

Looks like

             var subject: Observable<Element>
             subject
                .toArray()
                .map { array -> Element in
                    if array.count > 1{ throw CommonError.MoreThanOneResult}
                    if array.count < 1{ throw CommonError.NoResult}
                    return array.first!
                }

kzaher:
Hi @frogcjn ,

we don't have a specialized `single` operator right now but I think we would want to have it, both nonblocking and blocking version.

You can simulate it meanwhile in a way you've described :)

kzaher:
Hey guys,

we would like to port `single` operator with unit tests. Both nonblocking and blocking version.

It would probably make sense to split this up, so if somebody didn't port blocking operator, he/she can take a peek at that code and vice versa :)

I think that recently @yury, @hsubra89 has asked me recently about contributing, @bontoJR has already ported some nice operators, no this should be piece of cake, and @alcarvalho asked me before some time.

We also have some new people in contributors channel, so more people know and look at the codebase the better :)

If somebody is interested in helping us port this, he/she can respond to this thread :)

hsubra89:
I'm definitely interested in contributing. However, I don't currently have the bandwidth. I'll be able to start helping out sometime this weekend. :)

kzaher:
`single` was just released.

PR #245, opened 2015-11-09
Add `single` non-blocking operator

bontoJR:
This PR adds the single, non-blocking version, operator. Issue related: #235


RxSwift/Observables/Implementations/SingleAsync.swift
```diff
@@ -0,0 +1,80 @@
...
+        super.init(observer: observer)
+    }
+    
+    func on(event: Event<E>) {
+        switch event {
+        case .Next(let value):
+            
+            if let predicate = _parent._predicate {
+                do {
+                    print("val: \(value)")
```
kzaher: There is a `print` statement accidentally left, I'll push a change to fix this ...

PR #246, opened 2015-11-09
Blocking version of single operator

tottakai:
Adds the blocking no-predicate single operator.  Issue related: #235

### DIFF
--- a/RxBlocking/BlockingObservable+Operators.swift
+++ b/RxBlocking/BlockingObservable+Operators.swift
@@ -145,3 +145,56 @@ extension BlockingObservable {
         return element
     }
 }
+
+extension BlockingObservable {
+    /**
+     Blocks current thread until sequence terminates.
+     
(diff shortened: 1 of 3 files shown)


RxBlocking/BlockingObservable+Operators.swift
```diff
@@ -144,4 +144,57 @@ extension BlockingObservable {
...
+}
+
+extension BlockingObservable {
+    /**
+     Blocks current thread until sequence terminates.
+     
+     If sequence terminates with error before producing first element, terminating error will be thrown.
+     
+     - returns: Returns the only element of an sequence, and reports an exception if there is not exactly one element in the observable sequence.
+     */
```
kzaher: I think it would probably be better to say `error` then `exception` in context of Swift. But I can change that after merging. Not a big deal.

kzaher:
Awesome, thnx :)

PR #249, opened 2015-11-10
Blocking single operator with predicate

tottakai:
Adds the blocking predicate single operator. Issue related: #235

### DIFF
--- a/RxBlocking/BlockingObservable+Operators.swift
+++ b/RxBlocking/BlockingObservable+Operators.swift
@@ -155,47 +155,7 @@ extension BlockingObservable {
      - returns: Returns the only element of an sequence, and reports an exception if there is not exactly one element in the observable sequence.
      */
     public func single() throws -> E? {
-        var element: E?
-        
-        var error: ErrorType?
-        
-        let d = SingleAssignmentDisposable()
-        
-        let lock = RunLoopLock()
-        
-        lock.dispatch {
-            d.disposable = self.source.subscribe { e in
-                switch e {
-                case .Next(let e):
-                    if element == nil {
-                        element = e
-                    } else {
-                        error = RxError.MoreThanOneElement
-                    }
-                    return
-                case .Error(let e):
-                    error = e
-                    break
-                case .Completed:
-                    if element == nil {
-                        error = RxError.NoElements
-                    }
-                    break
-                }
-                
-                lock.stop()
-            }
-        }
-        
-        lock.run()
-        
-        d.dispose()
-        
-        if let error = error {
-            throw error
-        }
-
-        return element
+        return try single { _ in true }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


RxBlocking/BlockingObservable+Operators.swift
```diff
@@ -197,4 +197,64 @@ extension BlockingObservable {
...
+        var element: E?
+        
+        var error: ErrorType?
+        
+        let d = SingleAssignmentDisposable()
+        
+        let lock = RunLoopLock()
+        
+        lock.dispatch {
+            d.disposable = self.source.subscribe { e in
```
kzaher: Hi @tottakai ,

in case error occurs we should dispose subscription and exit the loop immediately. There shouldn't be any difference between error happening in `Next` or being received through `Error` event. So it seems to be that should be ....

```
+                switch e {
+                case .Next(let e):
+                    do {
+                        if ! try predicate(e) {
+                               return
+                        }
+                            if element == nil {
+                                element = e
+                            } else {
+                                throw RxError.MoreThanOneElement
+                            }
+                    } catch (let err) {
+                        error = err
+                        d.dispose()
+                        lock.stop()
+                    }
+                    return
+                case .Error(let e):
+                    error = e
+                    lock.stop()
+                case .Completed:
+                    if element == nil {
+                        error = RxError.NoElements
+                    }
+                    break
+                }
+
+                lock.stop()
+            }
+        }
````

File: RxBlocking/BlockingObservable+Operators.swift [hunk 2]
```diff
@@ -197,4 +197,64 @@ extension BlockingObservable {
...
+
+    /**
+     Blocks current thread until sequence terminates.
+     
+     If sequence terminates with error before producing first element, terminating error will be thrown.
+     
+     - parameter predicate: A function to test each source element for a condition.
+     - returns: Returns the only element of an sequence that satisfies the condition in the predicate, and reports an error if there is not exactly one element in the sequence.
+     */
+    public func single(predicate: (E) throws -> Bool) throws -> E? {
```
kzaher: It would be great if you could just make `single` version without predicate be an alias for this one 

```
public func single() throws -> E? {
    return single { _ in true }
}
```

I think that in this case there aren't any reasons why we would want to keep that code almost duplicated.

RxBlocking/BlockingObservable+Operators.swift
tottakai: Hi @kzaher,

like above, but `.Completed` has `if error == nil` check, so that the predicate error code is returned from `single`.

RxBlocking/BlockingObservable+Operators.swift
tottakai: The no-predicate version now uses the predicate version, just added `try` before `single`.

Issue #236, opened 2015-11-06
KVO enum type enhancement and context


frogcjn:
AVPlayer has an enum property. it's an integer enum type AVPlayerStatus.

So I must write in this format to observe

            _ = self.player.rx_observe("status").subscribeNext{ (value: Int?) -> Void in
                guard let value = value else {
                    print("not found")
                    return
                }
                guard let status = AVPlayerItemStatus(rawValue: value) else {
                    print("not found")
                    return
                }

or use .map operator map an Int? to AVPayerItemStatus?

instead of subscribe the AVPlayerItemStatus enum type directly

            _ = self.player.rx_observe("status").subscribeNext{ (status: AVPlayerItemStatus?) -> Void in
                guard let status = status else {
                    print("not found")
                    return
                }

This is because NSNumber cannot use "as" operator to transform to an Int enum type

    class KVOObservable<Element>{
        ...
    func subscribe<O : ObserverType where O.E == Element?>(observer: O) -> Disposable {
        let observer = KVOObserver(parent: self) { (value) in
            if value as? NSNull != nil {
                observer.on(.Next(nil))
                return
            }
            observer.on(.Next(value as? Element))    //!!! This line should be extended for raw enum values.
        }

--------------------------
Another issue is, is it better to add context in KVC?
in _RXKVOObserver.h

        @implementation _RXKVOObserver
        [self.target addObserver:self forKeyPath:self.keyPath options:options context:nil];//!!! better for adding a context?

--------------------------
is it better to move rx_observe into RxSwift instead of RxCocoa? 

Because rx_observe is for NSObject (Foundation framework) instead of Cocoa Touch (UIKit framework)

kzaher:
Hi @frogcjn :)

I have done some investigations regarding your request.

For us to be able to include something like 
```
_ = self.player.rx_observe("status").subscribeNext{ (status: AVPlayerItemStatus?) -> Void in
            guard let status = status else {
                print("not found")
                return
            }
```

We would need to have `AVPlayerItemStatus` conform to some protocol type. Even though `AVPlayerStatus` is declared as `AVPlayerStatus: Int` it does not conform to `IntegerType` or any other protocol that we could use to automate this.

The closest thing that we can do to support this is to define our own protocol and extension inside of RxCocoa:

```
public protocol EnumConvertibleFromInt {
    init?(rawValue: Int)
}

extension NSObject {
[...]

Issue #241, opened 2015-11-08
Is there a nicer way of observing object changes?


lyptt:
I've got a very basic example going with the latest RxSwift 2 version, admittedly I'm fairly new to the rx way of doing things:

```swift
combineLatest(firstNameField.rx_text, lastNameField.rx_text) { UserModel(firstName: $0, lastName: $1) }
	.subscribeNext { [unowned self] in
		self.user = $0
	}
	.addDisposableTo(disposeBag)

(self.rx_observe("user") as Observable<UserModel?>)
	.map { user in
		return user?.isValid() as Bool!
	}
	.bindTo(signInButton.rx_enabled)
	.addDisposableTo(disposeBag)
```

What I'm attempting to do is generate a new `UserModel` from two observables, and bind the result of that new model's isValid() function to a button's enabled state.

I've found that KVO works for this use case, but it requires using classes and dynamic variables to do so. Is there a better way of achieving this without KVO?

dtartaglia:
I'm pretty new as well, but I think if you make your user variable a `Variable` then you will be able to observe it without using KVO.

i.e.: `let user = Variable<UserModel?>()`

Then in your `subscribeNext` do `self.user.value = $0`

And observe changes to `user` by doing `user.subscribeNext { self.user.value.isValid() as Bool! }`

Something like that at least.

lyptt:
Oh cool! That looks like exactly what I need. I'll give that a go. Thanks!

kzaher:
Thnx @dtartaglia for answering.

Hi @lyptt, I would just like to add that you are trying to solve this problem imperatively :) Our example app is lacking good MVVM examples for now, but we'll improve it in future :)

lyptt:
@kzaher yes it's probably a bit more imperative than you should be with rx, I'm trying to slowly easy my way into it bit by bit! The examples you've currently got are a bit beyond my comprehension at the moment :)

kzaher:
Yeah, we have two problems.
* how to educate new users
* how to demonstrate advanced usages

Those two things don't mix well together :) We'll need to find some better ways to satisfy all of those cases :)

kzaher:
We've released some KVO changes that should help with this issue, but using KVO for non ObjC is discouraged.

We'll still need to work on examples, but closing this for now.

PR #258, opened 2015-11-14
KVO revamp.

kzaher:
Work related with https://github.com/ReactiveX/RxSwift/issues/236

It should help people use KVO more easily

```
object.rx_observe(AVPlayerStatus.self, "status").subscribeNext { status in
    ...
}
```

Improvements
* no more confusing deduction of type of observed element (now it's explicit first argument)
* support for Enums and RawRepresentable

It also introduces new protocol `KVORepresentable` that helps to map KVO type that is registered for observation and return type of observed elements.


kzaher:
Also related with https://github.com/ReactiveX/RxSwift/issues/241

ashfurrow:
Clever solution! :100:

PR #257, opened 2015-11-14
Move parameter line to correct function

a2:
The `maxConcurrent` parameter line was in the wrong function, documenting the parameter-less function.

PR #260, opened 2015-11-14
Move parameter line to correct function

a2:
The `maxConcurrent` parameter line was in the wrong function, documenting the parameter-less function.

Closes #257 and #259.


kzaher:
Cool, thnx :+1:

Issue #234, opened 2015-11-05
Is there a timeout operator?


frogcjn:
Is there a timeout operator?
http://reactivex.io/documentation/operators/timeout.html

AVAsset has a function called loadValuesAsynchronouslyForKeys. I'm going to use RxSwift to deal with it. But if an HTTP resource cannot be got, this method will wait until 60 seconds to return. So there is a need  to limit the time for this async fun to return.

frogcjn:
I can use

            .takeUntil(timer(2.0, MainScheduler.sharedInstance))

but there is no error will return. Just complete.

IanKeen:
You could create a 'timeout' observable that simply errors after a set duration and merge that with your request.. possibly not ideal but you could abstract it into something nice

frogcjn:
@IanKeen , this request is system triggered, which I cannot change the timeout property of the NSURLRequest.

IanKeen:
right, but you can combine it with another observable which you do control the timeout of (unsure if this is even the best way... just a suggestion on how you might tackle the problem)

frogcjn:
@IanKeen 

like

        .takeUntil(timer(2.0, MainScheduler.sharedInstance))
but there is no error will return. Just complete.

So I should have a single result ensure operator like this

                .toArray()
                .map { array -> Bool in
                    if array.count > 1{ throw CommonError.MoreThanOneResult}
                    if array.count < 1{ throw CommonError.NoResult}
                    return array.first!
                }

IanKeen:
I was thinking perhaps something like 

```
        return create { observer in
            let delayTime = dispatch_time(DISPATCH_TIME_NOW, Int64(2.0 * Double(NSEC_PER_SEC)))
            dispatch_after(delayTime, dispatch_get_main_queue()) {
                observer.onError(error)
            }
        }
```

frogcjn:
@IanKeen

Yes, you're right.

There also needs a flag to mark whether the function was returned by system.
        return create { observer in
           var returned = false
            let delayTime = dispatch_time(DISPATCH_TIME_NOW, Int64(2.0 * Double(NSEC_PER_SEC)))
            dispatch_after(delayTime, dispatch_get_main_queue()) {
                if !returned { observer.onError(error) }
            }
            someSystemAsyncFunc{
                  returned = true
                   observer.onNext(...)
                   observer.onCompleted()
            }
        }

kzaher:
Hi,

I believe timeout should be something like

```
[...]

PR #262, opened 2015-11-16
Adds ignoreElements() operator

ashfurrow:
Hey there! Here's an implementation of `ignoreElements()` – had a difficult time figuring out the test harness setup, but I really like it :heart_eyes_cat: Is there any documentation around it? 

Edit: Just for clarification, this is a small operator, but one that I needed in [this app](https://github.com/artsy/eidolon) for abstracting network request _completion_ from network response data.

### DIFF
--- a/RxSwift/Observables/Observable+Time.swift
+++ b/RxSwift/Observables/Observable+Time.swift
@@ -170,6 +170,23 @@ extension ObservableType {
     }
 }
 
+// MARK: ignoreElements
+
+extension ObservableType {
+
+    /**
+     Skips elements and completes (or errors) when the receiver completes (or errors). Equivalent to filter that always returns false.
+
+     - returns: An observable sequence that skips all elements of the source sequence.
+     */
+    @warn_unused_result(message="http://git.io/rxs.uo")
+    public func ignoreElements()
+        -> Observable<E> {
+            return filter { _ -> Bool in
+                return false
+            }
+    }
+}
 
 // MARK: delaySubscription
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+* Adds `ignoreElements` operator.
+
 ---
 
 ## [2.0.0-beta.3](https://github.com/ReactiveX/RxSwift/releases/tag/2.0.0-beta.3)
--- a/RxTests/RxSwiftTests/Tests/Observable+TimeTest.swift
+++ b/RxTests/RxSwiftTests/Tests/Observable+TimeTest.swift
@@ -1178,6 +1178,32 @@ extension ObservableTimeTest {
     }
 }
 
+// MARK: IgnoreElements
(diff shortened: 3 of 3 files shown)


kzaher:
Hi @ashfurrow ,

I understand that there is a `ignoreElements` registered on ReactiveX portal, but that operator should be fairly easy to make using filter if somebody needs it :) If it's ok with you, I would rather not include it for now.

I think next operator we would want to port is `timeout`. @yury said he would like to try so it would be good to sync with him regarding that :)

We always port all corresponding Rx.NET tests (fallback RxJS), to make sure our all implementations are  compliant.

As far as for test suite explanation, we basically use the same testing and operator architecture like Rx.NET and RxJS. There unfortunately isn't a lot of information on the internet on that, but yeah, those are one of the nicest unit tests one can write :) And testing architecture is extremely simple, but it's hard to explain it in PR comment :)

I'll try to explain it in some `md` when I find time, or we can have a video chat if you are interested in helping out with documenting it :)

ashfurrow:
OK, thanks for the response. Could I get a few clarifications?

- Do I need to write tests for newly ported operators? Or should I copy them from somewhere, or should I just not write them?
- Should I close this PR? I mean, it seems like kind of a waste since I've already written the code. 

Like, I understand there's a priority list of things to implement – this was an operator I needed so I implemented it ¯\\_(ツ)_/¯

RxSwift/Observables/Observable+Time.swift
```diff
@@ -170,6 +170,23 @@ extension ObservableType {
     }
 }
 
+// MARK: ignoreElements
+
+extension ObservableType {
+
+    /**
+     Skips elements and comples (or errors) when the receiver completes (or errors). Equivalent to filter that always returns false.
```
orta: comples ?

ashfurrow: Completes* I copied from a mother method, so I'll check that one too.

tottakai:
I started doing timeout over the weekend as it was mentioned in #234. Currently the operator with dueTime is done, including unit tests. Also started doing a overload which takes an other observable, to which it falls to when a timeout to the source observable occurs, currently thats in progress and lacking tests. 
Do you @yury want to help out, currently i have like a hour per evening time to work on it :), or maybe do the fallback observable version from scratch :)

yury:
@tottakai I really want to help. But I think, this week I will be unavailable also... Sorry

bontoJR:
Do you guys have the code somewhere? I can take a look and eventually complete it. :)

ashfurrow:
I'm still not clear on why we're not merging this pull request. It's part of the API, it's tested, and I need it.

ashfurrow:
I've pushed an amended commit that corrects the typo.

ashfurrow:
I've also corrected the return value of the documentation :tada:

kzaher:
Hey, thnx :)

I was little confused why you needed that. Thnx for explaining me (we've slacked).

Merging ...

PR #263, opened 2015-11-16
Adds UILabel rx_attributedText

ashfurrow:
Didn't find many examples of UI testing, guidance here would be :100:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+* Adds `rx_attributedText` to `UILabel`.
+
 ---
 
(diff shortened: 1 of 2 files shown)


kzaher:
Hi, 

We haven't done testing for writeonly/observer extensions because they are simple. But I think that setting up something like `ensureEventDeallocated` or `ensurePropertyDeallocated` test method would make sense probably.

We can name it `ensureControlObserverHasWeakReference` or something.

I think we want to test 2 things:

* that observer isn't holding a weak reference
* that sending next elements sets correct control value.

We've mostly written tests for properties and events, but I guess extensions observers also need their set of tests.

I think it should be `AnyObserver<NSAttributedString?>` because of `@NSCopying public var attributedText: NSAttributedString? ` :)

If you could point this PR to `develop` branch that would be great :)

ashfurrow:
Ha, whoops. Will re-open.

PR #264, opened 2015-11-16
Adds UILabel rx_attributedText.

ashfurrow:
Continued from #263.

I'll try and add tests later today.

### DIFF
--- a/RxTests/RxCocoaTests/Control+RxTests.swift
+++ b/RxTests/RxCocoaTests/Control+RxTests.swift
@@ -78,4 +78,24 @@ class ControlTests : RxTest {
         XCTAssertTrue(deallocated)
         XCTAssertTrue(completed)
     }
+
+    func ensureControlObserverHasWeakReference<C, T where C: NSObject>(@autoclosure createControl: () -> (C), _ observerSelector: C -> AnyObserver<T>, _ observableSelector: () -> (Observable<T>)) {
+        var deallocated = false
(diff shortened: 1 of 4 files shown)


ashfurrow:
OK, how does this look?

RxTests/RxCocoaTests/Control+RxTests.swift
```diff
@@ -77,4 +77,24 @@ class ControlTests : RxTest {
...
+
+    func ensureControlObserverHasWeakReference<C, T where C: NSObject>(@autoclosure createControl: () -> (C), _ observerSelector: C -> AnyObserver<T>, _ observableSelector: () -> (Observable<T>)) {
+        var deallocated = false
+
+        autoreleasepool {
+            let control = createControl()
+            let propertyObserver = observerSelector(control)
+            let observable = observableSelector()
+
+            let disposeBag = DisposeBag()
```
kzaher: Hey, looks great.

It would be great if you could move that `disposeBag` above `autoreleasepool`, because in that way we would be better testing is that control being deallocated in spite of subscription lasting.

I've just added similar improvement to control event testing code.

If you could do that and resolve merging conflicts, that would be :100: :)

ashfurrow:
Cool, rebased from `develop` and made the change with the dispose bag :bow:

kzaher:
Cool, thnx :)

PR #267, opened 2015-11-17
Making withLatestFrom subscribe to the right hand side first

mikelikespie:
This resolves an issue where you have two observables that yield values
immediately and it wasn't yielding a result


mikelikespie:
cc @yury who appears to be the withLatestFrom author

kzaher:
Ah, thnx :)

PR #269, opened 2015-11-18
Timeout operator

tottakai:
Timeout operator as per #234 

@bontoJR said that we would be interested in doing the overload that takes the fallback observable (other) as a parameter.

### DIFF
--- a/RxSwift/Observables/Implementations/Timeout.swift
+++ b/RxSwift/Observables/Implementations/Timeout.swift
@@ -1,0 +1,88 @@
+//
+//  Timeout.swift
+//  Rx
+//
+//  Created by Tomi Koskinen on 13/11/15.
+//  Copyright © 2015 Krunoslav Zaher. All rights reserved.
+//
+
+import Foundation
+
+class TimeoutSink<ElementType, Scheduler: SchedulerType, O: ObserverType where O.E == ElementType>
+    : Sink<O>
+    , LockOwnerType
+    , ObserverType
+    , SynchronizedOnType {
+    typealias E = ElementType
+    typealias Parent = Timeout<E, Scheduler>
+    
+    private let _parent: Parent
+    let _lock = NSRecursiveLock()
+
+    private let _timerD = SerialDisposable()
+    
+    init(parent: Parent, observer: O) {
+        _parent = parent
+        super.init(observer: observer)
+    }
+    
+    func run() -> Disposable {
+        _createTimeoutTimer()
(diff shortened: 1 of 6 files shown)


kzaher:
Hey @tottakai ,

will check it out over weekend, looks great on first read, but hard to really concentrate over week. 

Will check it out in more detail over weekend when I get some rest :)

kzaher:
Hi @tottakai ,

I think the this code that you've provided for version without second observable is accurate :)

I think it would be great to reuse code for that `timeout` operator for both versions, with and without second observable (Swift generates huge binary dumps, and don't think it would make any significant performance difference :)

You can check out .NET code [here](https://github.com/Reactive-Extensions/Rx.NET/blob/master/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/Timeout.cs#L139) (it's probably close to what we need).

In that way, we could make `timeout` operator without second observable to be just a convenience for `timeout(delay, failWith(RxError.Timeout)`.

I think unit tests for it are [from here](https://github.com/Reactive-Extensions/Rx.NET/blob/859e6159cb07be67fd36b18c2ae2b9a62979cb6d/Rx.NET/Source/Tests.System.Reactive/Tests/Linq/ObservableTimeTest.cs#L6392) [to here](https://github.com/Reactive-Extensions/Rx.NET/blob/859e6159cb07be67fd36b18c2ae2b9a62979cb6d/Rx.NET/Source/Tests.System.Reactive/Tests/Linq/ObservableTimeTest.cs#L6702).

I think it would be wonderful to have fully compliant unit tests, don't think we have any reasons to diverge here :)

It's not too far from your original code, so I'm hoping you are ok with making those changes. Or if you don't have time, me or somebody else can do them ofc, no pressure :)

RxSwift/Observables/Implementations/Timeout.swift
```diff
@@ -0,0 +1,88 @@
...
+    
+    func _synchronized_on(event: Event<E>) {
+        switch event {
+        case .Next:
+            forwardOn(event)
+            self._createTimeoutTimer()
+        case .Error:
+            forwardOn(event)
+            dispose()
+        case .Completed:
```
yury: Can be written:
```
case .Error, .Completed:
  ...
```

kzaher:
Ok, so this is a dupe of https://github.com/ReactiveX/RxSwift/pull/284 :)
I see @tottakai's commit there.

Closing this one then.

PR #284, opened 2015-11-24
Timeout operator by @tottakai

bontoJR:
I took the code from @tottakai and updated his version to be similar to the Rx.NET version.
I tried to update the original PR, but I can't find a way in the new interface to achieve that (I remember it was possible before) :neutral_face:

### DIFF
--- a/RxSwift/Observables/Implementations/Timeout.swift
+++ b/RxSwift/Observables/Implementations/Timeout.swift
@@ -46,7 +46,7 @@ class TimeoutSink<ElementType, Scheduler: SchedulerType, O: ObserverType where O
             _lock.performLocked() {
                 onNextWins = !self._switched
                 if onNextWins {
-                    self._id++
+                    self._id = self._id &+ 1
                 }
             }
             
@@ -60,7 +60,7 @@ class TimeoutSink<ElementType, Scheduler: SchedulerType, O: ObserverType where O
             _lock.performLocked() {
                 onEventWins = !self._switched
                 if onEventWins {
-                    self._id++
+                    self._id = self._id &+ 1
                 }
             }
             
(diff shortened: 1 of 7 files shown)


tottakai:
@bontoJR looks good, now i can watch football tonight ;) thx

RxSwift/Observables/Observable+Time.swift
```diff
@@ -245,3 +245,34 @@ extension ObservableType {
...
+    }
+    
+    /**
+     Applies a timeout policy for each element in the observable sequence, using the specified scheduler to run timeout timers. If the next element isn't received within the specified timeout duration starting from its predecessor, the other observable sequence is used to produce future messages from that point on.
+     
+     - parameter dueTime: Maximum duration between values before a timeout occurs.
+     - parameter other: Sequence to return in case of a timeout.
+     - parameter scheduler: Scheduler to run the timeout timer on.
+     - returns: The source sequence switching to the other sequence in case of a timeout.
+     */
```
kzaher: I think we are missing @warn_unused_result(message="http://git.io/rxs.uo") :)

RxSwift/Observables/Implementations/Timeout.swift
```diff
@@ -0,0 +1,122 @@
...
+
+    func on(event: Event<E>) {
+        switch event {
+        case .Next:
+            var onNextWins = false
+            
+            _lock.performLocked() {
+                onNextWins = !self._switched
+                if onNextWins {
+                    self._id++
```
kzaher: I think we should use `self._id = self._id &+ 1`, otherwise these bad boys could be party crashers :)

bontoJR:
fixed :relieved:

kzaher:
Thnx guys, looks great :)

Finally conjured a zen moment to simulate this in my head :)

:+1:

PR #290, opened 2015-11-26
Corrected the documentation for `flatMapFirst`

fpillet:


kzaher:
Hi @fpillet ,

you are correct original description is wrong :)

The correct description is here
https://github.com/Reactive-Extensions/RxJS/blob/6ba7ec08fbb7e2e4f9cc38be83be7eeec7577687/ts/rx.all.d.ts#L4402

We have version without incorporating index.

I believe that the first passage should be:
```
Transforms an observable sequence of observable sequences into an observable sequence which performs a exclusive waiting for the first to finish before subscribing to another observable.
```

Instead of 
```
- parameter selector: A transform function to apply to each element.
```
I believe it should be something like 
```
- parameter selector: A transform function to apply to element that was observed while no observable is executing in parallel.
```

Instead of 
```
- returns: An observable sequence whose elements are the result of invoking the one-to-many transform function on each element of the input sequence.
```
it should probably be something like
```
- returns: An observable sequence whose elements are the result of invoking the transform function on each element of source producing an Observable of Observable sequences and that at any point in time performs a exclusive waiting for the first to finish before subscribing to another observable.
```

Thnx for reporting this. If you could make PR with those changes to `develop`, that would be great. 

The same error in documentation is on driver unit, so we need to copy that documentation there also :(

kzaher:
I've talked with @fpillet , I can make the necessary changes. Thnx for reporting :)

PR #292, opened 2015-11-26
Update Examples to RxSwift 2.0

tailec:
If it gets approved, I will try to update the rest :)


kzaher:
Hi @tailec ,

I think your master branch is stale, this should already be fixed in master repo :) You can try to resolve those conflicts and check it out :)

tailec:
Ah sorry :)

PR #299, opened 2015-11-29
implementing NSLayoutConstraint.rx_constant

cellularmitosis:
Is this the right way to express this?  I'm still very new to RxSwift.

I've been writing up some programming challenges as I learn RxSwift, and I came across the need for `NSLayoutConstraint.rx_constant` in [Lesson 5.3](https://github.com/pepaslabs/LearningRxSwift/tree/master/lesson5.3_rx_constant).

This allows replacing something like this:

```swift
        topMarginConstraint.rx_observeWeakly(CGFloat.self, "constant")
            .subscribeNext { [weak self] (constant) -> Void in
                // do stuff...
            }
            .addDisposableTo(disposeBag)
```

with something like this:

```swift
        topMarginConstraint.rx_constant
            .subscribeNext { [weak self] (constant) -> Void in
                // do stuff...
            }
            .addDisposableTo(disposeBag)
```

...which seems a bit more idiomatic to me.

Feedback is welcome!

Thanks,
Jason

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -7,6 +7,10 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		2B385EDD1C0AA75A00461A5A /* NSLayoutConstraint+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B385ED81C0AA75A00461A5A /* NSLayoutConstraint+Rx.swift */; };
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @cellularmitosis ,

I think this should work, but I don't think this is conceptually a good idea :)

I'm not sure when would you want to observe constraints, since they should be "bound to" not "bound from" since user can't really change them directly through UI :)

I think it's ofc ok in situations when you are playing with the library and testing stuff out :)

kzaher:
We can probably close this one :)

PR #307, opened 2015-12-02
Add rx_animating to UIActivityIndicatorView

persidskiy:
Hi! I've just added support of UIActivityIndicatorView in RxCocoa. Take a look please :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 
 * Adds `ignoreElements` operator.
 * Adds `rx_attributedText` to `UILabel`.
+* Adds `rx_animating` to `UIActivityIndicatorView`.
 
 ---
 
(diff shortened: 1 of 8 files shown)


kzaher:
Hi @persidskiy ,

I think your proposal is inside of scope of this library. 

I love that you've added it to to example app, changelog and docs :)

I only have two remarks.

* integration tests are failing (./scripts/pre-release-tests.sh)
https://travis-ci.org/ReactiveX/RxSwift/builds/94486145
I think you need to add `#if os(iOS)` because of watch os. Take a look at `RxAlertViewDelegateProxy.swift`.

* could you also add unit tests analog to ones we have for `rx_attributedText`.

persidskiy:
@kzaher all is fine now. Hope, we can merge it :)

kzaher:
Cool, this looks great to me.

The only thing that I think it's missing is disposeBag here
```
boolSequence.subscribe(subject.rx_animating)
```

... but I can also add that after merge.

PR #312, opened 2015-12-03
Get rid of memory leaking in UIKit unit tests

persidskiy:


RxSwift/Subjects/Variable.swift
```diff
@@ -62,6 +62,7 @@ public class Variable<Element> : ObservableType {
     - parameter observer: Observer to subscribe to variable values.
     - returns: Disposable object that can be used to unsubscribe the observer from the variable.
     */
+    @warn_unused_result(message="http://git.io/rxs.ud")
```
persidskiy: I have no idea why Xcode does not catch `@warn` from protocol, so I've added it right into `Variable` implementation.

kzaher:
Cool, thnx :)

PR #278, opened 2015-11-21
make control property init method public

frogcjn:
make control property init method public

PR #313, opened 2015-12-04
Make ControlProperty initializer public

psophis:
When conforming to `ControlPropertyType` you are not able to create a `ControlProperty` to return in `asControlProperty()`.


yury:
This is duplicate of PR #278

PR #318, opened 2015-12-06
Add rx_hidden to UIView

edubarr:
Adds a bindable sink to the `hidden` property in `UIView`.


kzaher:
Hi @edubarr ,

I believe we've added those extensions like an hour ago :)

kzaher:
b08c1a5969962453dc2a67db1d743a33a2f0a7b2
¯\\\_(ツ)\_/¯

edubarr:
@kzaher Oops, sorry for not noticing. I guess that would explain the conflict then. No worries. :)

PR #337, opened 2015-12-14
Update new Playground API

nRewik:
Use `XCPlaygroundPage.currentPage.needsIndefiniteExecution = true` instead of  `XCPSetExecutionShouldContinueIndefinitely(true)` that was deprecated.


kzaher:
Awesome, thnx :)

PR #340, opened 2015-12-15
Reflect infinite sequences in the regex example

icanzilb:
The text gives an example of an infinite sequence here "Some sequences are finite, and some are infinite, like sequence of button taps" and just few lines below tries to explain sequences with a regex (which is great). However the regex was always having a terminating element (error or completion) and that contradicts the idea of an infinite sequence (which never terminates). 

What I did was to add a "?" to the terminating element in the regex example

### DIFF
--- a/Documentation/GettingStarted.md
+++ b/Documentation/GettingStarted.md
@@ -65,7 +65,7 @@ These diagrams are called marble diagrams.
 
 If we were to specify sequence grammar as regular expression it would look something like this
 
-**Next* (Error | Completed)**
+**Next* (Error | Completed)?**
 
 This describes the following:
 


kzaher:
Hi @icanzilb ,

That is a really interesting remark :)

I think that 
```
Next* (Error | Completed)
```
should be correct even in the case of "infinite sequence".

If we were to write it as 
```
Next* (Error | Completed)?
```

That would mean that a finite sequence of 

```
Next Next Next
```

can be generated without terminating with `Completed` or `Error`, which is not the case. 

I believe that the original regex does allow for generation of infinite `Next` sequences :)

icanzilb:
The original regex does match a final element and therefore cannot describe an infinite sequence. I find it confusing because I worked extensively with regexes but I guess for the majority of users the example is good enough :)

bontoJR:
@icanzilb good catch! You're right `(Error | Completed)` can't identify an infinite sequence b/c it should have either `Error` or `Completed` after the last `Next`.

kzaher:
The more I think about this, the more I think @icanzilb is right.

The situations that regex doesn't cover is `never()` and situation where there is a finite number of taps but nothing after.

I think that situation when there is an infinite number of taps should be covered with `Next* (Error | Completed)` :stuck_out_tongue: 
http://www.purplemath.com/modules/howcan1.htm

Thnx :)

icanzilb:
yay, thanks for merging

kzaher:
You are right, I'm just little tired :)

Thnx :+1:

PR #341, opened 2015-12-16
Set global deployment target for frameworks

petester42:
I noticed that the watchOS frameworks were pointed to watchOS 2.1 instead of 2.0 like it previously was. To make it easier to manage I updated the project to set the deployment target versions once in the project's build settings while also setting watchOS back to 2.0.

More details about the problem can be found [here](https://github.com/Moya/Moya/pull/344)


kzaher:
Less configuration, works for me :)

PR #346, opened 2015-12-16
[NSURLSession] Remove force unwrapped return from rx_JSON

nanoxd:
To keep the API similar, I've removed the force unwrapped return from rx_JSON like in 1f76dfcae4e97cc37602e43bb9668c2cca5037c3

### DIFF
--- a/Documentation/API.md
+++ b/Documentation/API.md
@@ -137,13 +137,13 @@ extension NSObject {
 ```swift
 extension NSURLSession {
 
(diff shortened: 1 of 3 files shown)


kzaher:
Huh, awesome :)

If you could just fix some of the tiny problems with compiling RxExample app that are causing our CI to fail, that would be :+1:

nanoxd:
Fixed :smile:

kzaher:
:+1:

PR #348, opened 2015-12-16
Remove use of unwrapped optionals from NSImageView

nanoxd:
* Removes use of unwrapped optionals from `NSImageView`
* Updates API documentation to correct method signatures


nanoxd:
Wrong branch :/

PR #350, opened 2015-12-18
Documentation Updates

nanoxd:
### Readme
* Updates NSURLSession's rx_response function to match current one
* Updates KVO section to use new api introduced in beta 3

### Getting Started
* Updates NSURLSession's rx_response function to match current one
* Use `String`'s `containString(string)` function in place of `rangeOfString(string)`

### Examples
* Replace `count(username) == 0` for `username.isEmpty`


kzaher:
Nice, thnx :)

PR #359, opened 2015-12-26
GroupBy operator

tottakai:


kzaher:
Hi @tottakai ,

This is a big PR :)

I'll need to find some time to digest all of this. I'm currently working on some parts of the system, but I'll try to review this later in the week.

This needs some focus :)

kzaher:
Hi @tottakai ,
 
sorry to keep you waiting. Lot of things happening :)
This PR has been haunting me for the past week :) Kept creeping in my head. This is not a simple PR to review.

I've been studying your PR, and I'm still kind of unsure should or should we not include this operator. I'm kind of inclined to include it since I see a lot of valuable work being done here, ~200 lines are operator code and the rest 1.2k are unit tests.

Having said that, I still have some concerns. Maybe discussing them here, giving this more visibility and including other people would help me decide more easily.

First of all, I'm curios in what scenario would this be useful. I've seen some examples with groupBy using keyboards and pressed keys, etc ..., but I'm not sure to how many people have those scenarios in their apps. I could find some really stretch edge cases maybe, but nothing really common. 

Maybe I just haven't ran into something like this, and there is some perfectly common scenario where to use this, so if somebody could present some case for this, it would make my life simpler :)

The reason I'm so reluctant to include new operators is because of swift compiler, compiler times, framework size and maintenance. So it's a matter of where should we add this operator. Should we add it here, or create some extension library ... etc 

Maybe I'm overlooking something obvious, so any usage examples to illustrate context better would be welcome.

Another question on my mind is that there are more versions of `groupBy` operator.
* groupBy
* groupByUntil

In case there is unbounded number of key values, `groupBy` is kind of "dictionary leak", the number of observables in memory just keeps pilling up. It's kind of similar situation to one with `replayAll` (although we did include it at the end because of people wanted to use that operator) ¯\\\_(ツ)\_/¯

That's kind of why I'm also interested in usage context in the platforms we are targeting. Should we have `groupBy`, or maybe it's better to have `groupByUntil` or both, or none?

I would really like to find some plausible answers to these questions :) For .NET team it was no brainer since they had the same LINQ operator, so I can perfectly see why they've needed this operator. I have no idea why they needed `groupByUntil` and how common is it.

As far as the code, I think it's mostly :+1:. 

There are some things that we would probably want to change. 

For example using `GroupedObservable` as struct that composes key and observable sequence.
Seems to me that inheritance seems odd and not idiomatic to Swift in this case, and maybe struct would be more appropriate. That struct could implement `ObservableType`, so the usage should be the same.

I don't think we need `GroupBySink` synchronized and implement `LockOwnerType`. Observing is done on single observable sequence and there shouldn't be any parallel events there.

I was little confused with unit tests. I would definitely want to have all .NET unit tests because they are really high quality tests and we would want to be compliant. It looks like you have included them, but I'm kind of confused with beginning of them.

I like what you've done with `testGroupBy_TwoGroup` test, it's a really nice illustration how that operator works.

Instead of `testGroupBy_OneGroup` and `testGroupBy_Error` it would probably be nice to use rest of original tests.

I wasn't sure if `testGroupBy_Error` you've coded as an equivalent of `GroupBy_Capacity_Outer_Error`, but it looks like it's not, because there is also `testGroupBy_OuterError`, so we probably don't need it.

Maybe you've meant `GroupBy_Capacity_Outer_Complete` to be equivalent to `testGroupBy_TwoGroup`, but the problem is that `GroupBy_Capacity_Outer_Complete` doesn't have that check that after completed no events are grouped, and I think it would be nice to use original implementation of `GroupBy_Capacity_Outer_Complete` instead of `testGroupBy_OneGroup` and `testGroupBy_Error` :)

`testGroupBy_OuterError` is also kind of missing that check that after error no element is grouped :)
```
OnNext(530, "    fOo    "),
OnError<string>(570, ex),
OnNext(580, "error"),
OnCompleted<string>(600),
```

... and other tests too :)

It would be valuable to create some kind of string extension in unit tests for this `x.lowercaseString.stringByTrimmingCharactersInSet(NSCharacterSet.whitespaceCharacterSet())`, so it's more compact, maybe just `trimWhitespace`?

I haven't seen `testGroupBy_Never` and `testGroupBy_Empty` in their tests, but I have nothing against leaving them in this PR.

I believe `testGroupBy_WithRealScheduler` is redundant. Those tests with real scheduler are only important if operator is using a scheduler as a parameter, or if we are testing some specific threading concerns, which we don't have here.

This is not a simple PR :)

tottakai:
Hi @kzaher, 

thanks for the thorough review, the comments for the code and tests seem fine. The GroupedObservable as struct seems reasonable, i initially didn't want to stray too far from the original concept but i see that different Rx implementations have adapted to their programming language styles. My use case is that i have a stream of objects that i want to group based on some enum/string on the objects, then calculate something based on the groups objects values by flatmapping the group and then either using scan or reduce. F.ex. calculate a value for the group.

kzaher:
Hi @tottakai ,

yeah, totally understand why you've made `GroupedObservable` abstract class. We need to add `_refCount.retain()` for each subscription.

 Unfortunately, we can't express abstract class in Swift explicitly. If we want to have `GroupedObservable` as an abstract class in public interface, we would we could theoretically add init subscription method, but that seems ugly.

If we would have `GroupedObservableType` as a protocol. Then we would need to add additional thunk. Something like:
```
struct AnyGroupedObservable<Key, Element> {}
``` 
... which is also ugly :(

We could have `GroupedObservable` as a struct, something like:
```
struct GroupedObservable<Key, Element> : ObservableConvertibleType {
    public let key: Key
    private let _source: Observable<Element>

   public func asObservable() {
       return _source
   }
}
```

But then we would need to use something like `using` operator inside `GroupBySink` that wraps `PublishSubject<Element>` and adds ref count logic, but that could add some performance penalty :(

Probably last solution looks to me like it has the nicest and simplest public API, but maybe there is some nicer solution.

tottakai:
Updated the tests according to the suggestions you gave earlier some time ago. I tried to make the `GroupedObservable` a struct with the `using` operator, the correct solutions escapes me :) I haven't used the `using` operator before, but i couldn't figure out how to tie the group to the writer and subscription. Maybe i just took a left where is should have taken a right turn :)

Also, it somewhat bothers me that if the `GroupedObservable` is a struct and the key could be a class, then a struct would contain a class, a value type containing a reference type.

Sorry for the longish delay, apparently time flies after the holidays.

kzaher:
Hi @tottakai ,

I'm not sure is there is there any obvious benefit of:

```
struct GroupedObservable<Key, Element> : ObservableConvertibleType {
    public let key: Key
    private let _source: Observable<Element>

   public func asObservable() {
       return _source
   }
}
```

vs

```
class GroupedObservable<Key, Element> : ObservableConvertibleType {
    public let key: Key
    private let _source: Observable<Element>

   public func asObservable() {
       return _source
   }
}
```

Everything here is immutable. You can't change key or contained observable sequence.

It seems to me that maybe `struct` communicates intent better, since we only want a value that doesn't have any identity, although I understand that value types should implement equals, etc (not sure implementing equal in this case makes sense or is beneficial to anyone).

As far as for left vs right turn :), I guess when I find some time to think about this, I can try to see have I missed something, but currently this is not high on the priority list IMHO :), so it might take a while.

ElwinHsiao:
Thanks for receive the pull request, i have a scenario that need this operator, that is: 
My app has a chat room that will receive push message from my server, the message is bound to a user, the product logic need to restrict the user's speak frequency, every user couldn't speak more than once in a minute, he/she must wait for 1 minute that he/she can speak again.
So i have to Observable emits the push message, then group the message by user, then use my custom operator 'minInterval' on every GroupedObservable to filter the overmuch message, after that i'll flatMap all the GroupedObservable and show the messages in the chat room.
By the way, is there any operator that behave likes my custom operator 'minInterval'. its function is:
apply a min Interval between the emissions, when emit a item, the next item must wait for as least the specified time interval. 
If there isn't a such operator, will you think about add one, because i found there are several scenarios need this, such as restrict the frequency of button taps.  It is different with the operators 'throttle/debounce' and 'sample', It won't emit the recently item of the source Observable, when a source emited a item, it just check the interval from the last successful emitted item and determine whether emit this one.

steeve:
hey folks,
any update on this PR ? because it looks great :)

seivan:
I've been using `reduce` to group things. The way I see it, it's a perfect fit unless the `groupBy` does something more than groupBy e.g flattening observables and what not.

@kzaher Does compile time matter if RxSwift is precompiled for the end user?

Issue #333, opened 2015-12-11
UITextView.rx_text doesn't catch iOS autocorrect text changes


timojaask:
`UITextView.rx_text` observes text changes via `textViewDidChange`. Unfortunately, when text changes due to iOS autocorrect, the `textViewDidChange` is not called, therefore `rx_text` never receives that new text.

You can test it by creating a text view, make sure autocorrection is enabled, observe `rx_text`, type a misspelled word and move focus away from the text view. The text will be automatically corrected, but the last change will never be observed via `rx_text`, because `textViewDidChange` doesn't fire when autocorrect happens.

kzaher:
Hi @timojaask ,

I've confirmed the problem. This is the only way I know how to solve this, but I don't like it.

https://github.com/ReactiveX/RxSwift/commit/e14ca2e94d4e995c552c2d7c417f52ad9c3b75d7

Any ideas how to solve this more elegantly?

sergdort:
Hi, guys
@kzaher  how about ```rx_observeWeakly``` of ```text``` keypath

kzaher:
Hi @sergdort ,

that's an interesting idea :) I've just tried it, and unfortunately it doesn't seem to work either :/

Also `NSNotificationCenter.defaultCenter().rx_notification(UITextViewTextDidChangeNotification, object:self)` doesn't fire in that case.

Is there some obvious API I'm missing?

sergdort:
Just looked at how ReactiveCocoa did this, looks like they did the same and probable they also have bug =)

kzaher:
Haha, yeah, I've also checked other reactive/binding frameworks :)

This hack is the best thing I could conjure so far. ¯\\\_(ツ)\_/¯ 
This does look like a bug in `UITextView`, but this is a pretty ugly one and would like to find some workaround.

sergdort:
How about observe ``textViewDidEndEditing`` instead of ``func textView(textView: UITextView, shouldChangeTextInRange range: NSRange, replacementText text: String) -> Bool``

Looks like this bug appears only when you moving focus away from the text view

kzaher:
Yeah, that also appears to be working ...

I'm wondering is there any other edge case like this one what `textViewDidEndEditing:` can't catch ...

[...]

PR #360, opened 2015-12-26
Fixes regarding UITextField rx_text autocorrection bug #333

sergdort:


kzaher:
Thnx @sergdort ,

Yeah, I think this is improvement :)

PR #371, opened 2015-12-30
Fixes for UITextView rx_text bug with autocorrect enabled #333

sergdort:


kzaher:
Hi @sergdort ,

I think we can probably pull in your solution after RC, before 2.0.0 release, assuming it works :) (since current solution works, and this solution doesn't remove anything from public API)

I'll still need to try it for myself.

I have noted a couple of issues we would want to tackle before merging this in develop branch, I'll note my comments in the code (and build server error in unit tests is one of them :)

RxCocoa/iOS/NSTextStorage+Rx.swift
```diff
@@ -0,0 +1,34 @@
...
+    import RxSwift
+#endif
+
+extension NSTextStorage {
+    
+    public var rx_delegate:DelegateProxy {
+        return proxyForObject(RxTextStorageDelegateProxy.self, self)
+    }
+    
+    public var rx_string:Observable<String> {
```
kzaher: The convention is for delegate proxies to expose a method that is named the same like corresponding delegate method.

You can find samples in other delegate proxies.

RxCocoa/iOS/NSTextStorage+Rx.swift
```diff
@@ -0,0 +1,34 @@
...
+        return rx_delegate
+            .observe("textStorage:didProcessEditing:range:changeInLength:")
+            .map({ a  in
+                let textStorage:NSTextStorage = castOrFatalError(a[0])
+                return textStorage.string
+            })
+            .distinctUntilChanged() // dont know why, but system call delegate twice on auto correction
```
kzaher: Last sentence was wrong :) We don't expose non required methods in delegate proxies :)

The `rx_text` method inside `UITextView+Rx.swift` should look something like 

```swift
let source: Observable<String> = Observable.deferred { [weak self] () -> Observable<String> in
            let text = self?.text ?? ""
            // basic event
            let textChangedEvent = self?.textStorage..observe("textStorage:didProcessEditing:range:changeInLength:") ?? Observable.empty()

 return textChangedEvent
                .map { a in
                    return self?.text ?? ""
                }
                .startWith(text)
                .distinctUntilChanged()
            }
}
```

RxCocoa/iOS/NSTextStorage+Rx.swift
sergdort: Do you mean `rx_string` ?

kzaher: This is snippet from CLLocationManager
```swift
/**
    Reactive wrapper for `delegate` message.
    */
    public var rx_rangingBeaconsDidFailForRegionWithError: Observable<(region: CLBeaconRegion, error: NSError)> {
        return rx_delegate.observe("locationManager:rangingBeaconsDidFailForRegion:withError:")
            .map { a in
                let region = try castOrThrow(CLBeaconRegion.self, a[1])
                let error = try castOrThrow(NSError.self, a[2])
                return (region: region, error: error)
            }
    }
```

so in this case this should probably be something like
```swift
public var rx_didProcessEditingRangeChangeInLength: Observable<(editedMask:NSTextStorageEditActions, editedRange:NSRange, delta:)> {
       ....
}
```

RxCocoa/iOS/NSTextStorage+Rx.swift
sergdort: I see, thanks. Will do

RxCocoa/iOS/NSTextStorage+Rx.swift
sergdort: Thanks :)

kzaher:
Great we've found a solution that isn't hacky.

PR #372, opened 2015-12-30
flatMap example needed print statement added in order to see what was coming from se…

finneycanhelp:
…quenceInt. Without adding the print statement and "(x:Int) -> Observable<String> in", it can be quite hard to see what is going on if not already familiar with the concept, IMHO.


kzaher:
Make sense.

PR #373, opened 2015-12-30
Adding Quality of Service (QOS) support to the Schedulers

hsoi:
When using `dispatch_get_global_queue()`, Apple recommends using quality of service class values to identify the well-known global concurrent queues. Here we add this support, mirroring the structure and implementation of the existing "priority" support.

The priority support remains, but is deprecated. 

I have need for "background" QOS/priority, which the existing "priority" support does not provide. So I figured if a change was to be made, might as well go all the way and adopt QOS as that's Apple's recommended approach for using `dispatch_get_global_queue()`. Especially in light of the fact RxSwift 2.0.0 is about to happen with some API changes... might as well see if this could be included in there.

Thank you. :smiley:

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -7,6 +7,10 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		79E9DE891C3417FD009970AF /* DispatchQueueSchedulerQOS.swift in Sources */ = {isa = PBXBuildFile; fileRef = 79E9DE881C3417FD009970AF /* DispatchQueueSchedulerQOS.swift */; };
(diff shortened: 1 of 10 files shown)


kzaher:
Hi @hsoi ,

what a sense for timing :) I've been converging to this version in develop branch for last few days, and I was just waiting for a last batch of tests to finish before hitting the launch button :)

I see you've figured out that `Package for Swift package manager isn't updated, please run ./scripts/package-spm.swift and commit the changes` :)

You are right, those values were deprecated, and it would make sense to make them.

I'll pull in your changes and make additional mocks in `Linux.Platform.swift` to fix linux build :)

How I wish this PR was made a day ago :)

hsoi:
Ha ha! Sorry I didn't get it in sooner ;-)

Thank you for taking care of the mocks, and thank you for RxSwift. I'm still getting up to speed with it, but loving it so far.

Cheers!

PR #380, opened 2016-01-03
Add file download to NSURLSession extensions

mwoollard:
Adds rx_download method following the pattern of rx_request and rx_data to download to file contents of NSURL or NSURLRequest.

### DIFF
--- a/RxCocoa/Common/Observables/NSURLSession+Rx.swift
+++ b/RxCocoa/Common/Observables/NSURLSession+Rx.swift
@@ -33,6 +33,10 @@ public enum RxCocoaURLError
     Deserialization error.
     */
     case DeserializationError(error: ErrorType)
+    /**
+     File error.
+     */
+    case FileError(error: ErrorType)
 }
 
 public extension RxCocoaURLError {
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @mwoollard ,

sorry for late response, lot of things happening.

I think that ideally we would want to keep our APIs small, well defined and supporting the idiomatic workflow.

I think this PR kind of conflicts with that `small` part :) I think that we would like to include code that is as small as possible and is useful to >90% of users. 

We didn't even include any internal way to build `NSURLRequest` because the API surface of `NSURLRequest` to cover was too big.

I'm also little concerned about useful part in this case. If I were to use some downloading API, I would also be interested in progress indication.

The rationale is that those specific functionalities that would complicate this library too much should go into their own separate external libraries, so if somebody needs them, he could find them there.

There are multiple external networking libraries that might be interesting to you. The ones that I had most contact and that are compatible with RxSwift library are:

https://github.com/RxSwiftCommunity/RxAlamofire - I believe this already contains all of the download upload APIs wrapped with progress indicators, etc ... (this is more of a low level networking library)
https://github.com/Moya/Moya - Moya is a more high level library to abstract away your networking layer

Hope you can understand our rationale, and it's ok with you if we close this PR.

¯\\\_(ツ)\_/¯

mwoollard:
The use case is for background download when the app is terminated whilst a transfer is ongoing and later restarted and how things get re-connected to those background transfers. It gets progress re-connected too as in the demo I included :-)

However I understand you may want to keep this complexity outside of the framework - you can still do the same by coding to use existing framework and re-connect things that way but not as straightforward. 

Thanks
Mark

kzaher:
Hey, yeah :)

Trying to pull as much code as possible from RxCocoa otherwise it will be so big it won't be shippable :)

Issue #361, opened 2015-12-26
How about to add rx_textColor for UITextField and UILabel


sergdort:
Now have a project where I need to change color of the text based on whether there any changes
Just wondering if community need this ?
I've did this like this :

```swift
extension UILabel {

   public var rx_textColor: AnyObserver<UIColor> {
      return AnyObserver { [weak self] event in
         MainScheduler.ensureExecutingOnScheduler()
         
         switch event {
         case .Next(let value):
            self?.textColor = value
         default:
            break
         }
      }
   }
}
```

kzaher:
Hi @sergdort ,

I think that we would ideally like to keep our API surface as small as possible, otherwise we'll have serious hoarding issues :)

There are ofc multiple reasons why we would want to keep API surface as small, and I think we all understand them :)

I would like to actually like to motivate people to do extensions in this way
```
extension UINavigationController {
    var rx_serviceState: AnyObserver<ServiceState?> { // <- this is some custom enum value 
        return AnyObserver { [weak self] event in
            switch event {
            case .Next(let maybeServiceState):
                // if nil is being bound, then don't change color, it's not perfect, but :)
                if let serviceState = maybeServiceState {
                    let isOffline = serviceState ?? .Online == .Offline

                    self?.navigationBar.backgroundColor = isOffline
                        ? Colors.OfflineColor
                        : Colors.OnlineColor
                }
            case .Error(let error):
                bindingErrorToInterface(error)
            case .Completed:
                break
            }
        }
    }
}
```

I'm thinking more in the direction how can we support people using RxCocoa with this pattern.

I'm thinking of adding `BindingObserver` maybe (still thinking about it), that could be nice and expressive, and help people write this kind of code:

```
extension UINavigationController {
    var rx_serviceState: AnyObserver<ServiceState?> { // <- this is some custom enum value 
        return BindingObserver { maybeServiceState, self in
                // if nil is being bound, then don't change color, it's not perfect, but :)
                if let serviceState = maybeServiceState {
                    let isOffline = serviceState ?? .Online == .Offline

                    self.navigationBar.backgroundColor = isOffline
                        ? Colors.OfflineColor
                        : Colors.OnlineColor
                }
        }
    }
}
```

In that way we've reduced some noise in VM/VC and moved code to view where it belongs.  `Observable<ServiceState?>` is also more natural to test.

This is kind of direction where I think we should be going :)

[...]

PR #383, opened 2016-01-04
Adding multiple rx wrappers to RxCocoa

stoegerit:
- Adding rx_title to NSButton
- Adding rx_animating to NSProgressIndicator
- Adding rx_textColor to NSTextfield

### DIFF
--- a/RxCocoa/OSX/NSButton+Rx.swift
+++ b/RxCocoa/OSX/NSButton+Rx.swift
@@ -31,4 +31,23 @@ extension NSButton {
             self?.state = state
         })
     }
+	
+    
+    /**
+     Reactive wrapper for `title` property.
+     */
+    public var rx_title: AnyObserver<String> {
+        return AnyObserver { [weak self] event in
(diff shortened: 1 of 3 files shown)


stoegerit:
I have no clue why travis ran into a timeout.

kzaher:
Hi @stoegerit ,

I've restarted Travis manually, for some reason sometimes it just doesn't start simulator. I believe that could be because our tests take such a long time. I've also noticed that if we have more then one of them running for PRs, usually one will fail, and one will pass.

I think I would be fine with pulling extensions for `rx_title` and `rx_animating`, but `public var rx_textColor: AnyObserver<NSColor!>` seems like a duplicate of https://github.com/ReactiveX/RxSwift/issues/361. 

The explanation is there inside the issue.

If you could add unit tests for those two extensions (rx_text and rx_animating), I would be happy to merge this.

kzaher:
This looks stale, I guess we'll close it

¯\\\_(ツ)\_/¯

PR #384, opened 2016-01-04
Fix http body logging of NSURLSession

hadashiA:
It is a log of when I POST body as json using NSURLSession+Rx.

```
curl -i -v -X POST -d "{"foo":true}"-H"Content-Type: application/json" "https://example.com/"
```

I found this small problems.
* There is no space after the `-d value`
* Value of `-d` is does not seem to be escaped


kzaher:
Makes sense, thnx.

PR #387, opened 2016-01-05
Rx wrappers for UIImagePickerController

sergdort:


kzaher:
Hi @sergdort ,

It looks nice. I believe that we would also need to add delegate unit tests. There is an example of how we tested CLLocationManager delegate wrappers.

That's an interesting usage of `AnyObserver` in example app :) I would probably play around little with that example to see how can we model this.

If you could just add those unit tests, I would be happy to merge this.

sergdort:
Hi @kzaher :) Thanks, will do

sergdort:
Regarding of usage `AnyObserver `. Last week I wrote a [post](https://medium.com/@SergDort/reactive-generic-segue-with-rxswift-e20a5219aeea#.mcmkydmuk) where I used something like that ;)

kbala:
Hi @sergdort 

Seems its more coding than imperative. Good if you implement like

- rx_takePhoto(allowsEditing:Bool = false)
- rx_choosePhoto(sourceType:UIImagePickerControllerSourceType, allowsEditing:Bool = false)

Also we need rx_didFinishPickingImage

:)

sergdort:
Hi, @kbala, As I understand from @kzaher perspective the idea of RxCocoa to provide only core functionality, because it keeps library small, and user can always create his own extensions. 

Another weird thing about `UIImagePickerController` is that it's also provide functionality to capture video =). 
And because of this we would need to add all methods which related to [Editing Information Keys](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIImagePickerControllerDelegate_Protocol/index.html#//apple_ref/doc/constant_group/Editing_Information_Keys)

But to implement them you just need to use `map` operator, which is easy, and every user of library can create them in `extension`

RxExample/RxExample/Examples/ImagePicker/ImagePickerController.swift
```diff
@@ -0,0 +1,97 @@
...
+   let soruceType:UIImagePickerControllerSourceType
+   let allowsEditing:Bool
+}
+
+class ImagePickerController: ViewController {
+
+   @IBOutlet var imageView: UIImageView!
+   @IBOutlet var cameraButton: UIButton!
+   @IBOutlet var galleryButton: UIButton!
+   @IBOutlet var cropButton: UIButton!
```
kzaher: @sergdort I'm not doing some work on this code, and it seems that the code you've written there is `tab=3 spaces` :)

This is kind of funny situation, just out of curiosity, do you have tab set to 3 spaces? Because it seems that spaces are divisible by 3 :)

sergdort: Yes, my bad, in my team we decided to use 3 spaces, and for this changes I forgot to change that in settings, Sorry :(
Will make changes

kzaher: I've already changed that code, no need :)

Just out of curiosity, why? :)

kbala:
Hi @sergdort Ok I can understand. Thanks :)

PR #391, opened 2016-01-06
DRY DispatchQueueSchedulerQOS

yury:


yury:
Not sure about `QOSClass` property name

PR #394, opened 2016-01-07
[ADD] rx_itemDeselected and rx_modelDeselected events

yannickl:
Hi,

For one of my project I need to check the selection and the deselection of UITableView cells. So I added the missing events.

Best


kzaher:
Hey,

I think it would make sense to have deselected versions also because of symmetry. There is a small error in description, but I'll probably make that mechanism more generic over weekend so I'm fine merging this.

yannickl:
Great, thank you. :+1: And sorry for the error in the description.

PR #402, opened 2016-01-12
Fixes for bug with UIImagePickerController example.

sergdort:
To dismiss image picker on cancel button tap


sergdort:
Hi, @kzaher 
Do you think this travis check fail some how [related to](https://github.com/ReactiveX/RxSwift/pull/383) ?

kzaher:
I've restarted the build manually, let's see what happens.

RxExample/RxExample/Examples/ImagePicker/UIImagePickerController+RxCreate.swift
```diff
@@ -31,6 +31,12 @@ extension UIImagePickerController {
     static func rx_createWithParent(parent: UIViewController?, animated: Bool = true, configureImagePicker: (UIImagePickerController) throws -> () = { x in }) -> Observable<UIImagePickerController> {
         return Observable.create { [weak parent] observer in
             let imagePicker = UIImagePickerController()
+            let dismissDisposable = imagePicker
+                .rx_didCancel
+                .subscribeNext({ [unowned imagePicker] in
```
kzaher: We would probably suggest using `weak` instead of `unowned` and it would probably make sense to use `animated: animated`, but I can pull this in and make the change myself.

PR #403, opened 2016-01-12
Adding concatMap operator

mikelikespie:
Description of operator is here: http://reactivex.io/documentation/operators/flatmap.html

Not sure if there's a way to avoid using CurrentThreadScheduler to start iteration w/o getting a stack overflow if there's a really long outer Observable.


kzaher:
Hi @mikelikespie ,

This is one of the most interesting PRs we've had in a while :)
I've forgot about adding `concatMap`, although we've had all necessary ingredients for a it a while ago :)

## Answer to question

As far as getting stack overflow, I think that in case of generating some large number of elements, it's best practice to use some scheduler to generate them, and in that way stack overflow can't happen.

I believe RxJava has slightly different internal architecture then Rx.NET and RxJS. Subscriber needs to request certain number of elements from producer before they can be produced and is more aligned with consumer/producer scenario.

Unfortunately, I'm not sure could we create some performant mechanism in Swift that would be equivalent. Atomic operations in Swift are very limited and replacing reference atomically is tricky because of ARC, but I'm assuming there will be more documented options in future.

## CR

I've seen you've ported unit tests equivalent to `flatMap`, and that's awesome.

It is my understanding that you've tried to make `concatMap` performant by queueing user space events using fast `SpinLock` and then elegantly using `AtomicCompareAndSwap` subscribe to inner observable sequences and serialize them.

I'm assuming this is more of a general proposal how we could also do other concat operator. 

There are unfortunately some problems with that approach :) Last week we've released 2.1.0. One of the changes there was replacing `OSSpinLock` with `NSRecursiveLock` for `Darwin` platforms because of https://lists.swift.org/pipermail/swift-dev/Week-of-Mon-20151214/000321.html

That means we can't use SpinLocks on iOS or OSX. The good news is that NSRecursiveLock is ~3x slower if I remember correctly. But if we are already using `NSRecursiveLock`, then these optimizations won't gain us much, and we'll be probably touching the kernel anyway.

Also one of the good bad news is that even if we are touching the kernel with those `NSRecursiveLock`, Swift is so slow that it's effect on performance is negligible :/

I've also ran performance tests target inside the project with this code

```
repeat {
compareTwoImplementations(benchmarkTime: true, benchmarkMemory: false, first: {
    let publishSubject = PublishSubject<Int>()

    publishSubject
        .concatMap { x in Observable.just(x) }
        .concatMap { x in Observable.just(x) }
        .concatMap { x in Observable.just(x) }
        .concatMap { x in Observable.just(x) }
        .subscribeNext { _ in

        }


    for i in 0..<100 {
        publishSubject.on(.Next(i))
    }

}, second: {
    let publishSubject = PublishSubject<Int>()

    publishSubject
        .map { x in Observable.just(x) }.concat()
        .map { x in Observable.just(x) }.concat()
        .map { x in Observable.just(x) }.concat()
        .map { x in Observable.just(x) }.concat()
        .subscribeNext { _ in

        }

    for i in 0..<100 {
        publishSubject.on(.Next(i))
    }
})
} while true
```

... and got these results

```
Fragmenting memory ...
Benchmarking ...
#1 implementation        0 bytes    0 allocations 12343 useconds
#2 implementation        0 bytes    0 allocations  2387 useconds
#1 implementation        0 bytes    0 allocations 12706 useconds
#2 implementation        0 bytes    0 allocations  2442 useconds
...
```

I think that for now implementing `concatMap` as `map` + `concat` is more efficient, and we can reuse existing code. Even if that recursive scheduler was't so expensive and we replace it, we still unfortunately wouldn't gain much without proper lockless queue implementation and other low level atomic primitives and even then Swift runtime and cache misses would drag us down.

I would like to include those unit tests and `concatMap` extension defined as `map + concat` without `ConcatMap.swift` if that's ok with you.

I have a couple of comments for unit tests that I've noted in the code.

Tests/RxSwiftTests/Tests/Observable+StandardSequenceOperatorsTest.swift
```diff
@@ -3181,6 +3181,598 @@ extension ObservableStandardSequenceOperatorsTest {
...
+            completed(2155),
+            ]
+        )
+        
+        XCTAssertEqual(xs.subscriptions, [
+            Subscription(200, 900)
+            ])
+        
+        
+        let innerSubscriptions = xs.recordedEvents[2..<xs.recordedEvents.endIndex].flatMap { $0.value.element?.subscriptions ?? [] }
```
kzaher: I'm not sure why do we start here from 2, it looks like we could start from 0, but maybe comment below is more important ....

Comparing a list of flat mapped subscription results could theoretically be different then expected, and still produce a valid result. I know this is a exponentially unlikely long shot, but my OCD is alarmed :), and would make unit test code more idiomatic.

Maybe it would be better to assert each subscription manually. I understand that it's little more code, but it's more exact IMHO ¯\_(ツ)_/¯

Tests/RxSwiftTests/Tests/Observable+StandardSequenceOperatorsTest.swift
```diff
@@ -3181,6 +3181,598 @@ extension ObservableStandardSequenceOperatorsTest {
...
+            next(950, 203),
+            next(975, 301),
+            next(1015, 302),
+            next(1035, 303),
+            next(1225, 304),
+            next(1275, 305),
+            next(1965, 306),
+            next(2135, 401),
+            next(2145, 402),
+            error(2155, testError),
```
kzaher: I think this test differs from `map + concat` implementation together with `testConcatMap_SelectorThrows` because of internal queueing of events. If implementation is replaced, this would require a small change in those two tests.

kzaher:
Ok, closing this one. It was an interesting PR and we'll maybe explore similar attempts at optimizing the library for 3.0 version.

PR #407, opened 2016-01-17
Remove warning for unused variable

thanegill:
Warning:
immutable value 'i' was never used; consider replacing with '_' or removing it


kzaher:
Oh, cool.

PR #409, opened 2016-01-18
Make UIBarButtonItem.rx_tap to shareable.

tarunon:
UIBarButtonItem has only one target, so UIBarButtonItem.rx_tap cannot share now.
```swift
let barButton = UIBarButtonItem()
barButton.rx_tap
  .subscribeNext { [weak self] _ in
    // Never not be called.
  }
  .addDisposable(disposeBag)
barButton.rx_tap
  .subscribeNext { [weak self] _ in
    ...
  }
  .addDisposable(disposeBag)
```

I move ControlEvent initializer to BarButtonItemTarget, and reuse it.
Then, UIBarButtonItem.rx_tap can share.

Please check this pr :3

### DIFF
--- a/RxCocoa/iOS/UIBarButtonItem+Rx.swift
+++ b/RxCocoa/iOS/UIBarButtonItem+Rx.swift
@@ -38,22 +38,11 @@ extension UIBarButtonItem {
     Reactive wrapper for target action pattern on `self`.
     */
     public var rx_tap: ControlEvent<Void> {
-        let source: Observable<Void> = Observable.create { [weak self] observer in
-
-            guard let control = self else {
-                observer.on(.Completed)
-                return NopDisposable.instance
-            }
-
-            let target = BarButtonItemTarget(barButtonItem: control) {
-                observer.on(.Next())
-            }
-            return target
-        }.takeUntil(rx_deallocated)
-        
-        return ControlEvent(events: source)
+        if let target = self.target as? BarButtonItemTarget {
+            return target.event
+        }
+        return BarButtonItemTarget(barButtonItem: self).event
     }
-    
 }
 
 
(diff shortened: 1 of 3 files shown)


RxCocoa/iOS/UIBarButtonItem+Rx.swift
```diff
@@ -38,22 +38,11 @@ extension UIBarButtonItem {
...
-                observer.on(.Next())
-            }
-            return target
-        }.takeUntil(rx_deallocated)
-        
-        return ControlEvent(events: source)
+        if let target = self.target as? BarButtonItemTarget {
+            return target.event
+        }
+        return BarButtonItemTarget(barButtonItem: self).event
```
kzaher: Hi @tarunon ,

I think you are correct, we do have a sharing issue here. Thnx for reporting this.

Now that I'm going over our target action code, we repeat this sharing bug in multiple places :(

UIKit extensions
* rx_value
* rx_controlEvent 

Cocoa extensions
* rx_value
* rx_controlEvent

... although it seems to me that Cocoa extensions at least use associated object to store observable sequence.

I believe that solution should be simple. 
* We need to make `rx_lazyInstanceObservable` more generic, and move it from (`NSControl` to private `NSObject` extension).
* Then use that `rx_lazyInstanceObservable` to fetch cached source or create it
* In this case it should be something like
```
public var rx_tap: ControlEvent<Void> {
        let source = rx_lazyInstanceObservable(&tapKey) { () -> Observable<Void> in
               Observable.create { [weak self] observer in

                guard let control = self else {
                    observer.on(.Completed)
                    return NopDisposable.instance
               }

                let target = BarButtonItemTarget(barButtonItem: control) {
                   observer.on(.Next())
               }
               return target
           }
           .takeUntil(rx_deallocated)
           .share()
        }
        
        return ControlEvent(events: source)
    }
```

The reason why I'm suggesting this solution is because if you create `BarButtonItemTarget` like you are suggesting, control.target would be set during calling `rx_tap` and not when observer subscribes, and in this way, we are keeping logic consistent across project.

I can ofc make all of those changes if you like or if you want to do them, that's also great.

Thnx for reporting this.

tarunon: Thanks for your review, and great solution!
Of course, I make all of those changes, and commit this pr :3
Please wait...

PR #410, opened 2016-01-18
Add example to HotAndColdObservables documentation

AndrewSB:
Found these (or similar) examples helpful while trying to understand the concept. Having them included would be a plus

### DIFF
--- a/Documentation/HotAndColdObservables.md
+++ b/Documentation/HotAndColdObservables.md
@@ -37,7 +37,7 @@ Observable.create { observer in
       observer.onNext(n)
     }
 
-  return Disposable {
+  return AnonymousDisposable {
     subscription.dispose()
   }
 }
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @AndrewSB ,

these are actually both examples of cold observable sequences :)

The reason why they are cold is because while there are no subscribers, no element is being produced (no heat is being generated), and once observer subscribes, element production begins.

```
NSURLSession
 +  .sharedSession()
 +  .rx_data(NSURLRequest(URL: NSURL(string: "http://www.timeapi.org/utc/now")!)
 +  .map { String(data: $0, encoding: NSUTF8StringEncoding) }
 +  .asObservable()
```
This starts production of data content from `http://www.timeapi.org/utc/now`, and in second example
```
Observable.interval(0.1, scheduler: MainScheduler.instance)
```

timer also starts when observer subscribes.

I think what probably confused you was the `sharedSession` naming.

Typical example of hot observable is something like
```
textField.rx_text
```
... because text box works and produces text (makes your CPU warm) no matter is there any observer subscribed or not, and first subscriber might not get initial text field text value (so it might lose some sequence elements).

AndrewSB:
😂 I thought I finally understood this. Boy was I wrong....

I'll update this PR in the next couple days and replace the hot observable example. Was hoping by figuring out an example or two I'd better understand Rx :)

kzaher:
This one is also stale.

PR #412, opened 2016-01-18
fix example `map`

skywinder:
Before output:

```
--- map example ---
Next(4799450059485597671)
Next(4799450059485597672)
Next(4799450059485597677)
Completed
```

after:

```
--- map example ---
Next(A)
Next(B)
Next(C)
Completed
```


carlosypunto:
Thanks @skywinder. The idea of the example is to show how `map` transform a type in other. In your correction isn't necessary the use of `map` because you give back exactly what you get. Could you suggest other diferent in/out types, if you believe it is not clever.

skywinder:
@carlosypunto agree with you. I updated PR. How about this example?

carlosypunto:
@skywinder, please make the PR in develop branch and if is posible with only one commit

skywinder:
Close in favor of #420

PR #414, opened 2016-01-18
Added reactive wrapper for the network activity indicator in UIApplication

MadsBogeskov:


kzaher:
Hi @MadsBogeskov ,

I think this is kind of at the border regarding what would we want to pull, but I think this is valuable enough to pull into this repo.

If you could just add two tests similar to the ones for `rx_enabled`, I would be happy to merge this.

MadsBogeskov:
Hi @kzaher,

Ok, i was actually in doubt of what you would want in your repo. I will add the tests later today :-)

Thanks for the look.

mwoollard:
I don't believe you need to access UIApplication.networkStatusIndicatorVisible from main thread (docs are unclear but I've not had issues doing so in the past) so maybe remove the test?

kzaher:
I think we'll unify all of that binding code, so all binding code will all have that form. 

It's maybe not technically necessary to call this from main thread, but unless we see documentation that it's thread safe and we see some obvious benefit to call this from background thread I think we can leave that check. I don't see any harm in it, and it would make the behavior more predictable :)

General rule of thumb is, if it's UI, access if only from main thread.

@mwoollard  Although, that is an interesting remark you've made. If you find somewhere documented that property is thread safe please let us know.

mwoollard:
I've not got a reference to any Apple produced docs on this, however interestingly Xamarin do bring this API out as thread safe. https://developer.xamarin.com/api/property/MonoTouch.UIKit.UIApplication.NetworkActivityIndicatorVisible/. My experience with Xamarin is they only do this for APIs they know are thread safe and take cautious approach to permitting non-UI thread access.

kzaher:
@mwoollard , I've been thinking about this, and I think that you are maybe right about this. I think that status bar is shared between applications and therefor the concept on main thread could not make sense. 

I think our abstraction with main thread is maybe little too specific there, but should be fine.

@MadsBogeskov could you maybe add unit tests so we can pull this in?

MadsBogeskov:
Sure :-) Ill do it asap!

via CloudMagic for Mac
[https://cloudmagic.com/k/d/mailapp?ct=dx&cv=7.6.19&pv=10.11.3&source=email_footer_2]
On Sun, Jan 31, 2016 at 7:09 PM, Krunoslav Zaher <notifications@github.com>
wrote:
@mwoollard [https://github.com/mwoollard] , I've been thinking about this, and I think that you are maybe right about
this. I think that status bar is shared between applications and therefor the
concept on main thread could not make sense.

I think our abstraction with main thread is maybe little too specific there, but
should be fine.

@MadsBogeskov [https://github.com/MadsBogeskov] could you maybe add unit tests so we can pull this in?

—
Reply to this email directly or view it on GitHub
[https://github.com/ReactiveX/RxSwift/pull/414#issuecomment-177566301] .[https://github.com/notifications/beacon/AAYbBVjNf0zk-nhs2lyRXCBaUnNEEGJIks5pfkVvgaJpZM4HHJoW.gif]

MadsBogeskov:
@kzaher just tried to implement the tests, however it doesn't seem to be possible to access `UIApplication.sharedApplication` from a unit test: `'sharedApplication()' is unavailable: Use view controller based solutions where appropriate instead.`

So I think that this PR perhaps should be removed, unless you guys have any idea.

kzaher:
Hm,

I think we could make an exception in this case since this is setting something on singleton object, and can't be unit tested easily.

PR #418, opened 2016-01-19
GeolocationExample

carlosypunto:


RxExample/RxExample/Services/GeolocationService.swift
```diff
@@ -0,0 +1,48 @@
...
+    private (set) var location: Driver<CLLocationCoordinate2D>
+    
+    private let locationManager = CLLocationManager()
+    
+    private init() {
+        locationManager.distanceFilter = kCLDistanceFilterNone;
+        locationManager.desiredAccuracy = kCLLocationAccuracyBestForNavigation;
+        
+        autorized = locationManager.rx_didChangeAuthorizationStatus
+            .startWith(CLLocationManager.authorizationStatus())
```
kzaher: I think this would need to have form
```
Observable.deferred {
    locationManager
       .rx_didChangeAuthorizationStatus
       .startWith(CLLocationManager.authorizationStatus())
}
.asDriver(....
```
... otherwise we are using value at time of declaration, and not subscription.

RxExample/RxExample/Services/GeolocationService.swift
```diff
@@ -0,0 +1,48 @@
...
+                    return true
+                default:
+                    return false
+                }
+            }
+        
+        location = locationManager.rx_didUpdateLocations
+            .asDriver(onErrorJustReturn: [])
+            .filter { $0.count > 0 }
+            .map { $0.last!.coordinate }
```
kzaher: Maybe do something like
```
.flatMap { $0.last?.coordinate.map(Observable.just) ?? Observable.empty() }
```
That `last!` seems kind of unsafe to me :)

RxExample/RxExample/Examples/GeolocationExample/GeolocationViewController.swift
```diff
@@ -0,0 +1,67 @@
...
+        geolocationService.autorized
+            .driveNext(driveAutorization)
+            .addDisposableTo(disposeBag)
+        
+        geolocationService.location
+            .driveNext(driveCoordinates)
+            .addDisposableTo(disposeBag)
+        
+        button.rx_tap
+            .bindNext(openAppPreferences)
```
kzaher: I think referencing a method like this would case a retain cycle :(

I believe the same is valid for `driveCoordinates` and `driveAutorization`.

Not sure I have some pretty suggestion if methods are defined on controller. This isn't a problem if extensions are defined on view because then nothing is referencing VC.

¯\\\_(ツ)\_/¯

kzaher: Maybe just extending UILabel with 

var rx_coordinate: AnyObserver< CLLocationCoordinate2D > {
...
}

and use two line text inside a single label for coordinates?

RxExample/RxExample/Services/GeolocationService.swift
carlosypunto: nice trick, thanks

RxExample/RxExample/Services/GeolocationService.swift
carlosypunto: I understand that `filter` no longer be necessary, right?

RxExample/RxExample/Services/GeolocationService.swift
carlosypunto: But I've found this problem, and I don't know how to solve?
image

carlosypunto: Any idea?

RxExample/RxExample/Examples/GeolocationExample/GeolocationViewController.swift
carlosypunto: I think I do not quite understand the phrase, "This isn't a problem if extensions are defined on view because then nothing is referencing VC." Why?

carlosypunto: What you think about to suse:

```

        weak var _self = self
        
        geolocationService.autorized
            .driveNext(_self!.driveAutorization)
            .addDisposableTo(disposeBag)
```

or

```

        unowned var _self = self
        
        geolocationService.autorized
            .driveNext(_self.driveAutorization)
            .addDisposableTo(disposeBag)
```

RxExample/RxExample/Services/GeolocationService.swift
carlosypunto: Solved!:

```
        let locationManager = self.locationManager
        locationManager.distanceFilter = kCLDistanceFilterNone;
        locationManager.desiredAccuracy = kCLLocationAccuracyBestForNavigation;
        
        autorized = Observable.deferred {
                locationManager
                    .rx_didChangeAuthorizationStatus
                    .startWith(CLLocationManager.authorizationStatus())
            }
            .asDriver(onErrorJustReturn: CLAuthorizationStatus.NotDetermined)
            .map {
                switch $0 {
                case .AuthorizedAlways:
                    return true
                default:
                    return false
                }
            }
```

RxExample/RxExample/Services/GeolocationService.swift
```diff
@@ -0,0 +1,53 @@
...
+class GeolocationService {
+    
+    static let instance = GeolocationService()
+    private (set) var autorized: Driver<Bool>
+    private (set) var location: Driver<CLLocationCoordinate2D>
+    
+    private let locationManager = CLLocationManager()
+    
+    private init() {
+        let locationManager = self.locationManager
```
carlosypunto: Maybe `unowned let locationManager = self.locationManager`

carlosypunto:
I apologize for so many commits, when is ready I will make a PR with one commit.

RxExample/RxExample/Examples/GeolocationExample/GeolocationViewController.swift
```diff
@@ -0,0 +1,69 @@
...
+    @IBOutlet weak private var button2: UIButton!
+    @IBOutlet weak private var latLabel: UILabel!
+    @IBOutlet weak private var lonLabel: UILabel!
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        let geolocationService = GeolocationService.instance
+        
+        unowned let _self = self
```
kzaher: I don't think we should be giving examples with `unowned`. I think it's probably ok in this case, but it's extremely easy to create subtle errors with `unowned` and crash the app.

Maybe it would be better to have instead of
```
private func driveAutorization(autorized: Bool) {
 +        if autorized {
 +            noGeolocationView.hidden = true
 +            view.sendSubviewToBack(noGeolocationView)
 +        }
 +        else {
 +            noGeolocationView.hidden = false
 +            view.bringSubviewToFront(noGeolocationView)
 +        }
 +    }
```
an extension on `UIView`

```
var rx_authorizationState: Bool<Bool> 
```

with that logic.

In that case you can just say `drive(noGeolocationView.rx_authorizationState)`.

The same solution can apply to `driveCoordinates`. We can have only one label with 
```
"Latitude: \(location.latitude)\nLongitude: \(location.longitude)"
```

RxExample/RxExample/Services/GeolocationService.swift
```diff
@@ -0,0 +1,58 @@
...
+                    return true
+                default:
+                    return false
+                }
+            }
+        
+        location = locationManager.rx_didUpdateLocations
+            .asDriver(onErrorJustReturn: [])
+            .filter { !$0.isEmpty }
+            .map { $0.last!.coordinate }
```
kzaher: I think we should replace this with `flatMap` and not use `last!`. Even though it looks safe I would try to not use `!` in any part of code when there are other solutions available.
```
.flatMap { x in
     return $0.last.map(Observable.just) ?? Observable.empty
}
```

carlosypunto:
Cosed in favor of #429

PR #421, opened 2016-01-20
ForwardIf operator

koke:
ForwardIf propagates the source observable sequence while the condition
observable sequence last value is true.

This is useful when an Observable performs side effects on subscription, so we
only want to subscribe when another observable is true.

An example case for this would be an Observable that performs network requests.
We only want it to hit the network if the subscribing view controller is
visible.

This example would look something like this:

```swift
service.refreshModel
  .forwardIf(visible)
  .subscribeNext(bindModel)
  .addDisposableTo(disposeBag)
```

Notes:

- `ForwardIf` is mostly a suggestion. I've thought about how to name this and I'm not 100% sold on any name. Other options: `subscribeWhen`, `subscribeWhile`, `forwardWhile`.
- Even if you decide not to include this as a part of RxSwift, I'd really appreciate a code review or any comments about the implementation, specially around locking and threading. It's my first approach to RxSwift code, and I'm not too sure if I did things right.

### DIFF
--- a/Rx.playground/Pages/Combining_Observables.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Combining_Observables.xcplaygroundpage/Contents.swift
@@ -236,6 +236,28 @@ example("merge 2") {
     subject1.on(.Next(80))
     subject1.on(.Next(100))
     subject2.on(.Next(1))
+
+}
+
+example("forwardIf") {
+    let subject1 = PublishSubject<Int>()
+    let subject2 = PublishSubject<Bool>()
+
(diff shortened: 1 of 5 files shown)


koke:
It looks like RxJS [pausable](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/pausable.md) does a similar thing. Looking at the implementation I understand it does the same thing, but the documentation doesn't guarantee that the source observable would only get subscribed when the condition is true.

koke:
Wrong base branch, will resubmit against develop, sorry

PR #422, opened 2016-01-20
ForwardIf operator

koke:
ForwardIf propagates the source observable sequence while the condition
observable sequence last value is true.

This is useful when an Observable performs side effects on subscription, so we
only want to subscribe when another observable is true.

An example case for this would be an Observable that performs network requests.
We only want it to hit the network if the subscribing view controller is
visible.

This example would look something like this:

```swift
service.refreshModel
  .forwardIf(visible)
  .subscribeNext(bindModel)
  .addDisposableTo(disposeBag)
```

Notes:

- `ForwardIf` is mostly a suggestion. I've thought about how to name this and I'm not 100% sold on any name. Other options: `subscribeWhen`, `subscribeWhile`, `forwardWhile`.
- It looks like RxJS [pausable](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/pausable.md) does a similar thing. Looking at the implementation I understand it does the same thing, but the documentation doesn't guarantee that the source observable would only get subscribed when the condition is true.
- Even if you decide not to include this as a part of RxSwift, I'd really appreciate a code review or any comments about the implementation, specially around locking and threading. It's my first approach to RxSwift code, and I'm not too sure if I did things right.

### DIFF
--- a/RxSwift/Observables/Implementations/ForwardIf.swift
+++ b/RxSwift/Observables/Implementations/ForwardIf.swift
@@ -1,0 +1,129 @@
+//
+//  ForwardIf.swift
+//  Rx
+//
+//  Created by Jorge Bernal Ordovas on 20/01/16.
+//  Copyright © 2016 Krunoslav Zaher. All rights reserved.
+//
+
+import Foundation
+
+class ForwardIfSinkIter<S: ObservableType, O: ObserverType where O.E == S.E> : ObserverType {
+    typealias Parent = ForwardIfSink<S, O>
+    typealias DisposeKey = CompositeDisposable.DisposeKey
+    typealias E = O.E
+
+    private let _parent: Parent
+    private let _disposeKey: DisposeKey
+
+    init(parent: Parent, disposeKey: DisposeKey) {
+        _parent = parent
+        _disposeKey = disposeKey
+    }
+
+    func on(event: Event<E>) {
+        switch event {
+        case .Next(let value):
+            _parent._lock.lock(); defer { _parent._lock.unlock() } // lock {
+                _parent.forwardOn(.Next(value))
+            // }
+        case .Error(let error):
+            _parent._lock.lock(); defer { _parent._lock.unlock() } // lock {
+                _parent.forwardOn(.Error(error))
+                _parent.dispose()
+            // }
+        case .Completed:
+            _parent._group.removeDisposable(_disposeKey)
+            _parent._lock.lock(); defer { _parent._lock.unlock() } // lock {
+                _parent.forwardOn(.Completed)
+                _parent.dispose()
+            // }
+        }
+    }
+
+}
+
+class ForwardIfSink<S: ObservableType, O: ObserverType where O.E == S.E>: Sink<O>, ObserverType {
+    typealias E = Bool
+    typealias SourceType = S.E
+    typealias DisposeKey = CompositeDisposable.DisposeKey
+
+    private let _lock = NSRecursiveLock()
+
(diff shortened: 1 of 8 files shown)


kzaher:
Hi @koke ,

Thnx for making this PR and explaining everything in such a detail. The don't think we should be pulling this code inside this repro although it seems to be functionally equivalent to `pausable`. I would be even weirder if we would intentionally pull in `pausable` operator under different name :)

I wouldn't like to pull pausable inside this repo for now. We would only like to have the most fundamental and useful operators here for now, otherwise this framework will grow so large nobody will want to ship it inside their mobile apps.

If you like, I can CR your code :)

Tests/RxSwiftTests/Tests/Observable+MultipleTest.swift
```diff
@@ -4621,4 +4621,85 @@ extension ObservableMultipleTest {
...
+        ])
+
+        let res = scheduler.start {
+            xs.forwardIf(ys)
+        }
+
+        XCTAssertEqual(res.events, [
+            next(250, 3),
+            next(260, 4),
+            next(360, 6),
```
kzaher: I think if you want to use your operator in these contexts
```
service.refreshModel
  .forwardIf(visible)
  .subscribeNext(bindModel)
  .addDisposableTo(disposeBag)
```
.. then IMHO this should be 
```
XCTAssertEqual(res.events, [
 +            next(250, 3),
 +            next(260, 4),
 +            next(350, 5), // <-- replaying last element received
 +            next(360, 6),
```

RxSwift/Observables/Implementations/ForwardIf.swift
```diff
@@ -0,0 +1,129 @@
...
+    private let _condition: Observable<Bool>
+
+    init(source: S, condition: Observable<Bool>) {
+        _source = source
+        _condition = condition
+    }
+
+    override func run<O: ObserverType where O.E == S.E>(observer: O) -> Disposable {
+        let sink = ForwardIfSink<S, O>(observer: observer, source: _source)
+        sink.disposable = sink.run(_condition)
```
kzaher: It is better to pass _condition inside initializer function probably :)

RxSwift/Observables/Implementations/ForwardIf.swift
```diff
@@ -0,0 +1,129 @@
...
+        case .Error(let error):
+            _lock.lock(); defer { _lock.unlock() } // lock {
+                forwardOn(.Error(error))
+                dispose()
+            // }
+        }
+    }
+
+    func subscribeInner(source: S) {
+        print("subscribeInner")
```
kzaher: :)

RxSwift/Observables/Implementations/ForwardIf.swift
```diff
@@ -0,0 +1,129 @@
...
+    init(observer: O, source: S) {
+        _source = source
+        super.init(observer: observer)
+    }
+
+    func on(event: Event<E>) {
+        switch event {
+        case .Next(let element):
+            if element {
+                subscribeInner(_source)
```
kzaher: You would also need to synchronize calling `subscribeInner` and `unsubscribeInner` since they are mutating self.

koke:
I understand, and thanks for the review. I had to rewrite this a bit to use in my project, since some of the classes like `Sink` and `Producer` are internal.

For reference, this is what I got so far: https://gist.github.com/koke/ff93f27f3c93b8a4f27c

Issue #404, opened 2016-01-13
How to write canEditRowAtIndexPath for tableview with rx_dataSource?


frogcjn:
```Swift            
            ...
            .bindTo(tableView.rx_itemsWithDataSource(dataSource))
            .addDisposableTo(controlsDisposeBag)
```
I have a tableView with an rx_datasource

but How can I implement tableView:canEditRowAtIndexPath function with this tableView?

sergdort:
Hi, @frogcjn  I guess (if you are using RxDataSources) you need just override your `RxTableView...DataSource` and implement method based on your logic.

thomasschuster:
I can confirm that. I just extended the `RxTableView...DataSource` and implemented the methods of the `UITableViewDataSource` protocol there.

kzaher:
Yeah, we'll probably turn those delegate methods into lambdas in data sources inside RxDataSources project.

sergdort:
Hi @kzaher bit later I can also implement some examples, to show usage, if you want.

kzaher:
@sergdort I think it might be useful to transform RxDataSources project example app to use navigation view controller and add customization example.

kzaher:
Looks like we can close this one also.

PR #425, opened 2016-01-21
#404 Added lambdas for RxDataSources

sergdort:


kzaher:
Hi @sergdort ,

I think this is correct PR, but to a old repo :) We have pulled out data sources [here](https://github.com/RxSwiftCommunity/RxDataSources).

We will maybe sync some changes back here in future, but it should be one unidirectional sync always :)

That means that we are planning to grow that codebase there :)

Hope you can make PR with these changes there.

sergdort:
Hi @kzaher , will do thanks :)

PR #431, opened 2016-01-23
Add rx wrapper for UITableViewDelegate tableVIew:accessoryButtonTappedForRowWithIndexPath: method

mwoollard:
Had a need for this is something I was working on - would seem to make sense to include this delegate method? Included its use in the simple table view example.

### DIFF
--- a/RxCocoa/iOS/UITableView+Rx.swift
+++ b/RxCocoa/iOS/UITableView+Rx.swift
@@ -149,6 +149,18 @@ extension UITableView {
     }
 
     /**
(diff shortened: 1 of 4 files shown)


kzaher:
Hi @mwoollard, I think we could include this if you could add unit tests for this extension.

mwoollard:
OK will do

mwoollard:
Have added unit test following pattern used for the other table view delegate / data source rx extensions.

kzaher:
Hi @mwoollard ,

If you could solve merge conflicts, I would be happy to merge this.

mwoollard:
Will do - may take a few days to get to it - lot going on right now !

mwoollard:
Should be all good now

PR #432, opened 2016-01-24
Added reactive wrapper for WKWebView

mokumoku:
I want to observe `WKWebView`'s read only properties.

### DIFF
--- a/Documentation/API.md
+++ b/Documentation/API.md
@@ -406,6 +406,23 @@ extension UIActivityIndicatorView {
 }
 ```
 
+```swift
+extension WKWebView {
+
+    public var rx_canGoBack: Observable<Bool> {}
+
+    public var rx_canGoForward: Observable<Bool> {}
+
+    public var rx_loading: Observable<Bool> {}
+
+    public var rx_estimatedProgress<Double> {}
+
+    public var rx_title<String?> {}
+
+    public var rx_URL<NSURL?> {}
+}
+```
+
 **OSX**
 
 ```swift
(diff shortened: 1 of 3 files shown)


kzaher:
Hi @mokumoku ,

I think you extensions totally make sense. There are just 2 problems.

The first one is that I'm not sure we would want to include them inside RxCocoa project since they are WebKit extensions.

```
import WebKit
```

The other is that I haven't found that those properties are officially KVO compatible, which means that some other notification mechanism should probably be implemented to observe changes.

If you are interested, I can add you to RxSwiftCommunity and we can start RxWebKit extensions project there. I understand if it's not something you are interested in or don't have time for.

mokumoku:
Hi @kzaher ,
Thank you for your response.

I could import `WebKit` naturally just like importing `Foundation` or `UIKit` in my project, Xcode 7.2 and Deployment Target iOS 8.0.
Now, I understand it is considerable importing `WebKit`to RxCocoa.

I am interested in RxWebKit!
I also want to find official KVO compatible ways to observe `WKWebView`'s properties there.

kzaher:
Hi @mokumoku ,

First of all, I've sent you an invite to join RxSwiftCommunity.

I've been thinking about this PR. I'm still unsure how would we organize all other frameworks there like:
* WebKit
* MapKit
* CoreData
...

Maybe separating them into their own repositories would be too much maintenance. Probably the simplest would be to create a single repository that contains pods (and Carthage targets) for all remaining apple frameworks.

Something like RxDarwin, or RxApple, or something else ...

In that way we could set up build server and integration tests for only one more repository.

As far as for KVO property observing, I don't think you can use KVO observing to officially observe them but you'll would need to use delegates instead.

These properties are KVO compliant:
* estimatedProgress
* loading
* title
* URL

For these we don't know:
* canGoBack
* canGoForward

I believe that `canGoBack` and `canGoForward` are related with navigation, so I think you would need to use `navigationDelegate` as a notification mechanism. At least, that is my understanding.

https://developer.apple.com/library/ios/documentation/WebKit/Reference/WKWebView_Ref/#//apple_ref/occ/instp/WKWebView/canGoBack

mokumoku:
Hi @kzaher ,

Thank you for your invitation.
I joined RxSwiftCommunity! What should I do next?

There are some repos in RxSwiftCommunity and I will add a new repo.
As you said, I think only one repository style is good from the viewpoint of maintenance or extensibility.

I also feel uneasy that we clone unnecessary framework if we have only one repository.
For example, now we install RxSwift via Carthage, we get 4 frameworks: RxSwift, RxCocoa, RxBlocking, RxTests though we want only RxSwift framework.
(if we install via CocoaPods, the problem doesn't occur.)

If we resolve the point, I agree your plan.

As far as KVO observing, I think you are right.

I will try to use `navigationDelegate` for `canGoBack` and `canGoForward`.

kzaher:
Hi @mokumoku ,

I believe you should be able to create new repository under RxSwiftCommunity. You can chose a name for this. Maybe `RxWebKit` would be reasonable.

You can check out how to set up CocoaPods and Carthage integration in one of the other sibling projects. I know it works for RxDataSources, but probably works for others as well. And that's it :)

It would be nice to add a small example app together with `RxWebKit`, short README.md that describes the project and license file. All the usual stuff :)

mokumoku:
Hi @kzaher ,

Thank you for your help.

I will create `RxWebKit` later.

I also try to make it CocoaPods and Carthage compatible using `RxDataSources` as a reference.

I close this Pull Request.

PR #444, opened 2016-01-27
Remove "Binding error to variable: " from error message from bindNext

thanegill:
There is no variable that is being bound to with `bindNext()`


thanegill:
Travis seems to have timed out.

kzaher:
Oh, can you just please target it towards `develop` branch?

Maybe it should say ""Binding error : \(error)"?

thanegill:
I swear I did. Guess not.

PR #449, opened 2016-01-27
adds rx_active observer for NSLayoutConstraint

mikelow:
with unit tests


mikelow:
@kzaher, tests were failing due to a sdk version issue (the `active` property is unavailable in OSX pre 10.10). I've inserted @available attributes to remedy the problem.

PR #456, opened 2016-01-29
Return generic SignedIntegerType Observable for timer instead of Int64

thanegill:
Seems to have been left out when the time creators where converted to contained extensions.


thanegill:
Closing as #457 will encompass these changes.

PR #459, opened 2016-01-30
Add rx_refreshing for UIRefreshControl

ishkawa:


kzaher:
Makes sense, thnx.

njovy:
``` 
extension UIRefreshControl {
  /**
   Reactive wrapper for `ValueChanged` control event.
   */
  public var rx_changed: ControlEvent<Void> {
    return rx_controlEvent(.ValueChanged)
  }
}
```
@ishkawa What about observing ```ValueChanged``` control event?

PR #461, opened 2016-01-30
RxBlocking fixes #460

### LINKED ISSUE #460: RxBlocking last() method does not call stop() on RunLoopLock if observer has value
//Steps to recreate:
1. Call: try observer.toBlocking().last()
2. CFRunLoop forever blocked

```
extension BlockingObservable {
   ...
    public func last() throws -> E? {
        ...
        let lock = RunLoopLock()
        ...
        lock.dispatch {
            d.disposable = self.source.subscribe { e in
                switch e {
                case .Next(let e):
                    element = e
                    return  <---- CFRunLoop never stopped
                case .Error(let e):
                    error = e
                default:
                    break
                }

                lock.stop()
            }
        }
        
        lock.run()        
        ...
        return element
    }
}
```

mark-randall:


kzaher:
Hi @mark-randall ,

I think our unit tests are failing because of this change.

Why do you think there shouldn't be a return there?

PR #466, opened 2016-02-02
Changes for currying

sergdort:


RxCocoa/iOS/UITableView+Rx.swift
```diff
@@ -46,16 +48,18 @@ extension UITableView {
...
-            return cell
+        return { configureCell in
+            let dataSource = RxTableViewReactiveArrayDataSourceSequenceWrapper<S> { (tv, i, item) in
+                let indexPath = NSIndexPath(forItem: i, inSection: 0)
+                let cell = tv.dequeueReusableCellWithIdentifier(cellIdentifier, forIndexPath: indexPath) as! Cell
+                configureCell(i, item, cell)
+                return cell
+            }
+                
+            return self.rx_itemsWithDataSource(dataSource)(source: source)
```
sergdort: I'm also wondering if we should use `[weak self]`, because of currying (since it just function which returns closure)

RxCocoa/Common/RxCocoa.swift
```diff
@@ -287,6 +287,15 @@ func castOrFatalError<T>(value: AnyObject!) -> T {
...
 
+func castOrFatalError<T>(value: Any!) -> T {
+    let maybeResult: T? = value as? T
+    guard let result = maybeResult else {
+        rxFatalError("Failure converting from \(value) to \(T.self)")
+    }
+    
+    return result
+}
+
```
kzaher: Is there any difference then just changing `AnyObject` parameter in `func castOrFatalError<T>(value: AnyObject!) -> T {`

to `Any!`?

RxCocoa/iOS/UITableView+Rx.swift
kzaher: I think it would be ok to use weak, but there is one more currying missing :) `(source: O)`

We also have UICollectionView

RxCocoa/Common/RxCocoa.swift
sergdort: I was not sure if you would like to add this, but I've noticed that there was a force unwrap :)
 And yes there is no difference except `Any!`

RxCocoa/iOS/UITableView+Rx.swift
sergdort: Oh, my... this currying :)
Will fixe it

sergdort:
I guess I have to make also these changes to `UICollectionView+Rx`
Should I close it for know, or just leave it ?

kzaher:
We would need to change this for UICollectionView and UITableView. Those are the only places where we use currying.

If you want to do it everywhere, that would be great :)

RxCocoa/Common/RxCocoa.swift
kzaher: Yeah, I think you can just change from `func castOrFatalError<T>(value: AnyObject!) -> T {` to `func castOrFatalError<T>(value: Any!) -> T {`

Currently don't see any reason why we would need two functions.

sergdort:
@kzaher  I've made those changes for `UICollectionView` as well

kzaher:
So far the only problem I'm seeing is that we still have 
```
+func castOrFatalError<T>(value: Any!) -> T {
```

I can take a closer look tonight.

sergdort:
Hi, @kzaher , I've just made this changes

PR #476, opened 2016-02-07
Adding duration parameter to animated properties, adding hidden and alpha animated properties

coryoso:


kzaher:
Hi @coryoso ,

I've made that animate function kind of an exceptional case since animated image binding is common these days.

As you can see, adding all combinations of animated parameters causes an explosion in the amount of code, so I'm not convinced that it is the direction we would like to take here.

I think it would be ok to expose.

````
transition.duration = 0.25
transition.timingFunction = CAMediaTimingFunction(name: kCAMediaTimingFunctionEaseInEaseOut)
```

as additional parameters with default values here
```
public func rx_imageAnimated(transitionType: String?)
```

... but wouldn't want to extend that interface style across the library for the mentioned reasons.

In statistically most likely cases people will want to use coordinated animations on more then one element, so using animatable properties and one of the ...

```
UIView.animateWithDuration*
```

... seems to me like a way to solve this problem more elegantly.

coryoso:
Makes sense, pull request can be closed

PR #478, opened 2016-02-08
Fix: Changes visibility of BinaryDisposable initializer to public

zittix:
Changes visibility of BinaryDisposable initializer to public

### DIFF
--- a/RxSwift/Disposables/BinaryDisposable.swift
+++ b/RxSwift/Disposables/BinaryDisposable.swift
@@ -34,7 +34,7 @@ public final class BinaryDisposable : DisposeBase, Cancelable {
     - parameter disposable1: First disposable
     - parameter disposable2: Second disposable
     */
-    init(_ disposable1: Disposable, _ disposable2: Disposable) {
+    public init(_ disposable1: Disposable, _ disposable2: Disposable) {
         _disposable1 = disposable1
         _disposable2 = disposable2
         super.init()


thanegill:
`StableCompositeDisposable.create` is the public interface for `BinaryDisposable`

zittix:
My bad, I didn't see this one. 

Closing this PR although, in my point of view, creating a BinaryDisposable from the init is much more coherent as other Disposables like AnonymousDisposable are created through an init and not through a factory method.

thanegill:
I agree. I think there needs to be a rewrite of the disposables. BinaryDisposable and CompositeDisposable seem to accomplish the same thing with StableCompositeDisposable.create being the public interface for BinaryDisposable. Seems a little convoluted.

Maybe @kzaher can chime in as tell us why we're wrong 😊.

kzaher:
Hi guys,

So the original plan was to have:

```
public final class StableCompositeDisposable {
    public static func create(disposable1: Disposable, _ disposable2: Disposable) -> Disposable {
        return BinaryDisposable(disposable1, disposable2)
    }
    public static func create(disposable1: Disposable, _ disposable2: Disposable, _  disposable3: Disposable) -> Disposable {
        return TernaryDisposable(disposable1, disposable2)
    }

   public static func create(disposable1: Disposable ... ) -> Disposable {
        return StableNaryDisposable(disposable1, disposable2)
    }
}
```

It was intended as an optimization that would enable to transparently choose correct private implementation.

Unfortunately Swift had issues with overloading, and that plan failed. The next plan was then to expose those specialized optimized disposables, although hoping for Apple to fix the overloads issue.

As time passed by, we also didn't internally have a need for `TernaryDisposable` or > 3 version, so this caused this interface to look weird like it looks right now.

I agree, this part could be done nicer, and we'll polish this in 3.0. I'm still hoping for original plan to work (in maybe Swift 3.0) and Swift compiler to correctly recognize those `create` overloads so we can hide implementations.

thanegill:
Why have a factory class and not initializers? Also what's wrong with a single `CompositeDisposable` that has a variadic initializer parameter, that could, if needed, have optimizations under the hood for the number of given disposables?

kzaher:
> Why have a factory class and not initializers?

Because the original plan was to go with factory and we couldn't because of Swift compiler and I was still hoping Apple would resolve this until we ship 2.0, but `BinaryDisposable` was exposed publicly (I guess we could now probably make that init public anyway, although it would be nicer to just be able to use that create factory).

> Also what's wrong with a single CompositeDisposable that has a variadic initializer parameter, that could, if needed, have optimizations under the hood for the number of given disposables?

There is nothing wrong with it, but I wanted to make this library really performance oriented and have all optimizations that we could, and that other Rx teams have.

The benefit of using stable disposable versions is additional optimization they bring to the table.

e.g.
`BinaryDisposable` already has allocated slots for two disposables, and doesn't need to allocate additional memory on heap to store those disposables (yes, even though Swift `Array` is a struct it does allocate memory on heap). Since stable disposables can be mutated only in a really simple way (dispose method), they can use extremely simple and fast atomics.

```
if AtomicCompareAndSwap(0, 1, &_disposed) {
            _disposable1?.dispose()
            _disposable2?.dispose()
            _disposable1 = nil
            _disposable2 = nil
}
```
... on the other hand ... `CompositeDisposable` needs to use real non reentrant locks
```

        let disposeBag = _disposables
        _disposables = nil
```

so that's way more expensive. Apple also says spin locks are unsafe in iOS (https://lists.swift.org/pipermail/swift-dev/Week-of-Mon-20151214/000321.html) so we are touching the kernel to dispose `CompositeDisposable` because we can't use `OSSpinLock` (that was replaced in RxSwift 2.1 version). 

There are maybe some other options available with dispatch semaphores, but didn't want to accidentally hit some other obscure problem. Until I investigate it more, I've made a simple choice.

thanegill:
That's a great explanation. 😊

I believe that there needs to be one place to initialize Disposables. Either with a factory class for public initializers but not both. What optimized disposable arity would you go up to? Three? Four?

kzaher:
I believe that so far the project only uses Binary version, so YAGNI says stop there :)

PR #481, opened 2016-02-10
Delay operator

tarunon:
Description of operator is here: http://reactivex.io/documentation/operators/delay.html

### DIFF
--- a/Documentation/API.md
+++ b/Documentation/API.md
@@ -61,7 +61,7 @@ Operators are stateless by default.
 
 #### Observable Utility Operators
 
-  * [`delaySubscription`](http://reactivex.io/documentation/operators/delay.html)
+  * [`delaySubscription` / `delay`](http://reactivex.io/documentation/operators/delay.html)
   * [`do` / `doOnNext`](http://reactivex.io/documentation/operators/do.html)
   * [`observeOn` / `observeSingleOn`](http://reactivex.io/documentation/operators/observeon.html)
   * [`subscribe`](http://reactivex.io/documentation/operators/subscribe.html)
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -1107,6 +1107,10 @@
 		D2FC15B31BCB95E5007361FF /* SkipWhile.swift in Sources */ = {isa = PBXBuildFile; fileRef = D22B6D251BC8504A00BCE0AB /* SkipWhile.swift */; };
 		D2FC15B41BCB95E7007361FF /* SkipWhile.swift in Sources */ = {isa = PBXBuildFile; fileRef = D22B6D251BC8504A00BCE0AB /* SkipWhile.swift */; };
 		D2FC15B51BCB95E8007361FF /* SkipWhile.swift in Sources */ = {isa = PBXBuildFile; fileRef = D22B6D251BC8504A00BCE0AB /* SkipWhile.swift */; };
+		EB8293841C698D1A00315EB6 /* Delay.swift in Sources */ = {isa = PBXBuildFile; fileRef = EB8293831C698D1A00315EB6 /* Delay.swift */; };
+		EB8293851C69A70700315EB6 /* Delay.swift in Sources */ = {isa = PBXBuildFile; fileRef = EB8293831C698D1A00315EB6 /* Delay.swift */; };
+		EB8293861C69A70700315EB6 /* Delay.swift in Sources */ = {isa = PBXBuildFile; fileRef = EB8293831C698D1A00315EB6 /* Delay.swift */; };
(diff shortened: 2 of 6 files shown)


kzaher:
Hi @tarunon ,

I think we would need to have a discussion first regarding is this operator common enough to include it in the main repo :)

Can you maybe give me some context why do you need this operator for?

tarunon:
Hello, @kzaher san,

I want this operator when provide a time difference to the multiple event handler.
It would be required in some cases. (e.g. monitoring of offset, sending URL request caused user action, adding animation effect, etc...)

Now, we should use `flatMap` and `delaySubscription` operators like next...
```swift
observable.doOn { event in
  // write first handler here.
}
.flatMap { value -> Observable<Value> in
  return Observable.just(value)
    .delaySubscription(3.0, someScheduler)
}
.subscribe { event in
  // write second handler here.
}
```
But I think it is lazy.
In other language (RxJava, RxJS, and other), we can use `delay` operator instead of it.

DanielAsher:
+1 for this feature.

kzaher:
Hi @tarunon , @DanielAsher ,

I haven't forgot about this PR and @tottakai 's group by. 

I have done some investigations what would it take to include this operator in this project, but still need to complete them. 

If we were to include this operator, we would need to change the proposed `delay` implementation significantly. The proposed implementation assumes that schedulers are serial, and wouldn't work for concurrent schedulers. You can check out source code for .NET version to find out more.

The unit tests would also need to be improved.

I'll need to find some time to dig in an evaluate how complex implementation would need to be :) That was the most important reason why we have `delaySubscription` and don't have `delay` right now.

On the other hand, workaround that @tarunon suggested will work for serial schedulers (like main scheduler)

```
observable.doOn { event in
  // write first handler here.
}
.flatMap { value -> Observable<Value> in
  return Observable.just(value)
    .delaySubscription(3.0, someScheduler)
}
.subscribe { event in
  // write second handler here.
}
```

... so I would say this is for now a valid workaround. You can ever create a custom operator that wraps that 

```
.flatMap { value -> Observable<Value> in
  return Observable.just(value)
    .delaySubscription(3.0, someScheduler)
}
```

DanielAsher:
Thanks so much for this detailed response and reiteration of a temporary solution. Not one of us presumes to have as complete a picture of the ramifications of small library modifications as you do. A `delay` for comteplation is quite apt :)

tarunon:
Hi @kzaher san,
Thanks for discussion.
I had thought the workaround is valid, but it was wrong.
```swift
.flatMap { value -> Observable<Value> in
  return Observable.just(value)
    .delaySubscription(3.0, someScheduler)
}
```
Because the `Completed` event did not delay. This workaround can delay only `Next` event...
`delay` operator must delay `Next` and `Completed`.
I'll improve my implementation and unit test.

tarunon:
Hi @kzaher san,
I improved my implementations, and I would like to resume our discussion.

kzaher:
Hi @tarunon ,

I'm not sure if you understood what I was trying to explain before. The path you've taken is not a correct one because your code assumes serial scheduler will be used, which isn't necessarily the case.

For example, here is a reference of RX.NET delay operator:

https://github.com/Reactive-Extensions/Rx.NET/blob/859e6159cb07be67fd36b18c2ae2b9a62979cb6d/Rx.NET/Source/System.Reactive.Linq/Reactive/Linq/Observable/Delay.cs

I understand this operator looks trivial, but there is significant complexity here because operator has to enforce event serialization. That was the reason why `delaySubscription` was ported and `delay` wasn't :)

I think that probably most optimal way is to use that convenience for now
```
.flatMap { value -> Observable<Value> in
  return Observable.just(value)
    .delaySubscription(3.0, someScheduler)
}
```
because usually delaying completed event usually isn't necessary, and people use serial schedulers.

If we want to include this operator in this repository I propose that maybe the simplest solution would be if I would implement the operator, and you helped me to port all unit tests from

https://github.com/Reactive-Extensions/Rx.NET/blob/master/Rx.NET/Source/Tests.System.Reactive/Tests/Linq/ObservableTimeTest.cs

We would need to port all unit tests that start with `Delay_TimeSpan`, and don't contain `Stopwatch` in their name.

tarunon:
Hi @kzaher san,
I'm sorry but I had a big misunderstanding.

I found the error of my implementation when using concurrency queue, and I repair my implementation now.
I think it is working correctly probably.

tarunon:
Hi @kzaher san,
I'll close this PR. 
I think there is that PR again repeated verification.
Thank you :)

kzaher:
Hey, sure :)

Hope we'll have this operator.

kzaher:
This has been merged to `develop` branch. Unsure why Github doesn't mark this as merged.

tarunon:
@kzaher 
Thank you notification.
I think it caused I closed already, But it is no problem.
I look forward to use this operator 😄

PR #484, opened 2016-02-15
Work around to an error with preprocessors in swift

AdamRitenauer:
Compilation of RxCocoa-iOS fails expecting an #endif at the end of UIApplication+Rx.swift, despite the #endif, in fact, being present.

Adding white space after the last #endif works around the problem.

See Open Radar here

http://www.openradar.appspot.com/18594933


kzaher:
Maybe only some versions of Xcode have this problem. 

We aren't WV the CI server, cross my heart :) ¯\\\_(ツ)\_/¯

PR #502, opened 2016-02-22
Added Observable for View animation completion that enables chaining

inkyfox:
Added Observable for View animation completion that enables chaining them.

- Simple Animation:
``` Swift
        UIView.rx_animation(1) { [weak self] in self?.summaryView.alpha = 0 }
            .subscribeNext { [weak self] in self?.showSomething() }
            .addDisposableTo(disposeBag)
```

- Animation Chaining:
``` Swift
    button.rx_tap.asObservable()
        .animate(1) { [weak self] in self?.imageView.alpha = 0 }
        .animate(1) { [weak self] in self?.imageView.alpha = 1 }
        .animate(1) { [weak self] in self?.imageView.alpha = 0 }
        .animate(1) { [weak self] in self?.imageView.alpha = 1 }
        .subscribeNext { [weak self] model in self?.showDetail(model) }
        .addDisposableTo(disposeBag)
```

### DIFF
--- a/RxCocoa/iOS/UIView+Rx.swift
+++ b/RxCocoa/iOS/UIView+Rx.swift
@@ -32,6 +32,28 @@ extension UIView {
             view.alpha = alpha
         }.asObserver()
     }
+    
+    static func rx_animation(duration: NSTimeInterval, animations: () -> Void) -> Observable<Bool> {
+        return Observable.create { observer in
+            UIView.animateWithDuration(duration, animations: animations) {
+                observer.on(.Next($0))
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @inkyfox ,

I also have similar convenience operators in my projects :)

I think that if we go that road and merge this PR, we'll increase RxCocoa public interface significantly because there are so many ways one can do animations.

I would rather group all animation conveniences into some kind of external project maybe.

inkyfox:
agreed. thank you!

PR #504, opened 2016-02-23
Adding 'rx_searchTapped' and 'rx_cancelTapped' ControlEvents to UISearchBar

fjcaetano:
I've created two ControlEvents that listens to UISearchBarDelegate methods `searchBarSearchButtonClicked:` and `searchBarCancelButtonClicked:`.

I also updated the API.md, the example project. I hope that's ok.

### DIFF
--- a/RxCocoa/iOS/UISearchBar+Rx.swift
+++ b/RxCocoa/iOS/UISearchBar+Rx.swift
@@ -16,7 +16,16 @@ import UIKit
 
 
 
-extension UISearchBar {
+  extension UISearchBar {
+    
+    /**
+     Factory method that enables subclasses to implement their own `rx_delegate`.
+     
+     - returns: Instance of delegate proxy that wraps `delegate`.
+     */
+    public func rx_createDelegateProxy() -> RxSearchBarDelegateProxy {
+      return RxSearchBarDelegateProxy(parentObject: self)
+    }
     
     /**
     Reactive wrapper for `delegate`.
@@ -51,7 +60,7 @@ extension UISearchBar {
     /**
      Reactive wrapper for `searchBarSearchButtonClicked:` delegate event.
      */
-    public var rx_searchTapped: ControlEvent<UISearchBar> {
+    public var rx_searchButtonClicked: ControlEvent<UISearchBar> {
         let source: Observable<UISearchBar> = Observable.deferred { [weak self] () -> Observable<UISearchBar> in
             return (self?.rx_delegate.observe("searchBarSearchButtonClicked:") ?? Observable.empty())
                 .flatMap { a -> Observable<UISearchBar> in
@@ -66,7 +75,7 @@ extension UISearchBar {
     /**
      Reactive wrapper for `searchBarCancelButtonClicked:` delegate event.
      */
-    public var rx_cancelTapped: ControlEvent<UISearchBar> {
+    public var rx_cancelButtonClicked: ControlEvent<UISearchBar> {
         let source: Observable<UISearchBar> = Observable.deferred { [weak self] () -> Observable<UISearchBar> in
             return (self?.rx_delegate.observe("searchBarCancelButtonClicked:") ?? Observable.empty())
                 .flatMap { a -> Observable<UISearchBar> in
--- a/RxCocoa/iOS/UISearchBar+Rx.swift
+++ b/RxCocoa/iOS/UISearchBar+Rx.swift
@@ -60,13 +60,10 @@ import UIKit
     /**
      Reactive wrapper for `searchBarSearchButtonClicked:` delegate event.
      */
-    public var rx_searchButtonClicked: ControlEvent<UISearchBar> {
-        let source: Observable<UISearchBar> = Observable.deferred { [weak self] () -> Observable<UISearchBar> in
+    public var rx_searchButtonClicked: ControlEvent<Void> {
+        let source: Observable<Void> = Observable.deferred { [weak self] () -> Observable<Void> in
             return (self?.rx_delegate.observe("searchBarSearchButtonClicked:") ?? Observable.empty())
-                .flatMap { a -> Observable<UISearchBar> in
-                    let result = a.first.flatMap { $0 as? UISearchBar }.flatMap { Observable.just($0) }
-                    return result ?? Observable.empty()
-            }
+                .map { _ in () }
         }
         
         return ControlEvent(events: source)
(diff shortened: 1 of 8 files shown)


kzaher:
Hi @fjcaetano ,

I think we could merge this to `develop` branch if we made some changes to the code to be more idiomatic. I'll note my suggestions.

RxCocoa/iOS/UISearchBar+Rx.swift
```diff
@@ -47,6 +47,36 @@ extension UISearchBar {
         
         return ControlProperty(values: source, valueSink: bindingObserver)
     }
-}
-
+    
+    /**
+     Reactive wrapper for `searchBarSearchButtonClicked:` delegate event.
+     */
+    public var rx_searchTapped: ControlEvent<UISearchBar> {
```
kzaher: We would want to make this method consistent with the way we are wrapping delegate methods regarding naming and return value.

That means this should be:
```
public var rx_searchButtonClicked: Observable<Void>
```
... and ...
```
public var rx_cancelButtonClicked: Observable<Void>
```

Using `Void` is what you want actually, because you want to abstract it as a generic event, it shouldn't be important that it comes from a `UISearchBar`.

RxExample/RxExample/Examples/GitHubSearchRepositories/GitHubSearchRepositoriesViewController.swift
```diff
@@ -50,8 +50,15 @@ class GitHubSearchRepositoriesViewController: ViewController, UITableViewDelegat
...
 
-        let searchResult = searchBar.rx_text.asDriver()
+        let searchText = searchBar.rx_text.asDriver()
             .throttle(0.3)
+        let searchTapped = searchBar.rx_searchTapped.asDriver()
+            .flatMap { $0.text.flatMap { Driver.just($0) } ?? Driver.empty() }
+        let cancelTapped = searchBar.rx_cancelTapped.asDriver()
+            .map { _ in "" }
+      
+        let searchResult = Driver.of(searchText, searchTapped, cancelTapped)
```
kzaher: Take a look at `withLatestFrom` operator and how we use it to get an idea how to write this in more idiomatic and simpler way :)

Tests/RxCocoaTests/Control+RxTests+UIKit.swift
```diff
@@ -616,10 +616,20 @@ extension ControlTests {
...
+  }
+  
+  func testSearchBar_DelegateEventSearchCompletesOnDealloc() {
+    let createView: () -> UISearchBar = { UISearchBar(frame: CGRectMake(0, 0, 1, 1)) }
+    ensureEventDeallocated(createView) { (view: UISearchBar) in view.rx_searchTapped }
+  }
+  
+  func testSearchBar_DelegateEventCancelCompletesOnDealloc() {
+    let createView: () -> UISearchBar = { UISearchBar(frame: CGRectMake(0, 0, 1, 1)) }
+    ensureEventDeallocated(createView) { (view: UISearchBar) in view.rx_cancelTapped }
```
kzaher: We would also want to add unit tests for 2 new extensions to verify that calling delegate methods works. Take a look at the tests we've added to `UITableViewDelegate` wrappers and `testTableView_ModelSelected_rx_itemsWithCellFactory`.

The reason why those tests don't exist for text property is because it's not trivial to trigger text changed on ui element in unit tests.

RxCocoa/iOS/UISearchBar+Rx.swift
fjcaetano: I'll rename it, and I get that you'd like the type to be consistent across the framework, but I honestly don't get why it couldn't be a `UISearchBar`, specially considering that that is what the original delegate returns.

Also, you wrote that it should be an `Observable`, why not `ControlProperty`?

RxExample/RxExample/Examples/GitHubSearchRepositories/GitHubSearchRepositoriesViewController.swift
fjcaetano: I realised the `searchTapped` is redundant with `searchText`, so I think it's better to remove it. But, if being so, is there a better way than to merge `cancelTapped` and `searchText`?

``` swift
        let searchText = searchBar.rx_text.asDriver()
            .throttle(0.3)
        let cancelTapped = searchBar.rx_cancelButtonClicked.asDriver()
            .map { _ in "" }
      
        let searchResult = Driver.of(searchText, cancelTapped)
            .merge()
```

Tests/RxCocoaTests/Control+RxTests+UIKit.swift
fjcaetano: For that I'd have to make RxSearchBarDelegateProxy public, is that ok?

fjcaetano:
@kzaher, I renamed the properties and wrote deeper tests, including DelegateProxy tests for the RxSearchBarDelegateProxy that had to be exposed.

I didn't change the observable type from ControlEvent to Observable nor its type because I'm waiting for your answer. If that's the way you decide to go, I'll make the requested changes, but I'd just like to understand that decision.

kzaher:
Hi @fjcaetano ,

I think that in this case we have a collision between two naming conventions. For delegates we've used just normal observable sequence. I think that in 3.0 we'll mark those delegate wrappers with their own unit, maybe `DelegateEvent` or something.

In this particular case I think it would probably make sense to leave them as `ControlEvents`.

RxCocoa/iOS/UISearchBar+Rx.swift
```diff
@@ -47,6 +56,36 @@ extension UISearchBar {
         
         return ControlProperty(values: source, valueSink: bindingObserver)
     }
-}
-
+    
+    /**
+     Reactive wrapper for `searchBarSearchButtonClicked:` delegate event.
+     */
+    public var rx_searchButtonClicked: ControlEvent<UISearchBar> {
```
kzaher: We would want to have `Void` instead of `UISearchBar` since this models an event.

RxCocoa/iOS/UISearchBar+Rx.swift
```diff
@@ -47,6 +56,36 @@ extension UISearchBar {
...
+        let source: Observable<UISearchBar> = Observable.deferred { [weak self] () -> Observable<UISearchBar> in
+            return (self?.rx_delegate.observe("searchBarSearchButtonClicked:") ?? Observable.empty())
+                .flatMap { a -> Observable<UISearchBar> in
+                    let result = a.first.flatMap { $0 as? UISearchBar }.flatMap { Observable.just($0) }
+                    return result ?? Observable.empty()
+            }
+        }
```
kzaher: We should be able to replace this implementation with just:
```
self.rx_delegate.observe("searchBarSearchButtonClicked:")
         .map { _ in
              return ()
         }
```

RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
```diff
@@ -44,9 +44,13 @@ class WikipediaSearchViewController: ViewController {
 
         let API = DefaultWikipediaAPI.sharedAPI
 
-        searchBar.rx_text
-            .asDriver()
+        let searchText = searchBar.rx_text.asDriver()
             .throttle(0.3)
+        let searchTapped = searchBar.rx_searchButtonClicked.asDriver()
+            .flatMap { $0.text.flatMap { Driver.just($0) } ?? Driver.empty() }
```
kzaher: Why do we need search tapped event here, shouldn't `searchText` already contain current search text value?

fjcaetano: You're right, we don't. I just added it to add an usage example, but I'll remove it.

PR #506, opened 2016-02-24
Added Rx wrapper for UIPageControl currentPage and numberOfPages properties

anuragajwani:


kzaher:
Hi @anuragajwani ,

I think we could include these extensions if they were slightly modified.

I believe these two properties should have implementation, public interface and unit tests equivalent to `rx_hidden`. 

If user can't change these values directly by interaction, but only programmatically, then I don't think observing them should be allowed.

PR #509, opened 2016-02-25
Make name parameter of rx_notification optional.

tomburns:
`NSNotificationCenter` allows observation without specifying a notification name via `addObserverForName`; this PR just promotes that optionality up to `rx_notification` as well.


tomburns:
Sample use cases for this include observing notifications from an NSMetadataQuery, where you supply the query as `object` without a specific notification name in order to observe all notifications related to that query.

kzaher:
Hi @tomburns ,

it looks like you are right, name is optional parameter.

PR #512, opened 2016-02-26
Initial support for scope button index changes in UISearchBar

stack:
I'll admit I'm very new to RxSwift, but the change seemed trivial. Any suggestions for improvement are welcome.

This should monitor the `selectedScopeButtonIndex` property of `UISearchBar`

### DIFF
--- a/RxCocoa/iOS/UISearchBar+Rx.swift
+++ b/RxCocoa/iOS/UISearchBar+Rx.swift
@@ -47,6 +47,27 @@ extension UISearchBar {
         
         return ControlProperty(values: source, valueSink: bindingObserver)
     }
+    
+    /**
+    Reactive wrapper for `scope` property.
+    */
+    public var rx_scope: ControlProperty<Int> {
+        let source: Observable<Int> = Observable.deferred { [weak self] () -> Observable<Int> in
+            let scope = self?.selectedScopeButtonIndex ?? 0
+            
+            return (self?.rx_delegate.observe("searchBar:selectedScopeButtonIndexDidChange:") ?? Observable.empty())
+                .map { a in
+                    return a[1] as? Int ?? 0
+                }
+                .startWith(scope)
+        }
+        
+        let bindingObserver = UIBindingObserver(UIElement: self) { (searchBar, scope: Int) in
+            searchBar.selectedScopeButtonIndex = scope
+        }
+        
+        return ControlProperty(values: source, valueSink: bindingObserver)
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @stack ,

I think that we could wrap that property also. 

If you could target this PR to `develop` branch and make the changes I'm about to suggest, I would be happy to merge this.

RxCocoa/iOS/UISearchBar+Rx.swift
```diff
@@ -47,6 +47,27 @@ extension UISearchBar {
         
         return ControlProperty(values: source, valueSink: bindingObserver)
     }
+    
+    /**
+    Reactive wrapper for `scope` property.
+    */
+    public var rx_scope: ControlProperty<Int> {
```
kzaher: I think for consistency sake, we should name this property, `rx_selectedScopeButtonIndex`.

RxCocoa/iOS/UISearchBar+Rx.swift
```diff
@@ -47,6 +47,27 @@ extension UISearchBar {
...
+        let source: Observable<Int> = Observable.deferred { [weak self] () -> Observable<Int> in
+            let scope = self?.selectedScopeButtonIndex ?? 0
+            
+            return (self?.rx_delegate.observe("searchBar:selectedScopeButtonIndexDidChange:") ?? Observable.empty())
+                .map { a in
+                    return a[1] as? Int ?? 0
```
kzaher: I think this should be something like 
```
let scopeIndex = try castOrThrow(NSNumber.self, a[1])
return scopeIndex.integerValue
```

kzaher: I'll also need to apply that `castOrThrow` to other parts of the library for consistency sake.

stack: Would this make sense since the number is actually just an NSInteger and not an NSNumber?

stack:
I'll be able to make these changes a little later today and I'll submit a new pull request. Thanks for your input.

kzaher:
Ok, cool,

We should also add unit tests for this extension like we have for other extensions, although there is a chance we've missed some, I now see not all our extensions have all necessary tests, argh ...

I'll add the necessary missing tests today/tomorrow, so you, or somebody else can add the same for this extension.

stack:
Closing in favor of pull request #515

PR #515, opened 2016-02-27
Support for UISearchBar selectedScopeButtonIndex

stack:
This is a cleaned up version of Pull Request #512. The attribute name has been updated and the value resolution now uses `castOrThrow`.


kzaher:
Hi @stack ,

I've improved unit tests for `UISearchBar`. I've pushed changes to `develop` branch. Class that contains tests is called `UISearchBarTests`.

That class currently contains tests for `rx_text` property. If you could add analogous tests for  `rx_selectedScopeButtonIndex` I would be happy to merge this.

stack:
I've added the tests. They're modeled after the ones you added. Thanks for that. 

I also fixed a typo in one of the test names.

PR #523, opened 2016-03-03
Fixed Grammar mistake then/than

afonsograca:
"Then" was being used as the comparison connector of some sentences. The correct word usage is "than". Also fixed an issue of number - "is" being used instead of "are"

### DIFF
--- a/Documentation/GettingStarted.md
+++ b/Documentation/GettingStarted.md
@@ -481,7 +481,7 @@ Disposed
 Ended ----
 ```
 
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @afonsograca ,

thnx for noticing this, if you could just target this PR to `develop` branch, I would be happy to merge this :)

afonsograca:
Will do. Closing this PR now and mentioning it in the new one.

PR #524, opened 2016-03-04
Update UIApplication+Rx.swift to fix mismatch #if statement

lamson:
Carthage build failed due to this issue.

`MyProject/Carthage/Checkouts/RxSwift/RxCocoa/iOS/UIApplication+Rx.swift:29:1: error: expected expression
MyProject/Carthage/Checkouts/RxSwift/RxCocoa/iOS/UIApplication+Rx.swift:29:7: error: expected #else or #endif at end of configuration block
A shell task failed with exit code 65:
** BUILD FAILED **`

### DIFF
--- a/RxCocoa/iOS/UIApplication+Rx.swift
+++ b/RxCocoa/iOS/UIApplication+Rx.swift
@@ -10,6 +10,7 @@ import Foundation
 
 #if os(iOS)
     import UIKit
+#endif
     
 #if !RX_NO_MODULE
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @lamson ,

I now know what happened, there wasn't a newline after `#endif`, that's why it failed with older Xcode versions.

I think we've already pulled those changes to `develop` branch, just didn't release new version, will do that soon.

PR #527, opened 2016-03-04
Adds 'rx_badgeValue' to UITabBarItem

ertrzyiks:
Hi,

i've added a property which allows to write `badgeValue` of `UITabBarItem` with `Observable` on iOS and tvOS

### DIFF
--- a/Documentation/API.md
+++ b/Documentation/API.md
@@ -455,3 +455,11 @@ extension NSTextField {
 
 }
 ```
+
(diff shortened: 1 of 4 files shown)


kzaher:
Hi @ertrzyiks ,

I'm fine with pulling this in. There is a small mistake with

`Bindable sink for `hidden` property.`, but I can fix that tiny thing :)

PR #539, opened 2016-03-10
Add rx_editing property in UITextField

chrisamanse:
Add `rx_editing` property for `UITextField`, a reactive wrapper for its `editing` property.

### DIFF
--- a/Documentation/API.md
+++ b/Documentation/API.md
@@ -229,6 +229,7 @@ extension UIButton {
 extension UITextField {
 
     public var rx_text: ControlProperty<String> {}
+    public var rx_editing: ControlProperty<Bool> {}
 
 }
 ```
--- a/RxCocoa/iOS/UITextField+Rx.swift
+++ b/RxCocoa/iOS/UITextField+Rx.swift
@@ -30,6 +30,23 @@ extension UITextField {
         )
     }
     
+    /**
+    Reactive wrapper for `editing` property.
+    */
+    public var rx_editing: ControlProperty<Bool> {
+        return UIControl.rx_value(
+            self,
+            getter: { textField in
+                textField.editing
+            }, setter: { textField, value in
+                if value {
+                    textField.becomeFirstResponder()
+                } else {
+                    textField.resignFirstResponder()
+                }
+            }
+        )
+    }
 }
 
 #endif


kzaher:
Hi @chrisamanse ,

I think that for now it would be best if we didn't include these extensions in the main repo for now :)

There are multiple reasons for that:
* `UIControl.rx_value` won't work with textField.editing because that's not a control value wrapper
* `firstResponder` seems to me like a concept that is more of a singleton nature. We could wrap it somehow, but I would like to think more about this and maybe find some more convenient way to express it.

chrisamanse:
Ok. For now, what should I use if I want to observe if a `UITextField` is editing and combine it with another Observable? Currently, I use this setup:

```swift
// Create and editing 
var editingTitleTextField = Variable(false)

titleTextField.rx_controlEvent(.EditingDidBegin)
    .map {
        return true
    }
    .bindTo(editingTitleTextField)
    .addDisposableTo(disposeBag)
titleTextField.rx_controlEvent(.EditingDidEnd)
    .map {
        return false
    }
    .bindTo(editingTitleTextField)
    .addDisposableTo(disposeBag)

// Change right bar button item depending on editing and validity
[editingTitleTextField.asObservable(), viewModel.validTaskTitle.asObservable()]
    .combineLatest { values -> BarButtonType in
        let editing = values[0]
        let valid = values[1]
        
        switch (editing, valid) {
        case (true, true):
            return .Add // BarButtonType enum
        case (true, false):
            return .Done
        default:
            return .None
        }
    }
    .distinctUntilChanged()
    .subscribeNext {
        let currentBarButton: UIBarButtonItem?
        switch $0 {
        case .Add:
            currentBarButton = self.addBarButton
        case .Done:
            currentBarButton = self.doneBarButton
        case .None:
            currentBarButton = self.editButtonItem()
        }
        
        self.navigationItem.rightBarButtonItem = currentBarButton
    }
    .addDisposableTo(disposeBag)

```

Is there a shorter way to do this?

kzaher:
Hi @chrisamanse,

This is how you can write this. It looks little verbose, but it's quite simple 

```swift
extension UIControl {
    var rx_editing: Observable<Bool> {
        return Observable.deferred { [weak self] in
            return self.map { responder in
                return Observable.of(
                        responder.rx_controlEvent(UIControlEvents.EditingDidBegin).map { _ in true },
                        responder.rx_controlEvent(UIControlEvents.EditingDidEnd).map { _ in false }
                    )
                    .merge()
                    .startWith(responder.isFirstResponder())
            } ?? Observable.empty()
        }
    }
}
```

chrisamanse:
@kzaher Thanks! It works perfectly.

PR #549, opened 2016-03-14
UISearchBar Cancel Event

stack:
I've added a ControlEvent for the UISearchBar's Cancel button, which is observed from the UISearchBarDelegate method `searchBarCancelButtonClicked:`


stack:
Apologies. Wrong branch.

PR #550, opened 2016-03-14
UISearchBar Cancel Event

stack:
I've added a ControlEvent for the UISearchBar's Cancel button, which is observed from the UISearchBarDelegate method `searchBarCancelButtonClicked:`


kzaher:
Hi @stack , @fjcaetano,

I think we should somehow unify this PR with https://github.com/ReactiveX/RxSwift/pull/504.

This PR contains the exact form for `rx_cancelTap`, and #504 contains delegate tests an `rx_searchButtonClicked`, but `rx_searchButtonClicked` should have implementation similar to `rx_cancelTap` from this PR :)

Not sure what to do here :)

stack:
I brought over the `rx_createDelegateProxy` method and tests from @fjcaetano, as well as the `rx_searchButtonClicked` method, only renamed to `rx_searchTap`

I've also upgraded the selectors in the UISearchBar extensions to the new Swift 2.2 selectors.

stack:
Looks like the new Swift 2.2 selectors are doing their job. `UISearchBarDelegate` does not have `searchBarCancelButtonClicked` on tvOS.

stack:
Some notes for tvOS:

- `searchBarCancelButtonClicked` does not exist in tvOS, so that has been skipped via precompiler macros.
- tvOS does not allow the instantiation of UISearchBar directly, which also means you cannot subclass it. I've skipped the subclass `rx_createDelegateProxy` method and tests on tvOS. See (https://gist.github.com/gonzalolarralde/fb9bd2b164e88a287ad8) for what is prohibited in tvOS.

RxCocoa/iOS/Proxies/RxSearchBarDelegateProxy.swift
```diff
@@ -14,19 +14,33 @@ import UIKit
...
     
-    class func setCurrentDelegate(delegate: AnyObject?, toObject object: AnyObject) {
+    public class func setCurrentDelegate(delegate: AnyObject?, toObject object: AnyObject) {
         let searchBar: UISearchBar = castOrFatalError(object)
         searchBar.delegate = castOptionalOrFatalError(delegate)
     }
+
+    // MARK: Delegate proxy methods
+    
+#if os(iOS)
```
kzaher: Hi,

I'm little confused, why is this specific to iOS only?

RxCocoa/iOS/UISearchBar+Rx.swift
```diff
@@ -68,6 +79,30 @@ extension UISearchBar {
         
         return ControlProperty(values: source, valueSink: bindingObserver)
     }
+    
+#if os(iOS)
+    /**
+    Reactive wrapper for delegate method `searchBarCancelButtonClicked`.
+    */
+    public var rx_cancelTap: ControlEvent<Void> {
```
kzaher: This should probably be `rx_cancelButtonClicked`

RxCocoa/iOS/UISearchBar+Rx.swift
```diff
@@ -68,6 +79,30 @@ extension UISearchBar {
...
+                return ()
+            }
+        return ControlEvent(events: source)
+    }
+#endif
+    
+    /**
+     Reactive wrapper for delegate method `searchBarSearchButtonClicked`.
+     */
+    public var rx_searchTap: ControlEvent<Void> {
```
kzaher: This should probably be `rx_searchButtonClicked`.

RxCocoa/iOS/Proxies/RxSearchBarDelegateProxy.swift
stack: UISearchBar cannot be directly instantiated on tvOS. This makes it un-subclassable. You can't call super.init(). In theory it could be available to tvOS, but it's untestable in the current methods we use to test.

RxCocoa/iOS/UISearchBar+Rx.swift
stack: Just checked this in.

RxCocoa/iOS/UISearchBar+Rx.swift
stack: Just checked this in

PR #559, opened 2016-03-19
Expanded on example operator to demonstrate how one could create a two-way binding between dissimilar types

justinswart:
I think this is a useful example for new users of the framework

### DIFF
--- a/RxExample/RxExample/Operators.swift
+++ b/RxExample/RxExample/Operators.swift
@@ -17,6 +17,8 @@ import RxCocoa
 infix operator <-> {
 }
 
+// Usage: textField.rx_text <-> viewModel.variable
+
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @justinswart ,

We've kind of added that two way binding because people were asking all the time how to do it :), but I don't think we should create a lot of examples with two way bindings because we would like to present concept in as pure and simple form as possible (unidirectional).

There will be probably a lot of small extensions people will create, but I'm afraid that including them in this repro would probably be too confusing for beginners.

¯\\\_(ツ)\_/¯

justinswart:
Understood and no problem :)  It was just something I found useful in my project with RxSwift, I'm sure someone looking for a similar solution will land up here via a Google search anyway!

Issue #472, opened 2016-02-04
RxTest fails to build in Xcode 7.3-beta2


koke:
Running the test suite (or using the RxTest pod) throws a couple build errors when using Xcode 7.3 beta 2

> XCTest+Rx.swift:67:7: Cannot invoke 'XCTAssertEqual' with an argument list of type '([AnyEquatable<Event<T>>], [AnyEquatable<Event<T>>], file: String, line: Uint)'
> XCTest+Rx.swift:95:7: Cannot invoke 'XCTAssertEqual' with an argument list of type '([AnyEquatable<Recorded<Event<T>>>], [AnyEquatable<Recorded<Event<T>>>], file: String, line: Uint)'

It seems the `file` parameter on `XCTAssertEqual` changed from `String` to `StaticString` in https://github.com/apple/swift/pull/888

koke:
I got a fix in https://github.com/ReactiveX/RxSwift/commit/60b4b965092c372edce7b18ed67f863d3a7a3e8b, but that breaks the builds on Xcode 7.2. I'm not sure if there's a workaround or this will have to be a new release that supports Swift 2.2 only.

kzaher:
Hi @koke ,

Thnx for reporting this. I'm hoping this is just an Apple but and they will fix this soon :(. 

I'm not sure it would be a good idea to pull this in immediately since it you are telling me this will break on Xcode 7.2, and this is the released version. Let's monitor this and see what will Apple do here. 

If they release a version that requires `StaticString`, then we'll make the same breaking change.

Bogidon:
Hi @koke and @kzaher,

How can I use 60b4b96 with Carthage? Is it in a branch? If not, could we create a branch called, for example, `Xcode7.3`?

kzaher:
Hi @Bogidon ,

I've created new branch xcode/7.3 If @koke makes PR to that branch, I would be happy to merge it.

tomburns:
@kzaher the xcode/7.3 branch gets further into the build process, but still fails:

```
/Users/.../RxSwift/RxSwift/DataStructures/Queue.swift:165:16: warning: 'anyGenerator' is deprecated: renamed to 'AnyGenerator'
/Users/.../RxSwift/RxSwift/DataStructures/InfiniteSequence.swift:26:16: warning: 'anyGenerator' is deprecated: renamed to 'AnyGenerator'
<unknown>:0: error: unable to execute command: Segmentation fault: 11
[...]

PR #565, opened 2016-03-22
Workaround for segmentation fault in Xcode 7.3

ishkawa:
Unfortunately, the compiler issue mentioned in #472 is not fixed in Swift 2.2.
This patch is just a workaround for the issue :pray:


koke:
:+1: Seems to fix the crash in Xcode 7.3 final

PR #567, opened 2016-03-22
Fixes for Xcode 7.3

koke:


kzaher:
I think we've talked around this in Slack while we were fixing 7.3 ....

PR #582, opened 2016-03-28
Documentation for skipUntil corrected

kkdevenda:
It seems a copy paste error as documentation for both `skipUntil` and `takeUntil` was same.
I have corrected the documentation for `skipUntil` keeping `takeUntil` intact.


kzaher:
Hi @kkdevenda ,

good catch :+1:

PR #583, opened 2016-03-29
Adding doOnSubscribe/doOnUnsubscribe Operators to ObservableType

rkawajiri:


kzaher:
Hi @rkawajiri ,

I think we should add those functionalities as additional optional closures to `doOn` operator.

We should probably name it `onDispose` instead of `onUnsubscribe` because I believe Java has `unsubscribe` that is equivalent to `dispose`. 

I don't think that we should allow for `onDispose` to throw exceptions because that would be like throwing exception from a destructor in C++, or decorating `deinit` with `throws` in Swift.

Being able to throw errors from `onDispose` can also result with this, which violates sequence grammar :)
```
+            next(240, 5),
+            completed(250),
+            error(1000, testError)
```

I think we should also test for one important property. Are `onSubscribe` and `onDispose` called before or after propagating calls to source sequence.

`onSubscribe` should be probably called before propagating subscription, and `onDispose` should be called after the source subscription is disposed, but I would need to think about this some more, because there are basically 4 cases :)

* before subscribe
* after subscribe
* before dispose
* after dispose 

I usually use `using` operator, so I'm not sure in what context would one use these extensions.

Kacper20:
Hello,

I have case in which this is needed(at least i think it is).
Here's sample of code were using in pre-release of our bluetooth rx library:
```swift
        let observable = peripheral.rx_didReadRSSI
            .take(1)
            .flatMap { (rssi, error) -> Observable<(Peripheral, Int)> in
                if let error = error {
                    return Observable.error(BluetoothError.PeripheralRSSIReadFailed(self, error))
                }
                return Observable.just(self, rssi)
            }
        return Observable.deferred {
            self.peripheral.readRSSI()
            return observable
        }
```
Writing this code like this creates possible race condition (emits event earlier than subscribtion is made).
It would be ideal to place `self.peripheral.readRSSI()` in `doOnSubscribe` block. 
Or maybe I'm missing something and there is a better way to what I want to achive?

Thanks very much for response.

marksands:
When will these operators be merged? @kzaher last I heard from you in the Slack channel you said they would be out with 2.6.0 😅

kzaher:
Hi @marksands ,

yeah man, I've really meant to :( On 2.6.0 release I've again lost so much time fixing compilation issues and making sure everything works with new version on Swift.

I've basically been up all night from Sunday to Monday solving compilation issues because there are so many combinations we of different iOSes, integration methods and Swift versions we need to support, and then there is parallel Swift 3.0 branch now and maintaining some of the RxSwiftCommunity projects :(

I've tried to explain problems with this PR in my last comments. The bottom line is that I'll need to think a bit how to best implement this in the spirit of other interfaces we have and I'll need to dig through RxJava probably, it's not just hitting merge button :)

I'll need to find some breathing room to think how to integrate this best. Hoping I'll catch some decent PTO time to think about this rested.
Shipping new functionality in hurry is not something I would want to do with this project :) Even when one is extremely careful, there are oversights and suboptimal decisions.

 ¯\\\_(ツ)\_/¯

kzaher:
We've incorporated `onSubscribe` and on `onDispose` with `do` operator inside `swift-3.0` branch.

PR #596, opened 2016-04-04
remove redundant state property of the PublishSubject

53ningen:
As near as I can tell, the state property `_stopped` of the `PublishSubject` class is redundant.
https://github.com/ReactiveX/RxSwift/pull/596/files#diff-53af0ab0a90f54260431a581ecb1822fL31

So I remove that property and modified `PublishSubject` implementation (like `BehaviorSubject`).
cf.) https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Subjects/BehaviorSubject.swift#L81

### DIFF
--- a/RxSwift/Subjects/PublishSubject.swift
+++ b/RxSwift/Subjects/PublishSubject.swift
@@ -28,7 +28,6 @@ final public class PublishSubject<Element>
     // state
     private var _disposed = false
     private var _observers = Bag<AnyObserver<Element>>()
-    private var _stopped = false
     private var _stoppedEvent = nil as Event<Element>?
     
     /**
@@ -58,7 +57,7 @@ final public class PublishSubject<Element>
     func _synchronized_on(event: Event<E>) {
         switch event {
         case .Next(_):
-            if _disposed || _stopped {
+            if _disposed || _stoppedEvent != nil {
                 return
             }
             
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @53ningen ,

you are correct, technically it's redundant, but it's performance optimization. That `_stoppedEvent != nil` is on a hot path for each `Next` element, and for some reason the same check for enum was extremely slower then just checking a boolean.

I think we could maybe run that profiler on Swift 3.0 and see where we are then. You are correct, it is confusing, and I should have made a comment here regarding why is that boolean there.

You are also right that it's inconsistent with `BehaviorSubject`. I should have then made the same optimizations on other subjects.

I would like to revise those optimizations for RxSwift 3.0 in a couple of months since we are going to do profiling again :)

Thnx for doing a CR.

53ningen:
@kzaher 
thank you for reviewing! :)

> you are correct, technically it's redundant, but it's performance optimization. That _stoppedEvent != nil is on a hot path for each Next element, and for some reason the same check for enum was extremely slower then just checking a boolean.

I didn't know that! thank you for kindly teaching me 🙇  


> you are correct, technically it's redundant, but it's performance optimization. That _stoppedEvent != nil is on a hot path for each Next element, and for some reason the same check for enum was extremely slower then just checking a boolean.

I know you are very busy, but when you have time, could you tell me how to check it.

thanks :)

PR #614, opened 2016-04-12
add rx_disposableBag

kingslay:
add rx_disposableBag

### DIFF
--- a/RxCocoa/Common/Observables/NSObject+Rx.swift
+++ b/RxCocoa/Common/Observables/NSObject+Rx.swift
@@ -14,8 +14,6 @@ import RxSwift
 #if !DISABLE_SWIZZLING
 var deallocatingSubjectTriggerContext: UInt8 = 0
 var deallocatingSubjectContext: UInt8 = 0
-var disposableBagContext: UInt8 = 0
-
 #endif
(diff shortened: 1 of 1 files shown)


sergdort:
@kingslay You probably can use [NSObject-Rx](https://github.com/RxSwiftCommunity/NSObject-Rx)

kingslay:
@sergdort but I think if the RxSwift can support rx_disposableBag is better

kzaher:
Hi @kingslay ,

I understand why do you want this to be merged, but I'm not sure that would be a good idea. 
I've thought about this before, and I didn't feel this functionality is well defined.

There is a also a lot of doubts how this should behave regarding thread safety and if you add this to `NSObject` as a public extension, then some class could reference that disposeBag externally and cause subscriptions to not be disposed.

This patternn is usually used from main thread, so the choice of synchronized vs assert accessing from main thread is also unclear to me, as is it's proper name. Should it be `rx_disposeBag`, `rx_deallocBag`, `rx_disposingBag`, etc ...  

There is a lot of convenient way of using and clearing subscriptions in a `disposeBag` and I wouldn't want to introduce magic behavior where it's not necessary.

I think it would be best not to include this for now.

PR #615, opened 2016-04-12
remove public access modifier from Queue<T>

53ningen:
Because RxSwift is not a collection library, I think `Queue<T>` is no need to be public.


kzaher:
Hi @53ningen ,

yep, yep, you are correct. Queue should be internal. I can't merge this to `develop` branch since these changes could cause backwards compatibility breakage.

I would like to include this in RxSwift 3.0 version. We are slowly adapting project for Swift 3.0 and Linux in `swift-3.0` branch, so if you could target this to that branch, I would be happy to merge this.

53ningen:
Thank you for taking your time to look at this!

I should have consulted with you before creating PR :l 
I created new PR #624. Closing this.

thanks :)

PR #617, opened 2016-04-13
Use guard ... _generators.last instead of count check and last!

yury:


kzaher:
Weird, it looks like something is broken, but it shouldn't be, I'll investigate this ...

PR #621, opened 2016-04-15
Uipagecontrol add rx

punty:
Add rx_currentPage property


kzaher:
Hi @punty ,

If you could add unit tests like we have for `rx_enabled` I would be happy to merge this.

punty:
I added the unit test. All checks have passed

PR #625, opened 2016-04-17
Added distinctUntilChanged operator for sequence that contains optional values

pikciu:
- added OptionalType protocol
- added distinctUntilChanged based on OptionalType
- added test for OptionalType - compare values

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -7,6 +7,10 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


kzaher:
Hi @pikciu,

Thnx for your PR. I think it would be best if we would include it in RxOptional maybe.

https://github.com/RxSwiftCommunity/RxOptional

If we go this road, I'm afraid we'll bloat the codebase. Not with this inclusion particular, but with all other ones equivalent ones that will be proposed.

Hope that's ok with you.

pikciu:
You are right. I am sorry to speak but I didn't know about RxOptional xD It's awesome! We can close this PR here ;)

PR #631, opened 2016-04-21
Solve ios example warning

Calvin-Huang:
Hello! I found there are some warning in storyboard then I try to solve it.
And remove UISearchDisplayController which was deprecated from iOS 8.0.

### DIFF
--- a/RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
+++ b/RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
@@ -13,13 +13,20 @@ import RxCocoa
 #endif
 
 class WikipediaSearchViewController: ViewController {
+    @IBOutlet var searchBarContainer: UIView!
+    
+    private let searchController = UISearchController(searchResultsController: UITableViewController())
+    
+    private var resultsViewController: UITableViewController {
+        return (self.searchController.searchResultsController as? UITableViewController)!
+    }
     
     private var resultsTableView: UITableView {
-        return self.searchDisplayController!.searchResultsTableView
+        return self.resultsViewController.tableView!
     }
 
     private var searchBar: UISearchBar {
-        return self.searchDisplayController!.searchBar
+        return self.searchController.searchBar
     }
 
     override func awakeFromNib() {
@@ -30,6 +37,13 @@ class WikipediaSearchViewController: ViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
+        
+        let searchBar = self.searchBar
+        let searchBarContainer = self.searchBarContainer
+        
+        searchBarContainer.addSubview(searchBar)
+        searchBar.frame = searchBarContainer.bounds
+        searchBar.autoresizingMask = .FlexibleWidth
         
         configureTableDataSource()
         configureKeyboardDismissesOnScroll()
--- a/RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
+++ b/RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
@@ -80,9 +80,13 @@ class WikipediaSearchViewController: ViewController {
 
     func configureKeyboardDismissesOnScroll() {
         let searchBar = self.searchBar
-
+        let searchController = self.searchController
+        
         resultsTableView.rx_contentOffset
             .asDriver()
+            .filter { _ -> Bool in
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @Calvin-Huang ,

Thnx for trying to solve those warnings. I have a couple of remarks, but if we solve them and target this PR to `develop` branch, I would be happy to merge this.

RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
```diff
@@ -13,15 +13,10 @@ import RxCocoa
 #endif
 
 class WikipediaSearchViewController: ViewController {
+    private let searchController = UISearchController(searchResultsController: UITableViewController())
+    
+    @IBOutlet weak var searchBarContainer: UIView!
     
-    private var resultsTableView: UITableView {
```
kzaher: Is there some reason why we are removing `resultsTableView` and `searchBar ` properties instead of just saying 
```
private var resultsViewController: UITableViewController {
     return (self.searchController.searchResultsController as?  UITableViewController)!
}
private var resultsTableView: UITableView {
        return resultsViewController.tableView!
}
private var searchBar: UISearchBar {
       return self.searchController.searchBar
}
```

I don't like the `!`, but it seems that we are doing that in the code anyway.

RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
```diff
@@ -58,36 +66,12 @@ class WikipediaSearchViewController: ViewController {
...
                 results.map(SearchResultViewModel.init)
             }
-            .drive(resultsTableView.rx_itemsWithCellIdentifier("WikipediaSearchCell", cellType: WikipediaSearchCell.self)) { (_, viewModel, cell) in
+            .drive(resultController.tableView!.rx_itemsWithCellIdentifier("WikipediaSearchCell", cellType: WikipediaSearchCell.self)) { (_, viewModel, cell) in
                 cell.viewModel = viewModel
             }
             .addDisposableTo(disposeBag)
     }
 
-    func configureKeyboardDismissesOnScroll() {
```
kzaher: Is there some reason why we deleted `configureKeyboardDismissesOnScroll` and `configureNavigateOnRowClick `?

RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
Calvin-Huang: If we add some variable for having more convenience to access properties, when the class grows up, next person who needs to maintain the project probably doesn't know there was some variables can be use.
So I prefer to eliminate adding some variables like this.

However, indeeds that is not the thing which must be done and there are exist code in here, I should follow the codes to minimum changes(If I wants to put these parts back, should I rebase? or just add some commits).

And after using UISearchController instead, SearchResultController had been setup as UISearchController constructed.
At beginning of `configureResultController ` method, it just ensure SearchResultController is UITableViewColler.
Then there must be a tableView below UITableViewController so I used the `!`.

RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
Calvin-Huang: Oh...Sorry, I didn't notice what `configureKeyboardDismissesOnScroll ` does.
It will close keyboard when users want to scroll down and check the search results.
This function must exists!

And...`configureNavigateOnRowClick ` method add events to open the search results, I remove this method and combine function into `configureResultController ` because I removed the instance variable `resultsTableView ` and setup actions should be placed at same place.

If we decide to put `resultsTableView ` back, `configureNavigateOnRowClick ` and `configureKeyboardDismissesOnScroll ` need to put back, too.

RxExample/RxExample/Examples/WikipediaImageSearch/Views/WikipediaSearchViewController.swift
kzaher: The main reason why I suggest properties is readability. We repeat `resultController.tableView!` in a lot of places instead of just `resultsTableView` :)

Repeating `!` throughout the code also pokes my eyes :)

I'm fine with adding additional commits.

Calvin-Huang:
Hi! @kzaher , I opened a new PR for changing merge target.
https://github.com/ReactiveX/RxSwift/pull/637

PR #636, opened 2016-04-21
Reactive observer that emit items whenever UIScrollView reaches bottom

diegomedina248:
Reactive observable that emit items whenever scroll view contentOffset.y is close to contentSize.height

From [here](https://github.com/tryswift/RxPagination/blob/f85363a3430a98aa45e38f18ff30e7f7a8946771/Pagination/UIScrollView%2BRx.swift)


kzaher:
Hi @diegomedina248 ,

I do agree that this is useful extension, we also have a similar one in RxExample project, but the general rule is that we can't maintain convenience extensions in this repo because it would grow too big.

Hope it's ok with you.

diegomedina248:
@kzaher Agree. Just thought to leave it here in case someone needs it in the future.

PR #638, opened 2016-04-22
Make lastIndexOf search from end of string to start

Nirma:
Thought that this might be a little more straight forward. 

How does it look?


kzaher:
Hi @Nirma ,

It seems to me this is longer that we used to have :), and should it be `while index >= startIndex` instead of `while index > startIndex`.

For this PR it's not clear to me why do you think the proposed solution would be better, could you explain a bit?

kzaher:
Oh, no, my bad, should be `>`, but still don't see the advantage :)

Nirma:
@kzaher Sorry for the nitpick. :P 
I thought that this might be a little more straight forward and there is also a bit of a speed boost!
 
test

PR #652, opened 2016-04-27
UILabel rx_text as `String?`

thanegill:
UILabels text attribute is a Optional String. With `rx_text` reflecting that you can now set a UILabels text back to `nil`, which is the default value. 😋


kzaher:
Hey @thanegill ,

I think you are correct, but this is a breaking change. I would like to include this for 3.0, so if you could point this to `swift-3.0` branch, I would be happy to merge this.

PR #656, opened 2016-04-30
Added hasObservers on each of the Subject classes

lizixroy:
The reason that I need hasObservers property on Subject type is the following:

Here at my company we need to implement hash table whose key is a String and value is BehaviorSubject<Something>. In order to keep memory usage in check, we need to clean up the hash table by periodically removing BehaviorSubjects that do not have any subscriptions. Therefore, we need a way to determine whether a BehaviorSubject has subscribers or not.

I noticed in RxJava we do have hasObservers at our disposal. So I basically ported this feature to RxSwift.

Details:

Added var hasObservers: Bool { get } to the SubjectType protocol.

Then made all the Subject classes conform to the protocol by providing
hasObservers as a computed property.

### DIFF
--- a/RxSwift/Subjects/BehaviorSubject.swift
+++ b/RxSwift/Subjects/BehaviorSubject.swift
@@ -22,6 +22,9 @@ public final class BehaviorSubject<Element>
     public typealias SubjectObserverType = BehaviorSubject<Element>
     typealias DisposeKey = Bag<AnyObserver<Element>>.KeyType
     
+    /**
+     Indicates whether the subject has any observers
+     */
     public var hasObservers: Bool {
         return _observers.count > 0
     }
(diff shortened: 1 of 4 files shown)


lizixroy:
Travis failed because of following reason:
ERROR: line `//  Copyright © 2016 kzaher. All rights reserved.` is invalid: //  Copyright © (\d+) Krunoslav Zaher. All rights reserved.

This line in my file says: "//  Copyright © 2015 Krunoslav Zaher. All rights reserved."

Do you need me to change 2015 to 2016?

kzaher:
Hi @lizixroy ,

I think it's failing with different error now. I guess we could add this for consistency sake, but I would advise against those kinds of checks because of possible races.

To merge this, we would need to fix build server error, and add unit tests, add comments, and use locks because accessing count isn't thread safe.

```
public var hasObservers: Bool {
    return _observers.count > 0
}
```

lizixroy:
@kzaher I have added comments, also added locks around  _observers.count to ensure thread safety as you suggested. 

Can you give me some suggestions about unit test? Where should I put test? How do you want it to be tested?

Thanks for your time!

Roy

lizixroy:
@kzaher When running any unit tests in the RxSwiftTests/Tests folder I got this error from Xcode: "The run destination iPhone 6s is not valid for tests you have chosen to perform.". What should I do to run these tests?

PR #659, opened 2016-05-02
add private to init of singleton

futantan:


kzaher:
Hi @futantan ,

if you could target this do `develop` branch, I would be happy to merge this.

PR #661, opened 2016-05-03
add private to init of singleton

futantan:


kzaher:
Hi @futantan ,

I think unit tests are showing that it's being used somewhere else :)

PR #664, opened 2016-05-04
Use unbounded replay subject in the Playground

fastred:
Before this change the output was as follows:

```
--- ReplaySubject example ---
Subscription: 1, event: Next(a)
Subscription: 1, event: Next(b)
Subscription: 2, event: Next(b)
Subscription: 1, event: Next(c)
Subscription: 2, event: Next(c)
Subscription: 1, event: Next(d)
Subscription: 2, event: Next(d)
```

It didn't match the description of `ReplaySubject`. `Subscription: 2, event: Next(a)` was missing since `bufferSize` was equal to 1. I think that an unbounded replay subject fits better here.

If we don't want to use an unbounded replay subject here, we should change the buffer size to `2`.

### DIFF
--- a/Rx.playground/Pages/Subjects.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Subjects.xcplaygroundpage/Contents.swift
@@ -58,7 +58,7 @@ example("PublishSubject") {
 */
 example("ReplaySubject") {
     let disposeBag = DisposeBag()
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @fastred ,

If you could target changes from this PR to `develop` branch, I would be happy to merge this.

kzaher:
This went stale.

PR #673, opened 2016-05-08
Fix to support iOS 7

acecilia:
By the way, I found [this](https://github.com/kylef/Hyperplay/tree/master/Pods/RxSwift) explanation very useful for adding RxSwift to my iOS7 project. I imagine that is an old file from RxSwift repo, but I could not find it or anything similar in the repo nowadays.


kzaher:
Hi @acecilia ,

yes, we aren't officially advising people to use this library or Swift in general on ios 7 because of past Swift compiler issues :)

But I guess if you are aware of the potential compiler issues, I'm fine with it :)

PR #680, opened 2016-05-13
Add bindable sinks for UITabBar properties

solidcell:


RxCocoa/iOS/UITabBar+Rx.swift
```diff
@@ -56,6 +56,17 @@ extension UITabBar {
     }
     
     /**
+     Bindable sink for `items` property.
+    
+     - parameter animated: A Boolean indicating whether changes should be animated.
+    */
+    public func rx_items(animated: Bool) -> AnyObserver<[UITabBarItem]?> {
```
solidcell: I'm wondering what others think about this signature.  I don't think it would be possible to pass this function around, since it overlaps with the `var rx_items`, however, I don't know if that's necessary or even helpful.

I tried to find as similar of a signature already in the codebase as possible, which was: https://github.com/ReactiveX/RxSwift/blob/2aebe14/RxCocoa/OSX/NSImageView+Rx.swift#L29

kzaher:
Hi @solidcell ,

sorry for not answering sooner. In general we can't wrap all properties out there, especially style properties, so we try to only wrap the most basic ones.

I'll need to document this somewhere, so I think this PR is a good reminder for that. 

What we do need a delegate wrapper (`UITabBarDelegateProxy`) for `UITabBarDelegate` in case you want to help us.

solidcell:
Ah I understand. I also have a branch with a delegate proxy for that so
I'll create a PR.

Curious, is it so that it doesn't become a maintenance burden?

solidcell:
#717 Has the delegate proxy.

As for documentation, I think that would be a good idea. I'd also like to see the documentation show priorities for things to work on so that we could focus efforts on the right things.  I'll open an issue for that so we can discuss ideas and format in one place.

solidcell:
#718 is an issue regarding contribution docs.

kzaher:
Yeah, I think we talked about this. This is related with #718. We can't add wrappers for every UIKit property, we'll need to asses this on case by case basis.

solidcell:
Yea, besides your [comment above](https://github.com/ReactiveX/RxSwift/pull/680#issuecomment-222252471), I think we discussed it over Slack.

PR #688, opened 2016-05-19
Added hasObservers on each of the Subject classes + unit tests (proper branch)

pRizz:
This is a continuation of this PR with unit tests: #656 
I communicated with that author, @lizixroy, on this work.

Here is a quote of the original reason for the PR:

> The reason that I need hasObservers property on Subject type is the following:
> 
> Here at my company we need to implement hash table whose key is a String and value is BehaviorSubject. In order to keep memory usage in check, we need to clean up the hash table by periodically removing BehaviorSubjects that do not have any subscriptions. Therefore, we need a way to determine whether a BehaviorSubject has subscribers or not.
> 
> I noticed in RxJava we do have hasObservers at our disposal. So I basically ported this feature to RxSwift.
> 
> Details:
> 
> Added var hasObservers: Bool { get } to the SubjectType protocol.
> 
> Then made all the Subject classes conform to the protocol by providing
hasObservers as a computed property.


pRizz:
Bump

kzaher:
Hey,

I will take a look at this, I've seen something I would like to verify before I can merge this.

PR #692, opened 2016-05-19
Fix Blurred demonstration of 'concat demo'. 

wddwycc:
This example of `concat()` cannot show users the effect of `concat()` which the document mentions:
'waiting for each sequence to terminate successfully before emitting elements from the next sequence'
So I fixed it.

### DIFF
--- a/Rx.playground/Pages/Mathematical_and_Aggregate_Operators.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Mathematical_and_Aggregate_Operators.xcplaygroundpage/Contents.swift
@@ -61,6 +61,7 @@ example("concat") {
     
     variable.value = subject2
     
+    subject2.onNext("I would be ignored")
     subject2.onNext("🐱")
     
     subject1.onCompleted()


scotteg:
I don't think this needs to be fixed 😊. These are `BehaviorSubjects`, which emit the previous element upon `subscribe`. The only reason why your added line is ignored is because of the following line. If you removed `subject2.onNext("🐱")`, your line would not be ignored, because it is the latest element on `subject2`, which will be emitted when `subject1` completes.

wddwycc:
@scotteg I know that `BehaviourSubject` has this feature. 
What I mean is the original demo doesn't show the notion that 'subject2 would start its mission right after subject1 completes' . If you remove 'concat()', the output would be the same. 

I add this line to make it clear : 
People knows that if 'concat()' doesn't exist, "I would be ignored" would be sent to subscribers. 
But now, because of 'concat', things change. This event is ignored because subject1 hasn't completed yet.

wddwycc:
BehaviourSubject would  'emit the previous element upon subscribe' is a perception when people look at this demo

scotteg:
Hmm. I don't think removing `concat()` yields the same output:

```
 // with concat()
--- concat example ---
Next(🍎)
Next(🍐)
Next(🍊)
Next(🐱)
Next(🐭)

// without concat()
--- concat example ---
Next(RxSwift.BehaviorSubject<Swift.String>)
Next(RxSwift.BehaviorSubject<Swift.String>)
Completed
```

That said, I think if you wanted to make them `PublishSubject`s--which don't replay--that may be a good improvement here; e.g.,
```
    let subject1 = PublishSubject<String>()
    let subject2 = PublishSubject<String>()
```

The resulting printout would then be:
```
--- concat example ---
Next(🍐)
Next(🍊)
Next(🐭)
```

kzaher:
Hi guys,

I think this example could be much simpler. Instead of focusing on particular events, we should be focusing on high level sequence concepts.

Does anybody see anything confusing about writing it this way?
```swift
example("concat") {
    let disposeBag = DisposeBag()
    
    Observable.of(
            Observable.of("🍎", "🍐"),
            Observable.of("🐶",  "🍊", "🐱")
        )
        .concat()
        .subscribe { print($0) }
        .addDisposableTo(disposeBag)
}
```

wddwycc:
@kzaher Yes, that's much clearer.

scotteg:
@kzaher Sure--and yes, I remember you specifically suggested this before 😃

wddwycc:
@kzaher I did some more test. 

    let disposeBag = DisposeBag()
    
    Observable.of(
        Observable.of("🍎", "🍐"),
        Observable.of("🐶",  "🍊", "🐱")
        )
        .merge()
        .subscribe { print($0) }
        .addDisposableTo(disposeBag)

    --- concat2 example ---
    Next(🍎)
    Next(🍐)
    Next(🐶)
    Next(🍊)
    Next(🐱)
    Completed

If you use merge, the result is the same, which means this is a default behaviour. So, here `concat()` still makes not sense. 

@scotteg When I said `remove concat`, I mean if we don't have the action of 'concat' instead of deleting that line of code. Sorry for my expression. 


So I still think my solution is better.

kzaher:
Hi @wddwycc ,

If you could target changes from this PR to `develop` branch, I would be happy to merge this.

wddwycc:
@kzaher done, see PR #703

PR #699, opened 2016-05-21
Add a code of conduct markdown

solidcell:
Any thoughts on adding a code of conduct markdown?

It still requires the email address to be updated to whomever should be the point of contact.

If it's wanted, I can update it with the correct email address to use.


kzaher:
I guess this looks immutable enough :)

I'll add my email address for now since I'm not sure what would be the other option ¯\\\_(ツ)\_/¯

PR #715, opened 2016-05-28
Add rx_title to UIViewController

kylef:


kzaher:
Hi @kylef ,

If you could add unit tests for this (like for `rx_enabled`) and point this to `develop` branch, I would be happy to merge this.

kylef:
Thanks @kzaher, I'll get to that on Monday. Sorry for not adding tests initially, I looked for tests for some other `rx_` extension and saw there was none so though these were generally not tested.

PR #720, opened 2016-05-29
Extend CompositeDisposable with variadic init

larryonoff:


larryonoff:
Hope that this time I understood you properly.

kzaher:
It's close, we'll change it a bit :)

PR #724, opened 2016-05-31
Rewrite an implementation to reduce compilation time

chunkyguy:
I've observed that there is some code that results in super long compilation times. 

I'm using the following build environment:
```
$ xcodebuild -version
Xcode 7.3.1
Build version 7D1014

$ swift -version
Apple Swift version 2.2 (swiftlang-703.0.18.8 clang-703.0.31)
Target: x86_64-apple-macosx10.9
```

And this is the script I used to profile the compiler:

```
xcodebuild\
		 -workspace Rx.xcworkspace\
		 -scheme RxSwift-iOS\
		 -configuration Release CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO OTHER_SWIFT_FLAGS="-Xfrontend -debug-time-function-bodies"\
		 clean build | grep [1-9].[0-9]ms | sort -nr | head -3 | cat
```

The worst performing line of code is in `Bag.swift`
```
public var count: Int {
    return _pairs.count + (_value0 != nil ? 1 : 0) + (_value1 != nil ? 1 : 0) + (_dictionary?.count ?? 0)
}
```

Which takes around **1005 ms** on my machine to compile
```
1005.0ms	/RxSwift/RxSwift/DataStructures/Bag.swift:154:27	get {}
```

Just by replacing the coalescing operator the same function takes around **13 ms** to compile

```
public var count: Int {
    var dictionaryCount = 0
    if let dc = _dictionary?.count {
        dictionaryCount = dc
    }
    return _pairs.count + (_value0 != nil ? 1 : 0) + (_value1 != nil ? 1 : 0) + dictionaryCount
}
```

```
12.8ms	/RxSwift/RxSwift/DataStructures/Bag.swift:154:27	get {}
```

### DIFF
--- a/RxSwift/DataStructures/Bag.swift
+++ b/RxSwift/DataStructures/Bag.swift
@@ -152,10 +152,7 @@ public struct Bag<T> : CustomDebugStringConvertible {
     - returns: Number of elements in bag.
     */
     public var count: Int {
-        var dictionaryCount = 0
-        if let dc = _dictionary?.count {
-            dictionaryCount = dc
-        }
+        let dictionaryCount: Int = _dictionary?.count ?? 0
         return _pairs.count + (_value0 != nil ? 1 : 0) + (_value1 != nil ? 1 : 0) + dictionaryCount
     }
     
--- a/RxSwift/DataStructures/Bag.swift
+++ b/RxSwift/DataStructures/Bag.swift
@@ -152,7 +152,11 @@ public struct Bag<T> : CustomDebugStringConvertible {
     - returns: Number of elements in bag.
     */
     public var count: Int {
-        return _pairs.count + (_value0 != nil ? 1 : 0) + (_value1 != nil ? 1 : 0) + (_dictionary?.count ?? 0)
+        var dictionaryCount = 0
+        if let dc = _dictionary?.count {
+            dictionaryCount = dc
+        }
+        return _pairs.count + (_value0 != nil ? 1 : 0) + (_value1 != nil ? 1 : 0) + dictionaryCount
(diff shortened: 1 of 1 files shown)


yury:
Wow!

kzaher:
Hey @chunkyguy ,

awesome work :)

I'm curious, does this also provide the same performance improvements?

```swift
    let dictionaryCount: Int = _dictionary?.count ?? 0
    return _pairs.count + (_value0 != nil ? 1 : 0) + (_value1 != nil ? 1 : 0) + dictionaryCount
```

It would be awesome if we could find short, simple and performant solution :)

chunkyguy:
@kzaher Yes, I just tried it, and it seems to be compiling equally fast.
So, I guess the problem is with type inferencing and not the coalescing operator, as I had initially assumed.

Another solution which works and might even be the fastest code (compiles in ~7.0ms) is
```
    public var count: Int {
        return _pairs.count + (_value0 != nil ? 1 : 0) + (_value1 != nil ? 1 : 0) + (_dictionary != nil ? _dictionary!.count : 0)
    }
```
Although I don't know if this follows the best Swift coding principles

kzaher:
I think it would make sense to go with not `!` solution :)

If you could add a commit with

```
    let dictionaryCount: Int = _dictionary?.count ?? 0
    return _pairs.count + (_value0 != nil ? 1 : 0) + (_value1 != nil ? 1 : 0) + dictionaryCount
```

I would be happy to merge this.

chunkyguy:
Done!

PR #726, opened 2016-06-04
UIResponder. Added resign and become bindings.

hernangonzalez:
Hello!

Does this look OK? I am not very experienced and was in need of a binding here. Let me know if this is good enough and I will add tests to it. 

Thank you!
Hernán


hernangonzalez:
Ups, sent to master. Will try again with develop.

PR #727, opened 2016-06-04
UIResponder become/resign bindings

hernangonzalez:
Hello!

Does this look OK? I am not very experienced and was in need of a binding here. Let me know if this is good enough and I will add tests to it.

Thank you!
Hernán

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -22,6 +22,9 @@
 		54D2138E1CE0824E0028D5B4 /* UINavigationItem+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = 54D2138C1CE081890028D5B4 /* UINavigationItem+Rx.swift */; };
 		54D213921CE08D0C0028D5B4 /* UINavigationItem+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = 54D2138C1CE081890028D5B4 /* UINavigationItem+Rx.swift */; };
 		54D213931CE08DDB0028D5B4 /* UINavigationItem+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = 54D2138C1CE081890028D5B4 /* UINavigationItem+Rx.swift */; };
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @hernangonzalez ,

hm ... Could you maybe provide some context to this? I didn't need those extensions, so I'm little curious about how do you use them, and in what cases.

hernangonzalez:
hi @kzaher I used this to connect actions of UI items still visible that would end up in actions that i want to dismiss the keyboard to. I.e if a user taps Done on the keyboard:
```
		/**
		Search is handled dynamically as user tipes. So we use the return key
		as done and dismiss the keyboard instead.
		*/
		searchBar.rx_searchButtonClicked
			.asDriver()
			.drive(searchBar.rx_resignFirstResponder)
			.addDisposableTo(rx_disposeBag)
```

kzaher:
Hey @hernangonzalez ,

I've been thinking about this. I think this is part of a more general problem, how do you bind '()' to call any method. If we were to add this extension, this would open doors to a bunch of extensions we could add :). There is a lot of methods out there, and all could potentially be event targets.

I think your code could be rewritten as:

```swift
searchBar.rx_searchButtonClicked
            .asDriver()
            .drive(weakify(searchBar, searchBar.dynamicType.resignFirstResponder))
            .addDisposableTo(rx_disposeBag)
```

It's not significantly longer, and there is a lot of weakify libraries out there :)

¯\\\_(ツ)\_/¯ 

I totally agree your extension is useful, and more readable then my suggestion, but not sure it would be a good idea to start pulling in similar extensions because this library would bloat too much.

hernangonzalez:
Sure @kzaher, I agree that is best to try and keep the lib slim. I'll give a shot to the weakify frameworks (just found https://github.com/klundberg/Weakify), and ping back if its not working for me.

Thanks,
H.

serejahh:
Hi @hernangonzalez 
You can create a PR with your changes here: https://github.com/RxSwiftCommunity/RxSwiftExt/

hernangonzalez:
Hi @serejahh will do that, thank you.

PR #730, opened 2016-06-07
Add API Documentation

solidcell:
This adds [jazzy](https://github.com/realm/jazzy) API documentation.

It's a WIP, as there are still some tasks to finish.  However, we should consider merging it before all of the tasks are finished, as they are not MVP and we could break it up into separate PRs for further refinement.

You can see the docs [here](https://solidcell.github.io/RxSwift).

Here's the list of remaining tasks that I can think of:
- [ ] Decide how we'd like the script to be run (new `release-prep.sh` which runs tests and updates docs?)
- [ ] Add links to docs for different platforms in README
- [x] Add a check for `jazzy` gem being installed (and error if not?)


kzaher:
Hi @solidcell ,

Thnx for helping us tackle that jazzy issue. Awesome work for wiring those scripts.

There are some things that we would need to change to pull this in.

* Could you just create a PR with contents of `./script` changes. We won't be hosting html files inside of this repo, we'll host them. They will be hosted on http://cocoadocs.org/docsets/RxSwift/2.5.0/, so we just need just that .jazzy.yml file, so we can help CocoaDocs generate those docs.

* The script currently doesn't transform directory structure to categories inside yml file, and it's not recursive. We would want to generate categories like:

```
- name: RxSwift
....
- name: RxSwift / Core
....
- name: RxSwift / Schedulers / Internal
....
- name: RxCocoa
...
- name: RxCocoa / iOS
....
- name: RxCocoa / iOS / Proxies 
....
```

If you could modify that script to generate that, it would be awesome, otherwise, I can modify it easily.

solidcell:
Yea for sure.  I'll change the grouping and remove the web files.

solidcell:
jazzy can't combine the output from multiple modules, so we'll have to have two different doc sets for `RxSwift` and `RxCocoa`.  However, they can both share the same `.jazzy.yml` which has all of the categories.

Let me know if there's anything else you'd like me to include or change.

solidcell:
Also, I force updated the changes, so it should be good to review.

solidcell:
Also, let me know if you were wanting literal spaces in the category names, such as `RxSwift / Schedulers / Internal`, or if it should stay as `RxSwift/Schedulers/Internal`.

PR #741, opened 2016-06-13
Add a convenient method for data source

serejahh:
I prefer to use the same reuse identifiers for cells to avoid hard coded values in the code. So I've add a convenient method to the UITableView extension.

### DIFF
--- a/RxCocoa/iOS/UITableView+Rx.swift
+++ b/RxCocoa/iOS/UITableView+Rx.swift
@@ -64,6 +64,22 @@ extension UITableView {
     }
     
     /**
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @serejahh ,

I can see why you find this useful. I think that it would be best to not go on this path in the main repo. 

These convenience methods could be useful for sure, but there is a lot of ways how one could map type to cell identifier, and I'm not sure it would be a good idea to add these conveniences to this repo and add additional noise to something which is already non trivial.

For example, people could want the cell to conform some protocol that has identifier statically defined, etc.

kzaher:
Sorry :) Hope you understand.

PR #742, opened 2016-06-14
copy edit Schedulers.md

mattfenwick:
This is a documentation-only change and doesn't affect the meaning of the docs: I just fixed the grammar and added some missing words.


kzaher:
Thnx :) 👍

PR #751, opened 2016-06-16
Replace `rand` with `arc4random`

brentleyjones:
Swift 2.3 won't compile with `rand`.

Is `_forceCompilerToStopDoingInsaneOptimizationsThatBreakCode` even needed anymore? If not, maybe it can be removed in a separate PR.


kzaher:
We'll remove that when we'll create RxSwift 3.0 branch. We probably still need it.

PR #758, opened 2016-06-21
Add NSProgressIndicator and NSTextView bindings for RxCocoa on Mac

bontoJR:
NSTextView is preliminary, it will require quite more work considering the complexity of that view in AppKit.


lunij:
Any reasons why this isn't merged yet?

kzaher:
We need to update this PR to new 3.0 API and resolve merge conflicts. And I need to check out OSX APIs a bit.

RxCocoa/OSX/NSProgressIndicator+Rx.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Copyright © 2016 Krunoslav Zaher. All rights reserved.
+//
+
+import Foundation
+import Cocoa
+#if !RX_NO_MODULE
+import RxSwift
+#endif
+
+
```
wesbillman: Do you want 2 newlines here?

RxCocoa/OSX/NSProgressIndicator+Rx.swift
```diff
@@ -0,0 +1,27 @@
...
+    /**
+    Bindable sink for `progress` property
+    */
+    public var rx_progress: AnyObserver<Double> {
+        return UIBindingObserver(UIElement: self) { progressView, progress in
+            progressView.doubleValue = progress
+        }.asObserver()
+    }
+
```
wesbillman: extra newline?

RxCocoa/OSX/NSTextView+Rx.swift
```diff
@@ -0,0 +1,51 @@
...
+            delegate.textSubject.startWith(self?.textStorage ?? NSTextStorage())
+            }.takeUntil(rx_deallocated)
+        
+        let observer = UIBindingObserver(UIElement: self) { control, value in
+            control.textStorage?.setAttributedString(value)
+        }
+        
+        return ControlProperty(values: source, valueSink: observer.asObserver())
+    }
+}
```
wesbillman: newline at end of file?

RxCocoa/OSX/Proxies/RxTextFieldDelegateProxy.swift
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ Delegate proxy for `NSTextField`.
+ 
+ For more information take a look at `DelegateProxyType`.
+ */
+public class RxTextFieldDelegateProxy
+    : DelegateProxy
+    , NSTextFieldDelegate
+, DelegateProxyType {
```
wesbillman: should this be intended?

RxCocoa/OSX/Proxies/RxTextFieldDelegateProxy.swift
```diff
@@ -0,0 +1,77 @@
...
+     */
+    public weak private(set) var textField: NSTextField?
+    
+    /**
+     Initializes `RxTextFieldDelegateProxy`
+     
+     - parameter parentObject: Parent object for delegate proxy.
+     */
+    public required init(parentObject: AnyObject) {
+        self.textField = (parentObject as! NSTextField)
```
wesbillman: you don't need `self` here. Should this be `textField = parentObject as? NSTextField` to avoid a crash?

RxCocoa/OSX/Proxies/RxTextFieldDelegateProxy.swift
```diff
@@ -0,0 +1,77 @@
...
+        super.init(parentObject: parentObject)
+    }
+    
+    // MARK: Delegate methods
+    
+    public override func controlTextDidChange(notification: NSNotification) {
+        let textField = notification.object as! NSTextField
+        let nextValue = textField.stringValue
+        self.textSubject.on(.Next(nextValue))
```
wesbillman: use `as?` to avoid crash?
you don't need `self` here
remove extraneous `nextValue`?
```swift
let textField = notification.object as? NSTextField
textSubject.on(.Next(textField.stringValue))
```

or 

```swift
if let textField = notification.object as? NSTextField {
    textSubject.on(.Next(textField.stringValue))
}
```

Depends on if you want to fire on nil strings like `textSubject.on(.Next(nil))` or not

RxCocoa/OSX/Proxies/RxTextFieldDelegateProxy.swift
```diff
@@ -0,0 +1,77 @@
...
+    }
+    
+    // MARK: Delegate proxy methods
+    
+    /**
+     For more information take a look at `DelegateProxyType`.
+     */
+    public override class func createProxyForObject(object: AnyObject) -> AnyObject {
+        let control = (object as! NSTextField)
```
wesbillman: do you want to force unwrap?

RxCocoa/OSX/Proxies/RxTextViewDelegateProxy.swift
```diff
@@ -0,0 +1,76 @@
...
+        self.textView = (parentObject as! NSTextView)
+        super.init(parentObject: parentObject)
+    }
+    
+    // MARK: Delegate methods
+    
+    public override func controlTextDidChange(notification: NSNotification) {
+        let textView = notification.object as! NSTextView
+        let nextValue = textView.attributedString()
+        self.textSubject.on(.Next(nextValue))
```
wesbillman: you don't need `self` here, same force unwrap and extra var comment as above

RxCocoa/OSX/Proxies/RxTextViewDelegateProxy.swift
```diff
@@ -0,0 +1,76 @@
...
+    }
+    
+    /**
+     For more information take a look at `DelegateProxyType`.
+     */
+    public class func setCurrentDelegate(delegate: AnyObject?, toObject object: AnyObject) {
+        let textField: NSTextField = castOrFatalError(object)
+        textField.delegate = castOptionalOrFatalError(delegate)
+    }
+}
```
wesbillman: Need newline here?

RxCocoa/OSX/Proxies/RxTextFieldDelegateProxy.swift
kzaher: I would say that I would be fine using `as!` here.  We use `castOrFatalError` on library level for situations like this. We don't want to design `RxTextFieldDelegateProxy.swift` to be used with some object that is not a subclass of `NSTextField`.  This is equivalent to `fatalError` caused by index out of range.

kzaher:
This is a bit stale.

I think it would be best to just close this for now since it's sitting here a while.

We can add OSX extensions, but let's do them one by one, maybe first opening an issue with specific extension request so we can evaluate it first, and after that a PR with all of the necessary unit tests and demo in OSX app since I really don't have a lot of experience with OSX.

gerbil802:
+1 On `NSTextView` extensions

PR #761, opened 2016-06-22
Swift 3.0

jamiepinkham:
This is an initial pass to get `RxSwift` working in Xcode 8 and Swift 3 beta. 🎉 

The majority of the changes relate to the migrator missing the `.Error` enum value, and a lot of changes to the argument values of a lot of functions to match the current `RxSwift` API. The migrator converted some from:

```swift
func someFunc(argument: Argument)
```
to

```swift
func someFunc(_ argument: Argument)
```

and in some cases this was not the desired output for RxSwift's API, and in other cases, it completely skipped some of those functions.

At this point the following targets do not build: 
* PerformanceTests
* RxTests-watchOS (no XCTest?!)
* RxExample-iOS-no-module (Segmentation Fault: 11 🎉 )
* AllTests

The `AllTest` bundle not building is due to the macro generated class `SentMessageTest_shared` and how it's methods are imported into Swift. I'm leaving this as a problem for @kzaher to solve, as I'm not fully aware of what's going on there, and would probably do more damage than help.

Some of the other build errors include issues with how we're dealing with some malloc hooks in PerformanceTests, `XCTest` not being available in `watchOS`?!, and the use of `_NSStringCoreType` which is unavailable in Swift.

Both of the example apps work as expected.

Lastly, the playground doesn't seem to appear to be working. I did get a playground working using `RxSwift` in the project, but the included one seems to have some issues.


bontoJR:
This is just massive! Amazing 🎉

kzaher:
Hi @jamiepinkham ,

great work.

I've seen some tiny issues, so if you'll have time, feel free to fix those also. If not, we can also fix them after the merge.

* on Driver we would want to have init with `(raw`
* code isn't changed in template (.tt) files. e.g. zip or combine latest n-ary versions and unit tests. You generate those files by running the generator (Preprocessor target) inside the project. It will generate .swift from .tt files.

There are also some changes that we'll need to do to later to modernize the method names, but even getting to this state is awesome. 🎉

PR #766, opened 2016-06-25
Made RxAllTests compile with Swift 3.0

sergdort:
Made compile AllTests 

There are some problems on iOS tests it crashes on `testDidRangeBeaconsInRegion `
Also if remove `autoreleasepool` from the tests it crashes on `AnonymousObserver ` `deinit`

There is also issue with `CFRunLoopPerformBlock` in run time it crashes on 
`CFRunLoopMode.defaultMode as! CFTypeRef` because it's swift struct now


sergdort:
Also hade to rename `voidJustCalledToSayObject` to `voidJustCalledObjectToSay`

Because now Swift converts selectors from objective c, kinda weird :)

So all selector was `voidJustCalled(toSay:)`

kzaher:
👍 

We'll we are getting closer :)

PR #767, opened 2016-06-27
Change _forwardToDelegate to weak reference

sora0077:
It causes EXC_BAD_ACCESS in `RxScrollViewDelegateProxy.scrollViewDidScroll` when deallocating object.


kzaher:
Hi @sora0077 ,

yeah, agreed, I've also ran into some non trivial issues with scroll view delegate. Thnx 👍

PR #770, opened 2016-06-28
Resolve minor static analysis issues

tonyarnold:
Not a major issue by any definition of the word, but we're seeing static analysis issues come through in our build logs when including RxSwift as a subproject. 

I could use some guidance here on your project's style and approach: there are 11 additional static analysis warnings as follows:

```
RxSwift/RxCocoa/Common/_RXObjCRuntime.m:599:16: warning: Potential null dereference.  According to coding standards in 'Creating and Returning NSError Objects' the parameter may be null
        *error = [NSError errorWithDomain:RXObjCRuntimeErrorDomain
        ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I see that you've annotated the `NSError *__autoreleasing *__nonnull`, however if you [read Apple's blog post on nullability](https://developer.apple.com/swift/blog/?id=25), they specifically say:

> The particular type NSError ** is so often used to return errors via method parameters that it is always assumed to be a nullable pointer to a nullable NSError reference.

So these issues are only going to go away if the code is changed to treat them as nullable. I'm happy to make this change (something as simple as an early return, or a parameter assertion), but I wasn't sure which approach you'd prefer. 

It's also worth noting that [people have filed radars against this issue](http://www.openradar.me/21766176), however I'd like to see this resolved sooner rather than later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 
 * Adds `rx_title` to `UIViewController`.
 * Adds `rx_scrollEnabled` to `UIScrollView`.
+* Resolve static analysis issues relating to non-use of an assigned value, and potential null dereferences in RxCocoa's Objective-C classes.
 
 
 ## [2.5.0](https://github.com/ReactiveX/RxSwift/releases/tag/2.5.0)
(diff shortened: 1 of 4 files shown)


kzaher:
Hi @tonyarnold ,

.... just noticed this one.

I think maybe the smart thing would be to add something like

```
#define RX_THROW(errorValue)  if (error) { *error = (errorValue); }
```

and use it.

```
RX_THROW([NSError errorWithDomain:RXObjCRuntimeErrorDomain
```

Let me know what you think.

tonyarnold:
OK, I've added a simple macro and documentation in e4219c8423776e2c597bb32f272268cc85036b7b. 

I considered using a static inline function, but the macro was simplest.

PR #773, opened 2016-06-28
UIGestureRecognizerDelegate proxy

sgl0v:
Implemented the UIGestureRecognizerDelegate proxy & unit tests.


kzaher:
Hi @sgl0v ,

unfortunately, I don't think there are any methods on `UIGestureRecognizerDelegate`, so I'm not sure why do you want to wrap this?

PR #775, opened 2016-06-28
Update for Xcode 8, Swift 2.3.

DouglasHeriot:
Bump OSX Deployment Target to 10.10 to avoid libarclite duplicate linking error:

    duplicate symbol __non_lazy_classes in:
    /Volumes/Sierra/Applications/Xcode-beta.app/Contents/Developer/Toolchains/Swift_2.3.xctoolchain/usr/lib/arc/libarclite_macosx.a(arclite.o)
    /Volumes/Sierra/Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_macosx.a(arclite.o)
    duplicate symbol _OBJC_METACLASS_$___ARCLite__ in:
    /Volumes/Sierra/Applications/Xcode-beta.app/Contents/Developer/Toolchains/Swift_2.3.xctoolchain/usr/lib/arc/libarclite_macosx.a(arclite.o)
    /Volumes/Sierra/Applications/Xcode-beta.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_macosx.a(arclite.o)
    ld: 2 duplicate symbols for architecture x86_64
    clang: error: linker command failed with exit code 1 (use -v to see invocation)

I would be using the swift-3.0 branch, but just waiting for some other 3rd party libraries I need to support Swift 3.

PR #776, opened 2016-06-29
Make Travis use Xcode 8 for building Swift 3

DouglasHeriot:
Not sure how/if the Swift-3 branch is getting CI at all without this?
Also updated validate-headers.swift script, as I noticed it failing with Xcode 8 in #775. There are other scripts that need updating too.


kzaher:
Hi @DouglasHeriot ,

thnx. Didn't know travis had xcode 8 already. I'll fix those headers that are failing.

PR #779, opened 2016-07-02
Fix RxBlocking in Swift 3

tomburns:
Remove force cast in favor of rawValue to fix crash at runtime

### DIFF
--- a/RxBlocking/RunLoopLock.swift
+++ b/RxBlocking/RunLoopLock.swift
@@ -39,7 +39,6 @@ class RunLoopLock {
     }
 
     func dispatch(_ action: () -> ()) {
-        
         CFRunLoopPerformBlock(currentRunLoop, CFRunLoopMode.defaultMode.rawValue) {
             if CurrentThreadScheduler.isScheduleRequired {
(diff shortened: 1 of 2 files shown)


tomburns:
Noticing now that several (but not all) tests that use RxBlocking appear to be deadlocking... 😕

tomburns:
Found that a source of "deadlock" was actually just incorrect conversion of the interval in `ConcurrentDispatchQueueScheduler.schedulePeriodic` and fixed it

tomburns:
Taking a look at why `ObservableConcurrentSchedulerConcurrencyTest2.testObserveOn_EnsureTestsAreExecutedWithRealConcurrentScheduler()` is still hanging, though it passes when run in isolation; any insight would be welcome as this one might have me stumped 🤒

tomburns:
Regardless of the remaining issues, I'm pretty sure these fixes are mergeable as-is; we can continue tackling the rest of the issues separately. Let me know if there's anything else I can do to wrap up this PR.

kzaher:
Hey, these changes look ok to me. Yeah, I think there are some issues left that we need to solve after this one.

I'll try to release 2.6 today, so I can concentrate on 3.0 after that.

PR #785, opened 2016-07-06
expecting RxCollectionViewDelegateProxy, not RxScrollViewDelegateProxy

seivan:
This can't be right?


kzaher:
Hi @seivan ,

`UICollectionView` inherits from `UIScrollView` so I think we'll unfortunately need to use `RxCollectionViewDelegateProxy`. Compiler is complaining :)

PR #787, opened 2016-07-07
Merged latest master

sergdort:
Also fixed some failed tests


RxCocoa/Common/Observables/Implementations/KVOObservable.swift
```diff
@@ -104,7 +104,8 @@ func observeWeaklyKeyPathFor(
     let propertyAttributes = property_getAttributes(property)
     
     // should dealloc hook be in place if week property, or just create strong reference because it doesn't matter
-    let isWeak = isWeakProperty(String(propertyAttributes) ?? "")
+    let propertyAttributesAsString = propertyAttributes == nil ? "" : String(cString: propertyAttributes!)
```
kzaher: I think we shouldn't use `!`. This was the only complaint I had, so I'll merge this and fix it after merge.

Thnx 👍

PR #788, opened 2016-07-07
Remove typealias Parent line in Filter.swift

kciter:
When I studying RxSwift internal sources, I found that unused code line. So, remove the code and request Pull-Request. 😄


kciter:
Oops.. my mistake..

PR #401, opened 2016-01-12
Add doOn shortcut operators

thanegill:
Adds three operators that are shortcuts for `doOn`. I use these commonly and have seem them in [other projects](https://github.com/artsy/eidolon/blob/04a0c8488525be4b918e212e92a29f143e0f5838/Kiosk/Observable%2BOperators.swift#L64-L94) as well.

PR #790, opened 2016-07-10
[Swift 3.0] Remove first parameter names for Observables.

sunshinejr:
I've noticed that few first parameter names were now mandatory in comparison to RxSwift for Swift 2.2. I went through all observables and I've tried to find which one were mandatory in version for Swift 3.0 and wasn't for Swift 2.2. If this is desired behavior then I'm sorry 🙊

### DIFF
--- a/RxSwift/Observables/Observable+Single.swift
+++ b/RxSwift/Observables/Observable+Single.swift
@@ -84,7 +84,7 @@ extension ObservableType {
     - returns: The source sequence with the side-effecting behavior applied.
     */
     // @warn_unused_result(message:"http://git.io/rxs.uo")
-    public func doOn(eventHandler: (Event<E>) throws -> Void)
+    public func doOn(_ eventHandler: (Event<E>) throws -> Void)
         -> Observable<E> {
         return Do(source: self.asObservable(), eventHandler: eventHandler)
     }
@@ -121,7 +121,7 @@ extension ObservableType {
      - returns: The source sequence with the side-effecting behavior applied.
      */
     // @warn_unused_result(message:"http://git.io/rxs.uo")
-    public func doOnNext(onNext: ((E) throws -> Void))
+    public func `do`(onNext: ((E) throws -> Void))
         -> Observable<E> {
         return self.doOn(onNext: onNext)
     }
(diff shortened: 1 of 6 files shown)


RxSwift/Observables/Observable+Single.swift
```diff
@@ -84,7 +84,7 @@ extension ObservableType {
     - returns: The source sequence with the side-effecting behavior applied.
     */
     // @warn_unused_result(message:"http://git.io/rxs.uo")
-    public func doOn(eventHandler: (Event<E>) throws -> Void)
+    public func doOn(_ eventHandler: (Event<E>) throws -> Void)
```
kzaher: I think that we might delete all `doOnNext`, `doOnError`, `doOnCompleted`, and leave just 

```
public func doOn(onNext: ((E) throws -> Void)? = nil, onError: ((ErrorProtocol) throws -> Void)? = nil, onCompleted: (() throws -> Void)? = nil)
```

Thoughts?

sunshinejr: I'm down for that, but shall we also consider changing it with `subscribe` methods?

sergdort: I bet guys from this [conversation](https://github.com/ReactiveX/RxSwift/pull/401) will complain :)

kzaher: Yeah, I think we should.

sunshinejr: Hmm, should I open issue about that and let's see how it goes? Seems like many people wanted the shortcut syntax 😄

kzaher: Maybe we can keep it simple and just use Swift 3.0 naming conventions first.
My intention isn't to remove any useful functionality, just to modernize the API. 

It seemed to me that Xcode autocomplete provided overloads with all combinations of optional parameters, but not sure what happened to that functionality.

sunshinejr: @kzaher I've modernized the `doOn` shortcut syntax as for the conversation (`doOn(next:)` etc.). I don't know about doing the same with `subscribeOn(next:)`, because there is also `subscribeOn` for schedulers and it might get confusing..

sunshinejr: @kzaher squashed and added backtick support for `do(onNext)` etc. Seems like it is working.

PR #795, opened 2016-07-14
Drive to signal intent that this is UI

seivan:
I am mostly asking if this would be accurate or correct.

### DIFF
--- a/Documentation/Why.md
+++ b/Documentation/Why.md
@@ -61,7 +61,8 @@ public func scrollViewDidScroll(scrollView: UIScrollView) { [weak self] // what 
 self.resultsTableView
     .rx_contentOffset
     .map { $0.x }
-    .bindTo(self.leftPositionConstraint.rx_constant)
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @seivan ,

I guess you could do something like this in this scenario also, but I'm not sure if it would bring much benefit since you can easily see that all properties are already satisfied since content offset can't fail and is already delivering elements on main thread.

seivan:
But shouldn't it be a driver? I mean it's about intent if I understood the difference since both types jump on to main and won't error out(?) 
I'm genuinely curious as I'm Trying to grasp this and write a mini clone lazy monad just to get a better hang.

kzaher:
Maybe you've meant to write something like
```
self.resultsTableView.rx_contentOffset.asDriver()
     .map { $0.x }
     .drive(self.leftPositionConstraint.rx_constant)
```

seivan:
Yeah, heh, that looks much better and suggests I gotta do another run down.

PR #796, opened 2016-07-19
Updates project with Xcode 8 beta 3.

noremac:
The release of Xcode 8 beta 3 necessitated a few small fixes.


kzaher:
Nice

Looks like this partially fixes the problem, but still can't build unit tests. Looks like also I'll need to dig again :)

noremac:
D'oh! I forget to check that, sorry!

Issue #800, opened 2016-07-25
[Swift 3.0] Rename disposable-related methods


mohsenr:
Rename Some methods to better match [Swift 3.0 API Design Guidelines](https://swift.org/documentation/api-design-guidelines/).

Methods will old name will still exist for a while with a deprecation annotation.

This is a summary of the proposed changes:

```diff

extension CompositeDisposable {
    
+   @available(*, renamed: "insert()")
    public func addDisposable(_ disposable: Disposable) -> DisposeKey?
    
+   public func insert(_ disposable: Disposable) -> DisposeKey?
    
+   @available(*, renamed: "removeDisposable(forKey:)")
    public func removeDisposable(_ disposeKey: DisposeKey)
    
+   public func removeDisposable(forKey disposeKey: DisposeKey)
    
}

extension DisposeBag {
    
+   @available(*, renamed: "insert()")
    public func addDisposable(_ disposable: Disposable)
    
+   public func insert(_ disposable: Disposable)
    
}

+@available(*, renamed: "Bag.disposeAll()")
public func disposeAllIn(_ bag: Bag<Disposable>)

+extension Bag where T: Disposable {
+    
+    public func disposeAll()
+    
+}


```

mohsenr:
The last renaming doesn’t work as the compiler can’t produce call to `Bag<Disposable>` when not using a concrete subtype. Instead, we have:

```diff
+@available(*, renamed: "disposeAll(in:)")
public func disposeAllIn(_ bag: Bag<Disposable>)

+public func disposeAll(in bag: Bag<Disposable>)
```

kzaher:
Hey,

yeah, I think your suggestions make sense.
The one change I would suggest is using
`public func remove(forKey disposeKey: DisposeKey)`
instead of 
`public func removeDisposable(forKey disposeKey: DisposeKey)`

... so there is no redundant `Disposable`.

mohsenr:
I modelled that on `Dictionary.removeValue(forKey:)` cause its the closest API for what we’re doing here.

Though you’re right that the shorter name feels better. Specially given that the type is not parameterised. Maybe we should parallel with `Array.remove(at:)`: so we’d have `remove(for:)` – unless you think the `Key` part is needed?

kzaher:
@mohsenr yeah, my reasoning was because of `array.remove(at:)`.

I'm fine with `remove(for:)` suggestion.

PR #801, opened 2016-07-25
[Swift 3.0] Rename disposable-related methods (implements #800)

mohsenr:


RxSwift/Disposables/CompositeDisposable.swift
```diff
@@ -102,15 +114,25 @@ public class CompositeDisposable : DisposeBase, Disposable, Cancelable {
...
     }
-
-    private func _removeDisposable(_ disposeKey: DisposeKey) -> Disposable? {
+    
+    /**
+     Removes and disposes the disposable identified by `disposeKey` from the CompositeDisposable.
+     
+     - parameter disposeKey: Key used to identify disposable to be removed.
+     */
+    public func removeDisposable(forKey disposeKey: DisposeKey) {
```
kzaher: I think we should have naming consistent with https://github.com/ReactiveX/RxSwift/issues/800

So probably `remove(for:)`.

PR #828, opened 2016-08-09
[Swift 3.0] Collect all immutable Disposable constructors in `Disposables` type

mohsenr:
Currently each disposable type has its own construction pattern. You need to know about `NopDisposabe.instance`, `StableCompositeDisposable.create`, and `AnonymousDisposable.init` separately.

This pattern hurts discoverability of existing disposables implementation.

This pull requests moves all of these constructors to a `Disposables` namespace type . With this move, all immutable disposable types can be removed from the public interface. It also makes creation of disposables more similar to how observables are created.

**Note*: This pull request should probably not be merged yet as there are still some changes anticipated. In particular:

* Internal usage of deprecated APIs hasn’t been migrated yet, so the code is peppered with warnings.
* It might be better to rename `Disposables.create` to `Disposables.make` to conform to [Swift API Guidelines](https://swift.org/documentation/api-design-guidelines/)


RxSwift/Disposables/Disposables.swift
```diff
@@ -0,0 +1,46 @@
...
+
+import Foundation
+
+public struct Disposables {
+    
+}
+
+public extension Disposables {
+    
+    static func empty() -> Disposable {
```
kzaher: Ok, so let's change this to `create` like we've discussed on slack. We also need to add inline documentation to all of these methods, and probably make them public? 

I think we are also missing a variadic version (3 disposables + ...).

mohsenr:
OK I _think_ we’re good to go. Let me know if I’ve missed anything.

PR #829, opened 2016-08-09
Update for Xcode 8 beta 5

rhx:
Newer versions of Swift-3.0 require parameters to be passed in the same order as specified in the interface.  This pull request fixes Package.swift so the Swift package manager won't complain when instantiating Package()

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,11 +3,6 @@ import PackageDescription
 #if os(OSX)
 let package = Package(
     name: "RxSwift",
-    exclude: [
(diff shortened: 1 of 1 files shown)


rhx:
Looking at the build output, it looks like all builds and tests pass, so I'm not sure why travis-ci isn't happy (sorry, I'm not a Travis CI expert).  The only thing that sticks out seems to be line 241 (error: branch 'danger_base' not found.), but I don't know if there is anything I can do in the Pull request to fix this.

Issue #826, opened 2016-08-08
[RxCocoa] Move from `rx_` prefix to a `rx.` proxy (for Swift 3 update ?)


JegnuX:
## Description

The `rx_` prefix of all reactive extensions to Cocoa are not really pretty and don't fit with the Swift coding guidelines. They are multiple choices to rid off of this syntax :
- `rx_tap` : The original syntax
- `rxTap` : Standard camel case with `rx` prefix but it feels pretty weird to have several method starting with `rx`. Remember, `rx` means `Reactive Extension` so it's read as `reactive extension tap`.
- `reactiveTap` : Still camel case but with a more readable and meaningful name. But it's clearly less convenient than the `rx` prefix
- `rx.tap` : ✨😲✨ Whouah, this one is so beautiful. A nice pretty namespace to all reactive extensions of an object. It's as readable as it's convenient to write in autocompleted environment.

##### Why `rx.` prefix is a good design imho ?

The main idea is to use a proxy which own the original object. Then we extend the proxy to implement a new reactive extension.

You can see this kind of pattern in the Switf Standard Library with the [LazySequence](https://github.com/apple/swift/blob/master/stdlib/public/core/LazySequence.swift) type. For instance if you have an array `myArray`, you can perform a `map` directly or through it's lazy proxy :
```
myArray.map { ... }
myArray.lazy.map { ... }
```

We could have the same syntax in RxCocoa : 
```
if myButton.enabled { ... }
myButton.rx.enabled.subscribeNext { ... }
```

I kinda like this syntax, and I [wrote a post](https://medium.com/@JegnuX/safe-collection-subsripting-in-swift-3771f16f883#.kdlofek1s) on medium about how to use this syntax to add a safe subscripting for CollectionType. Indeed, the common solution I used to see is to add an extension to be able to write something like 
```
if let element = myArray[safe: 2] { ...
```
While I prefer the proxy pattern to be able to write this :
```
if let element = myArray.safe[2] { ... 
```

### Shared code
I suggest we could transform the `Reactive` protocol to a generic `struct` : 
```swift 
public struct Reactive<Base: AnyObject> {
    public let base: Base
    public init(_ base: Base) {
        self.base = base
    }
}
```
Then we extend `NSObjectProtocol` to create this `rx` proxy :
```swift
public extension NSObjectProtocol {
    public var rx: Reactive<Self> {
        return Reactive(self)
    }
}
```
[...]

PR #832, opened 2016-08-15
[WIP] [Swift 3.0] Moving from `rx_` prefix to `rx.` "namespace"

JegnuX:
# Description
This PR is following the issue #826

# To-do list
As this PR is still `work-in-progress`, here is the to-do list of what's done and what's left : 
- [x] Remove `rx_` prefixes
- [x] Extend `Reactive` with protocol requirements
- [x] Update Unit Tests
- [x] Update Documentation
- [ ] Find something better to handle `RxTextInput` (maybe we should remove it ?)

# How can I help
- Review my changes
- See if I didn't miss anything
- Do something on the to-do list

To merge your changes into this PR, please fork https://github.com/JegnuX/RxSwift, and make a PR against https://github.com/JegnuX/RxSwift/tree/swift-3.0-removing-rx_-prefix

Before working on something, you may want to leave a comment here to avoid having several identical contributions.


Documentation/API.md
```diff
@@ -104,7 +104,7 @@ Creating new operators is also pretty straightforward.
 
 extension NSObject {
 
-    public var rx_deallocated: Observable<Void> {}
+    public var deallocated: Observable<Void> {}
```
kzaher: This should be `rx.deallocated` right?

Documentation/GettingStarted.md
```diff
@@ -208,7 +208,7 @@ Additional way to automatically dispose subscription on dealloc is to use `takeU
 
 ```swift
 sequence
-    .takeUntil(self.rx_deallocated)
+    .takeUntil(self.deallocated)
```
kzaher: `rx.deallocated`?

Documentation/API.md
JegnuX: Not really, `var rx.deallocated` is not valid 😜 
It's more the `extension NSObject` that should be changed.

But anyway, I didn't update documentation yet so this is "normal".

RxCocoa/Common/Reactive.swift
```diff
@@ -7,27 +7,34 @@
...
+ `Self` and not just specialized on common base type.
 
-    // 2. Extend Reactive protocol with constrain on Self
-    // Read as: Reactive Extension where Self is a SomeType
-    extension Reactive where Self: SomeType {
-        // 3. Put any specific reactive extension for SomeType here
-    }
+ */
 
+public struct Reactive<Base: AnyObject> {
```
kzaher: Is there any reason why we are limiting this to `: AnyObject`?

We use `source` terminology when adding extensions, it might make sense to use `public let source: Source` for consistency with other parts of libraries.

RxCocoa/Common/RxTextInput.swift
```diff
@@ -21,6 +21,19 @@ import Foundation
...
+        public var rx_text: ControlProperty<String> {
+            return self.rx.text
+        }
+    }
+
+    extension UITextView : RxTextInput {
+        public var rx_text: ControlProperty<String> {
+            return self.rx.text
+        }
+    }
```
kzaher: It would be great to solve this somehow.

I wish something like this would work:
```swift
/**
    Represents text input with reactive extensions.
    */
    public protocol RxTextInput : UITextInput {
        /**
         Reactive wrapper for `text` property.
        */
        static func text(extension: Self) -> ControlProperty<String>
    }

    extension UITextField : RxTextInput {
        public static func text(extension: Self) -> ControlProperty<String> {
            return `extension`.rx.text
        }
    }

    extension UITextView : RxTextInput {
        public static func text(extension: Self) -> ControlProperty<String> {
            return `extension`.rx.text
        }
    }
```

... but it doesn't :(

RxCocoa/Common/Reactive.swift
JegnuX: We can indeed remove the `AnyObject` requirement. I don't now if it really makes sense on a value type but who know. :)

For the `source` terminology I'm not convinced. I've came across the codebase looking for `let source =` declarations and everywhere, `source` is an `Observable`. Here it won't be on `Observable` but the object itself. It may confuse people if we mix up this two concepts behind the same term don't you think ?

PR #839, opened 2016-08-19
Additional extensions for RxTests

sergdort:
Hi, guys!

I found myself struggling with `XCTAssertEqual ` and `RxSwift + RxTests`

So for example with current API I can't don't this simple test:

```Swift
        
            next(150, [0, 1]),
            next(210, [1, 2]),
            next(240, [2, 3]),
            next(270, [3, 4]),
            next(300, [4, 5]),
            completed(400)
            ])
        
        let res = scheduler.start {
            xs.throttle(20, scheduler: scheduler)
        }
        
        let correct = [
            next(230, [1, 2]),
            next(260, [2, 3]),
            next(290, [3, 4]),
            next(320, [4, 5]),
            completed(400)
        ]
        
        XCTAssertEqual(res.events, correct) // This won't compile since there is no overload for `[Recorded<Event<[T]>>]`
```
Which is to me pretty obvious situation :)
Pleas correct me if I'm wrong :)


kzaher:
Hi @sergdort ,

sorry for not answering sooner, this is a limitation of Swift for now.

I wouldn't want to add a lot of APIs to solve current language issues.

The workaround is to use something like `EquatableArray`. The reason why we haven't added it to RxTests so far, but are only using it in Tests is because it's just a workaround for temporary language shortcomings.

sergdort:
Yep, I see you point :)
Then it will live in my extensions =)

PR #847, opened 2016-08-25
Moves the shared schemes from projects to the workspace.

mohsenr:
Given the ABI instability of Swift, one of the more robust way of importing frameworks is to embed the framework’s project directly in the client framework or workspace. The shared schemes in Rx currently add a lot of noise with schemes for 4 frameworks times 4 platforms.

By moving these shared schemes from the project to the workspace, anyone working on Rx itself has access to them, whilst anyone just embedding the project as a dependency won’t see them.


kzaher:
Hi @mohsenr ,

I think we need those schemes to be in project because of Carthage. Please correct me if I'm wrong.

mohsenr:
Not as far as I know. I'm actually using Rx with Carthage myself. Any other dependencies I use don't have any shared project schemes, so it can't be a requirement.

The scripts (e.g. for test) are also using the workspace.

kzaher:
Hi @mohsenr ,

this is what I get:
```
*** Fetching RxSwift
*** Checking out RxSwift at "22921de5c2b0e58f103bcaa45d0f0166db09ac7f"
*** xcodebuild output can be found in /var/folders/l2/yqd_k65j58v70n4lftbxl_fw0000gn/T/carthage-xcodebuild.nEcoJw.log
xcodebuild timed out while trying to read RxExample.xcodeproj 😭
```

with the following Cartfile

```
git "https://github.com/mohsenr/RxSwift.git" "xcode-schemes"
```

There is a show check box left of each scheme. If you uncheck it the corresponding scheme would be hidden. Would that solution work for you?

mohsenr:
Thanks for checking @kzaher, you’re right! I didn’t have that problem because I rely on Xcode to build the dependencies itself – it’s a lot smarter :-) .

Unchecking the “show” button doesn’t work reliably unfortunately, for some reason it resets itself when I update.

Oh well – there’s not much that can be done unless Carthage is updated, so I'll close this for now.

Issue #852, opened 2016-08-28
Initial value for the observable


nonameplum:
Hi,

I there a way to post an event when new subscriber subscribe to the observable?
Ideally for me would be to use e.g. `PublishSubject` that will post first initial value for every new subscriber. This initial value for every new subscriber may change over time.

Thanks in advance :)

Chlup:
I think that you are looking for `BehaviorSubject`.

nonameplum:
Not exactly, because `BehaviorSubject` will send for every new subscriber last event from the stream. I don't need this. I want to send for every new subscriber my own first event that may, but shall not be the last event from the stream. I think that something like `doOnSubscribe` that I could use to make `.onNext(initialEvent)` inside.

kzaher:
Hi @nonameplum ,

maybe something like:

```swift
let resultSequence = Observable.deferred {
    return originalSequence.startWith(yourInitialValue)
}
```

nonameplum:
Look quite good. One problem only that I have. I need to pass `yourInitialValue` as an array because I don't know how many events I should send (it depends during runtime). But start with can't be filled using an array :/. `StartWith` class is private so I have no idea how to get around this problem -> http://stackoverflow.com/questions/24024376/passing-an-array-to-a-function-with-variable-number-of-args-in-swift

kzaher:
This should be solved.

PR #854, opened 2016-08-29
Support array parameter for the startWith method.

nonameplum:
There is a lack of the possibility to call `startWith` with an array because of the current swift compiler implementation (Swift 2.2).

There is no way to e.g. create this code because of this:
```swift
let resultSequence = Observable.deferred { [unowned self] in
    let yourInitialValue = self.someEvents // Here we have array of events that should be sent at first
    return originalSequence.startWith(yourInitialValue) // Here will be compilation error because array can't be passed to the variadic parameter
}
```

Theres is also related conversation to this topic https://github.com/ReactiveX/RxSwift/issues/852

### DIFF
--- a/RxSwift/Observables/Observable+Single.swift
+++ b/RxSwift/Observables/Observable+Single.swift
@@ -168,6 +168,20 @@ extension ObservableType {
         -> Observable<E> {
         return StartWith(source: self.asObservable(), elements: elements)
     }
+    
+    /**
+    Prepends a sequence of values to an observable sequence.
+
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @nonameplum ,

There is a way to send those elements.

3.x version
```swift
let resultSequence = Observable.deferred { [unowned self] in
    return Observable.from(elements).concat(originalSequence)
}
```

2.x version
```swift
let resultSequence = Observable.deferred { [unowned self] in
    return elements.toObservable().concat(originalSequence)
}
```

PR #855, opened 2016-08-30
implement UIWebView+Rx.swift

AndrewSB:
adds UIWebView delegate methods to RxCocoa-iOS

PR #857, opened 2016-08-30
Implement UIWebView+Rx

AndrewSB:
continuation of #855

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -249,6 +249,7 @@ custom_categories:
   - UITextView+Rx
   - UIView+Rx
   - UIViewController+Rx
+  - UIWebView+Rx
 - name: RxCocoa/iOS/Protocols
   children:
   - RxCollectionViewDataSourceType
@@ -266,6 +267,7 @@ custom_categories:
   - RxTableViewDelegateProxy
   - RxTextStorageDelegateProxy
   - RxTextViewDelegateProxy
+  - RxWebViewDelegateProxy
 - name: RxCocoa/macOS
   children:
   - NSButton+Rx
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 
 ## Master
 
+* adds `UIWebView` extensions:
+  * `didStartLoad`
+  * `didFinishLoad`
+  * `didFailLoad`
+
 ## [3.0.1](https://github.com/ReactiveX/RxSwift/releases/tag/3.0.1) (Xcode 8 / Swift 3.0 compatible)
 
 #### Anomalies
@@ -141,6 +146,7 @@ any observers or `forwardToDelegate` wasn't implementing `UITableViewDataSource.
   * `textDidEndEditing`
 * Moves `CLLocationManager` and `UIImagePickerViewController` extensions from RxCocoa to RxExample project. #874
 * Adds matrix CI builds.
+=======
 
 ## [3.0.0.alpha.1](https://github.com/ReactiveX/RxSwift/releases/tag/3.0.0.alpha.1) (Xcode 8 beta 6 compatible 8S201h)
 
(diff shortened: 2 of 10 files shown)


kzaher:
Hi @AndrewSB ,

It looks like unit tests are failing for some reason. Do they pass when you run `./scripts/all-tests.sh` locally?

AndrewSB:
@kzaher I took a look at the tests locally, and was able to fix the issues travis *was* having 👍 

But it turns out that the three delegate methods I'm testing (`didStartLoad`, `didFinishLoad`, and `didFailLoad `) are async, so the `XCTAssert`'s are failing because the tests are being run synchronously. 
Do you have any experience with this kind of situation?

kzaher:
Hi @AndrewSB ,

you won't be able to load anything from the internet since that in unreliable to have in unit tests.

Please create a small html file inside test target. You can use blocking operators to make sure that method will be called and unit test waits until it receives some event.

AndrewSB:
@kzaher I've updated the tests to not hit the network, but now travis seems to be failing because of danger: `The command "if [ "${DANGER}" -eq 1 ]; then bundle exec danger --verbose; else xcodebuild -version; xcodebuild -showsdks; ./scripts/all-tests.sh "${TEST}"; fi" exited with 65.`

Not sure what's up

AndrewSB:
@kzaher anything else holding this back from being merged?

sergdort:
@AndrewSB  CI fail 😉

AndrewSB:
Ahh, I looked closer at the CI message, I thought it was an issue with the CI, not the code I wrote.

AndrewSB:
Not sure why this is failing,
It seems to be something with Travis?

```
Done. Your build exited with 1.
/Users/travis/build.sh: line 109: shell_session_update: command not found
```

kzaher:
Hi @AndrewSB ,

yeah, travis is behaving weird. I'm assuming it's because of xcodebuild and simulator startup time.

There are two things we would need to change to merge this.

I think the library code is ok, but unit tests look little weird right now, so hoping these guidelines would help you :)
I think it's simplest to just have something like https://github.com/ReactiveX/RxSwift/blob/master/Tests/RxCocoaTests/UITableView%2BRxTests.swift#L58-L62
We only need to make sure that when delegate method fires event is being dispatched through observable sequence.
The other test we need to add is delegate proxy subclassing test. https://github.com/ReactiveX/RxSwift/blob/master/Tests/RxCocoaTests/DelegateProxyTest%2BUIKit.swift#L108-L110

It would be also great if you could squash all of those commits into a couple of them (ideally one, but a couple is also ok).

AndrewSB:
I'm having some trouble understanding the delegate proxy subclassing test. I've created the `UIWebViewSubclass` that conforms to `TestDelegateControl`, but right now I'm crashing on `checkSelectorIsObservable`...

Why would my subclass respond to `testEventHappened:`? I don't implement that selector for my delegate proxy - should I?

AndrewSB:
Hmm, I was able to fix the test failure by adding this line of code `extension RxWebViewDelegateProxy: TestDelegateProtocol {}`. What is the purpose of the test? It seems contrived to unknowing little me 👶

AndrewSB:
Can you comment more directly on what you dislike of the style? Is it that I've created an `htmlString`, and I'm using that to pass the tests? If so, I can do this instead https://github.com/ReactiveX/RxSwift/pull/857/commits/378dd5ba6a46df189a9b2f042ee285175feb028d.
Or is it the use of `toBlocking().first()`? If that's the case, then I'm not sure how to rewrite the tests. If I write the test as a subscription that mutates a variable, and then send the webView a request to load, then my test fails (as the webView delegate methods are async). I need something that blocks for the tiniest amount of time

Once this is ready to go, I'll happily squash this into 2-3 commits 😄

ikesyo:
> Or is it the use of `toBlocking().first()?` If that's the case, then I'm not sure how to rewrite the tests. If I write the test as a subscription that mutates a variable, and then send the webView a request to load, then my test fails (as the webView delegate methods are async). I need something that blocks for the tiniest amount of time

You can use `XCTestExpectation`.

AndrewSB:
Can I ask another question? Why do some delegate proxies require delegate methods overridden, while others don't?

For instance, the web view delegate proxy doesn't require `webViewDidFinishLoading(_:)` implemented, but `RxScrollViewDelegateProxy` implements `scrollViewDidScroll`. Why is that?

AndrewSB:
@ikesyo I've used the super cool async `XCTestExpectation`
@kzaher I've rebased my commits

Let me know if there's anything else I should do!

AndrewSB:
@ikesyo I realized I could write the tests by just calling the delegate methods, as @kzaher said in https://github.com/ReactiveX/RxSwift/pull/985#issuecomment-259265107, so I removed `XCTestExpectation`, but thank you for showing it to me!

AndrewSB:
🙏

PR #896, opened 2016-09-18
Add `rx.` syntax on Types.

JegnuX:
Following #832, this pull request adds `rx.` dot syntax to types through static var.
Thus instead of `Reactive<UIControl>.value(...)` one can write `UIControl.rx.value(...)`.


devxoul:
I like this. @kzaher how do you think of this?

kzaher:
@JegnuX I'll also move this to RxSwift. You were right, it would be a better fit there.

PR #906, opened 2016-09-24
Change UIRefreshControl.rx.refreshing to ControlProperty

ishkawa:
I think isRefreshing of UIRefreshControl should be *observable*.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ## Master
 
 * Update Getting Started document, section on creating an observable that performs work to Swift 3.0.
+* Changes `refreshing` property of `UIRefreshControl` from `AnyObserver<Bool>` to `ControlProperty<Bool>`.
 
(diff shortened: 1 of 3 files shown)


kzaher:
Hi @ishkawa ,

I can see your reasoning here. 

The underlying principle of all cocoa extensions is that only those events that are produced by the user should be observable (following Apple conventions here, and they make sense IMHO), and programatic updates shouldn't be.

There were a lot of issues opened here because of this here, but the gist is that if there is some programmatic update, then one should observe the model value directly, and not it's cache in view layer, otherwise you will probably get infinite loops and stack overflows.

If you need this feature you can just create a variable or some other observable model and bind it to `UIRefreshControl`. If one needs refreshing value, it would and observe it directly from variable/model.

ishkawa:
> The underlying principle of all cocoa extensions is that only those events that are produced by the user should be observable (following Apple conventions here, and they make sense IMHO), and programatic updates shouldn't be.

Got it. Simple and robust principle 👍 
 I'm going to close this PR.

PR #910, opened 2016-09-26
Added bindable backgroundColor property to UIView

BenchR267:
As already described in the title, I added a bindable property to UIView which makes it possible to bind other observables to that property.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ## Master
 
+* Added backgroundColor observable extension to UIView.
+
 ## [3.0.0-beta.2](https://github.com/ReactiveX/RxSwift/releases/tag/3.0.0-beta.1) (Xcode 8 / Swift 3.0 compatible)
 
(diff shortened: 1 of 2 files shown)


BenchR267:
Is anyone ever reviewing this?

sergdort:
Hi @BenchR267 
I pretty sure your PR can be adressed [to this issue](https://github.com/ReactiveX/RxSwift/issues/361)

BenchR267:
@sergdort so in short: there are only a few given bindably properties and everything else should be extended in the project?

kzaher:
Hi @BenchR267 ,

I guess we need to think about how to define criteria what is considered an extension that we can support in RxCocoa and what isn't.

So far we've only included what I've considered basic data properties. It's not feasible to have one rx property for one UIKit property. 

I would like to either find some generic way to define those simple extensions or a simple rules that  limit the number of those extensions to some reasonable number.

I guess if there will be a lot of requests for some particular extension we can make exceptions.

BenchR267:
@kzaher Thanks for the explanation, makes totally sense. One bindable rx property for every UIKit property is not maintainable :)

Issue #922, opened 2016-09-29
Convert a Subject to a Driver


AndrewSB:
Is there a subject that behaves similarly to a driver? As in it can't take an error?

Currently I have some `PublishSubject`s that I only pass `.next` events to, and I pass them to views as `Driver`s by:

```
let scanning = PublishSubject<Bool>()
let scanningDriver = scanning.asDriver { _ in assertionFailure(); return .never() }
```

Is there anyway for me to compose the same logic without an assertionFailure?
IMHO it makes sense to leverage the compiler here, to make it impossible for me to `on(.error)` to my scanning `PublishSubject`, instead of me remembering that that shouldn't be allowed

kzaher:
Hi @AndrewSB ,

you can easily create your own concept based on `PublishSubject` that does that. Please take a look at how we've wrapped `BehaviorSubject` inside `Variable`.

I guess we can maybe add an `Event` to `RxCocoa`.

AndrewSB:
Hey @kzaher, yeah. I understand it would be just like `Variable`, another wrapper for a subject. I'd love to take a shot at implementing, I'll start a Pull Request soon!

AndrewSB:
Yeah, I just tried to create the `Event` subject, Xcode complained:
<img width="727" alt="screen shot 2016-10-06 at 6 20 14 pm" src="https://cloud.githubusercontent.com/assets/3814772/19176112/a2f5b93e-8bf1-11e6-9fa3-3f42cbbdbf8f.png">

patrickbdev:
@AndrewSB is there any update on this feature? This would be incredibly useful.

AndrewSB:
Yup @patrickbdev! This was superseded by #928. It looks like @DenHeadless is pretty interested in helping out too. Maybe we can take a shot at implementing it together 😄

PR #928, opened 2016-10-07
[WIP] Create a Subject that can't error

AndrewSB:
@kzaher started working on this from #922 

Facing two things currently:
1. Not a huge fan of the name `ErrorlessSubject`, would love suggestions to change it to anything else. You mentioned `Event`, but I had an issue implementing that (https://github.com/ReactiveX/RxSwift/issues/922#issuecomment-252131282)
2. I'm not sure how best to implement this. I'll take some time and look at the implementation of `Driver`, see if that takes me anywhere.
3. Should this be in the `RxSwift`, or `RxCocoa` target?

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -18,6 +18,10 @@
 		271A97441CFC9F7B00D64125 /* UIViewControler+RxTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 271A97421CFC99FE00D64125 /* UIViewControler+RxTests.swift */; };
 		46307D4E1CDE77D800E47A1C /* UIAlertAction+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = 46307D4D1CDE77D800E47A1C /* UIAlertAction+Rx.swift */; };
 		46307D4F1CDE77D800E47A1C /* UIAlertAction+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = 46307D4D1CDE77D800E47A1C /* UIAlertAction+Rx.swift */; };
+		46C882C61DDBD8CA0032BCA7 /* Event.swift in Sources */ = {isa = PBXBuildFile; fileRef = 46C882C51DDBD8CA0032BCA7 /* Event.swift */; };
+		46C882C71DDBD8CA0032BCA7 /* Event.swift in Sources */ = {isa = PBXBuildFile; fileRef = 46C882C51DDBD8CA0032BCA7 /* Event.swift */; };
+		46C882C81DDBD8CA0032BCA7 /* Event.swift in Sources */ = {isa = PBXBuildFile; fileRef = 46C882C51DDBD8CA0032BCA7 /* Event.swift */; };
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @AndrewSB ,

I would need to think about where should be put this, and how should it be called.

The reason why we've added `Variable` to RxSwift is because the chances of somebody new to the library finding `BehaviorSubject` was so slim.

I wouldn't want to deviate too much from other implementations.

kzaher:
Hi @AndrewSB ,

I've been thinking about this, and right now we have a nice pipeline for state propagation.

BehaviorSubject or Variable or ControlProperty (write part) -> Driver (read only view with certain guarantees)

And it really looks like we don't have such nice pipeline for event modeling. 

I think that we could add to RxCocoa `Publisher` that wraps `PublishSubject` and can't error out in the same way `Variable` wraps `BehaviorSubject`.

Something like
```
class Publisher<E> {
     func publish(event: E)

     func asObservable() -> Observable<E>
}
```

Maybe we can create a specialization for `SharedSequence` with `Publish` strategy and name it something like `EventBus`/`Event`/`EventSource`. Still unsure about the name. Open to suggestions.

Then one could have:
```swift
let publisher = Publisher<()>()
let events = publisher.asEventSource().map { }.filter { }.withLatestFrom(stateDriver) .....

publisher.publish()
```

I still need to make loosen constraints on SharedSequence so it would be possible to combine `Driver` and `EventSource`ish sequence.

Maybe it would make sense for RxSwift 4.0 to move `Variable`, `Driver`, `Event`, `ControlProperty`, `ControlEvent` from RxCocoa to separate RxUnits project and keep RxCocoa pure and focused on just extensions, but let's not think about this for now. I think it's fine for RxCocoa to contain them for now.

In the future if Swift 4.0 adds generic error parametrization and we add generic error type to `PublishSubject` it would still make sense to have something like `Publisher` for easier interaction with imperative parts of the app.

AndrewSB:
@kzaher that sounds absolutely lovely. `Publisher` and `Variable` sound great to me.

I'll wait for you to
> loosen constraints on SharedSequence so it would be possible to combine Driver and EventSourceish sequence.

and then continue with this PR. Give me a holler when you're done?

kzaher:
Hi @AndrewSB ,

I think we can add `Publisher` that is equivalent to `Variable` and `EventBus` that is equivalent to `Driver`.

While Googling, I've accidentally also found some other project that is using the same terminology :)

https://github.com/greenrobot/EventBus

kzaher:
I was maybe a bit too quick on naming, but don't have any better ideas.

DenHeadless:
Would really want this subject to exist. I've been searching for this kind of functionality recently, trying to replicate what Kickstarter guys did with their ViewModel event transformation like so: 
[https://github.com/kickstarter/ios-oss/blob/master/Library/ViewModels/CommentsViewModel.swift#L177-L215](url)

For events that have default value it's nice to use `Variable`. But for events that have not, for example tap a login button i'm forced to use PublishSubject, even though i know for sure that it will never error - i'm only sending next events, and transform them into outputs.

`Publisher` sounds nice, i would like to help make this, if i can. Was there any progress on this?

kzaher:
I think this is superseded by #1116

AndrewSB:
Yes, it is 😊

PR #956, opened 2016-10-24
Publicize SharedSequence arity static functions

AndrewSB:
I can't find these functions outside of the RxCocoa target 😮

That seems like a problem

### DIFF
--- a/RxCocoa/CocoaUnits/SharedSequence/SharedSequence+Operators+arity.tt
+++ b/RxCocoa/CocoaUnits/SharedSequence/SharedSequence+Operators+arity.tt
@@ -15,7 +15,7 @@ import RxSwift
 
 // <%= i %>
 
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @AndrewSB ,

We have public modifiers on extensions methods, so you should be able to see them. You've made changes to `SharedSequence+Operators+arity.tt`, which are template files and aren't actually being compiled but used to generate swift code.

You probably need to just move from `[].zip` to `Driver.zip([])` syntax.

AndrewSB:
Oh, so strange. I tried moving to that syntax earlier and my code wouldn't compile.
I have a feeling that it was related to me not having the correct version of RxSwift, I was having some issues with my dependencies.

I'm sorry about creating this issue before properly investigating myself 😄

PR #957, opened 2016-10-24
Fix NSTextInputClient conformance issues with NSTextField

tonyarnold:
This PR fixes a potential crasher on macOS by addressing 2 issues:

1. The `NSTextInput` protocol was deprecated in OS X 10.6 (Snow Leopard) - we should be using `NSTextInputClient` now.
2. Not all `NSTextField`s conform to `NSTextInputClient` (as you see when putting the non-deprecated API into the empty extension here). I've just constrained the extension on `Reactive` to only work with `NSTextField` instances that conform to `NSTextInputClient`.

### DIFF
--- a/RxCocoa/Common/TextInput.swift
+++ b/RxCocoa/Common/TextInput.swift
@@ -53,7 +53,7 @@ import Foundation
     import Cocoa
 
     /// Represents text input with reactive extensions.
-    public struct TextInput<Base: NSTextInput> {
+    public struct TextInput<Base: NSTextInputClient> {
         /// Base text input to extend.
         public let base: Base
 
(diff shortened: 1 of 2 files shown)


tonyarnold:
I'm still not sure that this should even be targeting `NSTextField` and not `NSTextView` which _actually_ conforms to `NSTextInputClient`.

kzaher:
Hi @tonyarnold ,

this is really weird. We had practically exact same code for about a year, but yeah, it's a bug.

For `__block IMP __nonnull targetImplementation = nil;`, the problem is that it shouldn't be annotated with `__nonnull`, I'll cherry-pick your first commit and fix the other error you've noticed.

kzaher:
Ok, I've cherry-picked the commit into `develop`.

tonyarnold:
Thanks @kzaher!

PR #960, opened 2016-10-25
Small changes and simplifications

moogle19:
I was going though the project to get a better insight of what I am using day to day 😃. 
During that I noticed some things, that (in my opinion) can be solved a bit nicer.

It would be nice if you would look at it and tell me what you think about it.

### DIFF
--- a/Platform/DataStructures/Bag.swift
+++ b/Platform/DataStructures/Bag.swift
@@ -127,14 +127,14 @@ struct Bag<T> : CustomDebugStringConvertible {
     mutating func removeKey(_ key: BagKey) -> T? {
         if _key0 == key {
             _key0 = nil
-            let value = _value0!
+            let value = _value0
             _value0 = nil
             return value
         }
 
         if _key1 == key {
             _key1 = nil
-            let value = _value1!
+            let value = _value1
             _value1 = nil
             return value
         }
@@ -143,15 +143,12 @@ struct Bag<T> : CustomDebugStringConvertible {
             return existingObject
         }
 
-        for i in 0 ..< _pairs.count {
-            if _pairs[i].key == key {
-                let value = _pairs[i].value
-                _pairs.remove(at: i)
-                return value
-            }
+
+        guard let index = _pairs.index(where: { $0.key == key }) else {
+            return nil
         }
 
-        return nil
+        return _pairs.remove(at: index).value
     }
 }
 
@@ -167,7 +164,7 @@ extension Bag {
     ///
     /// - parameter action: Enumeration closure.
     func forEach(_ action: (T) -> Void) {
-        if _onlyFastPath {
+        guard !_onlyFastPath else {
             if let value0 = _value0 {
                 action(value0)
             }
@@ -187,15 +184,9 @@ extension Bag {
             action(value1)
         }
 
-        for i in 0 ..< pairs.count {
-            action(pairs[i].value)
-        }
+        pairs.map({ $0.value }).forEach(action)
 
-        if dictionary?.count ?? 0 > 0 {
-            for element in dictionary!.values {
-                action(element)
-            }
-        }
+        dictionary?.values.forEach(action)
     }
 }
 
--- a/RxSwift/Disposables/RefCountDisposable.swift
+++ b/RxSwift/Disposables/RefCountDisposable.swift
@@ -34,10 +34,10 @@ public final class RefCountDisposable : DisposeBase, Cancelable {
      */
     public func retain() -> Disposable {
         return _lock.calculateLocked {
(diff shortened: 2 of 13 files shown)


Platform/DataStructures/Bag.swift
```diff
@@ -127,14 +127,14 @@ struct Bag<T> : CustomDebugStringConvertible {
     mutating func removeKey(_ key: BagKey) -> T? {
         if _key0 == key {
             _key0 = nil
-            let value = _value0!
+            let value = _value0
```
moogle19: No need to force unwrap, because return value is an optional

Platform/DataStructures/Bag.swift
```diff
@@ -143,15 +143,12 @@ struct Bag<T> : CustomDebugStringConvertible {
...
 
-        for i in 0 ..< _pairs.count {
-            if _pairs[i].key == key {
-                let value = _pairs[i].value
-                _pairs.remove(at: i)
-                return value
-            }
+
+        guard let index = _pairs.index(where: { $0.key == key }) else {
+            return nil
```
moogle19: Get the index of the pair, which needs to be removed with the build in index function instead of `for each` loop

Platform/DataStructures/Bag.swift
```diff
@@ -143,15 +143,12 @@ struct Bag<T> : CustomDebugStringConvertible {
...
         }
 
-        return nil
+        return _pairs.remove(at: index).value
```
moogle19: `remove(at:)` returns the removed element. Just use that as return value

Platform/DataStructures/Bag.swift
```diff
@@ -167,7 +164,7 @@ extension Bag {
     ///
     /// - parameter action: Enumeration closure.
     func forEach(_ action: (T) -> Void) {
-        if _onlyFastPath {
+        guard !_onlyFastPath else {
```
moogle19: Because it returns after that expression is true, I think a guard statement makes that more clear

fpillet:
I would be *very* careful with simplifications. In many cases, performance has been tuned and the code may simply be faster as it is written (although may not always be the case, but I know a lot of effort went into making the code as fast as possible for example).

It would be great to have some timings proving that the simplifications are not being made at the expense of execution time. RxSwift really is the backbone for many applications and, as such, is expected to be as fast as possible

Platform/DataStructures/Bag.swift
```diff
@@ -187,15 +184,9 @@ extension Bag {
...
-            action(pairs[i].value)
-        }
+        pairs.map({ $0.value }).forEach(action)
 
-        if dictionary?.count ?? 0 > 0 {
-            for element in dictionary!.values {
-                action(element)
-            }
-        }
+        dictionary?.values.forEach(action)
```
moogle19: Use optional chaining instead of `??` and `for each`

RxSwift/Disposables/RefCountDisposable.swift
```diff
@@ -34,10 +34,10 @@ public final class RefCountDisposable : DisposeBase, Cancelable {
      */
     public func retain() -> Disposable {
         return _lock.calculateLocked {
-            if let _ = _disposable {
+            if _disposable != nil {
```
moogle19: I think `!= nil` looks nicer and is better to read than `if let _`

RxSwift/Disposables/RefCountDisposable.swift
```diff
@@ -34,10 +34,10 @@ public final class RefCountDisposable : DisposeBase, Cancelable {
...
 
                 do {
-                    let _ = try incrementChecked(&_count)
+                    _ = try incrementChecked(&_count)
```
moogle19: No need for `let` here

RxSwift/Disposables/SerialDisposable.swift
```diff
@@ -64,14 +62,12 @@ public final class SerialDisposable : DisposeBase, Cancelable {
 
     private func _dispose() -> Disposable? {
         _lock.lock(); defer { _lock.unlock() }
-        if _isDisposed {
+        guard !_isDisposed else {
```
moogle19: Same as above. Because it is returning, `guard` is more expressive here.

moogle19:
@fpillet 
That's a good point. I haven't really thought about that.
I did some testing myself and found that most of my 'simplifications' were slower than the previous method.

For example `let _ = x` is much fast than `x != nil` (and by much I mean like 20x)

`for x in array` is also faster than `array.forEach`

Sorry for bothering you with this Pull Requst 😔 , but at least I know a bit more about Swift now! 😉

PR #961, opened 2016-10-25
Add Optional<Equatable> .distinctUntilChanged

AndrewSB:
This is implemented for other collections like Array and Dictionary, so I thought it would be intuitive to include for Observables.
I was quite confused when I couldn't write just `.distinctUntilChanged()` for one of my `Observable<Type?>`s, and instead had to fill it out with this `switch`

Not sure about the type erased `AnyOptional` protocol I defined, and where it should be.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ## Master
 
+* Adds a `.distinctUntilChanged()` operator for `Observable`s & `SharedSequence`s of `Element`s that are `Optional`s where the `Wrapped` type conforms to `Equatable`
+
 ## [3.0.0](https://github.com/ReactiveX/RxSwift/releases/tag/3.0.0) (Xcode 8 / Swift 3.0 compatible)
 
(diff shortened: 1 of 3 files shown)


kzaher:
Hi @AndrewSB ,

I understand your issue ....

I think this looks simple on surface and in 99% of cases it would work, but there are problems with this solution :)

I don't think we should define `AnyOptional` here and introduce into RxSwift namespace additional protocol that isn't related with core functionality and `Optional` namespace with `asOptional` property :)

I think situation would be more vague if we already had `AnyOptional` protocol in Swift standard library :)

I'm really trying to make this library as small as we can make it :) I also have convenience operators in my projects, but they probably are never going to end up here :) I have a feeling this might also fit into that category :)

There is already RxOptional, and this functionality is maybe more suitable for that library.

AndrewSB:
That makes complete sense @kzaher, I'm all for keeping RxSwift as small as possible.

I'll look into RxOptional, I'm not currently using it, but they probably make a number of my extensions redundant 👍 

Thanks @kzaher

PR #964, opened 2016-10-27
Add take operator to SharedSequence

Econa77:


kouki-dan:
+1

kzaher:
Hi @Econa77 ,

I'm curious why do you want to include `take` operator :) Using sharing strategies on terminating sequences could cause unexpected side-effects.

e.g.

```
let state: Driver<State>

state.drive(someUI)

.... then after some time

state.drive(somethingElse)
```

If `state` is something non-terminal, then sharing makes sense. If state is something that is known to terminate in finite time, maybe just modeling that as normal observable sequence that represents some operation is more expressive.

I'm curious what use case do you have?

kzaher:
No reply.

PR #973, opened 2016-11-01
Fixes Release compilation issue (#970).

### LINKED ISSUE #970: Patch release blocker `duplicate symbol __TMRbBp` (https://bugs.swift.org/browse/SR-3081)
We would like to release a patch for issues with RxCocoa on OSX, but we have ran into unexpected issues with Xcode 8.1

When running tests in `Release` or `Release-Tests` configuration we are getting  `duplicate symbol __TMRbBp`.

I'm curious about a few of things.

Does anyone have information when is patch for this coming out, or what's causing this?
Does anyone need an urgent patch release for RxCocoa?

fe9lix:
Fixes #970: Set SWIFT_REFLECTION_METADATA_LEVEL build settings to "none" as temporary workaround for Swift compiler issue (https://bugs.swift.org/browse/SR-3081).


kzaher:
Awesome find,

I've just made some small change to your solution so we can ship this. I've only turned it off for unit tests, and it seems to be enough.

I've merged your commit from terminal, but had to rebase after so github didn't catch the merge.
https://github.com/ReactiveX/RxSwift/commit/e4ff4680d757070c6d5c0518809172c57c9c262e

PR #975, opened 2016-11-02
Fix incorrect Slack URL in GitHub Issue template

revolter:
The Slack URL had the protocol duplicated. Also, it just redirects to the GitHub project page. Shouldn't it be http://rxswift-slack.herokuapp.com instead?


revolter:
@kzaher, Thanks for merging! What about my last question?

PR #985, opened 2016-11-07
[WIP] Add didScroll & didZoom delegate methods to UIScrollView+Rx

AndrewSB:
tests are currently failing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ## Master
 
+* Adds `didScroll` and `didZoom` `ControlEvent`s to `UIScrollView+Rx`
+
 ## [3.0.1](https://github.com/ReactiveX/RxSwift/releases/tag/3.0.1) (Xcode 8 / Swift 3.0 compatible)
 
 #### Anomalies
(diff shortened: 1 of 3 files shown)


kzaher:
Hi @AndrewSB ,

I think we can merge those extensions, but first please take a look how we've written unit tests for UITableView.
https://github.com/ReactiveX/RxSwift/blob/master/Tests/RxCocoaTests/UITableView%2BRxTests.swift#L58-L62

It seems that also unit test for `didScrollToTop` are missing.

AndrewSB:
@kzaher thank you for showing me those tests, they make so much more sense.

I'm having an issue getting the `testScrollViewDidScroll` to pass, and I have a feeling that it's something to do with how "`Delegate proxy is already implementing `scrollViewDidScroll:`, a more performant way of registering might exist.`".
Any ideas on how I should properly implement that?

Also, I've implemented one [test](https://github.com/AndrewSB/RxSwift/blob/f8993b3a432b349a3c9e004be2b468044de5ca3f/Tests/RxCocoaTests/UIScrollView%2BRxTests.swift#L86-L102) for the delegate, making sure it's a weak reference. Not sure if I should write more or not?

kzaher:
Hi @AndrewSB ,

yes, `didScroll` is already implemented for `contentOffset` property.

`didScroll` = `contentOffset.skip(1)`

The failure is explained here https://travis-ci.org/ReactiveX/RxSwift/jobs/176254298.

You can also just run unit tests locally and find out what the error is.

PR #986, opened 2016-11-07
add UITextField shouldReturn delegate message

AndrewSB:
@ikesyo: this is where I had to use https://github.com/ReactiveX/RxSwift/pull/857#issuecomment-258753754, for some reason the delegate message wouldn't forward until I added https://github.com/ReactiveX/RxSwift/pull/986/files#diff-830021a0e1ff14718758faddab24a1f4R32

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -264,6 +264,7 @@ custom_categories:
   - RxTabBarDelegateProxy
   - RxTableViewDataSourceProxy
   - RxTableViewDelegateProxy
+  - RxTextFieldDelegateProxy
   - RxTextStorageDelegateProxy
   - RxTextViewDelegateProxy
 - name: RxCocoa/macOS
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ## Master
 
+* Add `UITextField.shouldReturn` control event for delegate message `textFieldShouldReturn(_:)`
+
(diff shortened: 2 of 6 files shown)


kzaher:
Hi @AndrewSB ,

all methods on `UITextFieldDelegate` have return value as far as I can tell. That means there aren't any methods that are used as a notification mechanism (`void` return value) that are useful to observe.

AndrewSB:
@kzaher I was using this delegate method to receive a notification for when the Next/Done key was pressed.

Does it make sense to create a notification mechanism for that? If it does, should the RxDelegateProxy default to `super.<delegate method>` instead of `true` here https://github.com/ReactiveX/RxSwift/pull/986/files#diff-830021a0e1ff14718758faddab24a1f4R34?

kzaher:
Hi @AndrewSB ,

you can already observe next/done button by doing `.rx.controlEvent(.editingDidEndOnExit)` extension.

```
public static var editingDidEndOnExit: UIControlEvents { get } // 'return key' ending editing
```

AndrewSB:
@kzaher thank you for showing that to me, I had been using `.editingDidEnd`.

kioshimafx:
@kzaher hmm but .editingDidEndOnExit close keyboard after tap on return. Example: I need send message in chat, but i don't want close keyboard after tap send(return) button.

AndrewSB:
I think the keyboard closing makes sense, it's an apple API and it says
editing did end on exit. On exit, to me, says that they user exited the
editing session & the keyboard is now closed.
You should be looking for a different control event probably

switchawk:
The problem with `.editingDidEnd` (I think) is that it triggers also if another component becomes first responder. Without the user having pressed Enter.

PR #1007, opened 2016-11-18
update GettingStarted.md

demianturner:
Swift 3 compatibility

### DIFF
--- a/Documentation/GettingStarted.md
+++ b/Documentation/GettingStarted.md
@@ -1062,16 +1062,17 @@ Request determines is it a GET request, or a POST request, what is the request b
 This is how you can create a simple GET request
 
 ```swift
-let request = NSURLRequest(URL: NSURL(string: "http://en.wikipedia.org/w/api.php?action=parse&page=Pizza&format=json")!)
+let req = NSURLRequest(url: NSURL(string: "http://en.wikipedia.org/w/api.php?action=parse&page=Pizza&format=json") as! URL)
 ```
 
 If you want to just execute that request outside of composition with other observables, this is what needs to be done.
 
 ```swift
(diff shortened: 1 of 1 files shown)


Documentation/GettingStarted.md
```diff
@@ -1062,24 +1062,25 @@ Request determines is it a GET request, or a POST request, what is the request b
 This is how you can create a simple GET request
 
 ```swift
-let request = NSURLRequest(URL: NSURL(string: "http://en.wikipedia.org/w/api.php?action=parse&page=Pizza&format=json")!)
+let req = NSURLRequest(url: NSURL(string: "http://en.wikipedia.org/w/api.php?action=parse&page=Pizza&format=json") as! URL)
```
toshi0383: Shouldn't this be like this ?
```
let req = URLRequest(url: URL(string: "http://en.wikipedia.org/w/api.php?action=parse&page=Pizza&format=json")!)
```

Documentation/GettingStarted.md
```diff
@@ -1062,24 +1062,25 @@ Request determines is it a GET request, or a POST request, what is the request b
...
 If you want to just execute that request outside of composition with other observables, this is what needs to be done.
 
 ```swift
-let responseJSON = NSURLSession.sharedSession().rx.JSON(request)
+let responseJSON = URLSession.shared.rx.json(request: req as URLRequest)
 
 // no requests will be performed up to this point
 // `responseJSON` is just a description how to fetch the response
 
+
```
toshi0383: I think this diff is unnecessary.

Documentation/GettingStarted.md
kzaher: Yeah, this will need to be changed ... Good catch though. I'll just change this immediately after merge.

PR #1009, opened 2016-11-24
[RxCocoa] Rename 'refreshing' to 'isRefreshing'

devxoul:


kzaher:
Hi @devxoul ,

argh, I've missed that one. Good catch.

I think it would be great to add deprecated attribute and add an alias called `isRefreshing` to `refreshing`.

```
public var isRefreshing: UIBindingObserver<Base, Bool> {
      return refreshing
}
```

devxoul:
@kzaher, great idea! I marked it with `renamed` rather than `deprecated`.

sunshinejr:
How about the other way around? Rename `refreshing` to `isRefreshing` and make an alias? Reduces the amount of work in the future to just removing the alias. But this is really prefrerence I believe.

ikesyo:
> `@available(*, renamed: "isRefreshing")`

Just marking it `renamed` without `unavailable`: `@available(*, unavailable, renamed: "isRefreshing")` or `deprecated`: `@available(*, deprecated, renamed: "isRefreshing")` should be meaningless (compiler would not display any warning).

Furthermore, to keep backward compatibility, that should be `deprecated` rather than `unavailable` because the latter means a breaking API change. Introducing a breaking change requires major version update (RxSwift 4.0) for the sake of semantic versioning.

devxoul:
I marked it with `@available(*, deprecated, renamed: "isRefreshing")`. Thanks @ikesyo :+1:

Issue #966, opened 2016-10-27
Does somebody know why travis CI often returns error 65 after


kzaher:
```
▸ Touching AllTests-tvOS.xctest
▸ Signing build/Build/Products/Release-Tests-appletvsimulator/AllTests-tvOS.xctest
** TEST FAILED **
+++exitIfLastStatusWasUnsuccessful
+++STATUS=65
+++'[' 65 -ne 0 ']'
+++echo 65
65
+++exit 65
```

and how to make this more robust?

moogle19:
See: https://github.com/travis-ci/travis-ci/issues/4725

A workaround to get rid of the issue is to add `sleep 5` in `after_success`

kzaher:
I've implemented this and tried to reuse built in simulator instance. Hoping it will work in future.

PR #1014, opened 2016-11-26
Improve naming in GeolocationViewController

Czajnikowski:
I have moved "No Geolocation View" out of the `view` in `Geolocation.storyboard`, this required me to change implementation of `GeolocationViewController` a bit (adding/removing instead of hidding/showing a subview). I have also changed the binding names as I believe nouns describe better the meaning of `UIBingingObserver`s instead of `drive...` names.

Storyboard changes:

From "No Geolocation View" as a hidden subview:

zrzut ekranu 2016-11-24 o 16 16 56

Into scene dock:

zrzut ekranu 2016-11-26 o 17 57 33

### DIFF
--- a/RxExample/RxExample/Examples/GeolocationExample/GeolocationViewController.swift
+++ b/RxExample/RxExample/Examples/GeolocationExample/GeolocationViewController.swift
@@ -21,19 +21,6 @@ private extension Reactive where Base: UILabel {
     }
 }
 
-private extension Reactive where Base: UIView {
-    func subviewPresence(_ subview: UIView) -> UIBindingObserver<Base, Bool> {
-        return UIBindingObserver(UIElement: base) { view, show in
-            if !show {
-                subview.removeFromSuperview()
-            }
-            else {
-                view.addSubview(subview)
-            }
-        }
-    }
-}
-
 class GeolocationViewController: ViewController {
     
     @IBOutlet weak private var noGeolocationView: UIView!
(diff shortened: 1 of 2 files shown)


Czajnikowski:
I believe Travis failure is a glitch (same as this https://github.com/ReactiveX/RxSwift/issues/966 and https://github.com/fastlane/fastlane/issues/4197). Is there any way I can retry it with Travis?

RxExample/RxExample/Examples/GeolocationExample/GeolocationViewController.swift
```diff
@@ -14,23 +14,21 @@ import CoreLocation
...
-                view.isHidden = true
-                view.superview?.sendSubview(toBack:view)
+    func subviewPresence(_ subview: UIView) -> UIBindingObserver<Base, Bool> {
+        return UIBindingObserver(UIElement: base) { view, show in
+            if !show {
+                subview.removeFromSuperview()
             }
             else {
-                view.isHidden = false
-                view.superview?.bringSubview(toFront:view)
```
kzaher: Hi @Czajnikowski ,

It seems to me we should just remove `var driveAuthorization: UIBindingObserver<Base, Bool>` and use `subview.rx.hidden`

kzaher: Hi @Czajnikowski ,

are you suggesting to do that vs `rx.hidden` as an optimization? 

I would ideally want to reduce the amount of code as much as we can. You PR clearly improves the situation from where it was, but since we have started on this road, I would want to improve this part as much as we can.

Czajnikowski: Actually I have simplified it a bit by:
- totally removing hacky `UIBindingObserver` and using `rx.hidden`
- imperatively adding a `noGeolocationView` as a subview in `viewDidLoad` while still keeping it as a separate view in Storyboard

I believe it satisfies both my goal to improve readability of the example and your goal to keep code as exact as possible :)

PR #1023, opened 2016-12-07
Add RxTabBarControllerDelegateProxy

kitasuke:
- [x] Add RxTabBarControllerDelegateProxy
- [x] Passed AllTests

### DIFF
--- a/RxCocoa/iOS/Proxies/RxTabBarControllerDelegateProxy.swift
+++ b/RxCocoa/iOS/Proxies/RxTabBarControllerDelegateProxy.swift
@@ -1,5 +1,5 @@
 //
-//  RxTabBarControllerDelegate.swift
+//  RxTabBarControllerDelegateProxy.swift
 //  RxCocoa
 //
 //  Created by Yusuke Kita on 2016/12/07.
--- a/Tests/RxCocoaTests/UITabBarController+RxTests.swift
+++ b/Tests/RxCocoaTests/UITabBarController+RxTests.swift
@@ -1,0 +1,103 @@
+//
+//  UITabBarController+RxTests.swift
+//  Tests
+//
+//  Created by Yusuke Kita on 2016/12/07.
+//  Copyright © 2016 Krunoslav Zaher. All rights reserved.
+//
+
+#if os(iOS) || os(tvOS)
+
+import Foundation
+
+import RxSwift
+import RxCocoa
+import UIKit
+import XCTest
+
+class UITabBarControllerTests: RxTest {
+    let createSubject: () -> UITabBarController = { UITabBarController() }
+}
+
+/**
+ iOS only
+ */
+#if os(iOS)
+extension UITabBarControllerTests {
+    func testWillBeginCustomizing() {
+        let subject = createSubject()
+        
+        let viewControllers = [UIViewController()]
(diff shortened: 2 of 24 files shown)


Tests/RxCocoaTests/UITabBarController+RxTests.swift
```diff
@@ -0,0 +1,103 @@
...
+    func testWillBeginCustomizing() {
+        let subject = createSubject()
+        
+        let viewControllers = [UIViewController()]
+        var returnedViewControllers: [UIViewController]!
+        
+        _ = subject.rx.willBeginCustomizing
+            .subscribe(onNext: { vc in
+                returnedViewControllers = vc
+            })
```
kzaher: It would be great to add before `subject.delegate!.tabBarController!(subject, willBeginCustomizing: viewControllers)` `XCTAssertNotEqual(returnedViewControllers, viewControllers)`.

I'll need to add those checks everywhere where they aren't present. We actually had a bug because those checks weren't there :)

kitasuke: You're right. Maybe I could help adding those checks with different PR cause I know where they are's present.

kitasuke:
Thank you for your feedback! I fixed them, so please check them out.

Tests/RxCocoaTests/UITabBarController+RxTests.swift
kzaher: Yep, feel free to make another PR if you see those checks missing somewhere.

Tests/RxCocoaTests/UITabBarController+RxTests.swift
```diff
@@ -0,0 +1,113 @@
...
+    
+    func testWillEndCustomizing() {
+        let subject = createSubject()
+        
+        let viewControllers = [UIViewController()]
+        var returnedViewControllers = [UIViewController]()
+        var changed: Bool!
+        
+        XCTAssertNotEqual(returnedViewControllers, viewControllers)
+        XCTAssertNotEqual(changed, true)
```
kzaher: These methods should be after `subject.rx.willEndCustomizing` because we are testing that extensions haven't emitted anything initially.

RxCocoa/iOS/Proxies/RxTabBarControllerDelegateProxy.swift
```diff
@@ -0,0 +1,43 @@
...
+    
+    /// For more information take a look at `DelegateProxyType`.
+    public class func setCurrentDelegate(_ delegate: AnyObject?, toObject object: AnyObject) {
+        let tabBarController: UITabBarController = castOrFatalError(object)
+        tabBarController.delegate = castOptionalOrFatalError(delegate)
+    }
+    
+    /// For more information take a look at `DelegateProxyType`.
+    public override class func createProxyForObject(_ object: AnyObject) -> AnyObject {
+        let tabBarController = (object as! UITabBarController)
```
kzaher: I think this should be. 
```
let tabBarController: UITabBarController = castOrFatalError(object)
return tabBarController.createRxDelegateProxy()
```

kitasuke:
Thank you for your feedback again!

PR #1025, opened 2016-12-08
Update README.md

cyrilchandelier:
The example advertised in the Usage section of the README is not working with latest versions of RxSwift.

After the transition from `rx_text` to `rx.text`, the result given is an `String?` instead of `String` preventing the use of  `distinctUntilChanged()` without any comparer parameter. Using the `orEmpty` fix the issue.

Simply adding `orEmpty` isn't enough as the following lines generates a compilation issue:
```
if query.isEmpty {
    return .just([])
}
```

Since I don't know how to fix this issue and since it doesn't bring much value to the example, I removed it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,14 +78,10 @@ KVO observing, async operations and streams are all unified under [abstraction o
   </tr>
   <tr>
     <td><div class="highlight highlight-source-swift"><pre>
-let searchResults = searchBar.rx.text
+let searchResults = searchBar.rx.text.orEmpty
     .throttle(0.3, scheduler: MainScheduler.instance)
     .distinctUntilChanged()
     .flatMapLatest { query -> Observable<[Repository]> in
-        if query.isEmpty {
-            return .just([])
-        }
-
         return searchGitHub(query)
             .catchErrorJustReturn([])
     }


kzaher:
Hi @cyrilchandelier ,

thnx for reporting this.

I've tried does this compile

```
func searchGitHub(_ query: String) -> Observable<[Repository]> {
            return Observable.empty()
        }

        let searchResults = searchBar.rx.text.orEmpty
            .throttle(0.3, scheduler: MainScheduler.instance)
            .distinctUntilChanged()
            .flatMapLatest { query -> Observable<[Repository]> in
                if query.isEmpty {
                    return .just([])
                }
                return searchGitHub(query)
                    .catchErrorJustReturn([])
            }
            .observeOn(MainScheduler.instance)
```

and it seems to me that it does.

If you could create a PR that just adds `.orEmpty` I would be happy to merge it.

The reason why I would like to keep

```
if query.isEmpty {
      return .just([])
}
```

is because:
* a low of search engines fail with > 300 when you send "", so it's realistic use cause
* it demonstrates how to use `just` to return single value and combine synchronous logic with observable sequences.

cyrilchandelier:
Hi @kzaher,

I wanted to update my commit but tested first, unfortunately, the code doesn't compile on my side when the `query.isEmpty` block is here, with the following error:

```Type 'inout UITextField' does not conform to protocol 'ReactiveCompatible'```

kzaher:
Hi @cyrilchandelier ,

I've added example from README.md to [develop branch](https://github.com/ReactiveX/RxSwift/commit/44cee5305013247a1e83d1adb8630de68e0a8ac5) as a sanity check.

The build server successfully compiled everything. I'm not sure what is causing issues for you.

PR #1030, opened 2016-12-09
Extend the UIScrollView reactive wrappers

serluca:
It extends the current implementation of the UIScrollView reactive wrapper adding the support for the delegate methods `scrollViewDidEndDecelerating(_:)` and `scrollViewDidEndDragging(_:willDecelerate:)`.

### DIFF
--- a/RxCocoa/iOS/UIScrollView+Rx.swift
+++ b/RxCocoa/iOS/UIScrollView+Rx.swift
@@ -57,6 +57,12 @@ extension Reactive where Base: UIScrollView {
         let source = RxScrollViewDelegateProxy.proxyForObject(base).contentOffsetPublishSubject
         return ControlEvent(events: source)
     }
+	
+	/// Reactive wrapper for delegate method `scrollViewDidEndDecelerating`
+	public var didEndDecelerating: ControlEvent<Void> {
+		let source = delegate.methodInvoked(#selector(UIScrollViewDelegate.scrollViewDidEndDecelerating(_:))).map { _ in }
+		return ControlEvent(events: source)
+	}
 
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @serluca ,

if you would change unit test `testScrollViewDidEndDragging` to be equivalent to `testScrollViewDidEndDecelerating` I would be happy to merge this.

RxCocoa/iOS/UIScrollView+Rx.swift
```diff
@@ -57,6 +57,21 @@ extension Reactive where Base: UIScrollView {
...
+	/// Reactive wrapper for delegate method `scrollViewDidEndDecelerating`
+	public var didEndDecelerating: ControlEvent<Void> {
+		let source = delegate.methodInvoked(#selector(UIScrollViewDelegate.scrollViewDidEndDecelerating(_:))).map { _ in }
+		return ControlEvent(events: source)
+	}
+	
+	/// Reactive wrapper for delegate method `scrollViewDidEndDragging(_:willDecelerate:)`
+	public var didEndDragging: ControlEvent<Bool> {
+		let source = delegate.methodInvoked(#selector(UIScrollViewDelegate.scrollViewDidEndDragging(_:willDecelerate:))).map { value -> Bool in
+			let willDecelerate = value[1] as! Bool
```
kzaher: We are also using 
```
return try castOrThrow(Bool.self, value[1])
```

for these conversions.

If we aren't using it this way somewhere, we should probably change that other place also.

This shouldn't ever fail, but if it does we probably shouldn't bring the entire app down because of this.

serluca:
Hi @kzaher,

Thanks for reviewing my PR, I made the changes you proposed, it looks better now.

Issue #866, opened 2016-09-07
Request: Observable creation from Optional


tarunon:
I request extension of Optional like `Array.toObservable`.
https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Observables/Observable%2BCreation.swift#L200-L213
Optional is same of Array that has 1 element or empty. (It would be similar in `Array.flatMap`) 
So, I think we can create `Optional.toObservable` also.

I think this feature is useful, and has many profit.

The most simple example is:

``` swift
var a: Observable<Int?>
a.flatMap { $0.toObservable() } // -> Observable<Int>
```

kzaher:
Hi @tarunon ,

I believe there is already https://github.com/RxSwiftCommunity/RxOptional .

tarunon:
Hi @kzaher san,

Yes, I believed it also, but I could not find `Optional.toObservable` in RxOptional.
I opened this issue, because I think this feature would be needed standard support more than adding in RxOptional.

sergdort:
Hi guys, Isn't it `filterNil` in RxOptional?

tarunon:
@sergdort 
Thank you for help.
Yes, I think `.filterNil()` is equal to `.flatMap { $0.toObservable() }`, and `Observable.just(optional).filterNil()` is equal to `optional.toObservable()` also.
This feature is like of monad transform.

kzaher:
Hi @tarunon ,

I'm ~50:50 on this. I understand reasons for pro and con. For the interface to be consistent with 3.0.0 version we would need to have it as `Observable.from(optional)`.

If somebody needs this please tumb up this issue, or thumb down it. For this one, I'm fine with either outcome.

Busta117:
this is my solution, i made a set of method to write the bindings with that sintaxis but you can see here how i fixed it: 
``` swift
infix operator ~>
@discardableResult
public func ~> <T>(variable: Variable<T>, property: AnyObserver<T?>) -> Disposable {
    let bindDisposable = variable.asObservable()
        .map { $0 }
        .bindTo(property)
    return bindDisposable
}
```

and just use it like 
``` swift
var a: Observable<String?>
var label = UILabel()
(a ~> label).addDisposableTo(disposableBag) // this is the Disposable
```

tarunon:
Hi @Busta117 ,
I think your solution is near that convert `Observable<T>` to `Observable<T?>`.
This request is about transform `Optional<T>` to `Observable<T>`.

[...]

PR #1033, opened 2016-12-13
Add Optional overloads for `from` operator

tarunon:
Discussed on #866 
- [x] Add Optional overloads for `from` operator.
- [x] Passed all tests.

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -77,6 +77,7 @@ custom_categories:
   - Never
   - ObserveOn
   - ObserveOnSerialDispatchQueue
+  - Optional
   - Producer
   - Range
   - Reduce
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
   * `didStartLoad`
   * `didFinishLoad`
   * `didFailLoad`
+* Adds optional overloads for `from` operator.
 
 ## [3.0.1](https://github.com/ReactiveX/RxSwift/releases/tag/3.0.1) (Xcode 8 / Swift 3.0 compatible)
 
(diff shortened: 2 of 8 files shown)


kzaher:
Hi @tarunon ,

I think it's mostly ok, but there are a few changes that we would need to make.

I think there should be two versions (you can check out `just` operator implementation for more details).

```
public static func from(_ optional: E?) -> Observable<E>
public static func from(_ optional: E?, scheduler: ImmediateSchedulerType)
```

There are two reasons:
* optimizations, first overload without explicit scheduler will probably be most often used and will be significantly faster
* It will be more flexible when used as closure

```
func sum(first: Int, second: Int = 2) -> Int {
    return first + second
}

func sum2(first: Int) -> Int {
    return first + 2
}

func sum2(first: Int, second: Int) -> Int {
    return first + second
}

func doIt(action: (Int) -> Int) {
    
}

doIt(action: sum2) // works
//doIt(action: sum) // doesn't work
```

This was more important for `just` operator, but this one also seems close.

tarunon:
OK, I understand that. I'll add overloads without `Scheduler` version.

By the way, Observable.from(Array) uses default argument that `CurrentThreadScheduler`.
Should we change Observable.from(Array) also? If it is correct, I'll fix that new PR. ☺️

tarunon:
`Observable.from(Array)` has some of elements, maybe we'd use `CurrentThreadScheduler`. 🤔

kzaher:
Hi @tarunon ,

Yeah, if it has multiple elements then some scheduler has to be used so emitting of elements can be cancelled any time, and `CurrentThreadScheduler` is a safe default that doesn't make any assumptions about the environment.

PR #1035, opened 2016-12-14
Added missing orEmpty in an example in Readme

yuzushioh:
I realized when I was reading the RxExample project that `orEmpty` is missing from the sample code in README.  Since the string should not be optional in this case I added 'orEmpty' to it!


yuzushioh:
I didn't see #1025. Closing it.

PR #1037, opened 2016-12-14
Patched the subscribeProxyDataSource to flush the UITableView state

sgl0v:


kzaher:
Hi @sgl0v ,

We'll need to figure out how to solve those unit tests. I'll check them out this weekend.

PR #1039, opened 2016-12-16
Added optional parameter for 'sample' operator

CezaryKopacz:
Sometimes we want to get event even its value has not changed.


kzaher:
Hi @CezaryKopacz ,

`sample` operator with `onlyNew = false` is called `withLatestFrom`. If you remove that flag completely from internal implementation and leave only `true` value path, I would be happy to merge that complete cleanup.

CezaryKopacz:
Hi @kzaher ,
`withLatestFrom` is not obvious to use in my opinion, so that is the reason I see this in `sample` operator.
But anyway, I will remove that flag completely from internal implementation.

CezaryKopacz:
@kzaher I created new PR with cleanup so this PR is not valid anymore.

PR #1045, opened 2016-12-21
Refactor duplicates of generic parameters from Sink implementation.

tarunon:
I found 2-way declaration of generic type in implementations of `Sink`.
Case-A is
```swift
class XXXSink<Element, O: ObserverType> : Sink<O>, 
  ObserverType 
  where O.E == Element {
```
This implementation was used in `amb`,  `delay`, `sample`, and etc.

Case-B is
```swift
class XXXSink<O: ObserverType> : Sink<O>, 
  ObserverType {
  typealias Element = O.E
```
This implementation was used in `distinctUntilChanged`, `do`, `filter`, and etc.

I'm not sure why there are 2-way declaration of Generics type, but I think Case-B is better than Case-A there reasons. 
1. The generic parameters of Case-A are duplicated.
2. The generic parameter of Case-B gets more simply, and be more easy to understand.

I refactored implementation of Case-A as Case-B, and checked that passing tests.

### DIFF
--- a/RxSwift/Observables/Implementations/Amb.swift
+++ b/RxSwift/Observables/Implementations/Amb.swift
@@ -14,10 +14,10 @@ enum AmbState {
     case right
 }
 
-class AmbObserver<ElementType, O: ObserverType> : ObserverType where O.E == ElementType {
-    typealias Element = ElementType
-    typealias Parent = AmbSink<ElementType, O>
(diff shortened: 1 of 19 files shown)


kzaher:
There was a lot of refactorings, so I guess this were remainders of those refactorings.

PR #1048, opened 2016-12-23
Add default if empty

sergdort:
Adds defaultIfEmpty operator according to #662

PR #1049, opened 2016-12-23
Add switchIfEmpty operator to ObservableType

sergdort:
PR contains:

- optimise versions of `switchIfEmpty`
- tests for `switchIfEmpty`


Tests/RxSwiftTests/Observable+MultipleTest.swift
```diff
@@ -617,6 +617,43 @@ extension ObservableMultipleTest {
...
+        
+        XCTAssertEqual(res.events, [
+                next(220, 0),
+                next(230, 1),
+                next(240, 2),
+                next(250, 3),
+                completed(260)
+            ])
+        XCTAssertEqual(source.subscriptions, [
+                Subscription(200, 260)
```
sergdort: @kzaher  I wondering my assumption is correct and current implementation is incorrect:
If the `source` is empty then we are switching to another sequence,  does it mean thought that source should `unsubscribe` at the same time `260` as `switchSource`?

RxSwift/Observables/Observable+Multiple.swift
```diff
@@ -62,6 +62,14 @@ extension ObservableType where E : ObservableConvertibleType {
     }
 }
 
+// switchIfEmpty
+
+extension ObservableType {
+    public func switchIfEmpty<S: ObservableConvertibleType>(resultSelector: @escaping () throws -> S) -> Observable<E> where S.E == E {
```
kzaher: The same as for https://github.com/ReactiveX/RxSwift/pull/1048

RxSwift/Observables/Implementations/SwitchIfEmpty.swift
```diff
@@ -0,0 +1,127 @@
...
+//  Rx
+//
+//  Created by sergdort on 23/12/2016.
+//  Copyright © 2016 Krunoslav Zaher. All rights reserved.
+//
+
+import Foundation
+
+
+final class SwitchIfEmpty<SourceType, S: ObservableConvertibleType>: Producer<S.E> where S.E == SourceType {
```
kzaher: The same as for https://github.com/ReactiveX/RxSwift/pull/1048

RxSwift/Observables/Observable+Multiple.swift
kzaher: Maybe we should call it.

```
public func ifEmpty(switch: Observable<E>) -> Observable<E>
```

We wouldn't lose any generality because if one wants delayed evaluation it could use:

```
xs.ifEmpty(switch: Observable.defered {
          ....
          return result
})
```

And it seems we would be more consistent with RxJava 1.x. I still need to take a look at RxJava 2.x and see what they've done there.

RxSwift/Observables/Implementations/SwitchIfEmpty.swift
```diff
@@ -0,0 +1,127 @@
...
+}
+
+final class SwitchIfEmptySink<SourceType, S: ObservableConvertibleType, O: ObserverType>: Sink<O>
+    , ObserverType
+    , LockOwnerType
+    , SynchronizedOnType where S.E == O.E, SourceType == S.E {
+    typealias E = SourceType
+    typealias Selector = (Void) throws -> S
+    
+    let _lock = NSRecursiveLock()
```
kzaher: We don't need guard this or `SwitchIfEmptySinkIter` with `_lock`, `LockOwnerType`, `SynchronizedOnType`.

Each of the sequence should be already synchronized internally, and second sequence can only start once first one has completed.

RxSwift/Observables/Implementations/SwitchIfEmpty.swift
```diff
@@ -0,0 +1,127 @@
...
+                    _innerSubscription.disposable = d
+                    
+                    let observer = SwitchIfEmptySinkIter(parent: self, _self: d)
+                    let disposable = observable.subscribe(observer)
+                    d.setDisposable(disposable)
+                } catch let error {
+                    forwardOn(.error(error))
+                    dispose()
+                }
+            } else {
```
kzaher: We could create early return and lose one level of indentation.
```
guard !isEmpty else {
      forwardOn(.completed)
      dispose()
      return
}
... subscribe to other one
```

RxSwift/Observables/Implementations/SwitchIfEmpty.swift
```diff
@@ -0,0 +1,127 @@
...
+    typealias E = SourceType
+    typealias Selector = (Void) throws -> S
+    
+    let _lock = NSRecursiveLock()
+    private let _selector: Selector
+    fileprivate var isEmpty = true
+    fileprivate let _subscriptions: SingleAssignmentDisposable = SingleAssignmentDisposable()
+    fileprivate let _innerSubscription: SerialDisposable = SerialDisposable()
+    
+    init(selector: @escaping Selector, observer: O, cancel: Cancelable) {
```
kzaher: We can remove selector, and just use `Observable<E>` as switch source.

Please take a look at comment about public interface.

RxSwift/Observables/Implementations/SwitchIfEmpty.swift
```diff
@@ -0,0 +1,127 @@
...
+    func _synchronized_on(_ event: Event<E>) {
+        switch event {
+        case .next:
+            _parent.forwardOn(event)
+        case .error:
+            _parent.forwardOn(event)
+            _self.dispose()
+            _parent.dispose()
+        case .completed:
+            _parent.isEmpty = false
```
kzaher: We don't need to touch this since it shouldn't be used anyway after this observer has been created.

RxSwift/Observables/Implementations/SwitchIfEmpty.swift
kzaher: I think it should be unsubscribed at 210.

RxSwift/Observables/Implementations/SwitchIfEmpty.swift
sergdort: Hm. But would it try to switch again, 
e.g. In case where `source` and `switch` are both empty? Or I missing something?

Tests/RxSwiftTests/Observable+MultipleTest.swift
```diff
@@ -617,6 +617,43 @@ extension ObservableMultipleTest {
...
+            ])
+        XCTAssertEqual(source.subscriptions, [
+                Subscription(210, 260)
```
sergdort: My bad , I think there should be `seitchSource`

sergdort:
@kzaher I've addressed you comments

RxSwift/Observables/Implementations/SwitchIfEmpty.swift
```diff
@@ -0,0 +1,103 @@
...
+        case .next:
+            _parent.forwardOn(event)
+        case .error:
+            _parent.forwardOn(event)
+            _self.dispose()
+            _parent.dispose()
+        case .completed:
+            _parent.forwardOn(event)
+            _self.dispose()
+            _parent.dispose()
```
kzaher: I think one can just call `_parent.dispose()`

RxSwift/Observables/Observable+Multiple.swift
```diff
@@ -62,6 +62,21 @@ extension ObservableType where E : ObservableConvertibleType {
...
+// switchIfEmpty
+
+extension ObservableType {
+    /**
+     Switches to other sequence if source is empty
+     
+     - seealso: [DefaultIfEmpty operator on reactivex.io](http://reactivex.io/documentation/operators/defaultifempty.html)
+     - parameter other: Other sequence which it switches to if the source is empty
+     - returns: Switched observable sequence if source searqence is empty, otherwise emits element from the source sequence.
+     */
```
kzaher: I think we'll need to improve this. I'll try to check out can we reuse some of the texts from `http://reactivex.io/documentation/operators/defaultifempty.html`. 

Maybe something like:
```
Returns the elements of the specified sequence or `switchTo` sequence if the sequence is empty.
```

It's `parameter switchTo` now. 
Maybe `Observable sequence being returned when source sequence is empty.`

Maybe `- returns: Observable sequence that contains elements from switchTo sequence if source is empty, otherwise returns source sequence elements.`

Tests/RxSwiftTests/Observable+MultipleTest.swift
```diff
@@ -651,6 +651,89 @@ extension ObservableMultipleTest {
...
+            return source.ifEmpty(switchTo: switchSource)
+        }
+        
+        XCTAssertEqual(res.events, [])
+        XCTAssertEqual(source.subscriptions, [
+                Subscription(200, 1000)
+            ])
+        XCTAssertEqual(switchSource.subscriptions, [])
+    }
+}
```
kzaher: Hm, I don't see any unit tests for error cases.

PR #1055, opened 2016-12-29
Github Driver example small change to simulate networking work

gorozco58:
This is a small change to simulate networking work.

### DIFF
--- a/RxExample/RxExample/Examples/GitHubSignup/DefaultImplementations.swift
+++ b/RxExample/RxExample/Examples/GitHubSignup/DefaultImplementations.swift
@@ -106,13 +106,7 @@ class GitHubDefaultAPI : GitHubAPI {
         // this is also just a mock
         let signupResult = arc4random() % 5 == 0 ? false : true
         
-        return Observable<Bool>.create { observable in
-            DispatchQueue.global().asyncAfter(deadline: DispatchTime.now() + 3) {
-                observable.onNext(signupResult)
-            }
-            return Disposables.create()
-            }
-            .delay(0.4, scheduler: MainScheduler.instance)
-            .take(1)
+        return Observable.just(signupResult)
+            .delay(1.0, scheduler: MainScheduler.instance)
     }
 }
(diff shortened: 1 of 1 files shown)


RxExample/RxExample/Examples/GitHubSignup/DefaultImplementations.swift
```diff
@@ -105,8 +105,13 @@ class GitHubDefaultAPI : GitHubAPI {
     func signup(_ username: String, password: String) -> Observable<Bool> {
         // this is also just a mock
         let signupResult = arc4random() % 5 == 0 ? false : true
-        return Observable.just(signupResult)
-            .concat(Observable.never())
+
```
kzaher: Hi @gorozco58 ,

I think we should just do `delay(0.4, scheduler: MainScheduler.instance)` instead of `throttle(0.4, scheduler: MainScheduler.instance)`.

gorozco58: Cool I am gonna make it.

RxExample/RxExample/Examples/GitHubSignup/DefaultImplementations.swift
```diff
@@ -105,9 +105,14 @@ class GitHubDefaultAPI : GitHubAPI {
...
-        return Observable.just(signupResult)
-            .concat(Observable.never())
-            .throttle(0.4, scheduler: MainScheduler.instance)
+        
+        return Observable<Bool>.create { observable in
+            DispatchQueue.global().asyncAfter(deadline: DispatchTime.now() + 3) {
+                observable.onNext(signupResult)
+            }
+            return Disposables.create()
+            }
```
kzaher: Hi @gorozco58 ,

`delay` operator is here to simulate delay.

This code should be.

```
return Observable.just(signupResult)
     .delay(0.4, scheduler: MainScheduler.instance)
```

We don't need all of that other noise with dispatch queues.

PR #1059, opened 2017-01-03
Extend the UIScrollView reactive wrappers: `scrollViewDidEndScrollingAnimation(_:)`

sgr-ksmt:
It extends the current implementation of the UIScrollView reactive wrapper adding the support for the delegate methods `scrollViewDidEndScrollingAnimation(_:)`

Related : #1030

### DIFF
--- a/Tests/RxCocoaTests/UIScrollView+RxTests.swift
+++ b/Tests/RxCocoaTests/UIScrollView+RxTests.swift
@@ -172,13 +172,16 @@ extension UIScrollViewTests {
     }
 
     func testDidEndScrollingAnimation() {
-        // direct call `scrollViewDidEndScrollingAnimation`
+        var completed = false
+
         autoreleasepool {
             let scrollView = UIScrollView()
             var didEndScrollingAnimation = false
             
-            let subscription = scrollView.rx.didEndScrollingAnimation.subscribe(onNext: {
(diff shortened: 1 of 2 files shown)


Tests/RxCocoaTests/UIScrollView+RxTests.swift
```diff
@@ -170,6 +170,53 @@ extension UIScrollViewTests {
...
+            XCTAssertFalse(didEndScrollingAnimation)
+            
+            scrollView.delegate!.scrollViewDidEndScrollingAnimation!(scrollView)
+            
+            XCTAssertTrue(didEndScrollingAnimation)
+            subscription.dispose()
+        }
+        
+        // change contentOffset with animation
+        autoreleasepool {
```
kzaher: Hi, I would be happy to merge this if you removed this autorelease pool closure part of code.

We should be only testing that when delegate fires new element is produced.

Part inside this `autorelease` closure tests correctness of UIScrollView internal logic which is out of scope of this project.

sgr-ksmt: @kzaher 
Thank you for pointing out..! 🙇 
I removed `autoreleasepool` closure and fix test code. (added `completed` check)

PR #1065, opened 2017-01-08
AsyncSubject

victorg1991:
Add AsyncSubject implementation

### DIFF
--- a/RxSwift/Subjects/AsyncSubject.swift
+++ b/RxSwift/Subjects/AsyncSubject.swift
@@ -61,21 +61,30 @@ public final class AsyncSubject<Element>
         switch event {
         case .next(let value):
             _lastValue = value
+        
         case .error:
             _stoppedEvent = event
             _stopped = true
 
-            _observers.on(event)
+            let observers = _observers
             _observers.removeAll()
+            _lock.unlock()
+
+            observers.on(event)
+        
         case .completed:
             _stoppedEvent = event
             _stopped = true
 
-            if let lastValue = _lastValue {
-                _observers.on(.next(lastValue))
-            }
-            _observers.on(event)
+            let observers = _observers
             _observers.removeAll()
+            _lock.unlock()
+
+            if let lastValue = _lastValue {
+                observers.on(.next(lastValue))
+            }
+
+            observers.on(event)
         }
     }
 
--- a/RxSwift/Subjects/AsyncSubject.swift
+++ b/RxSwift/Subjects/AsyncSubject.swift
@@ -61,16 +61,20 @@ public final class AsyncSubject<Element>
 		switch event {
 		case .next(let value):
 			_lastValue = value
-		case .error, .completed:
+		case .error:
 			_stoppedEvent = event
 			_stopped = true
-		}
 
-		if let stoppedEvent = _stoppedEvent {
-			if let lastValue = _lastValue, case .completed = stoppedEvent {
+			_observers.on(event)
+			_observers.removeAll()
+		case .completed:
+			_stoppedEvent = event
+			_stopped = true
+
+			if let lastValue = _lastValue {
 				_observers.on(.next(lastValue))
 			}
-			_observers.on(stoppedEvent)
+			_observers.on(event)
 			_observers.removeAll()
 		}
 	}
--- a/RxSwift/Subjects/AsyncSubject.swift
+++ b/RxSwift/Subjects/AsyncSubject.swift
@@ -12,126 +12,126 @@ import Foundation
 ///
 /// Observers can subscribe to the subject to receive the last (or initial) value and all subsequent notifications.
 public final class AsyncSubject<Element>
-	: Observable<Element>
-	, SubjectType
-	, ObserverType
-	, SynchronizedUnsubscribeType
-	, Disposable {
-	public typealias SubjectObserverType = AsyncSubject<Element>
-	typealias DisposeKey = Bag<AnyObserver<Element>>.KeyType
+    : Observable<Element>
+    , SubjectType
+    , ObserverType
+    , SynchronizedUnsubscribeType
+, Disposable {
+    public typealias SubjectObserverType = AsyncSubject<Element>
+    typealias DisposeKey = Bag<AnyObserver<Element>>.KeyType
 
-	/// Indicates whether the subject has any observers
-	public var hasObservers: Bool {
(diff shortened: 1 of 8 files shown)


kzaher:
Hi @victorg1991 ,

interesting PR :) I would need to take a more closer look on this. I think it would be great to make this consistent with other Rx implementations. Did you maybe take a look at some of them?

victorg1991:
Hi @kzaher 
Thank you for your response :)
I followed http://reactivex.io/documentation/subject.html#AsyncSubject description
What do you think is missing?

sergdort:
Ha, the naming is quiet confusing :) At first I thought it emits elements to all subscribers asynchronously 😁

victorg1991:
I used the same naming that is used in other subjects implementations like https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Subjects/PublishSubject.swift

Anyway I'm open to suggestions :)

kzaher:
Hi @victorg1991 ,

since this is standard Rx concept, I will need to make sure our behavior matches other Rx implementations.

The easiest for me is to check Rx.NET and RxJS. It would be really great if you could take a look at them before I do and make sure our tests and conceptual implementation match theirs.

I would really want to make this behavior consistent.

This is something I will need to do for sure and it will take a lot less time if they are as close as possible so I can just simulate one system in my head, instead of trying to simulate two really similar ones and figure out are they the same.

This is not something we can always do, but when we can, I don't see why we wouldn't :)

I'm not saying up front that there is something wrong, but you can make my life sooo much easier, and we can merge this sooo much sooner if this is satisfied.

kzaher:
I'll check it in more detail this weekend. I'm just too swamped during the workdays.

victorg1991:
Ok gotcha! I will check it out this between today and tomorrow morning

Thanks! :D

victorg1991:
Hey @kzaher! I'm here again, sorry for the delay.

I checked it out RxNet, RxJava, RxJS v4 and RsJS v5. RxNet, RxJava and RxJS v4 implementations behave like mine. In RxJS v5 it behaves strangely but I think is due to a bug, I just sent them a PR that fixes this behaviour https://github.com/ReactiveX/rxjs/pull/2280

Let me know what you think :)

RxSwift/Subjects/AsyncSubject.swift
```diff
@@ -0,0 +1,138 @@
...
+	/// Notifies all subscribed observers about next event.
+	///
+	/// - parameter event: Event to send to the observers.
+	public func on(_ event: Event<E>) {
+		_synchronized_on(event)
+	}
+
+	func _synchronized_on(_ event: Event<E>) {
+		_lock.lock(); defer { _lock.unlock() }
+		if  _isDisposed {
```
kzaher: I think `_stopped` should be checked here otherwise completed followed by error will error out. And vice versa.

RxSwift/Subjects/AsyncSubject.swift
```diff
@@ -0,0 +1,138 @@
...
+		if let stoppedEvent = _stoppedEvent {
+			if let lastValue = _lastValue, case .completed = stoppedEvent {
+				_observers.on(.next(lastValue))
+			}
+			_observers.on(stoppedEvent)
+			_observers.removeAll()
+			return
+		}
+
+		return
```
kzaher: Do we need this?

RxSwift/Subjects/AsyncSubject.swift
```diff
@@ -0,0 +1,138 @@
...
+
+	func _synchronized_on(_ event: Event<E>) {
+		_lock.lock(); defer { _lock.unlock() }
+		if  _isDisposed {
+			return
+		}
+
+		switch event {
+		case .next(let value):
+			if !_stopped {
```
kzaher: This check should be moved near _disposed I think.

RxSwift/Subjects/AsyncSubject.swift
```diff
@@ -0,0 +1,138 @@
...
+				_lastValue = value
+			}
+		case .error, .completed:
+			_stoppedEvent = event
+			_stopped = true
+		}
+
+		if let stoppedEvent = _stoppedEvent {
```
kzaher: I think cases for error and completed should be separated
```
case .error
case .completed:
```

so there is no need for additional `if let` check.

I understand that your goal was to remove duplication of logic but in this case I think the code will have less conditionals and execute faster because this check will be performed for each event vs once per terminal event.

It maybe looks like a microoptimization, but last time I've checked `if let` construct was quite slow. That's why there is `_stopped` flag also.

victorg1991:
Hi @kzaher, thanks for taking your time reviewing this :)

I've addressed all your comments

kzaher:
Hi @victorg1991 ,

I'm taking a another look now. I've noticed that you are using tabs instead of 4 spaces. Project uses xcode defaults of 4 spaces so if I merge this and indent I'll erase all of your changes.

Tests/RxSwiftTests/AsyncSubjectTests.swift
```diff
@@ -0,0 +1,386 @@
...
+			next(110, 2),
+			next(220, 3),
+			next(270, 4),
+			next(340, 5),
+			next(410, 6),
+			next(520, 7),
+			next(630, 8),
+			next(710, 9),
+			next(870, 10),
+			completed(900),
```
kzaher: This test can be merged with `test_shouldIgnoreValuesAfterCompletes` since it contains a subset of checks.

Tests/RxSwiftTests/AsyncSubjectTests.swift
```diff
@@ -0,0 +1,386 @@
...
+		scheduler.start()
+
+		XCTAssertEqual(results1.events, [next(900, 10), completed(900)])
+
+		XCTAssertEqual(results2.events, [next(900, 10), completed(900)])
+
+		XCTAssertEqual(results3.events, [next(900, 10), completed(900)])
+	}
+
+	func test_shouldSendOnlyErrorWhenReceiveAnError() {
```
kzaher: This test can be merged with `test_shouldIgnoreValuesAfterError` since it contains a subset of checks.

victorg1991:
Hey @kzaher ! 
Sorry for the delay, 
I rebased my branch to take all the new changes and addressed all your comments 
Let me know what you think 
Thanks!

PR #1070, opened 2017-01-16
Playground: fix wording

ole:
I noticed some inaccuracies in the example playground:

* Some terms from Swift 2 hadn't been updated to Swift 3 (`SequenceType`, `generate()`).
* The section on `combineLatest` spoke of a "`combineLatest` extension on `Array`", which confused me because it isn't actually an extension on `Array`; it works on arrays (and other collections).

### DIFF
--- a/Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
@@ -110,7 +110,7 @@ example("combineLatest") {
     
     stringSubject.onNext("🆎")
 }
-//: There is also a `combineLatest` extension on `Array`:
+//: There is also a variant of `combineLatest` that takes an `Array` (or any other collection of `Observable` sequences):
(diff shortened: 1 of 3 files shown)


ole:
I'll check against `develop` and other another PR.

ole:
Okay, since `develop` and `master` are identical I guess I can reopen this one. Sorry for the confusion.

Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
```diff
@@ -110,7 +110,7 @@ example("combineLatest") {
     
     stringSubject.onNext("🆎")
 }
-//: There is also a `combineLatest` extension on `Array`:
+//: There is also a variant of `combineLatest` that takes an `Array` (or any other collection of `Observable` sequences):
 example("Array.combineLatest") {
```
kzaher: Yeah, and it should be `Observable.combineLatest` also.

PR #1077, opened 2017-01-20
[RxExamples] Cleanup/fixes in WikipediaImageSearch

radex:
Fixed two subtle mistakes in the wikipedia image search example:

- article urls weren't shared, so the article was fetched and parsed twice (for the title and for image urls field)
- typo in the title

plus I tweaked Swift code here and there to be cleaner.


kzaher:
... a long long time ago, in a galaxy far far away ...

Issue #1082, opened 2017-01-24
RetryWhen operator behaviour with completed trigger


bontoJR:
I was playing heavily with `retryWhen` operator and noticed a really weird behaviour, if the trigger completes, also the original observable completes, which, to be frankly, it's an unexpected behaviour.

If the outer observable errors out and the inner trigger completes, the operator should pass on the last error which might be then caught. 

Current behaviour

```
O: ----1----2----X----1----2----X
T: --------------T--------------C
R: ----1----2---------1----2----C
```

When I think it would make more sense to have:

```
O: ----1----2----X----1----2----X
T: --------------T--------------C
R: ----1----2---------1----2----X
```

This consideration comes on thinking about `retry(X)` and `retry` which would forward the error if the number of times of `retry(maxAttempts)` has been completed unsuccessfully. `retryWhen` is kind-of a super set of both `retry` operators, with a finite trigger sequence for the max attempt and an infinite one for unlimited retry, so it's a little bit illogical to terminate a sequence which previously had an error.

Open for discussions anyway. :)

icanzilb:
as always - any other Rx implementations that already feature `retryWhen()`, that we can have a look into?

bontoJR:
`retryWhen` from [rxJS](https://github.com/Reactive-Extensions/RxJS/blob/master/doc/api/core/operators/retrywhen.md) which operates in the way I am suggesting.

bontoJR:
I have to correct myself, the implementation of rxJS changed after my first check when I was working on this operator, so since November the behaviour has been changed from forwarding error to complete.

fpillet:
I gave a second though to that, following our discussion on Slack. To me it seems that the (undocumented) way it behaves now is the most flexible. The trigger sequence is the ultimate decision maker on what happens when the source sequence errors out.

Now since *you* control the trigger sequence, there is no good reason not to pass errors along to the outer observable when you want to error out. The ability to complete while silencing the last errors actually seems like a very nice feature! It spares you from adding a `catchError` to the sequence if all you want is retries up to a point and then just complete if nothing worked.

At the very least, this behaviour should be properly documented in RxSwift (maybe worth a post, Marin?) as it seems it could be a handy feature in some scenarios.

Unless I completely misunderstood  your issue, I like the current behaviour. This lets you do things like:

```
Silencing the last error:
O: ----1----2----X----1----2----X
T: --------------E--------------C
R: ----1----2---------1----2----C
```

and

```
[...]

PR #1083, opened 2017-01-24
Fixs retryWhen logic

bontoJR:
This PR addresses the problem brought up with #1082. I know there's still a discussion to be made, but  in case the change is accepted, there's already a PR ready for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ## Master
 ## [3.1.1](https://github.com/ReactiveX/RxSwift/releases/tag/3.1.1) (Xcode 8 / Swift 3.0 compatible)
 * Adds [`ifEmpty(default:)`]((http://reactivex.io/documentation/operators/defaultifempty.html)) operator
+* Fixes `retryWhen` to a more logical parent operator of `retry`, forwarding the error when the trigger completes (BREAKING CHANGE)
 
 ## [3.1](https://github.com/ReactiveX/RxSwift/releases/tag/3.1) (Xcode 8 / Swift 3.0 compatible)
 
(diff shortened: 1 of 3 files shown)


kzaher:
I think this PR is an improvement of operator behavior, but I can't merge it before 4.0 since this would be a breaking change.

bontoJR:
@kzaher absolutely right 👍

freak4pc:
Even though this PR is very old, do we still want it in RxSwift 4 ? If so we can perhaps rebase this off the swift4 working branch and get it merged ? @bontoJR

kzaher:
Hi, 

It seems to me that currently RxJava is the most popular implementation. As far as I can tell when error transformation block completes sequence doesn't terminate with latest error.

```java
Observable.just(0xf).concatWith(Observable.<Integer>error(new Throwable())).retryWhen(new Function<Observable<Throwable>, ObservableSource<?>>() {
            @Override
            public ObservableSource<?> apply(@NonNull Observable<Throwable> throwableObservable) throws Exception {
                return Observable.zip(throwableObservable, Observable.range(0, 3), new BiFunction<Throwable, Integer, Integer>() {
                    @Override
                    public Integer apply(@NonNull Throwable throwable, @NonNull Integer integer) throws Exception {
                        return integer;
                    }
                }).switchMap(new Function<Integer, ObservableSource<Long>>() {
                    @Override
                    public ObservableSource<Long> apply(@NonNull Integer o) throws Exception {
                        return Observable.timer(o.intValue() * 50, TimeUnit.MILLISECONDS, Schedulers.io());
                    }
                });
            }
        }).subscribe(new Consumer<Integer>() {
            @Override
            public void accept(@NonNull Integer integer) throws Exception {
                System.out.println(integer);
            }
        }, new Consumer<Throwable>() {
            @Override
            public void accept(@NonNull Throwable throwable) throws Exception {
                System.out.println(throwable);
            }
        }, new Action() {
            @Override
            public void run() throws Exception {
                System.out.println("Completed");
            }
        });
```

I kind of understand why that might be the case and since we are already consistent with RxJava it seems to me it might be better to leave the current behavior.

PR #1084, opened 2017-01-26
Add binding for `numberOfPages` on `UIPageControl`.

rdougan:
Allow binding of the `numberOfPages` property on `UIPageControl`.

I've also included a test.


rdougan:
Fixed.

PR #1090, opened 2017-02-06
Fix documentation of UnitTests

milkit:
In the UnitTests documentation, the virtual time disposing subscription is 300 because the code below is written.

```
let correctSubscriptions = [
    Subscription(200, 300)
]
```

This correction has no effect to the production, so master branch is base.


tomahh:
I think what the documentation wants to say is that `start` method will explicitly call dispose on the subscription at time 1000.

milkit:
@tomahh
Thank you for your response. I read code more precisely, I understood it.

PR #66, opened 2015-07-22
Feature/swift2.0 - Xcode 7 Beta 4 Fixes

garethknowles:
Fixes for errors introduced in Xcode 7 Beta 4.  
Mainly related to internal type alias usage in public functions.

Also fixed errors in the example project.
All unit tests currently passing.

PR #1091, opened 2017-02-06
[RxCocoa] Remove unnecessary variable assignment in response(request:) method in URLSession+Rx.swift

keitaito:
Hi RxSwift team!

It seems that `let t = task` variable assignment in `response(request:)` method in URLSession+Rx.swift is unnecessary, so I removed the line to make it clearer. I checked [the PR that added it][1], and I think it should be no problem to remove it. Please let me know if I am missing something with this variable assignment. I ran `./scripts/all-tests.sh` locally and it passed.

Can I please ask you to review this PR? Thank you!

[1]: https://github.com/ReactiveX/RxSwift/pull/66

### DIFF
--- a/RxCocoa/Foundation/URLSession+Rx.swift
+++ b/RxCocoa/Foundation/URLSession+Rx.swift
@@ -158,9 +158,7 @@ extension Reactive where Base: URLSession {
                 observer.on(.completed)
             }
 
-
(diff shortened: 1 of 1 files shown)


keitaito:
@kzaher Thanks for merging!

Issue #1087, opened 2017-01-31
Mutual exclusive delegate methods


djmadcat:
**Short description of the issue**:

When a delegate implements a new version of the method, the old version is not called.
So when we use `func methodInvoked(_ selector: Selector) -> Observable<[Any]>` method to intercept older method, it does not invoked.

**Expected outcome**:

Any method must be able to be intercepted.

**What actually happens**:

For example:
`UITextFieldDelegate` has two versions for `didEndEditing` method:
* old (iOS 2) `func textFieldDidEndEditing(_ textField: UITextField)`
* new (iOS 10) `func textFieldDidEndEditing(_ textField: UITextField, reason: UITextFieldDidEndEditingReason)`

What happens:
* User creates `Observable` by calling `delegate.methodInvoked(#selector(UITextFieldDelegate.textFieldDidEndEditing(_:)))`
* iOS (10!) asks delegate for `textFieldDidEndEditing:reason:` support by calling `respondsToSelector:`.
* `_RXDelegateProxy` collects all `UITextFieldDelegate` methods and finds it.
* iOS invokes `textFieldDidEndEditing:reason:` with `forwardInvocation:`.
* `textFieldDidEndEditing:` not being called.
* "forwardDelegate" methods also not being called.

**Self contained code example that reproduces the issue**:

```swift
class RxTextFieldDelegateProxy
    : DelegateProxy
    , DelegateProxyType
    , UITextFieldDelegate {

    /// For more information take a look at `DelegateProxyType`.
    class func setCurrentDelegate(_ delegate: AnyObject?, toObject object: AnyObject) {
        let o: UITextField = object as! UITextField
        o.delegate = delegate.map { $0 as! UITextFieldDelegate }
    }

    /// For more information take a look at `DelegateProxyType`.
    class func currentDelegateFor(_ object: AnyObject) -> AnyObject? {
        let o: UITextField = object as! UITextField
        return o.delegate
    }
}

extension UITextField {
    public func createRxDelegateProxy() -> RxTextFieldDelegateProxy {
        return RxTextFieldDelegateProxy(parentObject: self)
    }
}

extension Reactive where Base: UITextField {
    public var delegate: DelegateProxy {
        return RxTextFieldDelegateProxy.proxyForObject(self.base)
    }

    public func setDelegate(_ delegate: UITextFieldDelegate) -> Disposable {
[...]

Issue #1081, opened 2017-01-24
UICollectionViewCell gets draggable when elements are binded via rx.items(dataSource:)


ishkawa:
**Short description of the issue**:

UICollectionViewCell gets draggable when elements are binded via `rx.items(dataSource:)`, even if the delegate does not have implementation of `collectionView(_:canMoveItemAtIndexPath:)`.

**Expected outcome**:

UICollectionViewCell is not draggable by default.

**What actually happens**:

UICollectionViewCell is draggable by default.

**Self contained code example that reproduces the issue**:

```swift
import UIKit
import RxSwift
import RxCocoa

class Cell: UICollectionViewCell {
    @IBOutlet weak var label: UILabel!
}

class ViewController: UICollectionViewController {
    private let dataSource = DataSource()
    private let disposeBag = DisposeBag()
    
    override func viewDidLoad() {
        super.viewDidLoad()

        let values = Array(1...100)

        Observable
            .just(values)
            .bindTo(collectionView!.rx.items(dataSource: dataSource))
            .addDisposableTo(disposeBag)
    }

    class DataSource: NSObject, UICollectionViewDataSource, RxCollectionViewDataSourceType {
        typealias Element = [Int]

        var values = [Int]()

        func collectionView(_ collectionView: UICollectionView, observedEvent: Event<Element>) {
            if case .next(let element) = observedEvent {
                values = element
                collectionView.reloadData()
            }
        }

        func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
            return values.count
        }

        func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
            let cell = collectionView.dequeueReusableCell(withReuseIdentifier: "Cell", for: indexPath) as! Cell
            cell.label.text = String(indexPath.row)

            return cell
        }
    }
}
```

Entire project is available [here](https://github.com/ishkawa/sandbox/tree/master/RxCocoaReorderIssue). Run `carthage bootstrap` and open `RxCocoaReorderIssue/Demo.xcodeproj` to reproduce the issue.

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

3.1.0

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

[...]

PR #1096, opened 2017-02-12
Improves auto implementing observing mechanism for `DelegateProxy`. #1081 #1087

kzaher:
Hi @ishkawa, @djmadcat.

I've modified delegate proxy to make it more generic and only respond to those selectors that are being currently observed.

This feature is at the expense of more frequent delegate setting. I've also removed ugly specific hack for table view data source in favor of this PR.

It would be great if you could run this and see if it works for you. Give it a little smoke test.

I would like to release this potentially in a patch release so **I would appreciate if anyone who is relying on UI{Table,Collection}View extensions could also smoke test in their app and see does it cause any issues for them**.

### DIFF
--- a/RxCocoa/Common/DelegateProxy.swift
+++ b/RxCocoa/Common/DelegateProxy.swift
@@ -15,16 +15,16 @@
     #endif
 #endif
 
-var delegateAssociatedTag: UInt8 = 0
-var dataSourceAssociatedTag: UInt8 = 0
+var delegateAssociatedTag: UnsafeRawPointer = UnsafeRawPointer(UnsafeMutablePointer<UInt8>.allocate(capacity: 1))
+var dataSourceAssociatedTag: UnsafeRawPointer = UnsafeRawPointer(UnsafeMutablePointer<UInt8>.allocate(capacity: 1))
 
 /// Base class for `DelegateProxyType` protocol.
 ///
 /// This implementation is not thread safe and can be used only from one thread (Main thread).
 open class DelegateProxy : _RXDelegateProxy {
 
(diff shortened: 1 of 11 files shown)


RxCocoa/Common/DelegateProxy.swift
```diff
@@ -142,17 +143,18 @@ open class DelegateProxy : _RXDelegateProxy {
     - returns: Observable sequence of arguments passed to `selector` method.
      */
     open func methodInvoked(_ selector: Selector) -> Observable<[Any]> {
+        MainScheduler.ensureExecutingOnScheduler()
```
sergdort: Just out of curiosity, do you think it's mandatory to assert execution on the main, or I missing something :)

kzaher: Hi @sergdort , 

If `DelegateProxyType` interface would allow concurrent usage then we would also need it to provide some sort of synchronization mechanism that all extension methods would need to use.

That synchronization mechanism would also need to be native to the delegate owner.

So far 99% of classes that have delegates and are using delegate proxy are UI elements that use main dispatch queue as a synchronization mechanism.

The only case I know of that can use other synchronization mechanism is `CBCentralManager`. You can pass a specific dispatch queue to it's initializer but it's not exposed anywhere on it and I'm not exactly sure what restrictions does it create if you do.

This API only restricts calling `methodInvoked` from main scheduler and not subscribing to it.

As far as I can tell this is the simplest API that works all use cases and I'm not sure what would be the best way to make it thread safe and what benefit we would get from it.

sergdort: Cool, just remember that some guys had problems with `CoreBluetooth` because of this :)

PR #1097, opened 2017-02-13
Fixes retain cycle in GitHubSearchRepositoriesViewController

TheAdamBorek:
I've found the retain cycle issue inside `GitHubSearchRepositoriesViewController.swift`.
The fix is pretty simple, I've just added missing `[weak self]` in 2 places inside the viewController

- [x] `./scripts/all-tests.sh` passes 👌

### Before:
zrzut ekranu 2017-02-13 07 16 23

### After:
zrzut ekranu 2017-02-13 07 51 00

### DIFF
--- a/RxExample/RxExample/Examples/GitHubSearchRepositories/GitHubSearchRepositoriesViewController.swift
+++ b/RxExample/RxExample/Examples/GitHubSearchRepositories/GitHubSearchRepositoriesViewController.swift
@@ -42,10 +42,10 @@ class GitHubSearchRepositoriesViewController: ViewController, UITableViewDelegat
         }
 
 
+        let tableView: UITableView = self.tableView
         let loadNextPageTrigger = self.tableView.rx.contentOffset
-            .flatMap { [weak self] _ -> Observable<Void> in
-                guard let `self` = self else { return Observable.empty() }
-                return self.tableView.isNearBottomEdge(edgeOffset: 20.0)
+            .flatMap { _ in
+                return tableView.isNearBottomEdge(edgeOffset: 20.0)
                     ? Observable.just(())
                     : Observable.empty()
             }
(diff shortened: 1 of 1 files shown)


RxExample/RxExample/Examples/GitHubSearchRepositories/GitHubSearchRepositoriesViewController.swift
```diff
@@ -43,8 +43,9 @@ class GitHubSearchRepositoriesViewController: ViewController, UITableViewDelegat
```
kzaher: I think we can define local variable table here.

`let tableView = self.tableView`

RxExample/RxExample/Examples/GitHubSearchRepositories/GitHubSearchRepositoriesViewController.swift
```diff
@@ -43,8 +43,9 @@ class GitHubSearchRepositoriesViewController: ViewController, UITableViewDelegat
 
 
         let loadNextPageTrigger = self.tableView.rx.contentOffset
-            .flatMap { _ in
-                self.tableView.isNearBottomEdge(edgeOffset: 20.0)
+            .flatMap { [weak self] _ -> Observable<Void> in
+                guard let `self` = self else { return Observable.empty() }
+                return self.tableView.isNearBottomEdge(edgeOffset: 20.0)
```
kzaher: ... so we can just say here `return tableView.isNearBottomEdge(edgeOffset: 20.0)` without self.

RxExample/RxExample/Examples/GitHubSearchRepositories/GitHubSearchRepositoriesViewController.swift
```diff
@@ -80,7 +81,8 @@ class GitHubSearchRepositoriesViewController: ViewController, UITableViewDelegat
 
         // dismiss keyboard on scroll
         tableView.rx.contentOffset
-            .subscribe { _ in
+            .subscribe { [weak self] _ in
+                guard let `self` = self else { return }
                 if self.searchBar.isFirstResponder {
```
kzaher: We should be able to do the same here also, just use local variable.

RxExample/RxExample/Examples/GitHubSearchRepositories/GitHubSearchRepositoriesViewController.swift
TheAdamBorek: 👍

At the beginning I wanted to put the tableView inside the capture list to the closure. However, the tableView is forced unwrapped optional and when you put it in the capture list it become the optional.

That's also a reason why I set type explicit in `let tableView: UITableView = self.tableView`

PR #1099, opened 2017-02-13
Add countDownDuration for UIDatePicker

pawurb:
Hello. My first contribution here, I am just starting to play around with RxSwift. 

I noticed that the current `value` implementation does not make sense when UIDatePicker is in `.countDownTimer` mode. This PR adds another observable property that can be used for countDownDuration handling. I tested fork with my project and works fine.

### DIFF
--- a/RxCocoa/iOS/UIDatePicker+Rx.swift
+++ b/RxCocoa/iOS/UIDatePicker+Rx.swift
@@ -18,7 +18,7 @@ extension Reactive where Base: UIDatePicker {
     public var date: ControlProperty<Date> {
         return value
     }
-    
+
     /// Reactive wrapper for `date` property.
     public var value: ControlProperty<Date> {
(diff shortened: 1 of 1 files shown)


RxCocoa/iOS/UIDatePicker+Rx.swift
```diff
@@ -30,7 +35,18 @@ extension Reactive where Base: UIDatePicker {
             }
         )
     }
-    
+
+    /// Reactive wrapper for `countDownDuration` property.
+    public var countDownValue: ControlProperty<TimeInterval> {
```
kzaher: Hi @pawurb ,

I'm not sure what would be the proper naming conventions for this case. Let's only add `countDownDuration` property for now and just move this implementation there.

pawurb:
@kzaher fixed, I hope this is what you meant.

PR #1110, opened 2017-02-20
create UISearchResultsUpdating proxy

agvieira:
Hey guys, amazing project.
This is a try to implement a proxy for UISearchResultsUpdating protocol.

We don't really know if we did everything correctly or following the right way.
Feel free to comment and help us improve the code.

Thank you, cheers.

### DIFF
--- a/RxCocoa/iOS/UISearchController+Rx.swift
+++ b/RxCocoa/iOS/UISearchController+Rx.swift
@@ -20,6 +20,13 @@
         public func createRxDelegateProxy() -> RxSearchControllerDelegateProxy {
             return RxSearchControllerDelegateProxy(parentObject: self)
         }
+        
+        /// Factory method that enables subclasses to implement their own `searchResultsUpdater`.
+        ///
+        /// - returns: Instance of searchResultsUpdater proxy that wraps `searchResultsUpdater`.
+        public func createSearchResultsUpdaterProxy() -> RxSearchControllerResultsUpdaterProxy {
+            return RxSearchControllerResultsUpdaterProxy(parentObject: self)
+        }
     }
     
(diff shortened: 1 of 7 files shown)


RxCocoa/iOS/UISearchController+Rx.swift
```diff
@@ -65,6 +78,11 @@
                 .map {_ in}
         }
         
+        /// Reactive wrapper for `searchResultsUpdater` message.
+        public var updateSearchResultsFor: Observable<UISearchController> {
```
kzaher: I think this should be `ControlEvent` probably. I don't think we need `for`, just `updateSearchResults`.

Tests/RxCocoaTests/UISearchController+RxTests.swift
```diff
@@ -115,6 +115,24 @@ final class UISearchController_RxTests: RxTest {
...
+                }, onCompleted: {
+                    completed = true
+                })
+            searchController.searchResultsUpdater!.updateSearchResults(for: searchController)
+        }
+        
+        XCTAssertTrue(completed)
+        XCTAssertTrue(updated)
+    }
 }
```
kzaher: I think we would need to add subclassing tests.

Something equivalent to 

```
extension DelegateProxyTest {
    func test_UISearchBarDelegateExtension() {
        performDelegateTest(UISearchBarSubclass(frame: CGRect.zero))
    }
}
```

kzaher:
I think CI is failing for some legitimate reason.

```
Please run ./scripts/update-jazzy-config.rb
```

kzaher:
Hi guys,

CI is still failing because headers are invalid.

PR #1116, opened 2017-02-26
Publisher - wrapper for PublishSubject that cannot error

DenHeadless:
Hey!

So, based on discussion on #928, i hacked a quick `Publisher` class prototype. It's a wrapper on `PublishSubject`, that can't error. Similar to `Variable` being a wrapper on `BehaviourSubject`, it can be converted to `Driver`.

Because it's the first time i'm trying contributing to RxSwift, I'm basically in "I have no idea what i'm doing" mode, so sorry if code will seem rushed or out of place/guidelines of the repo.

I did not add any tests except for proof-of-concept one, since the design of this class needs to be discussed(obviously). I also did not include `EventBus/Event/EventSource` and `Publish` strategy, because i don't have a vision of how they need to connect with `Driver`s.

Current limitations: there is no way to get current value from Publisher - for some reason SpinLock, AtomicInt, and several other types are not compiling - maybe because this is RxCocoa target, not RxSwift. 

@kzaher What do you think about all of this?

/cc @AndrewSB

### DIFF
--- a/RxCocoa/Traits/EventHub/EventHub.swift
+++ b/RxCocoa/Traits/EventHub/EventHub.swift
@@ -1,0 +1,43 @@
+//
+//  EventHub.swift
+//  RxCocoa
+//
+//  Created by Krunoslav Zaher on 9/26/16.
+//  Copyright © 2016 Krunoslav Zaher. All rights reserved.
+//
+
+#if !RX_NO_MODULE
+    import RxSwift
+#endif
+
+/**
+ Unit that represents observable sequence with following properties:
+ 
+ - it never fails
+ - it delivers events on `MainScheduler.instance`
+ - all observers share sequence computation resources
+ - computation of elements is reference counted with respect to the number of observers
+ - if there are no subscribers, it will release sequence computation resources
+ 
+ `EventHub<Element>` can be considered a builder pattern for observable sequences that model imperative events part of the application.
+ 
+ To find out more about units and how to use them, please visit `Documentation/Units.md`.
+ */
+public typealias EventHub<E> = SharedSequence<PublishSharingStrategy, E>
+
+public struct PublishSharingStrategy : SharingStrategyProtocol {
+    public static var scheduler: SchedulerType { return publisherObserveOnScheduler }
+    
+    public static func share<E>(_ source: Observable<E>) -> Observable<E> {
+        return source.share()
+    }
+}
+
+extension SharedSequenceConvertibleType where SharingStrategy == PublishSharingStrategy {
+    /// Adds `asEventHub` to `SharingSequence` with `PublishSharingStrategy`.
+    public func asEventHub() -> EventHub<E> {
+        return asSharedSequence()
+    }
+}
+
+fileprivate var publisherObserveOnScheduler: SchedulerType = MainScheduler.instance
--- a/RxCocoa/Traits/Publisher.swift
+++ b/RxCocoa/Traits/Publisher.swift
@@ -1,0 +1,42 @@
+//
+//  Publisher.swift
+//  RxCocoa
+//
+//  Created by Krunoslav Zaher on 3/28/15.
+//  Copyright © 2017 Krunoslav Zaher. All rights reserved.
+//
+
+import Dispatch
+#if !RX_NO_MODULE
+    import RxSwift
+#endif
+
+/// Publisher is a wrapper for `PublishSubject`.
+///
+/// Unlike `PublishSubject` it can't terminate with error, and when publisher is deallocated
+/// it will complete it's observable sequence (`asObservable`).
+public final class Publisher<Element> {
+    
+    public typealias E = Element
+    public typealias SharingStrategy = PublishSharingStrategy
+    
+    private let _subject: PublishSubject<Element>
+    
+    public func publish(_ event: E) {
+        _subject.onNext(event)
+    }
+    
+    /// Initializes variable with initial value.
+    public init() {
+        _subject = PublishSubject()
+    }
+    
+    /// - returns: Canonical interface for push style sequence
+    public func asObservable() -> Observable<E> {
+        return _subject
+    }
+    
+    deinit {
+        _subject.on(.completed)
+    }
+}
--- a/Tests/RxCocoaTests/PublisherTests.swift
+++ b/Tests/RxCocoaTests/PublisherTests.swift
@@ -1,0 +1,147 @@
+//
+//  PublisherTests.swift
+//  Tests
+//
+//  Created by Krunoslav Zaher on 2/26/17.
+//  Copyright © 2017 Krunoslav Zaher. All rights reserved.
+//
+
+import Dispatch
+import RxSwift
+import RxCocoa
+import XCTest
+import RxTest
+
+class PublisherTest: RxTest {}
+
+extension PublisherTest {
+    func testPublisherSharing() {
+        let scheduler = TestScheduler(initialClock: 0)
+        
+        let observer1 = scheduler.createObserver(Int.self)
+        let observer2 = scheduler.createObserver(Int.self)
+        let observer3 = scheduler.createObserver(Int.self)
+        var disposable1: Disposable!
+        var disposable2: Disposable!
+        var disposable3: Disposable!
+        
+        let publisher = Publisher<Int>()
+        scheduler.scheduleAt(100) { publisher.publish(0) }
+        scheduler.scheduleAt(210) { publisher.publish(1) }
+        scheduler.scheduleAt(225) { publisher.publish(2) }
+        scheduler.scheduleAt(245) { publisher.publish(3) }
+        scheduler.scheduleAt(265) { publisher.publish(4)  }
+        
+        scheduler.scheduleAt(200) {
+            disposable1 = publisher.asObservable().subscribe(observer1)
+        }
+        
+        scheduler.scheduleAt(220) {
+            disposable2 = publisher.asObservable().subscribe(observer2)
+        }
+        
+        scheduler.scheduleAt(235) {
+            disposable1.dispose()
+        }
+        
+        scheduler.scheduleAt(260) {
+            disposable2.dispose()
+        }
+        
+        // resubscription
+        
+        scheduler.scheduleAt(260) {
+            disposable3 = publisher.asObservable().subscribe(observer3)
+        }
+        
+        scheduler.scheduleAt(285) {
+            disposable3.dispose()
+        }
+        
+        scheduler.start()
+        
+        XCTAssertEqual(observer1.events, [
+            next(210, 1),
+            next(225, 2)
+            ])
+        
+        XCTAssertEqual(observer2.events, [
+            next(225, 2),
+            next(245, 3)
+            ])
+        
+        XCTAssertEqual(observer3.events, [
+            next(265, 4)
+            ])
+    }
+}
+
+extension PublisherTest {
+    func testEventHubAlwaysObservingOnMainThread() {
+        var observedOnMainQueue = false
+        
+        let expectSubscribeOffMainQueue = expectation(description: "Did subscribe off main thread")
+        
+        let publisher = Publisher<Int>()
+        
+        _ = publisher.asEventHub().asObservable().subscribe(onNext: { value in
+            XCTAssertTrue(DispatchQueue.isMain)
+            observedOnMainQueue = true
+            XCTAssertEqual(value, 1)
+        })
+        
+        doOnBackgroundQueue {
+            let d = publisher.asObservable().subscribe { n in
+                
+            }
+            let d2 = publisher.asObservable().subscribe { n in
+                
+            }
+            doOnMainQueue {
+                d.dispose()
+                d2.dispose()
+                expectSubscribeOffMainQueue.fulfill()
+            }
+        }
+        
+        publisher.publish(1)
+        
+        waitForExpectations(timeout: 1.0) { error in
+            XCTAssertNil(error)
+        }
+        
+        XCTAssertTrue(observedOnMainQueue)
+    }
+}
+
+// MARK: map
+extension PublisherTest {
+    func testPublisherMap() {
+        let publisher = Publisher<Int>()
+        
+        let observer = PrimitiveMockObserver<Int>()
+        let _ = publisher.asEventHub().map { $0 + 1 }.asObservable().subscribe(observer)
+        
+        publisher.publish(0)
+        
+        XCTAssertEqual(observer.events, [next(1)])
+    }
+}
+
+// MARK: combine with Driver Sequence
+extension PublisherTest {
+    func testEventHubIsDriverCombinable() {
+        let publisher = Publisher<Void>()
+        
+        let driver = Driver.just(1)
+        let exp = expectation(description: "With latest from driver")
+        _ = publisher.asEventHub().map { $0 }.withLatestFrom(driver).asDriver(onErrorJustReturn: 0).drive(onNext: { value in
+            XCTAssertEqual(value, 1)
+            exp.fulfill()
+        })
+        
+        publisher.publish()
+        
+        waitForExpectations(timeout: 0.3, handler: nil)
+    }
+}
--- a/Tests/RxCocoaTests/PublisherTests.swift
+++ b/Tests/RxCocoaTests/PublisherTests.swift
@@ -12,10 +12,10 @@ import RxCocoa
 import XCTest
 import RxTest
 
-class PublisherTest: RxTest {}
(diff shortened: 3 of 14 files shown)


kzaher:
Hi @DenHeadless ,

> So, based on discussion on #928, i hacked a quick Publisher class prototype. It's a wrapper on PublishSubject, that can't error. Similar to Variable being a wrapper on BehaviourSubject, it can be converted to Driver.

I think `Publisher` should be only converted to `EventHub ` that has `publish` sharing strategy, and not to `Driver` because it has `shareReplayLatestWhileConnected` sharing strategy.

> I did not add any tests except for proof-of-concept one, since the design of this class needs to be discussed(obviously). I also did not include EventBus/Event/EventSource and Publish strategy, because i don't have a vision of how they need to connect with Drivers.

Yes, I think we need to chose a name for a unit that has `publish` sharing strategy.
Let's use `EventHub` as a `Driver` equivalent concept. That name kind of makes sense to me. `Event` -> because it's meant to model events over time, `Hub` -> implies sharing and that multiple observers can attach to a single source.
I think that name is good enough for now.

> Current limitations: there is no way to get current value from Publisher - for some reason SpinLock, AtomicInt, and several other types are not compiling - maybe because this is RxCocoa target, not RxSwift.

`Publisher` shouldn't have `value`. It's meant to dispatch events.

DenHeadless:
Okay, so what are next steps? Do i wait for you to loosen constraints on SharedSequence so that it could be combined with EventHub, or do i start implementing EventHub and publish strategy?

kzaher:
Hi @DenHeadless ,

I think we don't need to loosen the constraints on `SharedSequence` for this thing to work.

It would be great if you could clean up the code and finish the prototype.

DenHeadless:
Sure! I'll get it done on the weekend and ping you back when i'm done 👍

DenHeadless:
@kzaher So, i cleaned up a little bit, added `EventHub` `SharedSequence` with `PublishSharingStrategy`.

It's now combinable with Drivers, but that still produces Observable sequence, that does not guarantee(externally), that it will not error. Maybe you can help me with direction on what to do next? How we communicate to the user that EventHub is errorless?

P.S. Still without tests as i'm still trying to figure out how this stuff is supposed to work.

RxCocoa/CocoaUnits/EventHub/EventHub.swift
```diff
@@ -0,0 +1,43 @@
...
+ 
+ To find out more about units and how to use them, please visit `Documentation/Units.md`.
+ */
+public typealias EventHub<E> = SharedSequence<PublishSharingStrategy, E>
+
+public struct PublishSharingStrategy : SharingStrategyProtocol {
+    public static var scheduler: SchedulerType { return publisherObserveOnScheduler }
+    
+    public static func share<E>(_ source: Observable<E>) -> Observable<E> {
+        return source
```
kzaher: I think we are missing `share` sharing strategy.

kzaher:
Hi @DenHeadless ,

> How we communicate to the user that EventHub is errorless?

How do you communicate that it can error out?

> Maybe you can help me with direction on what to do next? 

I would suggest writing unit tests. You can take a look at how we've done this in case of `Driver`. `EventHub` has a slightly different strategy so one would need to create a equivalent tests for it.

RxCocoa/CocoaUnits/EventHub/EventHub.swift
```diff
@@ -0,0 +1,43 @@
...
+}
+
+extension SharedSequenceConvertibleType where SharingStrategy == PublishSharingStrategy {
+    /// Adds `asEventHub` to `SharingSequence` with `PublishSharingStrategy`.
+    public func asEventHub() -> EventHub<E> {
+        return asSharedSequence()
+    }
+}
+
+fileprivate var publisherObserveOnScheduler: SchedulerType = MainScheduler.instance
```
kzaher: We would need to use the same scheduler and system driver is using. So that `driveOnScheduler` works.

In 4.0 we will separate scheduler strategy from sharing strategy so we can reuse `SchedulerStrategy` for both of them.

DenHeadless:
> How do you communicate that it can error out?

Well, it's convertible to `Observable`, but when we try to convert it to `Driver`, we are forced to choose a way to recover. For example, both `ControlProperty` and `ControlEvent` have `asDriver` method without arguments, since there's no way they will receive error. I thought that having such method for `Publisher` makes complete sense. What do you think?

Other than that, i have no questions, and will proceed with unit tests.

RxCocoa/CocoaUnits/EventHub/EventHub.swift
DenHeadless: Is there something to be done here? Do i need to change scheduler?

RxCocoa/CocoaUnits/EventHub/EventHub.swift
DenHeadless: Added

DenHeadless:
I've added a couple of unit tests. Not sure they cover everything needed, but it's a start. 

If we should have more tests or some specific ones, please ping me, i'll add them.

DenHeadless:
@kzaher Test failure seems unrelated to PR, tests pass locally on my machine. What do you think, do i need to add something else or change implementation in some way?

DenHeadless:
Rebased on current develop and squashed all commits into one.

sergdort:
I wonder if it worth to be consistent with Java and call it [Relay](https://github.com/JakeWharton/RxRelay)?

RxCocoa/Traits/EventHub/EventHub.swift
```diff
@@ -0,0 +1,43 @@
...
+ - it delivers events on `MainScheduler.instance`
+ - all observers share sequence computation resources
+ - computation of elements is reference counted with respect to the number of observers
+ - if there are no subscribers, it will release sequence computation resources
+ 
+ `EventHub<Element>` can be considered a builder pattern for observable sequences that model imperative events part of the application.
+ 
+ To find out more about units and how to use them, please visit `Documentation/Units.md`.
+ */
+public typealias EventHub<E> = SharedSequence<PublishSharingStrategy, E>
```
kzaher: Let's rename this to `Publisher`.

RxCocoa/Traits/EventHub/EventHub.swift
```diff
@@ -0,0 +1,43 @@
...
+}
+
+extension SharedSequenceConvertibleType where SharingStrategy == PublishSharingStrategy {
+    /// Adds `asEventHub` to `SharingSequence` with `PublishSharingStrategy`.
+    public func asEventHub() -> EventHub<E> {
+        return asSharedSequence()
+    }
+}
+
+fileprivate var publisherObserveOnScheduler: SchedulerType = MainScheduler.instance
```
kzaher: It should use the same scheduler system as `Driver` does.

RxCocoa/Traits/Publisher.swift
```diff
@@ -0,0 +1,42 @@
...
+import Dispatch
+#if !RX_NO_MODULE
+    import RxSwift
+#endif
+
+/// Publisher is a wrapper for `PublishSubject`.
+///
+/// Unlike `PublishSubject` it can't terminate with error, and when publisher is deallocated
+/// it will complete it's observable sequence (`asObservable`).
+public final class Publisher<Element> {
```
kzaher: Let's name this like @sergdort suggests `PublishRelay`.

Maybe it would also make sense to create something like:
```
enum Relay {
    class Publish<Element> {
       ...;
    }
}
``` 

Consistent with https://github.com/JakeWharton/RxRelay.

We will also add `BehaviorRelay` and deprecate `Variable`. (but that's separate work).

RxCocoa/Traits/Publisher.swift
```diff
@@ -0,0 +1,42 @@
...
+    
+    public typealias E = Element
+    public typealias SharingStrategy = PublishSharingStrategy
+    
+    private let _subject: PublishSubject<Element>
+    
+    public func publish(_ event: E) {
```
kzaher: We can make this consistent with Relay and name it `accept`.

RxCocoa/Traits/Publisher.swift
```diff
@@ -0,0 +1,42 @@
...
+        _subject = PublishSubject()
+    }
+    
+    /// - returns: Canonical interface for push style sequence
+    public func asObservable() -> Observable<E> {
+        return _subject
+    }
+    
+    deinit {
+        _subject.on(.completed)
```
kzaher: Let's not do this any more.

RxCocoa/Traits/Publisher.swift
```diff
@@ -0,0 +1,42 @@
...
+///
+/// Unlike `PublishSubject` it can't terminate with error, and when publisher is deallocated
+/// it will complete it's observable sequence (`asObservable`).
+public final class Publisher<Element> {
+    
+    public typealias E = Element
+    public typealias SharingStrategy = PublishSharingStrategy
+    
+    private let _subject: PublishSubject<Element>
+
```
kzaher: Missing a comment.

Tests/RxCocoaTests/PublisherTests.swift
```diff
@@ -0,0 +1,147 @@
...
+            XCTAssertNil(error)
+        }
+        
+        XCTAssertTrue(observedOnMainQueue)
+    }
+}
+
+// MARK: map
+extension PublisherTest {
+    func testPublisherMap() {
```
kzaher: Let's remove this test. 

We are already testing `Driver`(SharedSequence). We'll rename `Driver` tests in future to be `SharedSequence` tests, so they will cover all of the operators.

Tests/RxCocoaTests/PublisherTests.swift
```diff
@@ -0,0 +1,147 @@
...
+        
+        publisher.publish(0)
+        
+        XCTAssertEqual(observer.events, [next(1)])
+    }
+}
+
+// MARK: combine with Driver Sequence
+extension PublisherTest {
+    func testEventHubIsDriverCombinable() {
```
kzaher: Let's also remove this test.

Tests/RxCocoaTests/PublisherTests.swift
```diff
@@ -0,0 +1,147 @@
...
+            ])
+        
+        XCTAssertEqual(observer3.events, [
+            next(265, 4)
+            ])
+    }
+}
+
+extension PublisherTest {
+    func testEventHubAlwaysObservingOnMainThread() {
```
kzaher: I don't think this unit tests tests anything useful. 

The logic should be:
* subscribe
* send element on background thread to `publisher`
* make sure element is received on main thread and fulfill expectation

RxCocoa/Traits/Publisher.swift
sergdort: Nice 👍  No `Variable` 🎉
Some people will be pissed off I guess =)

DenHeadless:
Addressed some comments from review, please ping me in slack or here if needed more adjustments.

kzaher:
Hi @DenHeadless ,

Unfortunately we can't use the name `Publisher` since it clashes with some other concepts in RxJava.

After a lot of brainstorming I think that it would be best to rename **Publisher** to **Signal**. Since there was a lot of necessary changes that we needed to do to be able to merge this, I've made the necessary changes.

It seems to me that everybody already intuitively knows how something like `Signal` should behave and it seems that a lot of reactive libraries are already using the terminology.

I guess there could be issues if somebody is using multiple reactive libraries inside their app, but I guess those could be resolved by using typealiases.

DenHeadless:
Great, let's use `Signal` =)  Excited to see this in RxSwift 4 🎉

freak4pc:
@kzaher Just my personal opinion but I don't think it's a good call using `Signal`. People coming from ReactiveSwift / RAC will be very confused. 

What about something like `Producer` or `InfallibleSubject`, or something of sort?

kzaher:
@freak4pc it is approx the same abstract concept in ReactiveSwift/RAC with approx similar properties as far as I can tell, so I'm hoping it won't confuse anyone.

I was hoping that would actually help people from different reactive communities more easily 
communicate.

IMHO `Producer` doesn't communicate `share` strategy that well and we both have similar names for other things internally and I'm pretty sure RxJava also has `Producer` :)

`InfallibleSubject` has it's own downsides.

¯\\\_(ツ)\_/¯

freak4pc:
> it is approx the same abstract concept in ReactiveSwift/RAC with approx similar properties as far as I can tell, so I'm hoping it won't confuse anyone.

I really don't think it is though? We have documentation stating that Signal and SignalProducer are both just Observable (with the difference of being cold/hot)

In RAC Signal can fail (unless NoError), and also it doesn't share by default. So again I think this is very confusing. This really isn't a regular "Single" per-se. I know this is an internal naming so it probably doesn't matter all that much but perhaps its worth rethinking. 

Just me 2cents 🤷‍♂️

kzaher:
> I really don't think it is though? We have documentation stating that Signal and SignalProducer are both just Observable (with the difference of being cold/hot). EDIT: Actually we don't use the RAC naming there but we do talk about the separation which would make users infer that possibly: https://github.com/ReactiveX/RxSwift/blob/master/Documentation/HotAndColdObservables.md

Let me try to provide some clarity.
**It's an observable sequence that has additional compile time provable traits. It's not a separate concept like in RAC that has different subscription semantics.**
https://github.com/ReactiveX/RxSwift/blob/master/Documentation/Traits.md

But, **the high level concept behaves in similar way. You have sharing with publish strategy.**

In this library when you do `signal.asObservable()` you haven't changed the way subscription and disposal works, you've just said to compiler, let's get the underlying observable sequence. This isn't the case with some other libraries because it's a different concept.

freak4pc:
As I said it seems to make less sense, since in RAC Signal is the "basic building block" for anything pretty much, and here in RxSwift we use the same naming for a Trait. I guess that's the part I find confusing as I don't think they correlate but as I said this is just personal taste I guess

kzaher:
I understand why you think this way, but I guess I'll try to provide additional explanations if somebody accidentally stumbles on this discussion.

Every unix/linux program uses signals. When you do `kill -9` or hit ctrl+c you are sending signal to a program. Signal doesn't have state, it's just a notification, so I'm hoping it should be intuitive from that perspective what are its properties. Signaling IMHO is more closer to sending events then sending state, although electrical signal is state for sure.

For every possible decision I can tell you thousand of reason for and against, but at the end of the day I need to make some decision.

As far as for other reactive libraries and ecosystem this is kind of what I've found.

RAC copied the naming from Elm
> Other naming differences are typically inspired by significantly better alternatives from Haskell or Elm (which is the primary source for the “signal” terminology).

It seems to me that Elm actually uses `share(replay: 1)` not `share` like RAC has since it's meant to represent behaviour (as in `BehaviorSubject`).  But at least I do agree with RAC on this point and it seems to me that `share` strategy is more intuitive for something called `Signal`.
http://package.elm-lang.org/packages/elm-lang/core/3.0.0/Signal
> A signal is a value that changes over time. For example, we can think of the mouse position as a pair of numbers that is changing over time, whenever the user moves the mouse.

... although I haven't worked with Elm and have no idea how that works under the hood. It seems that Elm is getting rid of signal terminology though.
http://elm-lang.org/blog/farewell-to-frp
> everything related to signals has been replaced with something simpler and nicer

According to its creators Rx has actually nothing to do with FRP, but in FRP ...
https://en.wikipedia.org/wiki/Functional_reactive_programming
> Modeling values that vary over continuous time, called "behaviors" and later "signals".

ReactiveKit (former Bond) has a bit weird strategy, but also some kind of publish as far as I can tell https://github.com/ReactiveKit/ReactiveKit/blob/master/Sources/Signal.swift
While interstellar has `share(replay:1)`
https://github.com/JensRavens/Interstellar/blob/master/Sources/Interstellar/Signal.swift

sergdort:
TBH I like name `Signal` I think it exactly what it does :) 

@freak4pc In `RAC` `Signal.pipe()` is essentially a publish subject as you get sequence and underlying observer where you can "push" events

yuzushioh:
Hi @DenHeadless @kzaher 

I was just wondering if this was solved in somewhere in the PR. I totally agree with the opinion that   `PublishReplay` could be convertible with `Driver` like ControlProperty and ControlEvent.  And it would be very useful if such methods will be available in RxSwift4.0

>Well, it's convertible to Observable, but when we try to convert it to Driver, we are forced to choose a way to recover. For example, both ControlProperty and ControlEvent have asDriver method without arguments, since there's no way they will receive error. I thought that having such method for Publisher makes complete sense. What do you think?

PR #1118, opened 2017-02-28
Add AnyVariable (get-only Variable)

inamiy:
This PR adds `AnyVariable` as a get-only `Variable`, which is highly useful when encapsulating `Variable` inside some view model, disallowing its binding from outside.

This type is almost the same as ReactiveSwift's [Property](https://github.com/ReactiveCocoa/ReactiveSwift/blob/1.0.1/Sources/Property.swift).

### DIFF
--- a/RxSwift/Subjects/Variable.swift
+++ b/RxSwift/Subjects/Variable.swift
@@ -10,7 +10,7 @@
 ///
 /// Unlike `BehaviorSubject` it can't terminate with error, and when variable is deallocated
 /// it will complete it's observable sequence (`asObservable`).
-public final class Variable<Element> {
+public final class Variable<Element>: ObservableConvertibleType {
 
     public typealias E = Element
     
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -168,6 +168,7 @@ custom_categories:
   - ScheduledItemType
 - name: RxSwift/Subjects
   children:
+  - AnyVariable
   - AsyncSubject
   - BehaviorSubject
   - PublishSubject
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -21,6 +21,13 @@
 		1AF67DA61CED430100C310FA /* ReplaySubjectTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1AF67DA51CED430100C310FA /* ReplaySubjectTest.swift */; };
 		1AF67DA71CED430100C310FA /* ReplaySubjectTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1AF67DA51CED430100C310FA /* ReplaySubjectTest.swift */; };
 		1AF67DA81CED430100C310FA /* ReplaySubjectTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1AF67DA51CED430100C310FA /* ReplaySubjectTest.swift */; };
+		1FC9E3EA1E64EE92006B00D7 /* AnyVariable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FC9E3E91E64EE92006B00D7 /* AnyVariable.swift */; };
+		1FC9E3EB1E64EF08006B00D7 /* AnyVariable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FC9E3E91E64EE92006B00D7 /* AnyVariable.swift */; };
+		1FC9E3EC1E64EF09006B00D7 /* AnyVariable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FC9E3E91E64EE92006B00D7 /* AnyVariable.swift */; };
+		1FC9E3ED1E64EF09006B00D7 /* AnyVariable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FC9E3E91E64EE92006B00D7 /* AnyVariable.swift */; };
+		1FC9E3EF1E64EF57006B00D7 /* AnyVariableTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FC9E3EE1E64EF57006B00D7 /* AnyVariableTest.swift */; };
+		1FC9E3F01E64EF57006B00D7 /* AnyVariableTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1FC9E3EE1E64EF57006B00D7 /* AnyVariableTest.swift */; };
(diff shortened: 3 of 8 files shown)


kzaher:
I guess it could also make sense to add `Value` unit to RxCocoa.

inamiy:
Hi @kzaher, thank you for mention :)

If I understand correctly:

- `Value` (get-only `Variable`) is not related to `Driver` or threading, as `Variable` isn't.
- `Value` doesn't require any Cocoa-related types, as `Variable` doesn't. So I think RxSwift (not RxCocoa) is a better place to live in.
- The type name `Value` has a bit of difficulty when one wants to get current element, i.e. `value.value`

RxSwift/Subjects/Variable.swift
```diff
@@ -10,7 +10,7 @@
 ///
 /// Unlike `BehaviorSubject` it can't terminate with error, and when variable is deallocated
 /// it will complete it's observable sequence (`asObservable`).
-public final class Variable<Element> {
+public final class Variable<Element>: ObservableConvertibleType {
```
inamiy: I got your point, but I just wonder if `ObservableConvertibleType` will ever be a useful protocol for 3rd party users :sweat_smile:

tarunon: Hi there, I'm interested that why `Variable` does not conform to `ObservableConvertibleType` and `ObserverType`.

In my opinion, I think that the type `A` that implemented `ObservableConvertibleType` make `B` as `Observable` and some other object should retain B object, not A object directly.
Then `Variable` case, the `Observable` is a internal `BehaviorSubject` so, `Variable` never be retained other object. 

But now, maybe some of Observable family has `ObservableConvertibleType` directly. Can we change this?

kzaher: Hi @tarunon ,

the mere fact that we are relying on the fact will somebody retain variable or not means that this is fragile API (the part about completing variable observable sequence on deinit). That's why I don't like it.

It is pretty normal to create a reference to extension source, being it `ObservableType` or `ObservableConvertibleType`. This is being done all the time, and it shouldn't be a problem.

That's why `Variable` doesn't conform to `ObservableConvertibleType`, because that increases the chances of somebody retaining it.

Every solution has it's drawbacks.

kzaher:
Hi @inamiy ,

> * Value (get-only Variable) is not related to Driver or threading, as Variable isn't.

I think it should be tied because it's a safer alternative. The only environment where that unit makes sense is a stateful one (like client app).

Since `Driver` is just a specialization of `SharedSequence`, so should probably `Value` be a typealias of some more general concept that always has latest value available and ensures being executed on some serial scheduler.

Since `SharedSequence` already ensures scheduler (when converting using `startWith`), then the only real difference is one additional `observeOn` when converting from `Variable` to ensure all properties are satisfied.

> Value doesn't require any Cocoa-related types, as Variable doesn't. So I think RxSwift (not RxCocoa) is a better place to live in.

I don't think that in general dependency chain should be the only reason to decide how public API should look like. If that was the case then we should also make `Queue` and `PriorityQueue` public.

Because:
* to introduce this unit it should have a general enough use cases
* it should have a private initializer
* it should be able to create it from `Variable`(asValue) and `Driver`/`SharedSequence`(`asValue`), and `SharedSequence` is defined in RxCocoa
* all code in RxSwift should be cross platform

I think it would be more suitable to add it to `RxCocoa`.

> The type name Value has a bit of difficulty when one wants to get current element, i.e. value.value

I don't think that it should have an API that allows access to its latest element, just that it should guarantee that initial value is always delivered immediately upon subscription.

Having a `value` property would mean that the concept is imperative in nature vs declarative like the rest of library. That would be a road to complexity IMHO.

inamiy:
@kzaher
I should have emphasized that **`Value` is just a wrapper of `Variable`** so all of the next scenarios should follow the same as `Variable`, i.e.

- `Value` is also a cross-platform, and should be place in RxSwift (not RxCocoa)
- Can add `Value+Driver` extension in RxCocoa if you want, but not mandatory.

BTW:

> I don't think that it should have an API that allows access to its latest element, just that it should guarantee that initial value is always delivered immediately upon subscription.

`Value` as "`Variable` without `var value { get set }`" sounds an interesting idea!
The naming problem will be solved if this idea is applied.

RxSwift/Subjects/Variable.swift
tarunon: Thanks for the discussion.
Finally, I can understand that why `Variable` doesn't conform to `ObservableConvertibleType`.

@inamiy 
About `ObserverType`, I think that is not a real problem. For example, `UIBindingObserver` has a same implement that `func bindTo(_ variable: Variable<E>) -> Disposable`'s one.

If my understanding is correct, adding `ObserverConvertibleType` may help us. 
I think it is effective not only `Variable`.

kzaher:
Hi @inamiy ,

> I should have emphasized that Value is just a wrapper of Variable so all of the next scenarios should follow the same as Variable, i.e.

Yep, got it :) I'm not trying to say is it useful or not, I'm trying to say that there are a lot more criteria that need to be satisfied for us to pull something in RxSwift project.

> Value is also a cross-platform, and should be place in RxSwift (not RxCocoa)

When I use term term cross platform, I'm not referring to iOS, macOS, tvOS, .... but to RxJS, RxJava, RxRust, RxGo ... `Value` would be optimized mostly for UI code. The place where it would fit best is it's own separate repository with all of the other extension units, but it's more convenient to have them in RxCocoa for now.

Everything that is entering RxSwift core library should **strive to be aligned with the core observable sequence interface/concept** (observable sequence compatible), **strive to follow cross platform implementations**, have no external dependencies, promote declarative style, be useful enough for us to maintain it here (maintaining Swift code is not cheap), align with future plans  (potentially adding `Value` to RxCocoa) ...

It would be extremely easy and popular choice to add a lot of convenience extensions in there for us to maintain, but it's extremely hard to get them out or actually maintain them.

inamiy:
@kzaher
OK, I understood what you meant by "cross-platform".
But how about `Variable`? It is also uncommon in ReactiveX, although still core RxSwift.framework.
To me, putting `Value` into RxCocoa is very inconsistent, and I'm not positive to fix in that way.

I would like to hear if anyone else is :+1: or :-1: to this PR.
I'm also okay with this PR closed and someone takes over to target RxCocoa instead of me.

RxSwift/Subjects/Variable.swift
inamiy: (Just for memo) `Variable` should not conform to `ObserverType` because terminal events will be ignored, i.e. `variable.on(.completed))` then `variable.on(.next(value))` will send value. (And using `fatalError` inside is not an option...)

kzaher:
Ok, cool.

inamiy:
FYI I open-sourced https://github.com/inamiy/RxProperty .

PR #1120, opened 2017-03-01
Feature/zip with

nxtstep:
This PR adds a convenience operator to `ObservableType`.`zip(with:)` that zips two observable sequences into one sequence on the same Rx Monad.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,6 @@ Carthage/Build
 
 .build/
 Packages/
+
+# AppCode
+.idea/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ## Master
 
 * Adds `AsyncSubject` implementation
+* Adds `zip(with:)` operator
 
 ## Anomalies
 * #1081, #1087 - Improves DelegateProxy `responds(to:)` selector logic to only respond to used selectors.
--- a/RxSwift/Observables/Observable+Multiple.swift
+++ b/RxSwift/Observables/Observable+Multiple.swift
@@ -40,6 +40,21 @@ extension Observable {
     }
 }
 
+extension ObservableType {
+
+    /**
+     Zips `self` with the second observable sequence into one observable sequence by using the selector function whenever all of the observable sequences have produced an element.
+
+     - seealso: [zip operator on reactivex.io](http://reactivex.io/documentation/operators/zip.html)
+
+     - parameter resultSelector: Function to invoke for each series of elements at corresponding indexes in the sources.
+     - returns: An observable sequence containing the result of combining elements of the sources using the specified result selector function.
+     */
+    public func zip<O2: ObservableConvertibleType, ResultType>(with second: O2, resultSelector: @escaping (E, O2.E) throws -> ResultType) -> Observable<ResultType> {
+        return Observable.zip(asObservable(), second.asObservable(), resultSelector: resultSelector)
+    }
+}
+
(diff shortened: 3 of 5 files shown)


kzaher:
Hi @nxtstep ,

it seems to me that `zip` is a symmetrical operator. I'm not sure why do you want this overload?

nxtstep:
@kzaher I thought this operator could be pretty convenient even though `zip` being a symmetric operator.

So let me expand a little on the usage of this convenience operator. I've seen and used it quite often in Android/Java projects where [the `zipWith` operator](http://reactivex.io/documentation/operators/zip.html#collapseRxJava 1․x) is available.

I intended it to be a shortcut for:
```
Observable.just(something)
        .flatMap { Observable.zip(Observable.just($0), Observable.create { ... }) { join($0, $1) }
```

to: 

```
Observable.just(something)
        .zip(with: Observable.create { ... }) { join($0, $1) }
```

IRL this becomes useful when you have more operators than only `Observable.just()` on an observable stream. I think it's a more elegant approach and serves readability.

sergdort:
I agree with @nxtstep,  but I think it's pretty simple to implement.
I would propose to add it to [RxSwiftExt](https://github.com/RxSwiftCommunity/RxSwiftExt) rather then to the Core library

kzaher:
Hi @nxtstep ,

I would rather not include it for now in the main repo. There is similar discussion for #1118. If that would fit it the Ext library, that would be great.

I'm wondering why would somebody write 
```swift
Observable.just(something)
        .flatMap { Observable.zip(Observable.just($0), Observable.create { ... }) { join($0, $1) }
```

instead of 
```swift
let second  = Observable.create { ... }
Observable.zip(Observable.just(something), second) { join($0, $1) }
```

What we can do to make this even more shorter is to add `zip` and `combineLatest` overloads that don't contain additional result closure.

nxtstep:
Hi @kzaher, @sergdort 

I have two general use-cases for a `zip(with:)` operator that can provide more an elegant/readable stream.

E.g.:
```
// Passing down the initial output from source1 through the the stream and use it as input for source2
repo.getEntity(key)
    .flatMap { otherRepo.fetchOther($0.otherKey)
        .zip(with: Observable.just($0)) { ($0, $1) }
    .map { convertModel($0) }
    ...
```

Prevent side-effect:
```
// Combined operation without side-effect
repo.save(entity)
    .flatMap { preferences.save($0.key)
        .zip(with: Observable.just($0)) { ($0, $1) }
    ...
```

To prevent this:
```
// Combined operation with side-effect
repo.save(entity)
    .do(onNext: { preferences.save($0.key)  <- Result ignored by original stream
        .subscribe { ... }  <- Scheduling is not captured by original stream
        .disposed(by: ... )  <-- Is not owned by original stream
    })
    ...
```

I think it's a very useful operator to provide more readability and can be used to prevent side-effects. And I would just merge it since it's not breaking anything; serves a purpose; and other Rx implementations do feature a `zipWith` operator as well.

kzaher:
Hm,

first one can be written more clearly IMHO by
```
repo.getEntity(key)
    .flatMap { input in
         otherRepo.fetchOther(input.otherKey)
                .map { ($0, input) }
    }
```

> and other Rx implementations do feature a zipWith operator as well.

Other implementations are sponsored by billion $ companies, don't have language that went through 4 major versions in 2 years and don't have compiler that spits out assembly code like bandwidth is irrelevant.

I would feel more comfortable for now to not include this, but will keep this in mind.

¯\\\_(ツ)\_/¯

PR #1123, opened 2017-03-06
[POC] Single + Maybe + Completeable

kzaher:
I'm playing with proof of concept for `Single`/`Maybe`/`Completeable` units from RxJava.

If somebody has more experience with RxJava it would be great to get some confirmations. 
I think this should be more or less it, but ¯\\\_(ツ)\_/¯ .

I don't think we'll be able to support all operators from those units from RxJava, but it seems to me like a really cheap and generic way to provide those units in RxSwift by using existing operators.

If somebody has a better idea how to do this, please share.


RxSwift/Units/Single+Maybe+Completeable.swift
```diff
@@ -0,0 +1,455 @@
...
+        }
+    }
+}
+
+// </Single>
+
+// <Maybe>
+
+public enum MaybeEvent<Element> {
+    /// One and only sequence element is produced. (underlying observable sequence emits: `.next(Element)`, `.completed`)
```
sergdort: Jus curious what value `MaybeEvent` provides in comparison with `Event`

sergdort: Oups there is on success 😄

RxSwift/Units/Single+Maybe+Completeable.swift
```diff
@@ -0,0 +1,455 @@
...
+/// Common functionality of all a push style sequences containing 0 or 1 element.
+public struct PrimitiveSequence<Proof, Element> {
+    fileprivate let source: Observable<Element>
+
+    fileprivate init(raw: Observable<Element>) {
+        self.source = raw
+    }
+}
+
+public enum SingleProof { }
```
sergdort: I'm not sure 🤔 with name 🙂
How about `Behavior`?

kzaher: I'm not sure what the best naming would be here. 

The purpose of that type was to use generic parameter `PrimitiveSequence` to mark that certain properties where proved.

We've used `SharingStrategy` on `SharedSequence` because it's purpose is to enable customization of sharing strategy.

Using proof was also little weird to me. I would like to find some better name.

sergdort: Should it be `Property` then?  🙂

RxSwift/Units/Single+Maybe+Completeable.swift
```diff
@@ -0,0 +1,455 @@
...
+/// Represents a push style sequence containing 1 element.
+public typealias Single<Element> = PrimitiveSequence<SingleProof, Element>
+
+public enum MaybeProof { }
+/// Represents a push style sequence containing 0 or 1 element.
+public typealias Maybe<Element> = PrimitiveSequence<MaybeProof, Element>
+
+public enum CompleteableProof { }
+/// Represents a push style sequence containing 0 elements.
+public typealias Completeable = PrimitiveSequence<CompleteableProof, Swift.Never>
```
sergdort: `Never` 👍

RxSwift/Units/Single+Maybe+Completeable.swift
```diff
@@ -0,0 +1,455 @@
...
+            }
+        }
+
+        return PrimitiveSequence(raw: source)
+    }
+
+    public func subscribe(_ observer: @escaping (MaybeEvent<ElementType>) -> ()) -> Disposable {
+        var stopped = false
+        return self.primitiveSequence.asObservable().subscribe { event in
+            guard !stopped else { return }
```
sergdort: To be honest I'm not a big fun of 'not' statement in guard, it's really hard for me to read =) not sure about anybody else 
IMO `if stoped { return }` is much more readable

RxSwift/Units/Single+Maybe+Completeable.swift
kzaher: Property looks better.

How about `SingleTrait`?

https://www.google.hr/search?q=trait&oq=trait&aqs=chrome..69i57j0l2j69i60j0l2.1088j1j7&sourceid=chrome&ie=UTF-8

sergdort: Yeah, `Trait` looks good 👍

AliSoftware:
These new units might deserve adding a paragraph in https://github.com/ReactiveX/RxSwift/blob/master/Documentation/Units.md 😉

kzaher:
@AliSoftware thnx for reminding, will add.

hzsweers:
`Completeable` is spelled differently than RxJava's `Completable`. Any reason for that?

PR #1133, opened 2017-03-13
Add debug() to PrimitiveSequence

muukii:
I've added debug() to PremitiveSequence.

Test is asSingle() only.
Is it also necessary to test other methods?

### DIFF
--- a/RxSwift/Units/PrimitiveSequence.swift
+++ b/RxSwift/Units/PrimitiveSequence.swift
@@ -508,6 +508,20 @@ extension PrimitiveSequence {
         -> PrimitiveSequence<Trait, Element> {
         return PrimitiveSequence(raw: source.retryWhen(notificationHandler))
     }
+
+    /**
+     Prints received events for all observers on standard output.
+
+     - seealso: [do operator on reactivex.io](http://reactivex.io/documentation/operators/do.html)
+
+     - parameter identifier: Identifier that is printed together with event description to standard output.
+     - parameter trimOutput: Should output be trimmed to max 40 characters.
+     - returns: An observable sequence whose events are printed to standard output.
+     */
+    public func debug(_ identifier: String? = nil, trimOutput: Bool = false, file: String = #file, line: UInt = #line, function: String = #function)
+        -> PrimitiveSequence<Trait, Element> {
+            return PrimitiveSequence(raw: Debug(source: self.source, identifier: identifier, trimOutput: trimOutput, file: file, line: line, function: function))
+    }
 }
 
 extension PrimitiveSequenceType where ElementType: SignedInteger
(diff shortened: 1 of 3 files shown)


kzaher:
Hi @muukii ,

I think it makes sense to add `debug` operator.

Could you please add it in a similar way like we've done for `map` operator? ... and I would be happy to merge this.

RxSwift/Units/PremitiveSequence+Debug.swift
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   Prints received events for all observers on standard output.
+
+   - seealso: [do operator on reactivex.io](http://reactivex.io/documentation/operators/do.html)
+
+   - parameter identifier: Identifier that is printed together with event description to standard output.
+   - parameter trimOutput: Should output be trimmed to max 40 characters.
+   - returns: An observable sequence whose events are printed to standard output.
+   */
+  public func debug(_ identifier: String? = nil, trimOutput: Bool = false, file: String = #file, line: UInt = #line, function: String = #function)
```
kzaher: Could we move this to `PrimitiveSequence.swift` and make it return `Single<E>`

Tests/RxSwiftTests/PrimitiveSequence+DebugTest.swift
```diff
@@ -0,0 +1,71 @@
...
+import RxSwift
+import RxTest
+
+class PrimitiveSequenceDebugTest : RxTest {
+
+}
+
+// MARK: debug
+extension PrimitiveSequenceDebugTest {
+  func testDebug_Completed() {
```
kzaher: We would just want to test that `debug` operator will return Single type and single element.

You can look at `func testSingle_map_producesSingleElement()` for content.

The reason why we aren't implementing other tests is because we are just reusing operators from normal observable sequence.

The only new thing we need to test is that if it receives one element, it will produce one element also.

We could theoretically add full suite of unit tests for those operators also, but that would simply cause explosion of unit tests and code, and don't think we would have much benefit, and would need to maintain another huge amount of code.

muukii:
@kzaher 
Thanks for your advice.
I've revised debug operator.
Please, Check it.

RxSwift/Units/PrimitiveSequence.swift
```diff
@@ -69,6 +69,24 @@ extension PrimitiveSequence: ObservableConvertibleType {
...
+  /**
+   Prints received events for all observers on standard output.
+
+   - seealso: [do operator on reactivex.io](http://reactivex.io/documentation/operators/do.html)
+
+   - parameter identifier: Identifier that is printed together with event description to standard output.
+   - parameter trimOutput: Should output be trimmed to max 40 characters.
+   - returns: An observable sequence whose events are printed to standard output.
+   */
+  public func debug(_ identifier: String? = nil, trimOutput: Bool = false, file: String = #file, line: UInt = #line, function: String = #function)
```
kzaher: Hi, could you please move it in extension scope with other operators.

Tests/RxSwiftTests/PrimitiveSequence+DebugTest.swift
```diff
@@ -0,0 +1,37 @@
...
+import RxSwift
+import RxTest
+
+class PrimitiveSequenceDebugTest : RxTest {
+
+}
+
+// MARK: debug
+extension PrimitiveSequenceDebugTest {
+  func testDebug_producesSingleElement() {
```
kzaher: Could you just move this unit tests with other unit tests in `PrimitiveSequenceTest`.

kzaher: Could you please move this tests to `PrimitiveSequenceTest` and group them with other tests.

Tests/RxSwiftTests/PrimitiveSequence+DebugTest.swift
```diff
@@ -0,0 +1,37 @@
...
+    let singleResult: Single<Int> = Single.just(1)
+      .debug()
+
+    let result = try! singleResult
+      .toBlocking().first()!
+    XCTAssertEqual(result, 1)
+  }
+
+  #if TRACE_RESOURCES
+  func testReplayNReleasesResourcesOnComplete() {
```
kzaher: I think these unit tests have wrong names. I don't think we even need to add them in this case since we are only reusing implementation, but if you want to add them, please rename them.

muukii:
@kzaher 
Thank you.
Fixed it.

PR #1131, opened 2017-03-10
[WIP] Feature - Materialize/Dematerialize operator

sergdort:
Hey guys I've created `materialize` operator regarding #1103 

The work is still in progress, and I'm not sure if we should use optimised version or just copy one from [RxSwiftExt](https://github.com/RxSwiftCommunity/RxSwiftExt)

PR #1138, opened 2017-03-17
Add `Event.isCompleted`

inamiy:
Currently `Event.isCompleted` is missing, and it will be useful especially when [`materialize()`](https://github.com/RxSwiftCommunity/RxSwiftExt/blob/2.1.0/Source/materialize.swift).

### DIFF
--- a/RxSwift/Event.swift
+++ b/RxSwift/Event.swift
@@ -36,7 +36,7 @@ extension Event : CustomDebugStringConvertible {
 }
 
 extension Event {
-    /// Is `Completed` or `Error` event.
+    /// Is `completed` or `error` event.
     public var isStopEvent: Bool {
         switch self {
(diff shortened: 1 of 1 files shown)


sergdort:
Hi @inamiy we are going to introduce `materialize` with #1131
There is `isStopEvent` not sure if helps

inamiy:
Hi @sergdort, thanks for letting me know 😉 
It's nice to have `materialize` to the core too!

But `isStopEvent` is not fully useful since it also returns `true` for `.error` event.
I want to know `.completed` only so that I can `filter`, `map`, etc more specifically.

(and BTW ReactiveCocoa supports both `isCompleted` and `isTerminating`)

kzaher:
I can see how this can be useful with `materialize`.

PR #1147, opened 2017-03-22
Add `var changed: Observable<E>` to `Variable`

inamiy:
There are often cases when one wants to **ignore `Variable`'s current value and only observe changed values**.
Though using `variable.asObservable().skip(1).subscribe(...)` is an easy solution, it is not as declarative as `variable.changed.subscribe(...)` (people often ask "why skip(1)?"), so I think adding `var changed: Observable<E>` to `Variable` helps code readability.

The basic idea is almost same as `ControlProperty.changed`.

### DIFF
--- a/RxSwift/Subjects/Variable.swift
+++ b/RxSwift/Subjects/Variable.swift
@@ -60,10 +60,17 @@ public final class Variable<Element> {
         _value = value
         _subject = BehaviorSubject(value: value)
     }
-    
+
+    /// Observable that synchronously sends current element and then changed elements.
+    ///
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @inamiy ,

I see how that extension property can be useful. 

I wouldn't want to merge this in this form because it's not exactly clear what should it's type be.

This is also related with #1116 or some other unit.

It's pretty simple to create that extension if needed and chose whichever type one wants.

inamiy:
Hi @kzahar,

OK, I understood we should have a better type for this hot and no-error observable.
I will close this PR for now until the new type is fixed.

PR #1150, opened 2017-03-24
Enable `asCompletable` on any `ObservableType`

mosamer:
This PR drops generic constraint on `ObservableType` extension containing `asCompletable() -> Completable`

It will allow `asCompletable()` to behave similar to `asSingle()` and `asMaybe()` by:
- Removing generic constraint of `ObservableType` extension limited usage for only `Swift.Never`.
- Error if received any next events


kzaher:
Hi @mosamer ,

can you maybe provide some context why do you want to convert any observable to `Completable`?

In case of `Observable<Never>` we have a compile time guarantee that the conversion will succeed.

The reason why we have an error case in conversion from `Observable<Element>` to `Single` is because there isn't any information recorded in the type system that could guarantee that the conversion will succeed.

mosamer:
Hi @kzaher,

A scenario I can think of is with POST network calls where response is not needed. For instance a save function which we are interested in knowing whether it completed _successfully_ or errors. Usually using `URLSession` extensions will return `Observable<Data>` and converting that to `Observable<Never>` might be cumbersome.

There was multiple questions on Slack channel to achieve this behaviour so I believe may be there exist some other scenarios as well.

kzaher:
Hi @mosamer ,

I think one can just do `flatMap { _ in Observable<Never>.empty() }`. One can even name this `igoreElements` or something.

fcy:
I'm looking for a way to make a network call which has side effects (so I don't need the returned value directly) but I want to know if the call completed successfully or not.

>I think one can just do flatMap { _ in Observable<Never>.empty() }

Will this suggestion work? Will it correctly map `onCompleted`  and `onError` from the source observable?

PR #1157, opened 2017-03-28
Introduce nested generics type at implementation of `Sink`.

tarunon:
Swift 3.1 is now available, and today we can use nested generics type.
I try to improve `Sink` implementation using nested generics class.  First I moved `MapSink` and `MapWithIndexSink` into there parent class.

## Pros
* Can eliminate many duplicate type defines and use same generics type and `typealias` between `Sink` and their parent.
* Can increase `Sink`s confidentiality. (now I used `private`)

## Cons
* Make many code differences, several other improvements may affected by this proposal.
* In the past, `Sink` was always written before their parent, cannot do that anymore. (Can write at `extension`, we need to write after their parent.)
* If there are shared `Sink` class, this proposal might be a not good way in shared `Sink` class.

I think this proposal improve many type defines. But It makes many code differences.
I think it is also good that waiting Swift 4.0.

### DIFF
--- a/RxSwift/Observables/Implementations/Map.swift
+++ b/RxSwift/Observables/Implementations/Map.swift
@@ -6,83 +6,47 @@
 //  Copyright © 2015 Krunoslav Zaher. All rights reserved.
 //
 
-final class MapSink<SourceType, O : ObserverType> : Sink<O>, ObserverType {
-    typealias Transform = (SourceType) throws -> ResultType
-
-    typealias ResultType = O.E
-    typealias Element = SourceType
-
-    private let _transform: Transform
-    
-    init(transform: @escaping Transform, observer: O, cancel: Cancelable) {
-        _transform = transform
-        super.init(observer: observer, cancel: cancel)
-    }
-
-    func on(_ event: Event<SourceType>) {
-        switch event {
-        case .next(let element):
-            do {
-                let mappedElement = try _transform(element)
-                forwardOn(.next(mappedElement))
-            }
-            catch let e {
-                forwardOn(.error(e))
-                dispose()
-            }
-        case .error(let error):
-            forwardOn(.error(error))
-            dispose()
-        case .completed:
-            forwardOn(.completed)
-            dispose()
-        }
-    }
-}
-
-final class MapWithIndexSink<SourceType, O : ObserverType> : Sink<O>, ObserverType {
-    typealias Selector = (SourceType, Int) throws -> ResultType
-
-    typealias ResultType = O.E
-    typealias Element = SourceType
-    typealias Parent = MapWithIndex<SourceType, ResultType>
-    
-    private let _selector: Selector
-
-    private var _index = 0
-
-    init(selector: @escaping Selector, observer: O, cancel: Cancelable) {
-        _selector = selector
-        super.init(observer: observer, cancel: cancel)
-    }
-
-    func on(_ event: Event<SourceType>) {
-        switch event {
-        case .next(let element):
-            do {
-                let mappedElement = try _selector(element, try incrementChecked(&_index))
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @tarunon ,

I like your proposal, I think we should definitely do it for RxSwift/Swift 4.0. 👍 

> Can eliminate many duplicate type defines and use same generics type and typealias between Sink and their parent.
> Can increase Sinks confidentiality. (now I used private)

🎉 

I'm little reluctant to do it right now in the middle of 3.0 release because we are risking breaking someones code with so many changes, and end users probably won't care about it at the moment :)

I would also like to understand what are performance implications of these changes before we make them.

freak4pc:
I agree with kzaher this would probably break RxSwift for anyone not using Xcode 8.3 ... that's tricky and even more tricky because 8.3 is broken in so many ways to begin with :)

tarunon:
I'll report measuring performance this pr when I have a time. Thank you discussion.

tarunon:
This is a report of measuring Performance between Nested Sink vs Existing MapSink.

## Conclusion
Nested `Sink` is slower than `MapSink`, I need investigate why init is slow when I use nested `Sink`, and improve implementation or recommend not adopting this pr.

My mac book spec is next.
```
MacBook Pro (15-inch, 2016), Processor: 2.9 GHz Intel Core i7, Memory: 16 GB 2133 MHz LPDDR3, Graphics: Radeon Pro 450 2048 MB Intel HD Graphics 530 1536 MB
```
And comparing this two revisions.
https://github.com/ReactiveX/RxSwift/commit/214c9c434075e037d5c0c49802c4905aa580390d
https://github.com/ReactiveX/RxSwift/pull/1157/commits/93cdc425a45609e4291de37a731058cdcf565d1f

## Compile time
|function name|NestedSink(ms)|MapSink(ms)|
|---|---|---|
|Map.init|0.46|0.59|
|MapSink.on|0.49|0.6|
|Map.run|0.43|0.34|

Maybe there is no significant difference.

## Run time (Map.init)
I use this test case.
```swift
func testMapInitMeasure() {
    
    let xs1 = scheduler.createColdObservable([
        next(10, 1),
        next(20, 2),
        next(30, 3),
        completed(40),
        ])
    measure {
        (0..<1_000_000).forEach { _ in
            let map = xs1.map { $0 }
        }
    }
}
```
|NestedSink(sec)|MapSink(sec)|
|---|---|
|1.464|1.379|

I think there are sufficiently significant differences.

## Run time (Subscribe)
I use this test case.
```swift
func testMapSubscribeMeasure() {
    
    let xs1 = scheduler.createColdObservable([
        next(10, 1),
        next(20, 2),
        next(30, 3),
        completed(40),
        ])
    let map = xs1.map { $0 }
    measure {
        (0..<10_000).forEach { _ in
            let res = scheduler.start { map }
        }
    }
}
```
|NestedSink(sec)|MapSink(sec)|
|---|---|
|0.628|0.622|

Maybe there is no significant difference.

kzaher:
Hi @tarunon ,

yeah, I would be surprised if there wasn't some kind of overhead :)

tarunon:
Hi @kzaher 

I'm wondering why `Map.init` has a overhead first time, but now I think it's caused Sink has generics type.
I'm not sure it is really reason, but if it is collect, we can exclude this overhead.

My nested sink has a generics type, 
```swift
private final class Sink<O : ObserverType> : RxSwift.Sink<O>, ObserverType
        where O.E == ResultType {
```
so I try to avoid using generics, I use this type instead,
```swift
private final class Sink: RxSwift.Sink<AnyObserver<ResultType>>, ObserverType  {
```
then, the overhead of `Map.init` has been reduced.

PR #1159, opened 2017-03-29
Rename data structures 

Interfere:
Currently a handful of primitives and data structures used by framework contains prefix Rx. Example: RxMutableBox.

Rename the rest to minimize a chance of name collisions.

### DIFF
--- a/Platform/DataStructures/PriorityQueue.swift
+++ b/Platform/DataStructures/PriorityQueue.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2015 Krunoslav Zaher. All rights reserved.
 //
 
-struct PriorityQueue<Element> {
+struct RxPriorityQueue<Element> {
     private let _hasHigherPriority: (Element, Element) -> Bool
     private let _isEqual: (Element, Element) -> Bool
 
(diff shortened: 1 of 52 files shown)


sergdort:
Hi, @Interfere 

To be honest I don't think that in Swift you can have naming collisions.
And syntax reminds me `Objc` where we can have naming collisions. But in swift you can always put modulate name in case you have it e.g `RxSwift.Bag`
And in fact almost all of the types are `internal` which means that this should not hurt client code

Interfere:
Hi, @sergdort 

You are right! The data structures are used internally. There is no trouble when you use it as a module. But! Until swift has no stable ABI, and swift modules are dynamic libraries which are to be loaded and linked at runtime, the easiest way to integrate any framework - or its parts - is to copy source files into your own app.

Sad but true, many developers don't use swift modules. So, to ease their life these data structures are to be renamed.

sergdort:
To be honest I don't think this a good practice to include any dependencies manually. There are Cocoapods, Carthage and SPM for this

kzaher:
Hi @Interfere ,

I would agree with @sergdort on this. I'm not aware of any library that is doing this.

The only reason why someone wouldn't use some supported method is if they need to support iOS 7, but Swift compiler doesn't work there anyway.

If somebody wants to just use it as git submodule inside the same app, he can rebase and cherry pick your renaming changes, I have nothing against that.

Sorry

PR #1165, opened 2017-03-30
Bindable sink for isEditable on UITextView

switchawk:
Added this sink. Happy to hear feedback =)


switchawk:
Noob question: where does it say what target fails to compile?

switchawk:
Doesn't compile on RxCococa-tvOS :-/  
What's the right thing to do here?

kzaher:
Hi @switchawk ,

I can understand that you need this, but I'm not sure this would be useful to a lot of people.
Maybe let's skip this for now.

¯\\\_(ツ)\_/¯

PR #1167, opened 2017-03-31
Add code to reproduce output shown in GettingStarted documentation

d108:
I added a scheduler that will actually produce the output shown in the documentation when the code sample is run on the main thread. The interval observable needs to run on a separate thread from the code that is being made to sleep so that the events can get sent. This may be helpful to those trying out the code samples.


kzaher:
Hi @d108 ,

oh, that example was contrived a bit, but I get your point.

PR #1168, opened 2017-04-02
Added bindable sink for isSecureTextEntry on UITextField

paulofierro:
Makes binding to this property a little easier to use.

### DIFF
--- a/RxCocoa/iOS/UITextField+Rx.swift
+++ b/RxCocoa/iOS/UITextField+Rx.swift
@@ -36,6 +36,13 @@ extension Reactive where Base: UITextField {
         )
     }
     
+    /// Bindable sink for `isSecureTextEntry` property.
+    public var isSecureTextEntry: UIBindingObserver<Base, Bool> {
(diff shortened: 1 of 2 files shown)


freak4pc:
Hey @paulofierro 
I actually think this isn't really needed as part of UITextField ... as you can see, around this project, the bindable sinks implementations on UIKit components are mainly around the "top used" things (value/text for example here, or isEnabled on a button) and allow people to implement whichever reactive extensions they need while working on their project.

Implementing a bindable sink for **every** UIKit property as the base that comes bundled with RxCocoa might be a bit of an overkill.. That's just my opinion though :)

paulofierro:
@freak4pc oh i see. As a newcomer to RxSwift I found it curious that I could easily add bindings to `isEnabled` but didn't understand why `isSecureTextEntry` was missing from `UITextField`. 

Adding a sink for every UIKit property does indeed seem to be overkill 🙂 Happy to add it as an extension in my project instead.

PR #1175, opened 2017-04-04
Add subscription closures for Single, Maybe and Completable (`onSuccess`, `onError`, `onCompleted`)

freak4pc:
Single, Completable and Maybe only have an event `subscribe` block but doesn’t have specific blocks for the different cases (such as onNext for a regular Observable). 

This PR tries to achieve parity with the [official ReactiveX docs](http://reactivex.io/documentation/single.html) by adding:
* `subscribe(onSuccess:onError:)` for Single
* `subscribe(onCompleted:onError:)` for Completable
* `subscribe(onSuccess:onError:onCompleted:)` for Maybe


RxSwift/Traits/PrimitiveSequence.swift
```diff
@@ -124,7 +124,31 @@ extension PrimitiveSequenceType where TraitType == SingleTrait {
...
                 observer(.error(error))
             case .completed:
-                rxFatalError("SingleProtocol")
+                rxFatalError("Singles can't emit a completion event")
+            }
+        }
+    }
+
+    /**
+     Subscribes a success handler, and an error handler this Single.
```
kzaher: Maybe it would be easier and consistent to just refer Single, Maybe and Completable as observable sequence or sequences.

They are technically just special cases of observable sequences.

It sounds little weird to me when one says: `Subscribes a success handler, and an error handler this Single.`

`Subscription object used to unsubscribe from the observable sequence.`

RxSwift/Traits/PrimitiveSequence.swift
```diff
@@ -124,7 +124,31 @@ extension PrimitiveSequenceType where TraitType == SingleTrait {
...
+
+     - parameter onSuccess: Action to invoke for each element in the observable sequence.
+     - parameter onError: Action to invoke upon errored termination of the observable sequence.
+     - returns: Subscription object used to unsubscribe from the observable sequence.
+     */
+    public func subscribe(onSuccess: ((ElementType) -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil) -> Disposable {
+        var stopped = false
+        return self.primitiveSequence.asObservable().subscribe { event in
```
kzaher: I think we should use single subscription method and not observable sequence subscription method.

RxSwift/Traits/PrimitiveSequence.swift
```diff
@@ -175,7 +199,7 @@ public extension PrimitiveSequenceType where TraitType == MaybeTrait {
     }
 
     /**
-     Subscribes `observer` to receive events for this sequence.
+     Subscribes `observer` to receive events for this Maybe.
```
kzaher: Same as first comment.

RxSwift/Traits/PrimitiveSequence.swift
```diff
@@ -195,6 +219,31 @@ public extension PrimitiveSequenceType where TraitType == MaybeTrait {
...
+     Subscribes a success handler, an error handler, and a completion handler for this Maybe.
+
+     - parameter onSuccess: Action to invoke for each element in the observable sequence.
+     - parameter onError: Action to invoke upon errored termination of the observable sequence.
+     - parameter onCompleted: Action to invoke upon graceful termination of the observable sequence.
+     - returns: Subscription object used to unsubscribe from the observable sequence.
+     */
+    public func subscribe(onSuccess: ((ElementType) -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil, onCompleted: (() -> Void)? = nil) -> Disposable {
+        var stopped = false
+        return self.primitiveSequence.asObservable().subscribe { event in
```
kzaher: Also here.

RxSwift/Traits/PrimitiveSequence.swift
```diff
@@ -248,14 +297,38 @@ public extension PrimitiveSequenceType where TraitType == CompletableTrait, Elem
...
             case .error(let error):
                 observer(.error(error))
             case .completed:
                 observer(.completed)
             }
         }
     }
+
+    /**
+     Subscribes a completion handler and an error handler for this Completable.
```
kzaher: and here.

RxSwift/Traits/PrimitiveSequence.swift
```diff
@@ -248,14 +297,38 @@ public extension PrimitiveSequenceType where TraitType == CompletableTrait, Elem
...
+
+     - parameter onCompleted: Action to invoke upon graceful termination of the observable sequence.
+     - parameter onError: Action to invoke upon errored termination of the observable sequence.
+     - returns: Subscription object used to unsubscribe from the observable sequence.
+     */
+    public func subscribe(onCompleted: (() -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil) -> Disposable {
+        var stopped = false
+        return self.primitiveSequence.asObservable().subscribe { event in
```
kzaher: and here

Tests/RxSwiftTests/PrimitiveSequenceTest.swift
```diff
@@ -793,16 +793,28 @@ extension PrimitiveSequenceTest {
     }
 
     #if TRACE_RESOURCES
-        func testAsSingleReleasesResourcesOnComplete() {
+        func testAsSingleReleasesResourcesOnComplete1() {
             _ = Observable<Int>.just(1).asSingle().subscribe({ _ in })
         }
 
+        func testAsSingleReleasesResourcesOnComplete2() {
```
kzaher: `testAsSingleReleasesResourcesOnComplete1` is a test does observable sequence ( some particular operator) releases all of it's resources without `dispose` being called explicitly.

In this case `asSingle` operator is being tested.

I think we need to create unit tests that test is block called for a particular sequence.

```
func testSingle_subscribeOnSuccess() {
var events: [SingleEvent<Int>] = []

_ = Single.just(1).subscribe(onSuccess: { element in
     events.add(.success(element))
}, onError: { error in
     events.add(.error(error))
})

XCTAssertEqual(events, [.success(1)])
}
```
...

I think we are actually missing those tests for observable sequence subscribe extensions, so feel free to add them also.

RxSwift/Traits/PrimitiveSequence.swift
freak4pc: So you mean you would want to change this to something like this?

```swift
/**
 Subscribes a success handler, and an error handler this Single.

 - parameter onSuccess: Action to invoke for each element in the observable sequence.
 - parameter onError: Action to invoke upon errored termination of the observable sequence.
 - returns: Subscription object used to unsubscribe from the observable sequence.
 */
public func subscribe(onSuccess: ((ElementType) -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil) -> Disposable {
    return self.primitiveSequence.subscribe { event in
        switch event {
        case .success(let element):
            onSuccess?(element)
        case .error(let error):
            onError?(error)
        }
    }
}
```

RxSwift/Traits/PrimitiveSequence.swift
freak4pc: I also dabbled with that thought. I'm not 100% sure. 
Single/Completable/Maybe won't actually ever be a sequence, they only emit a single event and terminate, right? They are quite unite "beasts" of themselves providing a sub-section of what a full Raw Observable is. 

That's why I thought it would be nice to refer to it as "Subscribes a success handler for this Single (which is actually a subset of a Sequence)" , so it's more clear it's referring to a Single. But I really don't have any specific preference. 

Do you prefer changing all of them to "sequence" like you suggested to keep a uniform language?

freak4pc:
Hey @kzaher , I believe I've addressed all of the comments you added here. Let me know if anything else needs attention :)  

About this note specifically 

>I think we are actually missing those tests for observable sequence subscribe extensions, so feel free to add them also.

I'm not sure where exactly do you mean? The actually `subscribe(onNext` methods ? If so, perhaps it should be part of a separate PR ?

Tests/RxSwiftTests/PrimitiveSequenceTest.swift
```diff
@@ -792,17 +792,53 @@ extension PrimitiveSequenceTest {
...
-        _ = Observable<Int>.error(testError).asSingle().subscribe({ _ in })
+            _ = Observable<Int>.error(testError).asSingle().subscribe({ _ in })
         }
 
         func testAsSingleReleasesResourcesOnError2() {
-        _ = Observable<Int>.of(1, 2).asSingle().subscribe({ _ in })
+            _ = Observable<Int>.of(1, 2).asSingle().subscribe({ _ in })
+        }
+
+        func testAsSingleReleasesResourcesOnError3() {
```
kzaher: Hi @freak4pc ,

Could you please remove modifications on `*ReleasesResources*` in all unit tests and just leave `*subscribeOn*` tests.

I don't think we have much use for `*ReleasesResources*` modifications.

kzaher:
> I'm not sure where exactly do you mean? The actual subscribe(onNext methods ? If so, perhaps it should be part of a separate PR ?

Yeah, sure.

kzaher:
I think that in the future it would also be great to add debug overloads that print line and file in case of error.

```print("Received unhandled error: \(file):\(line):\(function) -> \(e)")```

We could extract this into some internal shared method for reporting unhandled errors in debug mode.

File: Tests/RxSwiftTests/PrimitiveSequenceTest.swift (hunk 8 above)
Comment (freak4pc) [inline Tests/RxSwiftTests/PrimitiveSequenceTest.swift]: Yeah @kzaher , sure thing. 
I thought we would just want to test resources get released even when using the full subscription block at that case. No problem to get rid of those.

**Comment by freak4pc** [comment]
@kzaher That sounds good, I have a side to-do list of things I might want to try do PRs of in the future so I'll write that down :) 

Do you mean like adding a `.debug()` operator specifically or in what case really? Where would those things be printed? 

For now I'll just get rid of the ReleaseResources one and we should be good to go for this PR :)

**Comment by kzaher** [comment]
Hi @freak4pc ,

yeah, let's wrap up this PR. Let's just remove those redundant unit tests for now. I was speaking of 

```
#if DEBUG
    /**
    Subscribes an element handler, an error handler, a completion handler and disposed handler to an observable sequence.

    - parameter onNext: Action to invoke for each element in the observable sequence.
    - parameter onError: Action to invoke upon errored termination of the observable sequence.
    - parameter onCompleted: Action to invoke upon graceful termination of the observable sequence.
    - parameter onDisposed: Action to invoke upon any type of termination of sequence (if the sequence has
        gracefully completed, errored, or if the generation is cancelled by disposing subscription).
    - returns: Subscription object used to unsubscribe from the observable sequence.
    */
    public func subscribe(file: String = #file, line: UInt = #line, function: String = #function, onNext: ((E) -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil, onCompleted: (() -> Void)? = nil, onDisposed: (() -> Void)? = nil)
        -> Disposable
```
ObservableType+Extensions.swift

freak4pc:
No problem @kzaher, pushing the fix up in a bit. 
Just last clarification, what do you mean by using that to report internal error? Which error would we report in that case? I'm not sure I know of the specifics :)

freak4pc:
I just did an interactive rebase and dropped the commit with those unneeded unit tests. This should clean everything up :) Waiting on the CI for now. 

Thanks @kzaher !

freak4pc:
Hey @kzaher , I guess you meant specifically this but I'm not sure what kind of change you wanted to do : https://github.com/ReactiveX/RxSwift/blob/0b66f666ba6955a51cba1ad530311b030fa4db9c/RxSwift/ObservableType%2BExtensions.swift#L56

freak4pc:
I think all good to wrap this up @kzaher ?

PR #1176, opened 2017-04-05
Some little fix on RxExamples for XC 8.3

carlosypunto:


RxExample/RxExample/Examples/APIWrappers/APIWrappersViewController.swift
```diff
@@ -139,7 +139,7 @@ class APIWrappersViewController: ViewController {
 
         // also test two way binding
         let textValue = Variable("")
-        _ = textField.rx.textInput <-> textValue
+        (textField.rx.textInput <-> textValue).disposed(by: disposeBag)
```
kzaher: Hi @carlosypunto ,

We don't need `.disposed(by: disposeBag)` in this case and on line 164. I'm using this code to test will `<->` work without dispose bags.

RxExample/RxDataSources/DataSources/Differentiator.swift
```diff
@@ -73,7 +73,7 @@ struct SectionAssociatedData {
 extension SectionAssociatedData : CustomDebugStringConvertible {
     var debugDescription: String {
         get {
-            return "\(event), \(indexAfterDelete)"
+            return "\(event), \(String(describing: indexAfterDelete))"
```
kzaher: Could you also please make these changes in RxDataSources project?

carlosypunto: ok

RxExample/RxExample/Examples/APIWrappers/APIWrappersViewController.swift
carlosypunto: I see, then I think those disposables is not being disposed when dismissed. _View controller disposed with 392 resources
assertion failed: Resources weren't cleaned properly, 98 remained, 84 expected_

RxExample/RxDataSources/DataSources/Differentiator.swift
carlosypunto: That was already there

RxExample/RxExample/Examples/APIWrappers/APIWrappersViewController.swift
kzaher: Hi @carlosypunto ,

after our talk, I think that previous version should be ok for now. Could you please revert these changes, so I can merge this?

PR #1177, opened 2017-04-05
Fix markdown url for [combinatorial explosion].

zenangst:
Remove extra space that caused a glitch when parsing markdown.


zenangst:
Pointing to `develop` now.

PR #1178, opened 2017-04-05
Fixed wrong missing CHANGELOG detection in Dangerfile

freak4pc:
Just noticed every PR I push is marked as missing a CHANGELOG entry. Seems the Dangerfile is looking for a YAML changelog instead of a Markdown one.


kzaher:
Hi @freak4pc ,

looks like there is a conflict.

freak4pc:
Wasn't there when I pushed it :) Should be good to go now. 
Thanks @kzaher

PR #1179, opened 2017-04-05
Remove unnecessary duplication for release build

onevcat:
We'd better "Do not repeat yourself" :]

### DIFF
--- a/RxSwift/ObservableType+Extensions.swift
+++ b/RxSwift/ObservableType+Extensions.swift
@@ -21,7 +21,6 @@ extension ObservableType {
         return self.subscribeSafe(observer)
     }
 
-    #if DEBUG
     /**
     Subscribes an element handler, an error handler, a completion handler and disposed handler to an observable sequence.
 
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @onevcat ,

The prototypes are different for Debug and Release configurations. Debug version also has two additional parameters `file: String = #file, line: UInt = #line`.

kzaher:
> I've noticed that the signatures are different. However, I am not sure whether it would be a good idea or not.

I do agree that if Swift compiler was more like C compiler and it allowed us to exclude/include  arbitrary parts of code with it's preprocessor, this would be no brainer.

If we can easily remove `file: String = #file, line: UInt = #line`, then we should do it like you are suggesting, if not, then let's leave it as it is.

> The method of release configuration now is a subset of the Debug one. Most of us will develop software under Debug and ship in in Release. If a user is trying to pass file name or line number as additional parameters (although it is rarely needed), their code will not compile when building with Release

Yes, but also there will be no `Resources.total` and bunch of other debug related code.

> Furthermore, maintaining two versions of similar pieces of code would be hard (and easy to be forgot by accident) for framework developers too.

Yes, but this code doesn't really have a reason to ever change.

> I also tried a simple benchmark to check the runtime performance of these two, and it seems that there is no difference between two versions.

If it's not needed for release, and there is no reason for it to be there in release configuration, than we wouldn't want to ship it in release configuration. Swift compiler changes constantly and this also  embeds potentially sensitive information about source files into binary metadata.

> May I know what is the purpose behind this if it is intended? Thanks!

Don't understand this question.

onevcat:
Great explanation on it. I understand your concerns. Thanks.

PR #1184, opened 2017-04-06
Update throttle of SharedSequence to support `latest` parameter

diogot:
The `latest` parameter is in the documentation but it's not implemented.


kzaher:
Hi @diogot ,

could you please add one more unit test analogous to `func testAsDriver_throttle()` that uses this overload.

diogot:
Sure!

freak4pc:
@diogot You need to run `./scripts/package-spm.swift` and commit the updated test files for the SPM Package for your tests to pass

diogot:
Sorry, I'll do it.

PR #1190, opened 2017-04-12
Adds `isEnabled` property for UILabel extension

brunokoga:
Despite the fact that `UILabel` is not an `UIControl`, it has an `isEnabled` property.
Just out of curiosity, according to the [docs](https://developer.apple.com/reference/uikit/uilabel/1620530-isenabled), "This property determines only how the label is drawn. Disabled text is dimmed somewhat to indicate it is not active. This property is set to true by default".

### DIFF
--- a/RxCocoa/iOS/UILabel+Rx.swift
+++ b/RxCocoa/iOS/UILabel+Rx.swift
@@ -29,6 +29,13 @@ extension Reactive where Base: UILabel {
         }
     }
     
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @brunokoga ,

I can understand that you need this, but I'm not sure this would be useful to a lot of people.

¯\\\_(ツ)\_/¯

PR #1194, opened 2017-04-13
improve Package.swift

carlosypunto:
- More concise
- More readable: `.Target` case of Target.Dependency enum is treated as TargetDependencies and  makes a clear difference with `Target` class and Package

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -7,80 +7,44 @@ let supportsTests = true
 let supportsTests = false
 #endif
 
-#if os(Linux)
-let rxCocoaDependencies: [Target.Dependency] = [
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @carlosypunto ,

> More readable: .Target case of Target.Dependency enum is treated as TargetDependencies and makes a clear difference with Target class and Package

I've checked out the proposed changes, but it doesn't seem to me that they are more readable.

> More concise

There is less lines of code, but it seems to me there is more type definition noise in those lines.
That the form we have now looks more idiomatic.

https://github.com/apple/swift-package-manager/blob/master/Package.swift

PR #1196, opened 2017-04-14
`UINavigationController` delegate proxy

diogot:
Add extensions:
- `willShow`
- `didShow`

I had to make `RxNavigationControllerDelegateProxy` open, so I could subclass in `RxImagePickerDelegateProxy` that is in `RxExample`. Let me know if this is the recommended way to do it.

### DIFF
--- a/Tests/RxCocoaTests/DelegateProxyTest+UIKit.swift
+++ b/Tests/RxCocoaTests/DelegateProxyTest+UIKit.swift
@@ -41,6 +41,14 @@ extension DelegateProxyTest {
     }
 }
 
+// MARK: UITabBarController
+
+extension DelegateProxyTest {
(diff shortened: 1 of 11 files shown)


Tests/RxCocoaTests/DelegateProxyTest+UIKit.swift
```diff
@@ -41,6 +41,14 @@ extension DelegateProxyTest {
     }
 }
 
+// MARK: UITabBarController
```
kzaher: This is not a UITabBarController ;), but I can change this after the merge. Cool.

diogot: Sorry, missed that from copy and paste, hehe

PR #1208, opened 2017-04-20
Fix CLLocationManager extension example

### LINKED ISSUE #1136: NSInternalConsistencyException thrown when using `CLLocationManager.requestLocation` in `3.3.0`

Sample project demonstrating the issue can be found [here](https://github.com/jamiepinkham/RxCoreLocation/tree/master/RxCoreLocation)
We are using a `DelegateProxy` to wrap `CLLocationManagerDelegate`, found [here](https://github.com/jamiepinkham/RxCoreLocation/blob/master/RxCoreLocation/Location/RxCLLocationManagerDelegateProxy.swift), and have added support for calling `CLLocationManager.requestLocation` to get a one time location update as you can see [here](https://github.com/jamiepinkham/RxCoreLocation/blob/master/RxCoreLocation/Location/GeolocationService.swift#L116). In previous versions of RxSwift, this worked as expected. However, upon updating to `3.3.0` this code now crashes with an `NSInternalConsistencyException` because the `DelegateProxy` does not implement `locationManager:didFailWithError:`. Despite the method being declared as optional, the documentation states:
> When using this method, the associated delegate must implement the location​Manager(_:​did​Update​Locations:​) and location​Manager(_:​did​Fail​With​Error:​) methods. Failure to do so is a programmer error.

A quick fix for us was to implement the `locationManager:didFailWithError:` on the delegate proxy, but I'm not sure how to forward those calls such that the reactive extension's `didFailWithError` observable is triggered. I suspect it's similar to how it's implemented for `UIScrollView`'s delegate proxy.

I think this is somehow related to the changes made for these issues: https://github.com/ReactiveX/RxSwift/issues/1081 https://github.com/ReactiveX/RxSwift/issues/1087

I don't think this is a _bug_ per se in `RxSwift`, but a bug in my implementation of the DelegateProxy for `CLLocationManager`, and is also latent in the implementation in the DelegateProxy found in the example code. 

**Expected outcome**:

The delegate proxy for `CLLocationManager` should be updated to handle the `locationManager:didFailWithError:` delegate method.
We should update the examples to demonstrate the correct usage of required delegate methods in delegate proxies.

**What actually happens**:

Using the delegate proxy in the sample code, and calling `requestLocation` *without first obtaining location permissions* crashes with an `NSInternalConsistencyException`

**RxSwift**

`3.3.0`

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
8.2.1
```

**Installation method**:
  - [ ] CocoaPods
  - [x] Carthage
  - [ ] Git submodules

  - [ ] yes (which ones)
  - [x] no

  - [ ] just starting

phlippieb:
Replaces PR 1027, which targeted master instead of develop.
Should fix #1136. This uses a similar method as RxScrollViewDelegateProxy.swift.

### DIFF
--- a/RxExample/Extensions/RxCLLocationManagerDelegateProxy.swift
+++ b/RxExample/Extensions/RxCLLocationManagerDelegateProxy.swift
@@ -34,10 +34,12 @@ class RxCLLocationManagerDelegateProxy : DelegateProxy
     }
 
     public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        _forwardToDelegate?.locationManager(manager, didUpdateLocations: locations)
         didUpdateLocationsSubject.onNext(locations)
     }
 
     public func locationManager(_ manager: CLLocationManager, didFailWithError error: Error) {
+        _forwardToDelegate?.locationManager(manager, didFailWithError: error)
         didFailWithErrorSubject.onNext(error)
     }
 }
--- a/RxExample/Extensions/RxCLLocationManagerDelegateProxy.swift
+++ b/RxExample/Extensions/RxCLLocationManagerDelegateProxy.swift
@@ -15,14 +15,29 @@ import CoreLocation
 class RxCLLocationManagerDelegateProxy : DelegateProxy
                                        , CLLocationManagerDelegate
                                        , DelegateProxyType {
-    
+
+    internal lazy var didUpdateLocationsSubject = PublishSubject<[CLLocation]>()
+    internal lazy var didFailWithErrorSubject = PublishSubject<Error>()
+
     class func currentDelegateFor(_ object: AnyObject) -> AnyObject? {
         let locationManager: CLLocationManager = object as! CLLocationManager
         return locationManager.delegate
     }
-    
+
     class func setCurrentDelegate(_ delegate: AnyObject?, toObject object: AnyObject) {
         let locationManager: CLLocationManager = object as! CLLocationManager
-        locationManager.delegate = delegate as? CLLocationManagerDelegate
+        if let delegate = delegate {
+            locationManager.delegate = (delegate as! CLLocationManagerDelegate)
+        } else {
+            locationManager.delegate = nil
+        }
+    }
+
+    public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        didUpdateLocationsSubject.onNext(locations)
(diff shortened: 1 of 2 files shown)


RxExample/Extensions/RxCLLocationManagerDelegateProxy.swift
```diff
@@ -15,14 +15,29 @@ import CoreLocation
...
-        locationManager.delegate = delegate as? CLLocationManagerDelegate
+        if let delegate = delegate {
+            locationManager.delegate = (delegate as! CLLocationManagerDelegate)
+        } else {
+            locationManager.delegate = nil
+        }
+    }
+
+    public func locationManager(_ manager: CLLocationManager, didUpdateLocations locations: [CLLocation]) {
+        didUpdateLocationsSubject.onNext(locations)
```
kzaher: This needs to forward call to `_forwardToDelegate?.locationManager(manager, didUpdateLocations: locations)`

RxExample/Extensions/RxCLLocationManagerDelegateProxy.swift
```diff
@@ -15,14 +15,29 @@ import CoreLocation
...
+    }
+
+    public func locationManager(_ manager: CLLocationManager, didFailWithError error: Error) {
+        didFailWithErrorSubject.onNext(error)
```
kzaher: The same for this part.

RxExample/Extensions/RxCLLocationManagerDelegateProxy.swift
phlippieb: @kzaher Should it do that in addition to publishing to the subject, or instead of, or should it only publish to the subject if `_forwardToDelegate` is nil?

kzaher: @phlippieb it should do both.
* publish to publish subject -> for observing
* forward it to forwarding delegate -> because of forwarding delegate definition

phlippieb: @kzaher Roger that. Updated.

PR #1218, opened 2017-04-25
`do` operator make memory leaking

tarunon:
I found a serious memory leaking bug when using `flatMap`, `do` and `rx.deallocated`.

Minimum code is here,
```swift
Observable.just(NSObject())
  .flatMap { $0.rx.deallocated }
  .map { _ in }
  .subscribe { print($0) } // called `next`, `completed` immediately
  .disposed(by: disposeBag)

Observable.just(NSObject())
  .flatMap { $0.rx.deallocated }
  .do()
  .subscribe { print($0) } // nothing.
  .disposed(by: disposeBag)
```

Maybe it's caused a cycling reference between `Do`, `DoSink` and `Do.source`.

I try to fix cycling reference, and seems work well now.

I'd add test case for avoid these bug.

@kzaher Is it really bug? I wish hearing your opinion.

### DIFF
--- a/RxSwift/Observables/Implementations/Do.swift
+++ b/RxSwift/Observables/Implementations/Do.swift
@@ -8,18 +8,18 @@
 
 final class DoSink<O: ObserverType> : Sink<O>, ObserverType {
     typealias Element = O.E
-    typealias Parent = Do<Element>
+    typealias EventHandler = (Event<Element>) throws -> Void
     
-    private let _parent: Parent
+    private let _eventHandler: EventHandler
     
-    init(parent: Parent, observer: O, cancel: Cancelable) {
-        _parent = parent
+    init(eventHandler: @escaping EventHandler, observer: O, cancel: Cancelable) {
+        _eventHandler = eventHandler
         super.init(observer: observer, cancel: cancel)
     }
     
     func on(_ event: Event<Element>) {
         do {
-            try _parent._eventHandler(event)
+            try _eventHandler(event)
             forwardOn(event)
             if event.isStopEvent {
                 dispose()
@@ -51,7 +51,7 @@ final class Do<Element> : Producer<Element> {
     
     override func run<O: ObserverType>(_ observer: O, cancel: Cancelable) -> (sink: Disposable, subscription: Disposable) where O.E == Element {
         _onSubscribe?()
-        let sink = DoSink(parent: self, observer: observer, cancel: cancel)
+        let sink = DoSink(eventHandler: _eventHandler, observer: observer, cancel: cancel)
         let subscription = _source.subscribe(sink)
         _onSubscribed?()
         let onDispose = _onDispose


kzaher:
Hi @tarunon ,

I understand why this could look like a leak, but as far as I can tell it's not.

This object is part of sequence declaration (do -> flatMap -> just -> NSObject). While there is any live subscription it is perfectly normal that anything that definition references is kept in memory.

It would be a leak in case there are no subscriptions active, nothing references sequence definition and there were still objects left in memory.

tarunon:
Hi @kzaher ,
The actual problem is not memory leak, because it will be resolve when `dispose()` called.
Problem is a difference of behavior between `do` and `map`.

If my understanding is correct, there are 2 pattern `run` implementation and it make this difference.

1 is a like of `Map`, it has no reference of `self` in `run` implementation.
https://github.com/ReactiveX/RxSwift/blob/efd50121157049a3a2ee5aac65f7eb2d6538694f/RxSwift/Observables/Implementations/Map.swift#L132-L136

2 is a like of `Do`, it take `self` to `DoSink` in `run` implementation.
https://github.com/ReactiveX/RxSwift/blob/efd50121157049a3a2ee5aac65f7eb2d6538694f/RxSwift/Observables/Implementations/Do.swift#L52-L63

I think subscription has a reference of `Sink`s is correct, but they have no reference of `Observable` when I use `map`. 
I know that some of `Observable` implementation needs to reference of `self` (`publish`, `share`, `retry`), but seems `do` don't need to capture own self.

kzaher:
Hi @tarunon ,

I think the main problem is that you are treating sequence elements as reference type. That is 
something that you shouldn't do.

Simple rule would be, **never rely on sequence element deallocation** for your logic.

If sequence elements are reference types, they should be **value types, or immutable or declarative (value types having observable sequences)** (in either case observing deallocation for those types doesn't produce anything meaningful), and you should never count on `flatMap`ing with dealloc or something like that because the semantics of doing that are confusing and it's extremely fragile (like you've witnessed in this case). 

Even if we would change this, your coworker could create it's own operator that breaks this, but it's a completely valid operator that doesn't leak.

tarunon:
My minimum code is too bad, I'm sorry 🤕. (but it's a minimum code)

We have a hidden `rx.deallocated` implementation in `RxCocoa` (UIControl, UIBarButtonItem, UIGestureRecognizer, and DelegateProxy).
Someone can create `Observable` using functions, chain operators, and use` button.rx.tap` there (of course maybe its a `rx.deallocated`). It can not be controlled at compile time using Type. Because everything is "Observable".

I hope to make operators safer and reduce unnecessary cycling references.
I believe it will be useful to avoid confusion and deep bugs.

kzaher:
Hi @tarunon ,

> I hope to make operators safer and reduce unnecessary cycling references.

This doesn't remove any cyclic reference.

> I believe it will be useful to avoid confusion and deep bugs.

I'm actually worried that if people see this PR, and we merge it, people will think that we guarantee this it will lead to more bugs.

**If you are sending reference types as sequence elements, you shouldn't rely on `rx.deallocated` because you are relying on internal implementation details.**

What this PR does do is that it potentially removes one retain/release pair here `try _parent._eventHandler(event)`, so I guess that's an improvement.

PR #1222, opened 2017-04-27
Add "Playground" page to Rx.playground

devxoul:
## Background

I frequently use Rx.playground. I thought that it would be great if we have an empty playground page.

## Description

Use this page

* ... to prototype your observable streams and custom operators
* ... to write an answer for Rx questions
* ... for whatever you want!

screen shot 2017-04-27 at 3 45 05 pm

### DIFF
--- a/Rx.playground/Pages/Playground.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Playground.xcplaygroundpage/Contents.swift
@@ -1,0 +1,22 @@
+/*:
+ > # IMPORTANT: To use **Rx.playground**:
+ 1. Open **Rx.xcworkspace**.
+ 1. Build the **RxSwift-macOS** scheme (**Product** → **Build**).
+ 1. Open **Rx** playground in the **Project navigator**.
+ 1. Show the Debug Area (**View** → **Debug Area** → **Show Debug Area**).
+ */
+import RxSwift
+/*:
+ # Try Yourself
+ 
+ It's time to play with Rx 🎉
+ */
+playgroundShouldContinueIndefinitely()
+
+example("Try yourself") {
+  // let disposeBag = DisposeBag()
+  _ = Observable.just("Hello, RxSwift!")
+    .debug("Observable")
+    .subscribe()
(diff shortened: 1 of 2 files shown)


Rx.playground/Pages/Playground.xcplaygroundpage/Contents.swift
```diff
@@ -0,0 +1,22 @@
...
+import RxSwift
+/*:
+ # Try Yourself
+ 
+ It's time to play with Rx 🎉
+ */
+playgroundShouldContinueIndefinitely()
+
+example("Try yourself") {
+  let disposeBag = DisposeBag()
```
kzaher: Hi @devxoul ,

I think we can just write

```
// let disposeBag = DisposeBag()
_ = Observable.just("Hello, RxSwift!")
  .debug("Observable")
  .subscribe()
  // If dispose bag is used instead, sequence will terminate on scope exit
  // .disposed(by: disposeBag)
```

... or some comment worded better.

devxoul: In this case `just` will emit `completed` so disposeBag will be never used. How about just letting the function body empty? (with a single comment)

```swift
example("Try yourself") {
  // Observable.just("Hello, RxSwift!") ...
}
```

kzaher: The reason why I'm suggesting that is because this is intended for playing around, and you've already prepared the codebase with `playgroundShouldContinueIndefinitely` for playing with time based operators.

That means that if some new user replaces your code with
```
let disposeBag = DisposeBag()
_ = Observable<Int>.interval(0.2, scheduler: MainScheduler.instance)
  .debug("Observable")
  .subscribe()
  .disposed(by: disposeBag)
```

we'll get a bug report in best case, or user will be confused and just give up figuring out what has gone wrong.

devxoul: That makes sense. I updated the commit :)

Issue #1093, opened 2017-02-08
Implement concatMap operator


brendanarnold:
**Short description of the issue**:

No concatMap operator for serializing streams of streams - see [React docs for concatMap](http://reactivex.io/documentation/operators/flatmap.html). This coupled with the fact that there is no non-element repeat (i.e. an overload for repeat that resubscribes on completion) means that looping an observable in RxSwift is tricky - see this [Stackoverflow question for details](http://stackoverflow.com/questions/42115976/how-do-i-sequentially-loop-an-observable-in-rxswift)

**Expected outcome**:

There should be a way to serialize streams of streams when each stream completes

**What actually happens**:

Have not been able to figure out a way to do the above

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

3.1.0

**Platform/Environment**

  - [ ] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [ ] easy, 100% repro
  
**Xcode version**:

```
8.1
```

kzaher:
Hi @brendanarnold ,

Its kind of silly, but Swift doesn't have infinite sequence :) So you first need to create yours first.

```
struct InfiniteSequence<E> : Sequence {
    typealias Element = E
    typealias Iterator = AnyIterator<E>
    
    private let _repeatedValue: E
    
    init(repeatedValue: E) {
        _repeatedValue = repeatedValue
    }
    
    func makeIterator() -> Iterator {
        let repeatedValue = _repeatedValue
        return AnyIterator {
            return repeatedValue
        }
    }
}
```

... and then

```
InfiniteSequence(repeatedValue: Observable.just("Repeat me until end of time")
       .map { _ in "Really?" })
      .concat()
```

kzaher:
... because RxSwift doesn't have back pressure.

brendanarnold:
So RxSwift not having back pressure is a design decision? If so is there any more info on this decision? 

I think I understand the implications in implementing `concatMap` in that you would have to cache the 'queued' observables which could be tricky but for `repeat` the implementation is relatively straightforward since the above `InfiniteSequence` snippet is already in the source code and is used for the retry operator. In fact `repeat` is almost the same as `retry` in that it can also potentially loop forever.

tomahh:
[...]

PR #1224, opened 2017-04-28
Add Concat Map operator

sergdort:
Implemented `concatMap` operator for the Observable regarding #1093 

### Description of change

* Refactor `MergeLimitedSink` to make interface similar to `MergeSink`
* Add `MapConcatSink `
* Add `ConcatMap` producer
* Add `concatMap` method to the extension to the `ObservableType`


RxSwift/Observables/Implementations/Merge.swift
```diff
@@ -56,12 +56,32 @@ fileprivate final class MergeLimitedSinkIter<S: ObservableConvertibleType, O: Ob
...
+}
+
+fileprivate final class MergeLimitedBasicSink<S: ObservableConvertibleType, O: ObserverType>: MergeLimitedSink<S, S, O> where O.E == S.E {
+    
+    override func performMap(_ element: S) throws -> S {
+        return element
+    }
+}
+
+fileprivate class MergeLimitedSink<SourceType, S: ObservableConvertibleType, O: ObserverType>
```
sergdort: Not sure if it good idea to extend `MergeLimitedSink ` to be similar to `MergeSink`. 
In theory we could either subclass from `MergeSink` or create `ConcatMapSink` without subclassing from `MergeLimitedSink`

sergdort: Then we won't need to have  `MergeLimitedBasicSink`

RxSwift/Observables/Implementations/Merge.swift
```diff
@@ -408,6 +440,24 @@ final class FlatMapFirst<SourceType, S: ObservableConvertibleType>: Producer<S.E
...
+    private let _source: Observable<SourceType>
+    private let _selector: Selector
+    
+    init(source: Observable<SourceType>, selector: @escaping Selector) {
+        _source = source
+        _selector = selector
+    }
+    
+    override func run<O: ObserverType>(_ observer: O, cancel: Cancelable) -> (sink: Disposable, subscription: Disposable) where O.E == S.E {
+        let sink = MapConcatSink<SourceType, S, O>(selector: _selector, observer: observer, cancel: cancel)
```
sergdort: This is probable should be `ConcatMapSink`?

RxSwift/Observables/Implementations/Merge.swift
kzaher: Hi @sergdort ,

yeah, I think it's ok to make `MergeLimitedSink` a bit more generic and then reuse the logic.

I've reordered project structure a bit in `develop`. Can you please add unit tests for this and target `develop` branch?

We can reuse and modify `flatMap` unit tests a bit.

RxSwift/Observables/Implementations/Merge.swift
kzaher: Yep

sergdort:
Hi, @kzaher sorry for the delay, I was on holidays =)

I like that you moved public interface into implementation although I've noticed that sinks comes before producers, is it intentional if so I would like to know why :)
Because I feel like it much more natural to have classes and method in the order  they are used, sure this is kind of opinionated but ¯\_(ツ)_/¯ (Uncle Bob will be happy) 😆

sergdort:
I've also found some tests from [RxJS](https://github.com/Reactive-Extensions/RxJS/blob/master/tests/observable/concatmap.js#L145) I think we can reuse them

kzaher:
> I like that you moved public interface into implementation although I've noticed that sinks comes before producers, is it intentional if so I would like to know why :)

I think it was because all sequence declarations were the same, so they were practically irrelevant. Maybe there was some other rationale, but can't remember now.

> Because I feel like it much more natural to have classes and method in the order they are used, sure this is kind of opinionated but ¯_(ツ)_/¯ (Uncle Bob will be happy) 😆

I think it makes sense, we could move them in front of `Sink`s.

sergdort:
Seems like CI is passing :)

kzaher:
Hi @sergdort ,

did you maybe rebase?

sergdort:
I'm little bit confused :) Why do I need to rebase?

kzaher:
Hi @sergdort,

because I can't do a CR or merge this PR because of conflicts.

screen shot 2017-05-17 at 00 34 33

sergdort:
Done

PR #1229, opened 2017-05-01
fix compilation error on linux

hariseldon78:
I was getting this error compiling on linux with swift 3.1.1:
Sources/RxCocoa/URLSession+Rx.swift:144:76: error: 'Error' is not convertible to 'NSError'; did you mean to use 'as!' to force downcast?
                    print(convertResponseToString(response, error.map { $0 as NSError }, interval))
                                                                        ~~~^~~~~~~~~~
                                                                           as!
<unknown>:0: error: build had 1 command failures


hariseldon78:
will remake on develop branch

Issue #1246, opened 2017-05-10
Request: add rx.attributedText on UITextView


Arnoymous:
add rx.attributedText on UITextView

kzaher:
Hi @Arnoymous ,

I think we can add that. Were you thinking or making a PR?

carlosypunto:
@kzaher it's done, please check it

PR #1249, opened 2017-05-12
Add rx.attributedText to UITextView and UITextField

carlosypunto:
Requested in issue #1246

### DIFF
--- a/RxCocoa/iOS/UITextField+Rx.swift
+++ b/RxCocoa/iOS/UITextField+Rx.swift
@@ -39,7 +39,9 @@ extension Reactive where Base: UITextField {
     /// Bindable sink for `attributedText` property.
     public var attributedText: UIBindingObserver<Base, NSAttributedString?> {
         return UIBindingObserver(UIElement: self.base) { textField, attributedText in
-            textField.attributedText = attributedText
+            if textField.attributedText != attributedText {
+                textField.attributedText = attributedText
+            }
         }
     }
     
--- a/RxCocoa/iOS/UITextField+Rx.swift
+++ b/RxCocoa/iOS/UITextField+Rx.swift
@@ -36,6 +36,13 @@ extension Reactive where Base: UITextField {
         )
     }
     
+    /// Bindable sink for `attributedText` property.
+    public var attributedText: UIBindingObserver<Base, NSAttributedString?> {
+        return UIBindingObserver(UIElement: self.base) { textField, attributedText in
+            textField.attributedText = attributedText
+        }
+    }
+    
 }
 
 #endif
--- a/RxCocoa/iOS/UITextField+Rx.swift
+++ b/RxCocoa/iOS/UITextField+Rx.swift
@@ -37,12 +37,20 @@ extension Reactive where Base: UITextField {
     }
     
     /// Bindable sink for `attributedText` property.
-    public var attributedText: UIBindingObserver<Base, NSAttributedString?> {
-        return UIBindingObserver(UIElement: self.base) { textField, attributedText in
-            if textField.attributedText != attributedText {
-                textField.attributedText = attributedText
+    public var attributedText: ControlProperty<NSAttributedString?> {
+        return UIControl.rx.value(
+            base,
+            getter: { textField in
+                textField.attributedText
+            }, setter: { textField, value in
+                // This check is important because setting text value always clears control state
+                // including marked text selection which is imporant for proper input
+                // when IME input method is used.
+                if textField.attributedText != value {
+                    textField.attributedText = value
+                }
             }
-        }
+        )
     }
     
 }
--- a/RxCocoa/iOS/UITextView+Rx.swift
+++ b/RxCocoa/iOS/UITextView+Rx.swift
@@ -53,6 +53,45 @@ extension Reactive where Base: UITextView {
         
         return ControlProperty(values: source, valueSink: bindingObserver)
     }
+    
+    
+    /// Reactive wrapper for `attributedText` property.
+    public var attributedText: ControlProperty<NSAttributedString?> {
+        let source: Observable<NSAttributedString?> = Observable.deferred { [weak textView = self.base] in
+            let attributedText = textView?.attributedText
+            
+            let textChanged: Observable<NSAttributedString?> = textView?.textStorage
+                // This project uses text storage notifications because
+                // that's the only way to catch autocorrect changes
+                // in all cases. Other suggestions are welcome.
+                .rx.didProcessEditingRangeChangeInLength
+                // This observe on is here because attributedText storage
+                // will emit event while process is not completely done,
+                // so rebinding a value will cause an exception to be thrown.
+                .observeOn(MainScheduler.asyncInstance)
+                .map { _ in
+                    guard let textStorage = textView?.textStorage else { return nil }
+                    return textStorage.attributedSubstring(from: NSRange(location: 0, length: textStorage.string.utf8.count))
+                }
+                ?? Observable.empty()
+            
+            return textChanged
+                .startWith(attributedText)
+        }
+        
+        let bindingObserver = UIBindingObserver(UIElement: self.base) { (textView, attributedText: NSAttributedString?) in
+            // This check is important because setting text value always clears control state
+            // including marked text selection which is imporant for proper input
+            // when IME input method is used.
+            if textView.attributedText != attributedText {
+                print(textView.attributedText)
+                print(attributedText)
+                textView.attributedText = attributedText
+            }
+        }
+        
+        return ControlProperty(values: source, valueSink: bindingObserver)
+    }
 
     /// Reactive wrapper for `delegate` message.
     public var didBeginEditing: ControlEvent<()> {
--- a/RxCocoa/iOS/UITextView+Rx.swift
+++ b/RxCocoa/iOS/UITextView+Rx.swift
@@ -70,8 +70,7 @@ extension Reactive where Base: UITextView {
                 // so rebinding a value will cause an exception to be thrown.
                 .observeOn(MainScheduler.asyncInstance)
                 .map { _ in
-                    guard let textStorage = textView?.textStorage else { return nil }
-                    return textStorage.attributedSubstring(from: NSRange(location: 0, length: textStorage.length))
+                    return textView?.attributedText
                 }
                 ?? Observable.empty()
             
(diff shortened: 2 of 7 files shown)


RxCocoa/iOS/UITextView+Rx.swift
```diff
@@ -65,6 +65,45 @@ extension Reactive where Base: UITextView {
...
+                .startWith(attributedText)
+        }
+        
+        let bindingObserver = UIBindingObserver(UIElement: self.base) { (textView, attributedText: NSAttributedString?) in
+            // This check is important because setting text value always clears control state
+            // including marked text selection which is imporant for proper input
+            // when IME input method is used.
+            if textView.attributedText != attributedText {
+                print(textView.attributedText)
+                print(attributedText)
```
kzaher: I think these are accidental `print`s.

RxCocoa/iOS/UITextView+Rx.swift
```diff
@@ -65,6 +65,45 @@ extension Reactive where Base: UITextView {
...
+                // that's the only way to catch autocorrect changes
+                // in all cases. Other suggestions are welcome.
+                .rx.didProcessEditingRangeChangeInLength
+                // This observe on is here because attributedText storage
+                // will emit event while process is not completely done,
+                // so rebinding a value will cause an exception to be thrown.
+                .observeOn(MainScheduler.asyncInstance)
+                .map { _ in
+                    guard let textStorage = textView?.textStorage else { return nil }
+                    return textStorage.attributedSubstring(from: NSRange(location: 0, length: textStorage.string.utf8.count))
```
kzaher: I'm not sure what is the length here. Why do you think it's `textStorage.string.utf8.count`?

RxCocoa/iOS/UITextField+Rx.swift
```diff
@@ -36,6 +36,13 @@ extension Reactive where Base: UITextField {
         )
     }
     
+    /// Bindable sink for `attributedText` property.
+    public var attributedText: UIBindingObserver<Base, NSAttributedString?> {
+        return UIBindingObserver(UIElement: self.base) { textField, attributedText in
+            textField.attributedText = attributedText
```
kzaher: I think we need to find way to make sure this isn't updated in case attributed text is the same and test this.
```
if textField.attributedText != attributedText {
    textField.attributedText = attributedText
}
```

We had issues with IME.

```
// This check is important because setting text value always clears control state
// including marked text selection which is imporant for proper input 
// when IME input method is used.
```

Tests/RxCocoaTests/UITextField+RxTests.swift
```diff
@@ -13,35 +13,46 @@ import XCTest
...
+        textField.settedText = false
         textField.rx.text.on(.next("Text1"))
-        XCTAssertTrue(!textField.set)
+        XCTAssertTrue(!textField.settedText)
         textField.rx.text.on(.next("Text2"))
-        XCTAssertTrue(textField.set)
+        XCTAssertTrue(textField.settedText)
+    }
+    
+    func testLabel_attributedTextObserver() {
```
kzaher: I think this is copied code from `UILabel`, but nothing is updated, it doesn't test `UITextField`.

Tests/RxCocoaTests/UITextField+RxTests.swift
```diff
@@ -13,35 +13,46 @@ import XCTest
...
+        _ = Observable.just(text).bind(to: label.rx.attributedText)
+        
+        XCTAssertEqual(label.attributedText, text)
     }
 }
 
 final class UITextFieldSubclass : UITextField {
-    var set: Bool = false
-
+    var settedText = false
```
kzaher: I think maybe `didSetText` would be more descriptive.

Tests/RxCocoaTests/UITextView+RxTests.swift
```diff
@@ -116,15 +132,26 @@ final class UITextViewTests : RxTest {
 }
 
 final class UITextViewSubclass2 : UITextView {
-    var set: Bool = false
-
+    var settedText = false
+    var settedAttributedText = false
```
kzaher: The same like for `UITextField`. This should probably be `didSetText` and `didSetAttributedText`.

carlosypunto:
I think now is correct.
Maybe you want APIWrapper RxExample will log attributedText changes, I can add to PR changing [ this](https://github.com/ReactiveX/RxSwift/blob/master/RxExample/RxExample/Examples/APIWrappers/APIWrappersViewController.swift#L166-L170) with:

```
        let attributedTextValue = textView.rx.attributedText.asObservable()
        Observable.zip(textViewValue.asObservable(), attributedTextValue, resultSelector: { text, attributedString -> String in
            return "UITextView text: \(text)\nattributedText: \(attributedString?.description ?? "nil")"
        })
            .subscribe(onNext: { [weak self] x in
                self?.debug("UITextView text \(x)")
            })
            .disposed(by: disposeBag)
```

kzaher:
Yeah, we can do this in a separate PR.

Tests/RxCocoaTests/UITextField+RxTests.swift
```diff
@@ -13,35 +13,52 @@ import XCTest
...
+        textField.isSettedText = false
         textField.rx.text.on(.next("Text1"))
-        XCTAssertTrue(!textField.set)
+        XCTAssertTrue(!textField.isSettedText)
         textField.rx.text.on(.next("Text2"))
-        XCTAssertTrue(textField.set)
+        XCTAssertTrue(textField.isSettedText)
+    }
+    
+    func testLabel_attributedTextObserver() {
```
kzaher: Test title says `testLabel`. It should probably be just `test` because there is no label that we are testing inside this test.

Tests/RxCocoaTests/UITextField+RxTests.swift
```diff
@@ -13,35 +13,52 @@ import XCTest
...
         ensurePropertyDeallocated({ UITextField() }, "a", comparer: { $0 == $1 }) { (view: UITextField) in view.rx.value }
     }
-
+    
     func testSettingTextDoesntClearMarkedText() {
         let textField = UITextFieldSubclass(frame: CGRect.zero)
-
         textField.text = "Text1"
-        textField.set = false
+        textField.isSettedText = false
```
kzaher: I don't think `isSettedText` is formulated correctly. Maybe use `didSetText`.

Tests/RxCocoaTests/UITextField+RxTests.swift
```diff
@@ -13,35 +13,52 @@ import XCTest
...
     }
 }
 
-final class UITextFieldSubclass : UITextField {
-    var set: Bool = false
+private extension String {
+    var textFieldAttributedString: NSAttributedString {
+        let tf = UITextField()
+        tf.attributedText = NSAttributedString(string: self)
+        return tf.attributedText!
```
kzaher: I don't understand this code. `UITextField` is being created and then discarded immediately.

Is there something wrong with just using `NSAttributedText(string: "Hello!")`?

Tests/RxCocoaTests/UITextField+RxTests.swift
```diff
@@ -13,35 +13,52 @@ import XCTest
...
+final class UITextFieldSubclass : UITextField {
+    var isSettedText = false
+    
     override var text: String? {
         get {
             return super.text
         }
         set {
-            set = true
+            isSettedText = true
```
kzaher: This has the same grammar issue.

Tests/RxCocoaTests/UITextView+RxTests.swift
```diff
@@ -115,16 +131,35 @@ final class UITextViewTests : RxTest {
     }
 }
 
-final class UITextViewSubclass2 : UITextView {
-    var set: Bool = false
+private extension String {
+    var textViewAttributedString: NSAttributedString {
+        let tf = UITextView()
+        tf.attributedText = NSAttributedString(string: self)
+        return tf.attributedText!
```
kzaher: The same problem with redundant extension.

Tests/RxCocoaTests/UITextView+RxTests.swift
```diff
@@ -115,16 +131,35 @@ final class UITextViewTests : RxTest {
...
+    }
+}
 
+final class UITextViewSubclass2 : UITextView {
+    var isSettedText = false
+    var isSettedAttributedText = false
```
kzaher: This has the same grammar issue.

Tests/RxCocoaTests/UITextField+RxTests.swift
carlosypunto: The `NSAttributedText(string: "Hello!")` style can't have style info. When we set this to a Textfield or Textview,  they overwrite its internal NSAttributedString properties in the case  only text, then, if we get the control's `attributedStyle` is not equal to the setted. So this function returns and `NSAttributedText` with all properties with effect in these controls being it comparable with getted style from control.

Tests/RxCocoaTests/UITextView+RxTests.swift
carlosypunto: Same explanation than TextField case

RxCocoa/iOS/UITextView+Rx.swift
carlosypunto: In last commit is changed to `NSRange(location: 0, length: textStorage.length)` the full range

RxCocoa/iOS/UITextView+Rx.swift
carlosypunto: removed in las commit

Tests/RxCocoaTests/UITextField+RxTests.swift
carlosypunto: fixed in last commit

RxCocoa/iOS/UITextField+Rx.swift
```diff
@@ -36,6 +36,15 @@ extension Reactive where Base: UITextField {
         )
     }
     
+    /// Bindable sink for `attributedText` property.
+    public var attributedText: UIBindingObserver<Base, NSAttributedString?> {
+        return UIBindingObserver(UIElement: self.base) { textField, attributedText in
+            if textField.attributedText != attributedText {
```
sergdort: Hm. Why do we need this check?

PR #1253, opened 2017-05-16
Fix/1252 driver scheduling

### LINKED ISSUE #1252: SharedSequence should have an ObserveOn Scheduler and a SubscribeOn Scheduler

`Driver` (because of `SharedSequence`) uses the same scheduler for subscription and for observing, when it should use different ones for each.

Take the following statement:

```swift
Driver.just("View Controller Title").drive(navigationItem.rx.title).addDisposable(to: disposeBag)
```

**Expected outcome**:

The `navigationItem.title` should be set immediately, regardless if there are events queued in `MainScheduler.instance` (i.e. it should use `ConcurrentMainScheduler.instance` for subscription).

**What actually happens**:

If there are events queued in `MainScheduler.instance`, then the above won't set the `navigationItem.title` during the same run loop. This can cause issues with display, since in this example it will mean when the view controller is pushed on a navigation controller the title won't appear until the animation is complete.

RxSwift 3.4.1

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

  - [ ] just starting

TheAdamBorek:
This is the PR which fixes #1252 

As we discussed the issue on RxSwift's slack, the reason of the issue is all the Drivers share the same instance of `MainScheduler`.

The solution is to use different scheduler instances for Drivers.

`MainScheduler.instance` still returns a shared instance of the scheduler, so it is possible to have similar issue if people will use `observeOn(MainScheduler.instance)` on Observables. @kzaher I'm thinking if this is ok 🤔

### DIFF
--- a/RxCocoa/Traits/Driver/Driver.swift
+++ b/RxCocoa/Traits/Driver/Driver.swift
@@ -37,7 +37,7 @@
 public typealias Driver<E> = SharedSequence<DriverSharingStrategy, E>
 
 public struct DriverSharingStrategy: SharingStrategyProtocol {
-    public static var scheduler: SchedulerType { return driverObserveOnScheduler }
+    public static var scheduler: SchedulerType { return driverObserveOnScheduler() }
     public static func share<E>(_ source: Observable<E>) -> Observable<E> {
         return source.shareReplayLatestWhileConnected()
     }
@@ -58,12 +58,12 @@ extension SharedSequenceConvertibleType where SharingStrategy == DriverSharingSt
 */
 public func driveOnScheduler(_ scheduler: SchedulerType, action: () -> ()) {
     let originalObserveOnScheduler = driverObserveOnScheduler
-    driverObserveOnScheduler = scheduler
+    driverObserveOnScheduler = { return scheduler }
 
     action()
 
     // If you remove this line , compiler buggy optimizations will change behavior of this code
-    _forceCompilerToStopDoingInsaneOptimizationsThatBreakCode(driverObserveOnScheduler)
+    _forceCompilerToStopDoingInsaneOptimizationsThatBreakCode(scheduler)
     // Scary, I know
 
     driverObserveOnScheduler = originalObserveOnScheduler
@@ -87,4 +87,4 @@ func _forceCompilerToStopDoingInsaneOptimizationsThatBreakCode(_ scheduler: Sche
     }
 }
 
-fileprivate var driverObserveOnScheduler: SchedulerType = MainScheduler.instance
+fileprivate var driverObserveOnScheduler: () -> SchedulerType = { MainScheduler() }
--- a/Sources/AllTestz/main.swift
+++ b/Sources/AllTestz/main.swift
@@ -573,6 +573,7 @@ final class DriverTest_ : DriverTest, RxTestCase {
     ("testDriverFromOptionalWhenNil", DriverTest.testDriverFromOptionalWhenNil),
     ("testDriverFromSequence", DriverTest.testDriverFromSequence),
     ("testDriverFromArray", DriverTest.testDriverFromArray),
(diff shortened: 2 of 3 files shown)


brentleyjones:
Why can't we instead use `ConcurrentMainScheduler.instance` for subscribing but `MainScheduler.instance` for observing? The way this works it seems that each time it asks for a scheduler it makes a new one, so it's effectively `ConcurrentMainScheduler.instance` everywhere, even when calling `asDriver()`, which can run into issues since you want observing to queue up like `MainScheduler.instance`.

Here is my proposed solution, which I'll make into a PR if needed: https://github.com/ReactiveX/RxSwift/compare/master...brentleyjones:sharedSequence-subscriptionScheduler

TheAdamBorek:
@brentleyjones 🤔 actually Driver used to have `driverSubscribeOnScheduler` at least in Swift 2.3
It seems it was removed. @kzaher can you give as a small reason why? ;)

zrzut ekranu 2017-05-16 16 59 42

However, even in Swift 2.3 I used to have issues with `Driver` that it sent it's events on async queue :/
I would have to debug more to find out what was the root of the problem to create sample & short test case...

brentleyjones:
`Driver` will still send events on async queue, if they go through `asDriver()`. My fix will only fix "pure" `Driver`s, where you start with `.just()` (or another constructor) and never change away from `Driver`. Once you call `asDriver()` it uses `observeOn`, and you are back to possibly being async.

TheAdamBorek:
Yep, @brentleyjones I've just figured that out also :D

I changed my test to use Driver which is transformed from `Observable.create` to not have `just` dependency which has `subscribeOn` applied.

So... basically your changes won't fix my issue 😉

kzaher:
Hi @TheAdamBorek @brentleyjones ,

I think there could be some issues if you are using `ConcurrentMainScheduler.instance` and have a lot of operator combinations (because call stack could be deep and potentially cause stack overflow). For example, I think that you could have a stack overflow in case you are sending a large array on `ConcurrentMainScheduler.instance`. 

Having said that, I think we might investigate adding another customization point in RxSwift 4.0 like we had in previous version `subscribeOnScheduler`.

I don't think this would cause any performance concerns since `MainScheduler` only has an integer and a reference to main dispatch queue. Creating new instance once per `Driver` entry point should be cheap.

@brentleyjones the reason why we can't add this now is because adding this to `SharingStrategyProtocol` isn't a backwards compatible change. But yes, we can maybe optimize this a bit for 4.0 version.

kzaher:
@brentleyjones could you for now try this PR and tell us if this solves your issue and maybe make a PR for 4.0 version?

brentleyjones:
@kzaher I'll check it out after lunch :)

brentleyjones:
@kzaher It works. I do wonder if using `MainScheduler()` for `onObserve` will introduce any subtle bugs though?

kzaher:
I've been thinking about what are the other consequences of changing this.

The nice thing about the previous solution with shared instance was that there wasn't any reentrancy possible in `Driver` (that "Warning: Recursive call ....") but now that will not be the case.

On the other hand it did cause these async dispatches unpredictably.

I'm still for merging this thing. I'm just trying to consider all of the effects of this change.

One will probably be us needing to add another `observeOn()` operator to `SharedSequence` because now a use case for that has appeared.

We will also try to improve that reentrancy code. I'll try to think about how to best adapt it.

TheAdamBorek:
So @kzaher do you want to me to fix anything in the PR? Are code style, tests written ok?

kzaher:
Hi @TheAdamBorek ,

I was just thinking a bit. Merged this. Made some changes to unit tests 8e6d467d1052aa21ffa31de36c095cc7f7433b4c.

PR #1254, opened 2017-05-16
Add concat operator to Completable

hagmas:
Hello, it is a PR for adding concat on Completable.

### DIFF
--- a/RxSwift/Traits/PrimitiveSequence.swift
+++ b/RxSwift/Traits/PrimitiveSequence.swift
@@ -627,20 +627,6 @@ extension PrimitiveSequenceType where TraitType == MaybeTrait {
     }
 }
 
-extension PrimitiveSequence where Trait == CompletableTrait, Element == Never {
-    /**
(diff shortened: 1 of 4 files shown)


kzaher:
Hi @hagmas ,

will check it out soon.

RxSwift/Traits/PrimitiveSequence.swift
```diff
@@ -627,6 +627,20 @@ extension PrimitiveSequenceType where TraitType == MaybeTrait {
     }
 }
 
+extension PrimitiveSequence where Trait == CompletableTrait, Element == Never {
```
kzaher: Hi @hagmas ,

I think you need to make this into `PrimitiveSequence`. Unit tests are failing because of this IMHO.

hagmas:
Hi @kzaher 
Thank you for reviewing. The error was due to concrete same-type requirements which is not supported with Swift3.0. Please take a look again.

Thank you.

PR #1256, opened 2017-05-17
Add ViewController LifeCycle

muukii:
I've added methods that UIViewController's lifecycle.

### DIFF
--- a/RxCocoa/iOS/UIViewController+Rx.swift
+++ b/RxCocoa/iOS/UIViewController+Rx.swift
@@ -22,32 +22,32 @@
             }
         }
         
-        public var viewDidLoad: ControlEvent<Void> {
-            return ControlEvent(events: sentMessage(#selector(UIViewController.viewDidLoad)).map { _ in })
+        public var viewDidLoad: ControlEvent<[Any]> {
+            return ControlEvent(events: sentMessage(#selector(UIViewController.viewDidLoad)))
         }
         
-        public var viewWillAppear: ControlEvent<Void> {
-            return ControlEvent(events: sentMessage(#selector(UIViewController.viewWillAppear(_:))).map { _ in })
+        public var viewWillAppear: ControlEvent<[Any]> {
+            return ControlEvent(events: sentMessage(#selector(UIViewController.viewWillAppear(_:))))
         }
         
-        public var viewDidAppear: ControlEvent<Void> {
-            return ControlEvent(events: sentMessage(#selector(UIViewController.viewDidAppear(_:))).map { _ in })
+        public var viewDidAppear: ControlEvent<[Any]> {
+            return ControlEvent(events: sentMessage(#selector(UIViewController.viewDidAppear(_:))))
         }
         
-        public var viewWillDisappear: ControlEvent<Void> {
-            return ControlEvent(events: sentMessage(#selector(UIViewController.viewWillDisappear(_:))).map { _ in })
+        public var viewWillDisappear: ControlEvent<[Any]> {
+            return ControlEvent(events: sentMessage(#selector(UIViewController.viewWillDisappear(_:))))
         }
         
-        public var viewDidDisapppear: ControlEvent<Void> {
-            return ControlEvent(events: sentMessage(#selector(UIViewController.viewDidDisappear(_:))).map { _ in })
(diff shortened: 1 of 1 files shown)


muukii:
@jamiepinkham 

I think ControlEvent ensures `no-failure` and `called from main-thread`.
Therefore, we can use it with fewer worries than Observable.
Do you think there are any problems by using it?

jamiepinkham:
Using `ControlEvent` is fine. The problem I see is you're throwing away information (the arguments passed to `sentMessage`), which could be useful to consumers of the observable.

muukii:
@jamiepinkham 
Oh, I see. sorry.
Indeed, it may be better not to discard the information.
Handling of information should be left to consumers.
Don't you think so?

jamiepinkham:
@muukii Precisely.

muukii:
@jamiepinkham 
Thanks, I'll fix it.

RxCocoa/iOS/UIViewController+Rx.swift
```diff
@@ -14,13 +14,40 @@
...
+        
+        public var viewDidAppear: ControlEvent<[Any]> {
+            return ControlEvent(events: sentMessage(#selector(UIViewController.viewDidAppear(_:))))
+        }
+        
+        public var viewWillDisappear: ControlEvent<[Any]> {
+            return ControlEvent(events: sentMessage(#selector(UIViewController.viewWillDisappear(_:))))
+        }
+        
+        public var viewDidDisapppear: ControlEvent<[Any]> {
```
brentleyjones: We should return the proper `Element` types here. For example, this one is `Bool`. And `viewDidLoad` is `Void`.

muukii: @brentleyjones 
Thanks, I understand.

jamiepinkham: something like this: 

```    
public var viewDidAppear: ControlEvent<Bool> {
        return ControlEvent(events: self.sentMessage(#selector(UIViewController.viewDidAppear(_:))).map { args in
            assert(args.count == 1)
            assert(args.first as? Bool != nil)
            return args.first as! Bool
        })
    }
```

muukii:
@jamiepinkham @brentleyjones 
Thanks for the advice.
I've fixed it.

kzaher:
Hey guys,

I also have code that looks something like this in my codebase.

I'm not sure we it would be a good idea to include all of these extensions in the repo and ship them by default because:

* We would really want to ship as few "magical" methods as possible (methods that rely on swizzling)
* We would need to find some criteria that would include these extensions, and not a lot of other ones that are probably also useful. Otherwise we don't have a simple transparent way of deciding what should we include. So far we have included "main" data binding properties of common UIKit controls, delegate wrappers and a handful of really commonly used extensions that were so common it would be silly to not include them.
* Usually these methods are used for different purposes. Without `boolean` argument if one wants to construct some kind of `appearing: ControlProperty<Bool>`, and with if one would potentially want to perform some kind of animations. There were even comments in this PR about boolean argument :)
* these methods (`view{Will,Did}{Appear,Dissapear}`) are already fragile as it is, so I usually have to combine them with `viewDidLoad` and `deallocated` to get some kind of reasonable behavior. (sometimes they will fire ok, or fire twice, or won't fire at all)
* since they aren't really useful by themselves, and they have to be probably combined with `viewDidLoad` and `deallocated` to get some more robust behavior, there aren't really that useful standalone as it might originally seem IMHO.
* I can see why they could be used a lot, but I'm assuming if one is working on a product, then there already are some local extensions in the project so there isn't a lot to gain from including this. If one is working in an agency, there are probably some shared per agency extension libraries.

Sincerely, I'm more inclined to not including these.

muukii:
@kzaher 

I've understood your opinion. And, I think so too.
Therefore, This APIs should be included for each project.
Thanks!

kzaher:
Now there is https://github.com/devxoul/RxViewController

muukii:
@kzaher Thanks!

onmyway133:
@kzaher I used `sentMessage`, it is so convenient. After I found out that it used swizzling under the hood, I think I need to change back to use plain `Subject` to achieve this

PR #1259, opened 2017-05-18
Add elementAt operator to SharedSequence

yuzushioh:
Hi! I added `elementAt` operator to `SharedSequence` in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ## Master
 
-* Adds `from` operator on "SharedSequence"
+* Adds `from` and `elementAt` operator on "SharedSequence"
 * Adds `merge` operator on "Completable"
 * Adds `using` operator on "PrimitiveSequence" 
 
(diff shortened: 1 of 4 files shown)


kzaher:
Hi @yuzushioh ,

could you maybe provide some context why do you need this operator on `SharedSequence`?

yuzushioh:
Hi @kzaher 

In my ViewModel, I have a stored property `pagination` type of `Driver<Pagination>`. a type `Pagination` is a struct and  has properties such as `MaxSortId`. I want to extract a pagination for a specific page.  In most of my view models I use `Driver` instead of `Observable`

I am doing like this when I extract a pagination element for third page.

```swift
// Trigger is `Driver<Void>`
let extractedPagination = trigger.flatMap {  _ -> Driver<Pagination> in 
    let paginationObservable = pagination.asObservable().elementAt(2)
    return paginationObservable.asDriver(onErrorDriveWith: .empty())
}
```

Instead, it would be great if I can do

```swift
// Trigger is Driver<Void>
let extractedPagination = trigger.flatMap {  _ -> Driver<Pagination> in 
    return pagination.elementAt(2)
}
```

This is why I added `elementAt` operator to `SharedSequence`.

kzaher:
Hi @yuzushioh ,

I'm not sure using `elementAt` is a good idea with `Driver`. `Driver` has `shareReplay` sharing strategy. That has a consequence that calculation could start from anywhere, and using `.elementAt(2)` could return you any element with index >= 2.

yuzushioh:
Hi @kzaher 

I thought about it and you are right about this.
> using .elementAt(2) could return you any element with index >= 2.

I will think about other ways to do my thing. Thanks 👍

PR #1266, opened 2017-05-23
Add rx.isHighlighted to UIImageView

ThomasLeblond:
Add rx.isHighlighted to UIImageView

### DIFF
--- a/RxCocoa/iOS/UIImageView+Rx.swift
+++ b/RxCocoa/iOS/UIImageView+Rx.swift
@@ -40,6 +40,13 @@ extension Reactive where Base: UIImageView {
             imageView.image = image
         }
     }
+    
+    /// Bindable sink for `isHighlighted` property.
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @ThomasLeblond ,

Hm. I think that if we add this, we would need to add highlighted image, etc. I'm not sure a lot of people use it, or know about this.

ThomasLeblond:
Hi @kzaher,

It's really useful when you want a check image for a form.

```swift
let emailValid = emailTextField.rx.text.orEmpty
            .map { !$0.isEmpty && $0.isMatch(RegexPattern.email) }
            .shareReplay(1)

emailValid
            .bind(to: emailImageView.rx.isHighlighted)
            .disposed(by: disposeBag)
```

sergdort:
I agree with @kzaher 
This is fairly simple to implement. And I'm not sure if this is that common to use. TBH I've used this property couple times only

kzaher:
Let's skip this for now then :)

PR #1270, opened 2017-05-26
add NSScrollView.rx.backgroundColor

DivineDominion:
I am just getting my feet wet providing macOS reactive extensions. The test succeeds, but feedback on the approach is very welcome so I know what to look for with other controls.


kzaher:
Hi @DivineDominion ,

It doesn't seem to me that interface is made for observing. Maybe KVO will work, but in general it doesn't seem to me it should be used this way.

DivineDominion:
Oh, okay. Can you explain why a bit? -- I can imagine something among this: a Sink would work to consume changes, but since the scroll view doesn't update its color on its own, providing an Observable is not a good choice.

kzaher:
Hi @DivineDominion ,

it just follows from my own personal experience regarding how to do unidirectional architectures. 

You are allowed to bind to any UI property (bindings are just optimized view renderings for stateful UIs), but you are only allowed to observe direct user interaction events and convert them into commands that mutate the state.

Otherwise you are storing model inside the UI, which is ok if you are writing a "model" for UIControl, but not ok if you are writing you business logic.

Issue #1272, opened 2017-05-28
[Proposition] Errorless Observers


jeremiegirault:
**Short description of the issue**:

Except using `Driver`, current RxSwift design does not allow to define and consume error-less Observable streams. 
Internally Driver relies on Observable and runtime checks, which can fail and error. 
Most errors are prevented by making the driver creation feature `internal` to the framework. Therefore, most of the time the conversion from Observable to Driver is painful and brings lots of unnecessary features (main thread, shared subscription).
A lot of RxSwift Primitive are highly dependent on errors (e.g. binding an `Observable` to a `Subject` can make it silently fail). See #1116 for example of typical use cases.

I expect to be able to create chains of observers which will either be able to error or not and consume them accordingly.

**Proposition**:

I used ReactiveCocoa a lot in the past and my experience tells me that having typed errors is not so enjoyable (flatMapping on `Error` types while combining streams with operators is painful).

However I Feel that swift typing and generic system would allow us to create `Observer`s, `Observable`s and `Event`s having a specific kind (trait) (`Error`-enabled or not).
This specific kind would prevent or allow binding at compile time using this logic :
- `ErrorlessObservable` can bind to `ErrorObservable`
- `ErrorlessObservable` can bind to `ErrorlessObservable`
- `ErrorObservable` can bind to `ErrorObservable`
Combining Observables would follow the following logic:
- `ErrorObservable.flatMap(ErrorlessObservable) -> ErrorObservable`
- `ErrorObservable.flatMap(ErrorObservable) -> ErrorObservable`
- `ErrorlessObservable.flatMap(ErrorlessObservable) -> ErrorlessObservable`
- `ErrorlessObservable.flatMap(ErrorObservable) -> ErrorObservable`
This would enforce the need to create observables handling errors in such cases and avoid silently ignoring errors (especially when binding).

In this context, `Observer` would be a specialization of the `ErrorlessObserver` (Observer can be cast to ErrorlessObserver)
On the opposite `ErrorlessEvent` is a specialisation of `Event` by removing the `.error` case. (ErrorlessEvent can be cast to Event).
With `Observable` I'm a bit lost, I don't know understand the full implementation of RxSwift. 
I guess we will need two kind, or an associatedType defining the required `Observable` type. Then we would have operators allowing us to lift the error-less kind to error-enabled or handling errors. 
I feel that in most case the operator implementation would need to implement the error-less path except when the operator can fail.

[...]

PR #1273, opened 2017-05-28
Preserve type on shared sequence custom binder

jeremiegirault:
In my codebase error are handled. At the moment I'm using `Driver` to guarantee that the binding is errorless (See #1272) Therefore on `Observer` side I have guarantees that `Observable` streams won't fail.
That's why it is a problem for me that the `drive` with binding function gives me an `Observable` (which means potential error on `Observer` side). Moreover the `drive` function is provided by driver which contradicts the interface given.

Therefore, I just changed the `drive` with binding function prototype in `RxCocoa/Traits/Driver/Driver+Subscription.swift`. That led me to an interesting compile error track and I solved everything.

I ran all the testing scripts, and it seem to run fine.


kzaher:
Hi @jeremiegirault ,

Merging this would be a big breaking change that would require incrementing the major version :)

There is a reason why all of the interfaces use `ObserverType`/`Observable`. We wanted to provide freedom for users to define their own traits and bind them to UI. Or just use plain `Observable` if they have proven their traits manually or they are plain obvious.

` public func drive<R>(_ transformation: (Observable<E>) -> R) -> R {` this has `Observable` for the same reason.

I think it's raising these questions is a valid point, and I didn't make those decisions lightly, but I think that in the long run having `ObserverType` interfaces will be better for the project and everything will fall into place once Swift provides us with the necessary capabilities.

jeremiegirault:
I would disagree and would like to argue on this point :

If we strictly look at the definition of `drive(_ () -> R) -> R`, it is only defined on `Driver`
```
extension SharedSequenceConvertibleType where SharingStrategy == DriverSharingStrategy {
    public func drive<R>(_ transformation: (Observable<E>) -> R) -> R {
        MainScheduler.ensureExecutingOnScheduler(errorMessage: driverErrorMessage)
        return transformation(self.asObservable())
    }
}
```
This means that on the Observable/binding side we will never `drive` `Observables`, only `Drivers`.
On the opposite side, as an `Observer`, `Driver` could be upcast to `Observable` so it is not an issue. But the opposite is not true. Providing an `Observable` where a `Driver` could be expected means that the observer have to downcast the Observable (which would always be a driver). 
Therefore, I do not see the point of only using `Observable` restricting usages; both from a typing point of view, and from a "reactive trait convertibility" point of view.

Would you think it would be better if the parameter was a `SharedSequenceConvertibleType` to abstract the `Driver` ? It would still fit the `ObservableConvertibleType` protocol and hide the `Driver` implementation from the Observer/Consumer point of view.

Actually I think it would improve the solution, what do you think ?

kzaher:
Hi @jeremiegirault ,

Even if we agree this would be a breaking change :) 

The only thing I can do is to provide you with explanation why this interface is designed this way. I've had doubts about how to design this interface best, so the one you are suggesting was unfortunately my second choice.

RxCocoa consists of two parts, RxCocoa and RxTraits. We've packed them together for now is because it was more convenient to distribute RxTraits this way (inside RxCocoa library). The existing RxTraits (Driver, SharedSequence and soon to be Publisher) are designed for usage in stateful environment anyway.

Conceptually speaking, I wanted to make RxCocoa only dependent on RxSwift. That way if somebody finds it inconvenient to use our RxTraits, they can create their own. Its basic "don't put all eggs in one basket" thinking.

Like you've noticed, it's little inconvenient now to extend SharedSequence and Driver so if someone find them too inconvenient, they could design their reactive traits and use them with our RxCocoa extensions, although in the end that wasn't that big of a problem.
We have also ported RxTraits to Kotlin and will probably publish them soon. They look a bit different there, but we've managed to make them work somehow :)

The reason why `public func drive<R>(_ transformation: (Observable<E>) -> R) -> R` looks this way is because it's adapter interface.

`(Observable<E>) -> R` can be found in RxCocoa project, and `Driver` in RxCocoa/Traits.

jeremiegirault:
I still don't think that it is a breaking change.
However maybe we could find a compromise with this solution that I implemented: replacing the whole PR with an extension allowing the new feature while keeping the original design ?

```
// RxCocoaAdditions.swift
extension SharedSequenceConvertibleType where SharingStrategy == DriverSharingStrategy {
    public func drive<R>(_ transformation: (Self) -> R) -> R {
        MainScheduler.ensureExecutingOnScheduler(errorMessage: driverErrorMessage)
        return transformation(self)
    }

    public func drive<R1, R2>(_ with: (Self) -> (R1) -> R2, curriedArgument: R1) -> R2 {
        MainScheduler.ensureExecutingOnScheduler(errorMessage: driverErrorMessage)
        return with(self)(curriedArgument)
    }
}
```

PR #1282, opened 2017-06-03
Compile under Swift 4 

volodg:
This applies minimal changes to allow RxSwift to compile under Swift 4

### DIFF
--- a/RxCocoa/iOS/UITableView+Rx.swift
+++ b/RxCocoa/iOS/UITableView+Rx.swift
@@ -168,15 +168,6 @@ extension Reactive where Base: UITableView {
 extension UITableView {
  
     /**
-    Factory method that enables subclasses to implement their own `delegate`.
-    
-    - returns: Instance of delegate proxy that wraps `delegate`.
-    */
-    public override func createRxDelegateProxy() -> RxScrollViewDelegateProxy {
-        return RxTableViewDelegateProxy(parentObject: self)
-    }
-
-    /**
     Factory method that enables subclasses to implement their own `rx.dataSource`.
     
     - returns: Instance of delegate proxy that wraps `dataSource`.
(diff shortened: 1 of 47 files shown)


ldiqual:
@volodg I get this error still:

```
RxCocoa/Foundation/NSObject+Rx.swift:458:57: Value of optional type 'objc_property_t?' (aka 'Optional<OpaquePointer>') not unwrapped; did you mean to use '!' or '?'?
```

kzaher:
Hi @volodg ,

I think there are some issues with `public typealias RxAbstractInteger = FixedWidthInteger` and `SwiftSupport/SwiftSupport.swift`.

You can try running `./scripts/all-tests.sh` locally.

RxCocoa/iOS/UITableView+Rx.swift
```diff
@@ -168,15 +168,6 @@ extension Reactive where Base: UITableView {
 extension UITableView {
  
     /**
-    Factory method that enables subclasses to implement their own `delegate`.
-    
-    - returns: Instance of delegate proxy that wraps `delegate`.
-    */
-    public override func createRxDelegateProxy() -> RxScrollViewDelegateProxy {
```
kzaher: Hi @volodg ,

I've missed these changes. I looks like these cause real issues.

Could you please explain why you've removed these `createRxDelegateProxy` method overrides?

volodg: Missed it, already fixed.
"override" now forbidden inside of extension, so I moved this logic into UIScrollView

kzaher: Oh crap :(

I mean, your changes are a bandaid, so they are great first step, but hoping we can find some more extensible solution :(

volodg: Of cause, hope you can )

kzaher: Let's try to find some better solution for this, I'll also try to figure out some better way.

PR #1283, opened 2017-06-06
Added UIViewController Reactive extension for tvOS

oanhof:
I added the UIViewController Reactive extension for tvOS and added the Unit Test.


kzaher:
Hi @oanhof ,

ups, we are missing unit tests for this extension. Could you help us and also add them please?

oanhof:
I added `UIViewControler+RxTests.swift` file to the tvOS tests target. So the same test will run for iOS and tvOS. That should be enough right?

PR #1290, opened 2017-06-12
Tone down warning about swift 3.0 requirement

hfossli:
With Swift 4.0 out in preview I think it is safe to say most people have moved past swift 2.x.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,8 +5,6 @@
 
 ## About Rx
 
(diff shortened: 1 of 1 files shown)


hfossli:
Can be previewed here https://github.com/hfossli/RxSwift/blob/945b178ae81d821ff85e137704ef4cf8e4954931/README.md

PR #1291, opened 2017-06-13
forward RxTextFieldDelegateProxy.controlTextDidChange to underlying delegate

DivineDominion:
Although this method is not part of the `NSTextFieldDelegate` protocol, and although the docs say this is part of a `NSControl`, in fact this method is available on `NSObject`s, view controllers, and the like and is the standard typing observation mechanism.

Since the `RxTextFieldDelegateProxy` hooked itself onto this without forwarding, existing delegates in the rest of the app become deactivated.

### DIFF
--- a/RxCocoa/macOS/NSTextField+Rx.swift
+++ b/RxCocoa/macOS/NSTextField+Rx.swift
@@ -40,6 +40,7 @@ public class RxTextFieldDelegateProxy
         let textField: NSTextField = castOrFatalError(notification.object)
         let nextValue = textField.stringValue
         self.textSubject.on(.next(nextValue))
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @DivineDominion ,

could you also please add a unit test for that?

DivineDominion:
Sure, will do.

DivineDominion:
@kzaher looking at existing RxCocoa tests, this seemed sufficient. If you want me to split it up into multiple test cases (Rx only, delegate only, Rx + delegate), just say so.

kzaher:
@DivineDominion yep, cool.

PR #1294, opened 2017-06-14
Tidy up clutter in top of README.md

hfossli:
Moved note about swift 2.x branch to "requirements" section.

Removed "About RxSwift" because I find it excessive considering it is just below the super title `ReactiveX for Swift`


hfossli:
Preview can be seen here https://github.com/hfossli/RxSwift/blob/b48f55d4c902371bf9e3077fe700df29c0e58e12/README.md

It didn't look good for the eye with warning icons on top of readme. Especially considering excessive considering it was targeting a very small number of visitors to this site.

Issue #1297, opened 2017-06-15
Programmatically triggering UIBarButtonItem action in tests causes a crash


lyricsboy:
**Short description of the issue**:

In some unit tests, I use code like the following to simulate tapping on a UIBarButtonItem through its target-action pair. If that action is wired up using Rx, I get a crash in my tests related to memory.

```swift
    func testBarButtonItem_actionExecution() {
        let button = UIBarButtonItem()
        weak var tapExpectation = expectation(description: "tap")
        button.rx.tap.subscribe(onNext: {
            tapExpectation?.fulfill()
        }).disposed(by: disposeBag)
        _ = button.target?.perform(button.action) // crashes here
        waitForExpectations(timeout: 1, handler: nil)
    }
```

**Expected outcome**:

Triggering the action fires the Rx observable and everything is happy. This was actually working until recently, when I switched they way I was integrating RxSwift from Carthage-built to directly incorporated into my project's workspace via a Carthage-managed git submodule.

**What actually happens**:

Crash in the execution of BarButtonItemTarget.action, swift_unknownRetain on the lines with the arrow:

```
RxCocoa`@objc BarButtonItemTarget.action(AnyObject) -> ():
    0x115984f10 <+0>:  pushq  %rbp
    0x115984f11 <+1>:  movq   %rsp, %rbp
    0x115984f14 <+4>:  pushq  %r15
    0x115984f16 <+6>:  pushq  %r14
    0x115984f18 <+8>:  pushq  %r12
    0x115984f1a <+10>: pushq  %rbx
    0x115984f1b <+11>: movq   %rdx, %r14
    0x115984f1e <+14>: movq   %rdi, %rbx
    0x115984f21 <+17>: incq   0x51058(%rip)             ; __profc__TZFC7RxCocoa22RxWebViewDelegateProxy20createProxyForObjectfPs9AnyObject_PS1__ + 5504
    0x115984f28 <+24>: movq   0x4eba9(%rip), %r15       ; direct field offset for RxCocoa.BarButtonItemTarget.callback : Swift.ImplicitlyUnwrappedOptional<() -> ()>
    0x115984f2f <+31>: movq   (%rbx,%r15), %r12
    0x115984f33 <+35>: movq   %r14, %rdi
    0x115984f36 <+38>: callq  0x11599f708               ; symbol stub for: swift_unknownRetain
->  0x115984f3b <+43>: movq   %rbx, %rdi
[...]

PR #1299, opened 2017-06-16
Adds a test for tap action on UIBarButtonItem

lyricsboy:
Based on discovery in #1297

### DIFF
--- a/Tests/RxCocoaTests/UIBarButtonItem+RxTests.swift
+++ b/Tests/RxCocoaTests/UIBarButtonItem+RxTests.swift
@@ -32,12 +32,12 @@ extension UIBarButtonItemTests {
     
     func testBarButtonItem_actionExecution() {
         let button = UIBarButtonItem()
-        weak var tapExpectation = expectation(description: "tap")
+        var onNextCalled = false
         let disposable = button.rx.tap.subscribe(onNext: {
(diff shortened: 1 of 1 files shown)


Tests/RxCocoaTests/UIBarButtonItem+RxTests.swift
```diff
@@ -29,4 +29,15 @@ extension UIBarButtonItemTests {
...
+    
+    func testBarButtonItem_actionExecution() {
+        let button = UIBarButtonItem()
+        weak var tapExpectation = expectation(description: "tap")
+        let disposable = button.rx.tap.subscribe(onNext: {
+            tapExpectation?.fulfill()
+        })
+        defer { disposable.dispose() }
+        _ = button.target?.perform(button.action, with: nil)
+        waitForExpectations(timeout: 1, handler: nil)
```
sergdort: I think you can make it work without expectations e.g

```
let button = UIBarButtonItem()
         var onNextCalled = false
         let disposable = button.rx.tap.subscribe(onNext: {
             onNextCalled = true
         })
         _ = button.target?.perform(button.action, with: nil)
```

lyricsboy:
@sergdort good call. I forgot that it would be synchronous. Updated!

Issue #1287, opened 2017-06-09
Feature request: Adding Generics parameter in DelegateProxy


tarunon:
Hello, 
This proposal is about improving `DelegateProxy` type supports.

`DelegateProxy` had weak type support, so we needed to use many force casting when implement `DelegateProxy` subclass.
This proposal goal is adding Generics type in `DelegateProxy`, and removing unnecessary force casting from implementation of `DelegateProxy` subclass.
This proposal will make many code differences and API changes, I hope to discussion it targeting Swift4 and RxSwift4.0.

## Plan
Adding associatedtype on `DelegateProxyType` and change `DelegateProxy` implementation like
```swift
public protocol DelegateProxyType : AnyObject {
    associatedtype ParentObject: AnyObject
    associatedtype Delegate: AnyObject
...
} 
open class DelegateProxy<P: AnyObject, D: AnyObject>: _RXDelegateProxy, DelegateProxyType { 
    public typealias ParentObject = P
    public typealias Delegate = D
...
}
```
 Implements `DelegateProxy` subclass is specify Generics parameter like 
```swift
 public class RxScrollViewDelegateProxy
    : DelegateProxy<UIScrollView, UIScrollViewDelegate>
    , UIScrollViewDelegate {
...
}
```

I try to this way in my [fork](https://github.com/tarunon/RxSwift/tree/feature/type_safe_delegateproxy), It's seems work well. 

## Pros
- `DelegateProxy` is now support `DelegateProxyType`, It mean can remove some of force casting and type assertion from internal implementation of `DelegateProxy`.
- We don't need to write many force casting anymore when we implement `DelegateProxy` subclass.
- `installForwardDelegate` is safer than before, `Delegate` type will be check in compile time.

## Cons
- It's unfortunately we still need to write force casting when we implement `DelegateProxy` subclasses subclass. 
- `DelegateProxy` has a Generics type, then we cannot use `DelegateProxy` at extension function return type because of Swift 3.0 don't support overriding function that has generics type on extension.  I use `Delegate` type instead of `DelegateProxy` for this workaround.  In future swift, maybe this workaround will be unnecessary.
- `DelegateProxy` is abstract class.

kzaher:
Hi @tarunon ,

I've tried to figure out some more "nicer" typesafer solution a while ago, but it didn't work out.

I've glanced at your suggestion, but there are tons of changes, so please correct me if I'm wrong.

I would expect the general solution to have this form.

```
[...]

PR #1300, opened 2017-06-16
Implement DelegateProxy factory for Swift 4.0

tarunon:
It's based on discuss of #1287 

Notes:
- I made this more simply than discussions one.
- Maybe we don't need `func createProxyForObject(_ :)` specific implementation anymore.
- The performance is less than overriding way, but we cannot use overriding way anymore...
- I think I will separate type around `factory` from `DelegateProxy` when I implement `DelegateProxy` with generics type. Because we cannot make static property using generics type.

### DIFF
--- a/RxCocoa/Common/DelegateProxyFactory.swift
+++ b/RxCocoa/Common/DelegateProxyFactory.swift
@@ -18,7 +18,7 @@ For example, in RxScrollViewDelegateProxy
 
 
     class RxScrollViewDelegateProxy: DelegateProxy {
-        static var delegateProxyFactory = DelegateProxyFactory { (parentObject: UIScrollView) in
+        static var factory = DelegateProxyFactory { (parentObject: UIScrollView) in
             RxScrollViewDelegateProxy(parentObject: parentObject)
         }
     ...
@@ -27,7 +27,7 @@ For example, in RxScrollViewDelegateProxy
 If need to extend them, chain `extended` after DelegateProxyFactory.init
 
     class RxScrollViewDelegateProxy: DelegateProxy {
-        static var delegateProxyFactory = DelegateProxyFactory { (parentObject: UIScrollView) in
+        static var factory = DelegateProxyFactory { (parentObject: UIScrollView) in
                 RxScrollViewDelegateProxy(parentObject: parentObject)
             }
             .extended { (parentObject: UITableView) in
@@ -38,23 +38,33 @@ If need to extend them, chain `extended` after DelegateProxyFactory.init
  
  */
 public class DelegateProxyFactory {
-    var factories: [ObjectIdentifier: ((AnyObject) -> AnyObject)]
+    private var _factories: [ObjectIdentifier: ((AnyObject) -> AnyObject)]
     public init<Object: AnyObject>(factory: @escaping (Object) -> AnyObject) {
-        factories = [ObjectIdentifier(Object.self): { factory(castOrFatalError($0)) }]
+        _factories = [ObjectIdentifier(Object.self): { factory(castOrFatalError($0)) }]
     }
     
+    /**
+     Extend DelegateProxyFactory for specific object class and delegate proxy.
+     Define object class on closure argument.
+    */
     public func extended<Object: AnyObject>(factory: @escaping (Object) -> AnyObject) -> DelegateProxyFactory {
-        guard factories[ObjectIdentifier(Object.self)] == nil else {
+        MainScheduler.ensureExecutingOnScheduler()
+        guard _factories[ObjectIdentifier(Object.self)] == nil else {
             rxFatalError("The factory of \(Object.self) is duplicated. DelegateProxy is not allowed of duplicated base object type.")
         }
-        factories[ObjectIdentifier(Object.self)] = { factory(castOrFatalError($0)) }
+        _factories[ObjectIdentifier(Object.self)] = { factory(castOrFatalError($0)) }
         return self
     }
     
-    func createProxy(for object: AnyObject) -> AnyObject {
+    /**
+     Create DelegateProxy for object.
+     DelegateProxyFactory should have a factory of object class (or superclass).
+    */
+    public func createProxy(for object: AnyObject) -> AnyObject {
+        MainScheduler.ensureExecutingOnScheduler()
         var mirror: Mirror? = Mirror(reflecting: object)
         while mirror != nil {
-            if let factory = factories[ObjectIdentifier(mirror!.subjectType)] {
+            if let factory = _factories[ObjectIdentifier(mirror!.subjectType)] {
                 return factory(object)
             }
             mirror = mirror?.superclassMirror
--- a/RxCocoa/Common/DelegateProxyFactory.swift
+++ b/RxCocoa/Common/DelegateProxyFactory.swift
@@ -1,0 +1,66 @@
+//
+//  DelegateProxyFactory.swift
+//  RxCocoa
+//
+//  Created by tarunon on 2017/06/17.
+//  Copyright © 2017 Krunoslav Zaher. All rights reserved.
+//
+
+#if !os(Linux)
+    
+#if !RX_NO_MODULE
+    import RxSwift
+#endif
+
+/**
+Define `DelegateProxy.init` for a specific object type.
+For example, in RxScrollViewDelegateProxy
+
+
+    class RxScrollViewDelegateProxy: DelegateProxy {
+        static var delegateProxyFactory = DelegateProxyFactory { (parentObject: UIScrollView) in
+            RxScrollViewDelegateProxy(parentObject: parentObject)
+        }
+    ...
+
+
+If need to extend them, chain `extended` after DelegateProxyFactory.init
+
+    class RxScrollViewDelegateProxy: DelegateProxy {
+        static var delegateProxyFactory = DelegateProxyFactory { (parentObject: UIScrollView) in
+                RxScrollViewDelegateProxy(parentObject: parentObject)
+            }
+            .extended { (parentObject: UITableView) in
+                RxTableViewDelegateProxy(parentObject: parentObject)
+            }
+    ...
+ 
+ 
+ */
+public class DelegateProxyFactory {
+    var factories: [ObjectIdentifier: ((AnyObject) -> AnyObject)]
+    public init<Object: AnyObject>(factory: @escaping (Object) -> AnyObject) {
+        factories = [ObjectIdentifier(Object.self): { factory(castOrFatalError($0)) }]
+    }
+    
+    public func extended<Object: AnyObject>(factory: @escaping (Object) -> AnyObject) -> DelegateProxyFactory {
+        guard factories[ObjectIdentifier(Object.self)] == nil else {
+            rxFatalError("The factory of \(Object.self) is duplicated. DelegateProxy is not allowed of duplicated base object type.")
+        }
+        factories[ObjectIdentifier(Object.self)] = { factory(castOrFatalError($0)) }
+        return self
+    }
+    
+    func createProxy(for object: AnyObject) -> AnyObject {
+        var mirror: Mirror? = Mirror(reflecting: object)
+        while mirror != nil {
+            if let factory = factories[ObjectIdentifier(mirror!.subjectType)] {
+                return factory(object)
+            }
+            mirror = mirror?.superclassMirror
+        }
+        rxFatalError("DelegateProxy has no factory of \(object). Call 'DelegateProxy.extend' first.")
+    }
+}
+    
+#endif
--- a/RxCocoa/Common/DelegateProxyType.swift
+++ b/RxCocoa/Common/DelegateProxyType.swift
@@ -69,22 +69,25 @@ every view has a corresponding delegate virtual factory method.
 
 In case of UITableView / UIScrollView, there is
 
-    extension UIScrollView {
-        public func createRxDelegateProxy() -> RxScrollViewDelegateProxy {
-            return RxScrollViewDelegateProxy(parentObject: base)
-        }
+    RxScrollViewDelegateProxy has factories that contains RxScrollViewDelegateProxy(parentObject: parentObject)
     ....
 
 
-and override in UITableView
+and extend it
 
-    extension UITableView {
-        public override func createRxDelegateProxy() -> RxScrollViewDelegateProxy {
-        ....
+    RxScrollViewDelegateProxy.extend { (parentObject: UITableView) in
+       RxTableViewDelegateProxy(parentObject: parentObject)
+    }
 
 
 */
 public protocol DelegateProxyType : AnyObject {
+    /// DelegateProxy factory
+    static var factories: [((AnyObject) -> AnyObject?)] { get set }
+    
+    /// Extend DelegateProxy for specific subclass
+    static func extend<Object: AnyObject>(with factory: @escaping ((Object) -> AnyObject))
+    
     /// Creates new proxy for target object.
     static func createProxyForObject(_ object: AnyObject) -> AnyObject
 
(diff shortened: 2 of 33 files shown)


RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -210,6 +213,14 @@ extension DelegateProxyType {
             proxy.setForwardToDelegate(nil, retainDelegate: retainDelegate)
         }
     }
+    
+    public static func extend<Object: AnyObject>(with factory: @escaping ((Object) -> AnyObject)) {
+        factories.append({ ($0 as? Object).map(factory) })
+    }
+    
+    public static func createProxyForObject(_ object: AnyObject) -> AnyObject {
+        return factories.reversed().reduce(AnyObject?.none) { $0 ?? $1(object) }!
```
kzaher: Hi @tarunon ,

I think this implementation makes an assumption that `extend` function callers perform calls in correct order.

This seems fragile to me and it's not exactly clear from the public interface there is this requirement. I would expect that implementation would resolve class hierarchies correctly.

So this should also work:
```
RxScrollViewDelegateProxy.extend { (parentObject: UITableViewSubclassSubclass1) in
         ExtendTableViewDelegateProxy(parentObject: parentObject)
}
RxScrollViewDelegateProxy.extend { (parentObject: UITableViewSubclass1) in
         ExtendTableViewDelegateProxy(parentObject: parentObject)
}
```

Since you are also saying 
> I think I will separate type around factory from DelegateProxy when I implement DelegateProxy with generics type. Because we cannot make static property using generics type.

Let's then maybe change this storage right now and move everything into one private static dictionary somewhere.
```
static var proxyMakers: [AnyClass: (AnyObject) -> AnyObject] // or some more idiomatic name
```

I think the resolution algorithm should look something like:
```
public static func makeProxy(for: AnyObject) -> AnyObject {
     MainScheduler.ensureExecutingOnScheduler(some meaningful message)
     var type = type(of: for)
     while type != nil {
          defer {
              type = type.parent
          }
          if let result = proxyMakers[type].map { $0(for) } {
               return result
          }

     }

     fatalError("Some meaningful error that describes we couldn't resolve delegate proxy for "\(for). Please use xxx to register delegate proxy")
} 
```

I don't think this should cause any performance concerns.

kzaher: We could probably name it something more idiomatic, like `createProxy(for object: AnyObject)`

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -210,6 +213,14 @@ extension DelegateProxyType {
             proxy.setForwardToDelegate(nil, retainDelegate: retainDelegate)
         }
     }
+    
+    public static func extend<Object: AnyObject>(with factory: @escaping ((Object) -> AnyObject)) {
+        factories.append({ ($0 as? Object).map(factory) })
```
kzaher: Please take a look at `createProxyForObject` method suggestion first.

We need to crash the app in case extend is called from non main scheduler. // thread safety
We need to crash the app in case somebody registers extension for the same class twice.

RxCocoa/Common/DelegateProxyType.swift
kzaher: We need to add unit tests for factory resolution algorithm.

kzaher:
I've somehow messed up unit tests on swift4.0 branch. I've tried to fix them, but didn't see anything obvious, and everything worked for me locally. 

So frustrating. If somebody can help out with unit tests on swift4.0 branch, that would be awesome.

I really have a limited internet access in hospital, so it's a bit too much for me to download swift runtime on linux each time and xcode 8.2/9.0 beta. ¯\\\_(ツ)\_/¯

RxCocoa/Common/DelegateProxyType.swift
tarunon: That's seems cool! I try it, thanks.

RxCocoa/Common/DelegateProxyType.swift
tarunon: Maybe if I follow swift coding guideline, `createProxyForObject(_:)` should be `createProxy(for object:)`.
I will do so once, please let me know if it's wrong.

RxCocoa/Common/DelegateProxyType.swift
tarunon: I think it have a lot of things in `factory`, then I decided that divide implementation and hide actual `Dictionary`.

tarunon:
Hi @kzaher ,
The most common reason of failing build of unit tests on swift 4.0 is [SE-0110](https://github.com/apple/swift-evolution/blob/master/proposals/0110-distingish-single-tuple-arg.md), but it is still go on discussion. 
If we fix it immediately, we should replace many things like a closure that has tuple argument.
```swift
// as is 
let subscription = tableView.rx.willDisplayCell
    .subscribe(onNext: { (cell, indexPath) in // error here
        resultIndexPath = indexPath
        resultCell = cell
    })

// to be
let subscription = tableView.rx.willDisplayCell
    .subscribe(onNext: { event in
        resultIndexPath = event.indexPath
        resultCell = event.cell
    })
```

As a result of the discussion, it may be unnecessary change, but if it is okay, I can make a new PR.

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -69,22 +69,29 @@ every view has a corresponding delegate virtual factory method.
...
-        ....
+    RxScrollViewDelegateProxy.extendProxy { (parentObject: UITableView) in
+       RxTableViewDelegateProxy(parentObject: parentObject)
+    }
 
 
 */
 public protocol DelegateProxyType : AnyObject {
+    /// DelegateProxy factory
+    static var delegateProxyFactory: DelegateProxyFactory { get }
```
kzaher: Maybe just name it `factory`. `delegateProxyFactory` seems a bit redundant :)

RxCocoa/Common/DelegateProxyFactory.swift
```diff
@@ -0,0 +1,66 @@
...
+            }
+            .extended { (parentObject: UITableView) in
+                RxTableViewDelegateProxy(parentObject: parentObject)
+            }
+    ...
+ 
+ 
+ */
+public class DelegateProxyFactory {
+    var factories: [ObjectIdentifier: ((AnyObject) -> AnyObject)]
```
kzaher: We use `_` conventions for private. This could be `private` as far as I can tell.

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -210,6 +217,16 @@ extension DelegateProxyType {
             proxy.setForwardToDelegate(nil, retainDelegate: retainDelegate)
         }
     }
+    
+    public static func extendProxy<Object: AnyObject>(with factory: @escaping ((Object) -> AnyObject)) {
+        MainScheduler.ensureExecutingOnScheduler()
```
kzaher: It seems to me that this is in wrong place. Somebody could made a direct call to `delegateProxyFactory` and avoid this check.

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -210,6 +217,16 @@ extension DelegateProxyType {
...
+        _ = delegateProxyFactory.extended(factory: factory)
+    }
+    
+    public static func createProxyForObject(_ object: AnyObject) -> AnyObject {
+        MainScheduler.ensureExecutingOnScheduler()
```
kzaher: It seems to me that this is in wrong place. Somebody could made a direct call to `delegateProxyFactory` and avoid this check.

Tests/RxCocoaTests/DelegateProxyTest+UIKit.swift
```diff
@@ -265,6 +259,36 @@ final class UIScrollViewSubclass
     }
 }
 
+final class ExtendScrollViewDelegateProxy2
+    : RxScrollViewDelegateProxy
+, TestDelegateProtocol {
```
kzaher: In a lot of places there is `, XXX`. It would be great to properly align all of these places.

Tests/RxCocoaTests/DelegateProxyTest+UIKit.swift
```diff
@@ -265,6 +259,36 @@ final class UIScrollViewSubclass
...
+    
+    required init(parentObject: AnyObject) {
+        self.control = (parentObject as! UIScrollViewSubclass2)
+        super.init(parentObject: parentObject)
+    }
+}
+
+final class UIScrollViewSubclass2
+    : UIScrollView
+, TestDelegateControl {
```
kzaher: another alignment ...

Tests/RxCocoaTests/DelegateProxyTest.swift
```diff
@@ -56,6 +56,70 @@ extension TestDelegateControl {
 // MARK: Tests
 
 final class DelegateProxyTest : RxTest {
+    
+    static var extendDelegateProxy: Void = {
```
kzaher: What is the purpose of this part? We should be able to remove this, right?

RxCocoa/Common/DelegateProxyFactory.swift
```diff
@@ -0,0 +1,66 @@
...
+    
+    public func extended<Object: AnyObject>(factory: @escaping (Object) -> AnyObject) -> DelegateProxyFactory {
+        guard factories[ObjectIdentifier(Object.self)] == nil else {
+            rxFatalError("The factory of \(Object.self) is duplicated. DelegateProxy is not allowed of duplicated base object type.")
+        }
+        factories[ObjectIdentifier(Object.self)] = { factory(castOrFatalError($0)) }
+        return self
+    }
+    
+    func createProxy(for object: AnyObject) -> AnyObject {
```
kzaher: This should be public probably, and documented.

RxCocoa/Common/DelegateProxyFactory.swift
```diff
@@ -0,0 +1,66 @@
...
+ 
+ 
+ */
+public class DelegateProxyFactory {
+    var factories: [ObjectIdentifier: ((AnyObject) -> AnyObject)]
+    public init<Object: AnyObject>(factory: @escaping (Object) -> AnyObject) {
+        factories = [ObjectIdentifier(Object.self): { factory(castOrFatalError($0)) }]
+    }
+    
+    public func extended<Object: AnyObject>(factory: @escaping (Object) -> AnyObject) -> DelegateProxyFactory {
```
kzaher: This should have documentation.

RxCocoa/Common/DelegateProxyFactory.swift
kzaher: We need to add unit tests for this method.

I saw that you've added `test_DelegateProxyHasNoSpecificFactory`. 

I would like us to cover this a bit better using unit tests.

```
ClassSometimesProxy1:ClassSometimesProxy2:Class1: Class2
InitialProxy1: InitialProxy2
```

```
test1
calling `createProxy ` for object of class1 -> returns initial proxy1 type

test2
calling `createProxy` for object of class 2 -> returns initial proxy2 type

test3
calling `createProxy` for object of ClassSometimesProxy2 -> return initial proxy1 type
calling `createProxy` for object of ClassSometimesProxy1 -> return initial proxy1 type
then call extend for `ClassSometimesProxy2`
calling `createProxy` for object of ClassSometimesProxy2 -> return of type ClassSometimesProxy2
calling `createProxy` for object of ClassSometimesProxy1 -> return of type ClassSometimesProxy2

test4
calling `createProxy` for object of ClassSometimesProxy2 -> return initial proxy1 type
calling `createProxy` for object of ClassSometimesProxy1 -> return initial proxy1 type
then call extend for `ClassSometimesProxy1`
calling `createProxy` for object of ClassSometimesProxy2 -> return initial proxy1 type
calling `createProxy` for object of ClassSometimesProxy1 -> return of type ClassSometimesProxy1
```

kzaher: We might need to create separate classes for test3 and test4 since they are mutating static data.

Tests/RxCocoaTests/DelegateProxyTest.swift
tarunon: I'm little bit confusing, we should call `DelegateProxy.extendProxy` in runtime, but it should be only one time.
`static var` is always lazy property and I use this one in `setup` function.
Is it better that adding `main` function in test? `dispatch_once_t` was depricated on Swift 3.0.

tarunon: Maybe I understand now, It's better that change 
```swift
func performDelegateTest<Control: TestDelegateControl>( _ createControl: @autoclosure() -> Control) { // as is
func performDelegateTest<Control: TestDelegateControl, ExtendProxy: DelegateProxy & DelegateProxyType>( _ createControl: @autoclosure() -> Control, extendProxyType: ExtendProxy.self) { // to be
```

kzaher: Oh, sorry, this part is ok. I was just a bit confused.

RxCocoa/Common/DelegateProxyFactory.swift
tarunon: Maybe can add `ClassSometimesProxy3` and check in 1 test case is ok.

kzaher: @tarunon I think we need the following classes or equivalent

`Class1`: `Class2`
`InitialDelegateProxy1`: `InitialDelegateProxy2`

`ClassSometimesProxy1_a` : `ClassSometimesProxy2_a` : `Class1`
`ClassSometimesDelegateProxy2_a`

`ClassSometimesProxy1_b`:`ClassSometimesProxy2_b`: `Class1`
`ClassSometimesDelegateProxy2_b`

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -210,6 +218,17 @@ extension DelegateProxyType {
...
     }
+    
+    /// Extend DelegateProxy for specific subclass
+    /// See 'DelegateProxyFactory.extendedProxy'
+    public static func extendProxy<Object: AnyObject>(with creation: @escaping ((Object) -> AnyObject)) {
+        _ = factory.extended(factory: creation)
+    }
+    
+    /// Creates new proxy for target object.
+    public static func createProxyForObject(_ object: AnyObject) -> AnyObject {
```
kzaher: This should be named the same.

RxCocoa/iOS/Proxies/RxCollectionViewDataSourceProxy.swift
```diff
@@ -35,7 +35,11 @@ final class CollectionViewDataSourceNotSet
 public class RxCollectionViewDataSourceProxy
     : DelegateProxy
     , UICollectionViewDataSource
-    , DelegateProxyType {
+, DelegateProxyType {
```
kzaher: Indentation

RxCocoa/iOS/Proxies/RxSearchControllerDelegateProxy.swift
```diff
@@ -18,12 +18,10 @@
 public class RxSearchControllerDelegateProxy
     : DelegateProxy
     , DelegateProxyType
-    , UISearchControllerDelegate {
-
-    /// For more information take a look at `DelegateProxyType`.
-    public override class func createProxyForObject(_ object: AnyObject) -> AnyObject {
-        let pickerView: UISearchController = castOrFatalError(object)
-        return pickerView.createRxDelegateProxy()
+, UISearchControllerDelegate {
```
kzaher: Indentation.

RxCocoa/iOS/Proxies/RxTextStorageDelegateProxy.swift
```diff
@@ -16,12 +16,10 @@
     public class RxTextStorageDelegateProxy
         : DelegateProxy
         , DelegateProxyType
-        , NSTextStorageDelegate {
-
-        /// For more information take a look at `DelegateProxyType`.
-        public override class func createProxyForObject(_ object: AnyObject) -> AnyObject {
-            let pickerView: NSTextStorage = castOrFatalError(object)
-            return pickerView.createRxDelegateProxy()
+    , NSTextStorageDelegate {
```
kzaher: Indentation.

Issue #887, opened 2016-09-16
RxSwift kills Xcode memory debugger's Leak Agent


nighthawk:
Using `Observable.map` kills Xcode 8's Leak Agent, which means that the Xcode's nice memory debugging tool can't be used in most projects that use RxSwift.

Filed as radar 28331935, but maybe this can be addressed from the RxSwift side?

Minimum sample that's causing the crash:

```swift
class ViewController: UIViewController {

  private let disposeBag = DisposeBag()
  
  override func viewDidLoad() {
    super.viewDidLoad()
    
    // Run, then hit the 'memory graph' button. Nothing will happen.
    // Console will tell you that LeakAgent did crash.
    //
    // Note: Removing the `map` part makes LeakAgent not crash.
    Observable.just("foo")
      .map { _ in
        return "bar"
      }
      .subscribe(
        onNext: { something in
          // good
        }, onError: { error in
          // error
        }
      )
      .addDisposableTo(disposeBag)
    
  }
}
```

**Xcode version**:

```
Version 8.0 (8A218a)
```

kzaher:
Hi @nighthawk ,

if somebody has connections inside Apple that could help us with, that would be great. I'm not sure that we can do something to fix this in this repo, but if we are doing something wrong we can fix it for sure.

DenHeadless:
Just spent an hour trying to make Visual Debugger work. Nothing helps - not tweaking build settings, not anything else. Then i found this issue and it seems to be legit. 
When trying to activate memory debugger, i see only white screen with "No selection" title. Console.app has this to say:

```
kernel AMFI: allowing exception handler for 'LeakAgent' (90508) because the process is not restricted.
kernel AMFI: allowing exception handler for 'LeakAgent' (90508) because the process is not restricted.
Xcode Memory Graph Debugger: no reply dictionary received from LeakAgent request
```

And here's crash, that can be found in Console/Users.log

```
Process:               LeakAgent [90508]
Path:                  /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator.sdk/Developer/Library/PrivateFrameworks/DVTInstrumentsFoundation.framework/LeakAgent
Identifier:            LeakAgent
Version:               61120.1
Code Type:             X86-64 (Native)
Parent Process:        DTServiceHub [84513]
Responsible:           LeakAgent [90508]
User ID:               501

[...]

PR #1302, opened 2017-06-19
adding unavailable messages for Swift 4.0

tarunon:
Hello @kzaher ,
I'm sorry it's my mistaking, I forgot adding unavailable messages for old API.
If some guy override `createRxDelegateProxy`, then they can see unavailable messages.

### DIFF
--- a/RxCocoa/Deprecated.swift
+++ b/RxCocoa/Deprecated.swift
@@ -132,3 +132,109 @@ extension ObservableType {
         })
     }
 }
+
+#if os(iOS) || os(tvOS)
+    import UIKit
+
+    extension NSTextStorage {
+        @available(*, unavailable, message: "createRxDelegateProxy is now unavailable, check DelegateProxyFactory")
+        public func createRxDelegateProxy() -> RxTextStorageDelegateProxy {
+            fatalError()
+        }
+    }
+
+    extension UIScrollView {
+        @available(*, unavailable, message: "createRxDelegateProxy is now unavailable, check DelegateProxyFactory")
+        public func createRxDelegateProxy() -> RxScrollViewDelegateProxy {
+            fatalError()
+        }
+    }
+
+    extension UICollectionView {
+        @available(*, unavailable, message: "createRxDataSourceProxy is now unavailable, check DelegateProxyFactory")
+        public func createRxDataSourceProxy() -> RxCollectionViewDataSourceProxy {
+            fatalError()
+        }
+    }
+
+    extension UITableView {
+        @available(*, unavailable, message: "createRxDataSourceProxy is now unavailable, check DelegateProxyFactory")
+        public func createRxDataSourceProxy() -> RxTableViewDataSourceProxy {
+            fatalError()
+        }
+    }
+
+    extension UINavigationBar {
+        @available(*, unavailable, message: "createRxDelegateProxy is now unavailable, check DelegateProxyFactory")
+        public func createRxDelegateProxy() -> RxNavigationControllerDelegateProxy {
+            fatalError()
+        }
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @tarunon ,

I'm not sure does this compile with Swift 4.0. I'll need to check that out first.

tarunon:
Hi @kzaher , 
I see, ok.
May I help you with something? I know we cannot compile test case on swift 4.0. 
Maybe I can fix it.

kzaher:
@tarunon if you know something doesn't work on 4.0, feel free to make a PR.

tarunon:
Hello @kzaher ,
There is good news for RxSwift Test-case
https://lists.swift.org/pipermail/swift-evolution/Week-of-Mon-20170619/037616.html
> Therefore, we will “back out” the SE-0110 change regarding function arguments from Swift 4.

I think its good that waiting next Swift4 beta releasing. I'll make PR then.

kzaher:
@tarunon ok

tarunon:
Hello, @kzaher 
Xcode9 beta3 is now available, but unfortunately it contains (maybe) bugs when using Optional closure that has tuple argument. I reported it [bugs.swift.org](https://bugs.swift.org/browse/SR-5433) today.
I think we can waiting next beta release, or can make PR that avoid using Optional closure in `subscribe`, `drive`, and `do` temporarily.

Swift 4.0 test cases seems work well when I use Mac version, but it's fail when I use iOS Simulator cause DeallocatedProxy captured by VM object(VM: libobjc.A.dylib __DATA).  Maybe seems same of https://github.com/ReactiveX/RxSwift/issues/887

kzaher:
Hi @tarunon ,

> Xcode9 beta3 is now available, but unfortunately it contains (maybe) bugs when using Optional closure that has tuple argument. I reported it bugs.swift.org today.
I think we can waiting next beta release, or can make PR that avoid using Optional closure in subscribe, drive, and do temporarily.

Oh geez. I think it's best to wait. Let me first set up Xcode9 on Travis.

tarunon:
Thank you for alpha release and fix test of swift 4.0 branch, @kzaher 
It seems fine now.

I'm thinking about this pr meaning, DelegateProxy has a lot of changes, and it will be major release.
I worry that this PR may be meaningless. 
Please let me know if you think this pr does not need.

kzaher:
Hi @tarunon ,

I've first had to fight with compiler bugs to release something usable. 

The alpha release was the first one that passed all of the tests and it doesn't have any known defects but we still need to polish everything.

I now have the time to take a look at this PR.

RxCocoa/iOS/NSTextStorage+Rx.swift
```diff
@@ -11,6 +11,14 @@
     import RxSwift
 #endif
     import UIKit
+
+    extension NSTextStorage {
+        /// Use 'RxTextStorageDelegateProxy.extend' instead.
+        @available(*, unavailable, message: "createRxDelegateProxy is now unavailable, check DelegateProxyFactory")
+        public func createRxDelegateProxy() -> RxTextStorageDelegateProxy {
+            return RxTextStorageDelegateProxy(parentObject: self)
```
kzaher: I think we would need to implement all of these `unavailable` methods as `fatalError()`. I believe using unavailable methods result in compiler errors.

In the root of RxCocoa we have `Deprecated.swift` file. Could you please move all of these extensions to that file so they don't create deprecation noise :).

tarunon: every deprecated method was moved, I clean up my commit. Thanks!

Issue #1309, opened 2017-06-26
RxExample Calculator:  continuous input operations should use the last operation


JolieYang:
**Short description of the issue**:

  _In RxExamples->Calculator demo input such as  `9 + / 3`_

**Expected outcome**:

_should update operator then process as `9 / 3`_

**What actually happens**:

 _didn't update operator then process as `9 + 3`_

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

  _RxSwift3.5.0_

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
8.3.2
```

:warning: Fields below are optional for general issues or in case those questions aren't related to your issue, but filling them out will increase the chances of getting your issue resolved. :warning:

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [x] yes (which ones) Xcode9 beta2,Xcode8.3.2
  - [ ] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [x] I have a small code base
  - [ ] I have a significant code base

kzaher:
Hi @JolieYang ,

I've left some comments on your PR.

PR #1310, opened 2017-06-26
:bug: #1309 continuous input operation should update operator

JolieYang:
In RxExamples->Calculator demo input such as 9 + / 3 should process as `9 / 3` but actuallay is `9 + 3`.


kzaher:
Hi @JolieYang ,

I think you are right. I think the model in example app is wrong.

I think this is probably the correct model.

```
enum CalculatorState {
    case oneOperand(screen: String)
    case oneOperandAndOperator(operand: Double, operator: Operator)
    case twoOperandsAndOperator(operand: Double, operator: Operator, screen: String)
}
```

JolieYang:
@kzaher you're right,I will change the model

PR #1317, opened 2017-07-05
Add missing collectionview delegate control events

renrawnalon:
This PR:

- Adds control events that were not implemented for UICollectionViewDelegate.
- Updates the method prototype syntax in UICollectionView+Rx comments

Please let me know if you need anything modified.

Thanks!

### DIFF
--- a/RxCocoa/iOS/UICollectionView+Rx.swift
+++ b/RxCocoa/iOS/UICollectionView+Rx.swift
@@ -179,8 +179,8 @@ extension UICollectionView {
 }
 
 extension Reactive where Base: UICollectionView {
-    public typealias DisplayCollectionViewCellEvent = (cell: UICollectionViewCell, indexPath: IndexPath)
-    public typealias DisplayCollectionReusableViewEvent = (reuseableView: UICollectionReusableView, elementKind: String, indexPath: IndexPath)
+    public typealias DisplayCollectionViewCellEvent = (cell: UICollectionViewCell, at: IndexPath)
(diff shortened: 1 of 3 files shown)


RxCocoa/iOS/UICollectionView+Rx.swift
```diff
@@ -179,6 +179,8 @@ extension UICollectionView {
 }
 
 extension Reactive where Base: UICollectionView {
+    public typealias DisplayCollectionViewCellEvent = (cell: UICollectionViewCell, indexPath: IndexPath)
```
kzaher: I think this should be `.. at: IndexPath)`.
I understand that we currently have `indexPath` in other places in this project, but it seems to me that it would be nice to change this in 4.0, so why not start now.

RxCocoa/iOS/UICollectionView+Rx.swift
```diff
@@ -179,6 +179,8 @@ extension UICollectionView {
...
+    public typealias DisplayCollectionReusableViewEvent = (reuseableView: UICollectionReusableView, elementKind: String, indexPath: IndexPath)
```
kzaher: I think that for consistency sake this should be named `DisplayCollectionViewSupplementaryViewEvent = (supplementaryView: UICollectionReusableView, ...` and `at: IndexPath)` because this is it's name in all delegate methods.
`UICollectionReusableView` is it's type.

renrawnalon:
@kzaher 

Thanks for checking this PR. The requested changes sound reasonable. I'll keep this in mind when making contributions in the future.

I've made the changes you requested, and I also updated variable names in the UICollectionView test file for consistency. Please let me know if anything else needs changing.

Issue #1305, opened 2017-06-21
Reactive UIPickerView


MrAsterisco:
**Short description of the issue**:

  The UIPickerView RX extension is limited and doesn't allow binding of the UIPickerView data source.

**Expected outcome**:

  I would expect to have RX extensions to use the UIPickerView just like I'm used with UITableViews.

**What actually happens**:

 The RX extension only provides an `itemSelected` property which, of course, doesn't work alone (because, there is no way to bind an array of items to the picker).

**Self contained code example that reproduces the issue**:

I would expect something like:

```swift
viewModel.items
 .bind(to: pickerView.rx.items) { row, element in
    // return the display name of the element	
 }
 .disposed(by: self.disposeBag)
```

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

3.5.0

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
8.3.3
```

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [ ] yes (which ones)
  - [x] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [x] I have a small code base
  - [ ] I have a significant code base

kzaher:
Hi @MrAsterisco ,

I think it would be theoretically possible, but I would need to investigate that a bit.

kzaher:
Hi @MrAsterisco ,

I think I now remember this one. It wasn't clear how should data model look like. It seems to me that each component could have it's own type, so it's not clear how should generic method that supports multiple components look like.

It also seems that there are 3 ways how one could implement view factory.

```
    // these methods return either a plain NSString, a NSAttributedString, or a view (e.g UILabel) to display the row for the component.
    // for the view versions, we cache any hidden and thus unused views and pass them back for reuse. 
    // If you return back a different object, the old one will be released. the view will be centered in the row rect  
[...]

PR #1318, opened 2017-07-05
[WIP] Reactive UIPickerView extensions

sergdort:
#### Issue

#1305 

#### Changes

* Adds RxPickerViewDataSourceType
* Adds RxPickerViewDataSourceProxy - to forward datasource calls
* Adds UIPickerView extension methods `func setDelegate(_ delegate: UIPickerViewDelegate)` so we can forward delegate calls
* Adds UIPickerView extension `rx.items` to be able to bind `Observable` sequence to UIPickerView (same way as UI{Table/Collection}View)

#### TODO 🚨 :

- [x] Add unit tests 
- [x] Add documentation
- [x] Add changes to the CHANGELING
- [x] Add example 

#### Example

untitled


RxCocoa/iOS/Proxies/RxPickerViewDataSourceProxy.swift
```diff
@@ -0,0 +1,84 @@
...
+    import UIKit
+#if !RX_NO_MODULE
+    import RxSwift
+#endif
+
+let pickerViewDataSourceNotSet = PickerViewDataSourceNotSet()
+
+final class PickerViewDataSourceNotSet: NSObject, UIPickerViewDataSource {
+    func numberOfComponents(in pickerView: UIPickerView) -> Int {
+        rxAbstractMethod()
```
kzaher: I think we should return `0` here.

RxCocoa/iOS/Proxies/RxPickerViewDataSourceProxy.swift
```diff
@@ -0,0 +1,84 @@
...
+    }
+    
+    func pickerView(_ pickerView: UIPickerView, numberOfRowsInComponent component: Int) -> Int {
+        rxAbstractMethod()
```
kzaher: ... and also here.

RxCocoa/iOS/UIPickerView+Rx.swift
```diff
@@ -40,6 +61,20 @@
                 }
             return ControlEvent(events: source)
         }
+        
+        public func items<O: ObservableType,
+                          Adapter: RxPickerViewDataSourceType & UIPickerViewDataSource & UIPickerViewDelegate>(adapter: Adapter)
+            -> (_ source: O)
+            -> Disposable where O.E == Adapter.Element {
+                return { source in
+                    let delegateSubscription = self.setDelegate(adapter)
```
kzaher: I think we should probably also use `subscribeProxyDataSource` to set delegate.

sergdort:
Regarding tests, I think I can take a look on tests for `UITableView`

@kzaher Do you think we should add `Adapters` to the main repo?

RxCocoa/iOS/Proxies/RxPickerViewDataSourceProxy.swift
sergdort: Making both of these methods return 0 will just fallback to the empty UI, right?
Should we at least put the assertions there?

kzaher: I believe we should present nothing when data source isn't set. It seems to me that asserting would crash the app in case somebody first observed some appropriate method and then set data source.

kzaher:
I think we can put one example adapter in RxExample app, but I don't think we should include any adapter in RxCocoa for now. Let's think about putting array of arrays that conform to `CustomStringConvertible` in RxCocoa, but I would like to play with it a bit first once we implement it in RxExample app. We can then easily move it to RxCocoa if we find it useful enough.

RxCocoa/iOS/UIPickerView+Rx.swift
sergdort: Not sure what do you mean here :)
Why this could be an issue?

kzaher: Maybe we should rename `subscribeProxyDataSource` to something else. It has some nice mutation checks. Take a look at "Proxy changed from the time it was first set.\nOriginal...".

I've remembered all of the small checks we are doing in that method but forgot about the main functionality of that method which is subscribing 😂.

I think we can live for now without those small checks, but it would be nice to figure out how to refactor them in future.

RxCocoa/iOS/Proxies/RxPickerViewDataSourceProxy.swift
```diff
@@ -0,0 +1,84 @@
...
+    func numberOfComponents(in pickerView: UIPickerView) -> Int {
+        return 0
+    }
+    
+    func pickerView(_ pickerView: UIPickerView, numberOfRowsInComponent component: Int) -> Int {
+        return 0
+    }
+}
+
+public class RxPickerViewDataSourceProxy
```
kzaher: I think this should be `open`.

RxCocoa/iOS/UIPickerView+Rx.swift
```diff
@@ -40,6 +61,20 @@
                 }
             return ControlEvent(events: source)
         }
+        
+        public func items<O: ObservableType,
```
kzaher: Missing comments.

Tests/RxCocoaTests/UIPickerView+RxTests.swift
```diff
@@ -37,4 +37,137 @@ final class UIPickerViewTests: RxTest {
...
+        case .next(let items):
+            self.items = items
+            pickerView.reloadAllComponents()
+            pickerView.setNeedsLayout()
+            pickerView.layoutIfNeeded() //force layout in test
+        default:
+            break
+        }
+    }
 }
```
kzaher: We are missing `DelegateProxyTests`. Take a look at `DelegateProxyTest+UIKit.swift`.

RxCocoa/iOS/UIPickerView+Rx.swift
kzaher: It would be great to add a simple example in RxExample app. Perhaps something similar to `SimpleTableViewExampleSectionedViewController`.

Tests/RxCocoaTests/UIPickerView+RxTests.swift
sergdort: Don't we already have tests [here](https://github.com/ReactiveX/RxSwift/blob/8735a751eb0714214fe772ee84ae039a99313feb/Tests/RxCocoaTests/DelegateProxyTest%2BUIKit.swift#L74) ?

RxCocoa/iOS/Proxies/RxPickerViewDataSourceProxy.swift
sergdort: Hm. I see that `RxTableViewDataSourceProxy` and `RxCollectionViewDataSourceProxy` are `public`

Tests/RxCocoaTests/UIPickerView+RxTests.swift
kzaher: I think we are missing data source extension test as [this](https://github.com/ReactiveX/RxSwift/blob/8735a751eb0714214fe772ee84ae039a99313feb/Tests/RxCocoaTests/DelegateProxyTest%2BUIKit.swift#L34).

kzaher:
I've been thinking about this. I think we can also add extension to RxCocoa that would enable somebody to do something like this.

```
items
 .bind(to: pickerView.rx.itemTitles) { (row, element) -> String in
    // return the display name of the element	
 }
 .disposed(by: self.disposeBag)

items
 .bind(to: pickerView.rx.itemAttributedTitles) { (row, element) in
    // return the display name of the element	
 }
 .disposed(by: self.disposeBag)

items
 .bind(to: pickerView.rx.items) { (row, element) -> UIView in
    // return the display name of the element	
 }
 .disposed(by: self.disposeBag)
```

`items` can only be simple array of some elements, like we have to UITableView. That should cover > 90% of cases.

If somebody wants more than one component that will require custom adapter.

kzaher:
Hi @sergdort ,

this looks great so far.

I think that the last thing that is missing for this to really nice is `modelSelected` as we have for `UITableView`, and unit tests for those APIs.

We can also add 
```
/// Data source with access to underlying sectioned model.
public protocol SequenceViewDataSourceType {
    /// Returns model at index path.
    ///
    /// In case data source doesn't contain any items when this method is being called, `RxCocoaError.ItemsNotYetBound(object: self)` is thrown.

    /// - parameter indexPath: Model index path
    /// - returns: Model at index path.
    func model(at index: Int) throws -> Any
}
```

It also looks like unit tests are failing.

Want me to add it or do you have time?

sergdort:
Yeah, I'll add it. And I believe I have to a tests for `DataSource`

RamwiseMatt:
Great work on this feature, sergdort! I've been eyeing this for the past couple of days. Is there any way it could be merged into the swift4 branch?

sergdort:
Hi, @RamwiseMatt not sure, seems like we have 2 `swift 4` branches. But I feel like once it gonna be merged into develop we can merge/rebase develop in to `swift 4`.

freak4pc:
I think the Swift 4 version has to be maintained separately until we release RxSwift 4 ... Right now it's in a very beta state (Both this feature and the Swift 4 branch)

Tests/RxCocoaTests/UIPickerView+RxTests.swift
```diff
@@ -37,4 +38,271 @@ final class UIPickerViewTests: RxTest {
...
+        
+        _ = Observable<[Int]>.never().bind(to: pickerView.rx.itemTitles) { _ in
+            return ""
+        }
+        
+        do {
+            let _: Int = try pickerView.rx.model(at: IndexPath(item: 0, section: 0))
+            XCTFail()
+        } catch let error as RxCocoaError {
+            XCTAssertTrue(error.isItemsNotYetBound)
```
sergdort: Not sure if it's not overkill to have assertion on specific error rather just that it throws

sergdort:
Hi, @kzaher I feel like comments addressed :)

freak4pc:
🎉🎉🎉🎉🎉

Awesome job @sergdort !!

reshadf:
Is this PR still worked on? :)

PR #1324, opened 2017-07-12
Fixed a problem where the TableViewWithEditingCommands example died.

magicmon:
If "Allow Arbitrary Loads" is set to No in Info.plist, the "TableViewWithEditingCommands" example will exit as soon as you run it.

So, need to add a catchError inside the getExampleUserResultSet() function.


devxoul:
I think an API should able to emit errors. It'd better to add an exception handler on a view controller, IMO.

Issue #1323, opened 2017-07-12
Deadlock in interaction between `flatMap`, `Single`, and `shareReplayLatestWhileConnected`


dpassage:
Using an admittedly odd combination of `flatMap` and `shareReplayLatestWhileConnected`, I managed to get my code into a deadlock condition.

Here's a short, contrived example:

```swift
    @IBAction func go(_ sender: Any) {
        print("go again!")
        disposables = DisposeBag()

        Observable.just(1)
            .flatMap { (int) -> Observable<[Int]> in
                return Observable.create { (observer) -> Disposable in
                    print("creating!")
                    // in real code, int is really an NSManagedObjectContext,
                    // and the dispatch is really moc.perform with a fetch
                    DispatchQueue.global().async {
                        observer.onNext([int])
                    }
                    sleep(1)
                    return Disposables.create()
                }
            }
            .shareReplayLatestWhileConnected()
            .subscribe { (e) in
                print(e)
            }
            .disposed(by: disposables)
    }
```

Put that in a single-VC iOS sample app, hook it up to a button, and hit go!

The result will be that the main thread is holding the lock in `ShareReplay1WhileConnected.subscribe()` and waiting on the lock in `MergeSink.on` to deliver the `.completed` event coming out of the `just(1)` at the top.

Meanwhile, the background dispatch queue will be holding the lock in `MergeSinkIter.on` (which is the same as in the parent `MergeSink` object) while waiting for the lock in `ShareReplay1WhileConnectedConnection.on`. to deliver the little array.

In real, non-contrived code, `Observable.just(1)` is actually a `ReplaySubject` which delivers an `NSManagedObjectContext` wrapped into a `Single`, and the dispatch block is really a call to `NSManagedObjectContext.perform` in which we do a fetch. In the real code, I was getting the deadlock about one time in 5 at my desk. The `sleep` above is just there to make it happen reliably; I'm sure `NSManagedObjectContext.perform` is actually doing a bunch of work here, plus there are a bunch of other threads in my real app.

It appears that the root cause is the background thread sneaking in to deliver the results of the computation _between_ the `next` and `completed` events coming out of the `.just`.

Once I found it, I was able to work around it with proper application of `.observeOn`, to ensure that the `connect` down in the `ShareReplay1WhileConnected` returns immediately instead of waiting for the events. But it was a bear to find.

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

RxSwift 3.5.0

[...]

PR #1327, opened 2017-07-13
Fixes deadlock with `shareReplayWhileLatest`. #1323

kzaher:
Hi @dpassage,

I think this should solve #1323 . Could you also verify that?


freak4pc:
I can confirm, testing with @dpassage's example that this solves the Main Thread deadlock

sergdort:
R.I.P `ShareReplay1 ` 😢

freak4pc:
He was a brave warrior.

dpassage:
Yes, fix works for me in the sample code; I'll try it in my production code later today. Thanks, @kzaher !

dpassage:
Works in the production code, too!

I approve of this pull request.

seems legit

PR #1330, opened 2017-07-14
added binding observer isFirstResponder on UIControl.rx

DragonCherry:
I found it's necessary when some condition should invoke a control to be a responder.


DragonCherry:
I will find another way to handle such situation..

PR #1336, opened 2017-07-14
Add image sink to NSButton and NSComboBox extensions

apparentsoft:
Copied implementation from NSImageView+Rx


RxCocoa/macOS/NSComboBox+Rx.swift
```diff
@@ -0,0 +1,105 @@
...
+    public var delegate: DelegateProxy {
+        return RxComboBoxDelegateProxy.proxyForObject(base)
+    }
+
+    /// Reactive wrapper for `indexOfSelectedItem` property.
+    public var indexOfSelectedItem: ControlProperty<Int> {
+        let delegate = RxComboBoxDelegateProxy.proxyForObject(base)
+
+        let source = Observable.deferred { [weak comboBox = self.base] in
+            delegate.selectionIndexSubject.startWith(comboBox?.indexOfSelectedItem ?? -1)
```
kzaher: This should return empty sequence in combobox is `nil`.

Tests/RxCocoaTests/NSComboBox+RxTests.swift
```diff
@@ -0,0 +1,66 @@
...
+        var completed = false
+
+        autoreleasepool {
+            let comboBox = NSComboBox()
+            let delegate = ComboBoxDelegate()
+            comboBox.delegate = delegate
+            var rxDidChange = false
+
+            _ = comboBox.rx.indexOfSelectedItem
+                .skip(1) // Initial value
```
kzaher: This should test both initial value is correct and changing of value works.

Tests/RxCocoaTests/NSComboBox+RxTests.swift
```diff
@@ -0,0 +1,66 @@
...
+//  Created by Jacob Gorban on 07/17/17.
+//  Copyright © 2017 Krunoslav Zaher. All rights reserved.
+//
+
+import RxSwift
+import RxCocoa
+import AppKit
+import XCTest
+
+final class NSComboBoxTests: RxTest {
```
kzaher: This is missing `DelegateProxyTest+Cocoa.swift` tests.

freak4pc:
Hey @apparentsoft ! Thank you very much for the help here, 
Did you have any time to go through @kzaher's comments here? Let us know if you need any help pushing this through! 👍

apparentsoft:
Hi @freak4pc. I didn't have time yet. I was on vacation for 2 weeks and I'm just back and swamped with work, as you might imagine. It is also my first (and only, so far) contribution to RxSwift, so it already took me quite a while to get it compile and submit right, and I was basing most of the code on existing code anyway. If you have the capacity to help with these changes, I'll appreciate this. If not, I'll get to it but I can't commit to as to when this will happen. 
Thank you for caring.

freak4pc:
Hey @apparentsoft, First of all I hope you had an awesome vacation :) 

There's really no rush on this as far as I can see and I wouldn't want to take this off your hand, but - if I'll see I have some time to tackle this I will, but I think i'll leave it open for a week or two before to perhaps see if you have any time for it :) 

Highly appreciate the time you point into this !

Tests/RxCocoaTests/NSComboBox+RxTests.swift
apparentsoft: @kzaher I was basing this test on your NSTextField+RxTests.swift file, which doesn't test for initial value. It's also pretty meaningless. The test is named testComboBox_ComboBoxSelectionDidChange_ForwardsToDelegates and doesn't actually check the values of the selected items (nor are they even set in the test).

apparentsoft:
@kzaher I added the other test that you asked and the empty case.

apparentsoft:
@freak4pc I've made most of the changes and commented on another one. Krunoslav doesn't seem to respond in the last week, so if you've got reviewing/merging privileges as well, maybe you'll take a look.

freak4pc:
@apparentsoft I don't mind reviewing later but I don't feel comfortable merging here. 
@kzaher will still have to green light this

apparentsoft:
@kzaher Hi, Krunoslav. Will you take a look at it and incorporate into the repository. Should also work with Swift 4 (unless the whole architecture for these things changed). I went over your changes and fixed where needed.

freak4pc:
Hey Jacob @apparentsoft. This PR is a couple of years old and it seems to have very little community interesting. Do you still want to pursue this or it's OK to close? Thanks.

apparentsoft:
We can close it, I guess.

-- Jacob Gorban
Apparent Software
Twitter: @jacobgorban
Facebook: http://www.facebook.com/apparentsoft

PR #1341, opened 2017-07-18
rxcocoa build fix on xcode 9

anonymcek:
I am not sure about other classes and their visibility but the RxPickerViewSequenceDataSource being private doesn't work on XCode 9. Actually I don't understand how it could build at all without being public.

### DIFF
--- a/RxCocoa/iOS/DataSources/RxPickerViewAdapter.swift
+++ b/RxCocoa/iOS/DataSources/RxPickerViewAdapter.swift
@@ -31,12 +31,12 @@ private class RxPickerViewArrayDataSource<T>: NSObject, UIPickerViewDataSource, 
     }
 }
 
-public class RxPickerViewSequenceDataSource<S: Sequence>
+class RxPickerViewSequenceDataSource<S: Sequence>
     : RxPickerViewArrayDataSource<S.Iterator.Element>
     , RxPickerViewDataSourceType {
-    public typealias Element = S
+    typealias Element = S
 
-    public func pickerView(_ pickerView: UIPickerView, observedEvent: Event<S>) {
+    func pickerView(_ pickerView: UIPickerView, observedEvent: Event<S>) {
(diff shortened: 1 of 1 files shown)


RxCocoa/iOS/DataSources/RxPickerViewAdapter.swift
```diff
@@ -31,12 +31,12 @@ private class RxPickerViewArrayDataSource<T>: NSObject, UIPickerViewDataSource,
     }
 }
 
-private class RxPickerViewSequenceDataSource<S: Sequence>
+public class RxPickerViewSequenceDataSource<S: Sequence>
```
kzaher: Hi @anonymcek ,

I don't think any of these elements should be `public`. 

Can you please try just removing `private` identifier from `RxPickerViewSequenceDataSource` and from `RxPickerViewArrayDataSource`?

anonymcek: `RxPickerViewArrayDataSource` can stay public, should I remove it anyway?

kzaher: I don't think that anyone understands anymore what is by design and what works by pure accident here so I think it would be safer to also remove `private` from `RxPickerViewArrayDataSource`.

sergdort:
I wonder if it's because of new `Swift` access control changes of `private` and `fileprivate`

I feel like it should work with `fileprivate`

anonymcek:
it is referenced from `RxUIPickerView+Rx` so it wouldn't be visible with `fileprivate` either

anonymcek:
does anyone have a clue what is wrong with that tests?

sergdort:
Hm this is so weird.. Because we do not reference `RxPickerViewSequenceDataSource ` from `UIPickerView+Rx` ¯\_(ツ)_/¯

PR #1345, opened 2017-07-19
	Improves locking behavior of `merge` and `switch` operators.

kzaher:
Hi @MartinMoizard,

I've found an issue in an optimized array version of `Observable.merge`. Could you please verify this solves your issue?


MartinMoizard:
Hi @kzaher,

Thanks a lot for the quick fix, I tested it and it works! 👏

PR #1349, opened 2017-07-21
Remove nested subscribe "tip"

passsy:
I removed the section about nesting subscribe because it leads to more problems than it fixes.

### Nested subscribes are ok:

- **The lifecycle of the Observables can be different.** Disposing the text change Observable doesn't mean I wan't to cancel the network request in one of my services. The lifecycle of my service and the ViewController can be totally different. This also means I have different disposeBags
- **Handling errors is different for Observables which never complete and Observables which complete/error normally. Don't combine them.** When flatmapping the network request to the text change stream one must be very careful to catch all possible errors. Worst case scenario: The network request fails and the textField stops emitting text changes.

### DIFF
--- a/Documentation/Tips.md
+++ b/Documentation/Tips.md
@@ -22,29 +22,3 @@ extension ObservableType where E: MaybeCool {
   * Rx operators are as general as possible, but there will always be edge cases that will be hard to model. In those cases you can just create your own operator and possibly use one of the built-in operators as a reference.
 
   * Always use operators to compose subscriptions.
-
-  **Avoid nesting subscribe calls at all cost. This is a code smell.**
-
-  ```swift
-  textField.rx.text.subscribe(onNext: { text in
-      performURLRequest(text).subscribe(onNext: { result in
-          ...
-      })
-      .addDisposableTo(disposeBag)
-  })
-  .addDisposableTo(disposeBag)
-  ```
-
-  **Preferred way of chaining disposables by using operators.**
-
-  ```swift
-  textField.rx.text
-      .flatMapLatest { text in
-          // Assuming this doesn't fail and returns result on main scheduler,
-          // otherwise `catchError` and `observeOn(MainScheduler.instance)` can be used to
-          // correct this.
-          return performURLRequest(text)
-      }
-      ...
-      .addDisposableTo(disposeBag) // only one top most disposable
-  ```


freak4pc:
@peagasilva Thanks for your note and opinion. This is an interesting discussing but I, personally, don't totally agree with removing this note. 

I think it needs to be revised from "at all cost", and I do think that 99% of the time it is an unwanted code smell and not a smart decision. The "general" use case does not call of them and most of the time there is a better way for doing things then performing these sort of side-effect on a subscriber's onNext / onEvent. 

Getting rid of this note completely is something I wouldn't feel comfortable doing, personally. 

I would feel better with "Nested subscribers should generally be avoided / proceed with caution" kind of thing.

sergdort:
Hi guys! Very interesting discussion indeed =)

In my opinion nested subscribes are not OK! They are essentially "nested callbacks", and this is one of the problems that Rx is trying to solve.

@passsy I feel like in your argument here:

> When flatmapping the network request to the text change stream one must be very careful to catch all possible errors. Worst case scenario: The network request fails and the textField stops emitting text changes.

the problem is not in flat mapping but wrong error handling strategy.
I think when it comes to UI everything should be modelled as "infinite sequence loop" as an example please take a look [RxFeedback](https://github.com/kzaher/RxFeedback) or more examples in the main repo 

Regarding of:

> The lifecycle of the Observables can be different.

I feel like you have wrong assumption of `Observable`. 
Please correct me if I'm wrong but in my opinion`Observable` does not have a lifecycle, it's just a type that encapsulates a side effects that gonna be hapend on the subscription. 

But yeah, would love to hear other people input =)

freak4pc:
@sergdort I think lifecycle is a good definition for it as much as a Promise's lifecycle is tied to it's resolve/reject calls. 

When an Observable emits error or completed events, it's lifecycle is eventually over as an object, regardless if it was deallocated at that point or not IMO

kzaher:
Hi guys,

if you subscribe to an observable sequence that has finite duration, like the one that wraps HTTP command calls and don't add anything to `disposeBag` but just do `_ = sequence.subscribe()` there won't be any resource leakage. 
If observable sequence is implemented correctly, it will dispose all resources on `completed` or `error` events but for as long as HTTP call is alive, those resources will be allocated. That could be a valid behavior if some cases.

Having said that ...

> The lifecycle of the Observables can be different. Disposing the text change Observable doesn't mean I wan't to cancel the network request in one of my services. The lifecycle of my service and the ViewController can be totally different. This also means I have different disposeBags

... if that is the case it is better to create your own operator that doesn't forward `dispose` call.

Something like

```swift
extension Observable {
    func mutationOperation() -> Observable<E> {
        return Observable.create { observer in
            _ = self.subscribe(observer)
            return Disposables.create()
        }
    }
}

  textField.rx.text
       .flatMap { text in
           return performURLRequest(text)
               .mutationOperation()
       }
```

> Handling errors is different for Observables which never complete and Observables which complete/error normally. Don't combine them. When flatmapping the network request to the text change stream one must be very careful to catch all possible errors. Worst case scenario: The network request fails and the textField stops emitting text changes.

There are ways how you can ask the compiler to help you with this. Please take a look at `Driver`. One it it's purposes is to help you get compile time errors in case you haven't considered error handling.

Everybody is free to use Rx in any way they but calling subscribe a lot of times is a code smell to me.

freak4pc:
Closing due to inactivity and @kzaher's answer. Rx is Rx and the way to choose it should be up to the user, but this is still left as a fair warning that is relevant for 99% of common use cases

passsy:
Rx is Rx, right. What surprises me is that other Rx implementations RxJava, RxJs, Rx.Net don't have such a warning.

kzaher:
Hi @passsy ,

Every repository contains it's maintainers opinions about how to use it. They can be either explicit and can be found in the docs, or implicit and found in the API design and code.

Even tough the basic concept is the same there will always be slight variations in opinions between different teams working on the same thing.

I can't point you to docs in some other repository that give the same advice but I can suggest you to take a look at code in all of the other Rx repositories. 
In 99.99% of times subscribe is only called inside operator code (maybe there are some subscribe convenience or testing extensions where it's also being called).

PR #1354, opened 2017-07-27
Rx blocking remove duplicate code

sgleadow:
I've just discovered `asBlocking()` and it's so useful for unit testing observables, thanks.

At a first glance, the code in each of `toArray`, `first`, `last` and `single` is really similar. This PR combines them all into one to share the `toArray` implementation.

I initially thought testing the `predicateVals` callbacks was an unnecessary implementation detail, however, I've now realised it's basically testing that the sequence is terminated as soon as the conditions fail, so it wont keep observing forever.

UPDATE: I added a general purpose array converter that has the same behaviour around terminating as soon as it's got the values it needs, or breaks an expected precondition (eg. `single`), but it can still share the same code.

~Do you think `asBlocking()` is used on eg. infinite or ongoing observables, where the lazy evaluation is important?~

~Assuming that's the case, I'm playing with some ways of keeping the minimal code and also not evaluating the entire sequence, but haven't nailed it yet, in which case this PR might not be useful.~

### DIFF
--- a/RxBlocking/BlockingObservable+Operators.swift
+++ b/RxBlocking/BlockingObservable+Operators.swift
@@ -67,50 +67,7 @@ extension BlockingObservable {
     ///
     /// - returns: First element of sequence. If sequence is empty `nil` is returned.
     public func first() throws -> E? {
-        var element: E?
-
-        var error: Swift.Error?
-
-        let d = SingleAssignmentDisposable()
-
-        defer {
-            d.dispose()
-        }
-        
-        let lock = RunLoopLock(timeout: timeout)
-
-        lock.dispatch {
-            let subscription = self.source.subscribe { e in
-                if d.isDisposed {
-                    return
-                }
-
-                switch e {
-                case .next(let e):
-                    if element == nil {
-                        element = e
-                    }
-                    break
-                case .error(let e):
-                    error = e
-                default:
-                    break
-                }
-
-                d.dispose()
-                lock.stop()
-            }
-
-            d.setDisposable(subscription)
-        }
-
-        try lock.run()
-
-        if let error = error {
-            throw error
-        }
-
-        return element
+        return try toArray().first
     }
 }
 
@@ -121,47 +78,7 @@ extension BlockingObservable {
     ///
     /// - returns: Last element in the sequence. If sequence is empty `nil` is returned.
     public func last() throws -> E? {
-        var element: E?
-
-        var error: Swift.Error?
-
-        let d = SingleAssignmentDisposable()
-
-        defer {
-            d.dispose()
-        }
-        
-        let lock = RunLoopLock(timeout: timeout)
-
-        lock.dispatch {
-            let subscription = self.source.subscribe { e in
-                if d.isDisposed {
-                    return
-                }
-                switch e {
-                case .next(let e):
-                    element = e
-                    return
-                case .error(let e):
-                    error = e
-                default:
-                    break
-                }
-
-                d.dispose()
-                lock.stop()
-            }
-
-            d.setDisposable(subscription)
-        }
-        
-        try lock.run()
-        
-        if let error = error {
-            throw error
-        }
-        
-        return element
+        return try toArray().last
     }
 }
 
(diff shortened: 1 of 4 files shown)


sgleadow:
I've now used the common implementation to easily extract a `toError() -> Error?` function, which is the reason I was looking at this code in the first place.

In unit tests where I want to check error scenarios, being able to return and assert on the error seems cleaner than using one of the existing functions and `do`/`catch`.

This may be pushing into some patterns and approaches that I should have raised in an issue for discussion first (but I already have an extension in my own codebase for the error extraction, so thought I'd contribute back)

freak4pc:
I personally like this idea. Didn't have time to dig into if this is causing any degradation, but you should fix the test failures for this to be considered

sgleadow:
@freak4pc thanks and done.

There's a warning about not adding to the CHANGELOG - I can do this, but it feels like I'm not in a good position to, is it usually done by a core maintainer on release?

freak4pc:
Hey @sgleadow , This is your branch, feel free to add a CHANGELOG entry here :)

Tests/RxBlockingTests/Observable+BlockingTest.swift
```diff
@@ -17,11 +17,11 @@ class ObservableBlockingTest : RxTest {
 
 extension ObservableBlockingTest {
     func testToArray_empty() {
-        XCTAssert(try! Observable<Int>.empty().toBlocking().toArray() == [])
+        XCTAssertEqual(try! Observable<Int>.empty().toBlocking().toArray(), [])
```
freak4pc: Why the styling change here? 
Moving from XCTAssert with == to XCTAssertEqual ?

kzaher:
Hi @sgleadow ,

I think blocking code could be improved indeed. I would like to separate this and first refactor what we have.

Could you please focus this PR on refactoring first?

I think the internal refactorings make sense but I don't think that we should add `toError` interface that looks this way.

Tests/RxBlockingTests/Observable+BlockingTest.swift
sgleadow: The code was already was mixing both styles, so I guess one is better, but more importantly, you seem to get much better error messages when comparing objects rather than comparing the result of a boolean operation... so you get _[1,2,3] is not equal to [3,4,5]` with XCTAssetEqual rather than _assert failed_ or _false is not equal to true_ with XCTAssert.

I wasn't 100% sure of the exact difference in wording in Xcode, so I grabbed a screenshot.

screen shot 2017-07-31 at 7 54 05 am

sgleadow:
@kzaher Thanks for the feedback.

I've removed the `toError` function and simplified the implementation now it's not required to capture the error this way.

I'll raise a separate issue to discuss retrieving the error from a blocking observable separately.

kzaher:
@sgleadow On top of CHANGELOG.md there is a master section that represents upcoming release pending changes.

RxBlocking/BlockingObservable+Operators.swift
```diff
@@ -182,61 +60,76 @@ extension BlockingObservable {
...
+            return elements.first
+        default:
+            throw RxError.moreThanOneElement
+        }
+    }
+}
+
+extension BlockingObservable {
+    fileprivate func convertToArray(max: Int? = nil, predicate: @escaping (E) throws -> Bool = { _ in true }) throws -> [E] {
+        var elements: [E] = Array<E>()
```
kzaher: Interesting, I think `var elements = [E]()` would also work, but ok :)

RxBlocking/README.md
```diff
@@ -18,6 +18,15 @@ extension ObservableType {
...
     public func last() throws -> E? {}
 }
+
+extension Observableype {
+    public func single() throws -> E? {}
+    public func single(_ predicate: @escaping (E) throws -> Bool) throws -> E? {}
+}
+
+extension ObservableType {
+    public func toError() -> Swift.Error? {}
```
kzaher: This is still noted even though this PR doesn't contain these changes.

RxBlocking/README.md
```diff
@@ -18,6 +18,15 @@ extension ObservableType {
 extension ObservableType {
     public func last() throws -> E? {}
 }
+
+extension Observableype {
```
kzaher: I've just realized all of these extensions are on BlockingObservable vs ObservableType :)

RxBlocking/README.md
sgleadow: Oops, sorry, I missed that.

PR #1356, opened 2017-07-31
Add first operator to ObservableType.

viniciussoares:
- **first**: emits only the very first item emitted by the ObservableType,  or a default item if the ObservableType completes without emitting anything.
- **firstOrError**: emits only the very first item emitted by the ObservableType or  throws a `RxError.noElements` if the Observable is empty.

**Use case:**

They are very useful when you need to handle data from [multiple sources](http://blog.danlew.net/2015/06/22/loading-data-from-multiple-sources-with-rxjava/).


kzaher:
Hi @viniciussoares ,

I see what you want to accomplish. I'm little concerned about a few things.

* `first*` method works in entirely different way then method on regular sequences which returns an optional. 
* we need to add two more methods to cover this use case.
* we are creating overloads to anticipate all possible cases what can happen when there isn't an element vs providing a way for user to decide in a more flexible way. 
* `first` extension on `BlockingObservable` already returns `E?`.

IMHO we should either return `Single<Element?>` or `Maybe<Element>` as a result of `first`. I'm probably a bit more inclined to `Single<Element?>` since that's the form of API that normal sequence uses.

I think that we could potentially add an interface that has this form, so one can do:
```
return xs.first().map { try $0 ?? { throw RxError.noElements }() }
```
or 
```
return xs.first().map { $0 ?? defaultValue }
```
or
```
return xs.first().flatMap { $0.map(Single.just) ?? defaultFallbackSingle }
```

viniciussoares:
Hello @kzaher ,

Yeah, now that you pointed it out, it makes total sense to go with Single<Element?>, have only one operator to cover this use case and maintain a clean and consistent api.

I'll update the PR :)

freak4pc:
Hey @viniciussoares, just wanted to touch base and see if you're still tackling this or need any assistance?

viniciussoares:
Hi @freak4pc

I've merged the operators and now `first` returns a optional.

The only problem i had was with the tests, i couldn't find a way to assert optional values.

Tests/RxSwiftTests/PrimitiveSequenceTest.swift
```diff
@@ -1046,6 +1046,131 @@ extension PrimitiveSequenceTest {
...
+        let scheduler = TestScheduler(initialClock: 0)
+
+        let xs = scheduler.createHotObservable([
+            next(150, 1),
+            completed(250),
+            error(260, testError)
+            ])
+
+        let res: TestableObserver<Int> = scheduler.start { () -> Observable<Int> in
+            let single: Single<Int> = xs.first().map { $0 ?? -1 }
```
kzaher: I think we would should make this `let single: Single<Int?> = xs.first()`.

Tests/RxSwiftTests/PrimitiveSequenceTest.swift
```diff
@@ -1046,6 +1046,131 @@ extension PrimitiveSequenceTest {
...
+
+        let xs = scheduler.createHotObservable([
+            next(150, 1),
+            next(210, 2),
+            completed(250),
+            error(260, testError)
+            ])
+
+        let res = scheduler.start { () -> Observable<Int> in
+            let single: Single<Int> = xs.first().map { $0 ?? -1 }
```
kzaher: The same in other tests.

Tests/RxSwiftTests/PrimitiveSequenceTest.swift
```diff
@@ -1046,6 +1046,131 @@ extension PrimitiveSequenceTest {
...
+
+        XCTAssertEqual(res.events, [
+            error(210, testError)
+            ])
+
+        XCTAssertEqual(xs.subscriptions, [
+            Subscription(200, 210)
+            ])
+    }
+}
```
kzaher: We are missing 
```
#if TRACE_RESOURCES
        func testFirstReleasesResourcesOnComplete()
        func testFirstReleasesResourcesOnError() 
#endif
```

Tests/RxSwiftTests/PrimitiveSequenceTest.swift
viniciussoares: I've tried to use optionals, but i get this error `Value of optional type '[Recorded<Event<Int?>>]' not unwrapped; did you mean to use '!' or '?'?`

kzaher: I think you might take a look at how we've created
```
public func XCTAssertEqual<T: Equatable>(_ lhs: [Recorded<Event<T>>], _ rhs: [Recorded<Event<T>>], file: 
```
extensions and add
```
public func XCTAssertEqual<T: Equatable>(_ lhs: [Recorded<Event<T?>>], _ rhs: [Recorded<Event<T?>>], file: 
```
to `XCTest+AllTests.swift`.

PR #1360, opened 2017-08-03
Use NSControl.StateValue in NSButton.rx.state

tarunon:
Use NSControl.StateValue instead of Int, NSButton.state's type was changed.

### DIFF
--- a/RxCocoa/macOS/NSButton+Rx.swift
+++ b/RxCocoa/macOS/NSButton+Rx.swift
@@ -20,25 +20,31 @@ extension Reactive where Base: NSButton {
         return controlEvent
     }
 
-    /// Reactive wrapper for `state` property`.
-    public var state: ControlProperty<Int> {
-        return NSButton.rx.value(
-            base,
-            getter: { control in
-                #if swift(>=4.0)
-                    return control.state.rawValue
-                #else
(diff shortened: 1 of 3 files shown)


RxCocoa/macOS/NSButton+Rx.swift
```diff
@@ -20,25 +20,31 @@ extension Reactive where Base: NSButton {
...
 
-    /// Reactive wrapper for `state` property`.
-    public var state: ControlProperty<Int> {
-        return NSButton.rx.value(
-            base,
-            getter: { control in
-                #if swift(>=4.0)
-                    return control.state.rawValue
-                #else
+    #if swift(>=4.0)
```
tarunon: [Q] RxSwift 4.0 will support swift 3.x ? I think can remove this term.

kzaher: @tarunon I don't think we'll support <= 3.1. Potentially 3.2 but we'll need to evaluate what would it take to do that first.

tarunon: @kzaher Thank you answer.
If support swift3.2, we will need this term.
I think it's good to keep it now, and if you decide don't support 3.2, please remove this term.

PR #1367, opened 2017-08-07
Enumerated

kzaher:
We would like to clean up our public interface a bit.

Instead of having:
* `mapWithIndex`
* `flatMapWithIndex`
* `skipWhileWithIndex`
* `takeWhileWithIndex`

we would have single `enumerated` operator equivalent that could be combined with:
* `map`
* `flatMap`
* `skipWhile`
* `takeWhile`

We will mark those operators deprecated in 4.0. I'm curious, is anyone extensively using them?

### DIFF
--- a/RxSwift/Deprecated.swift
+++ b/RxSwift/Deprecated.swift
@@ -35,6 +35,75 @@ extension Observable {
     }
 }
 
+extension ObservableType {
+    /**
+
+    ** @available(*, deprecated, message: "Please use enumerated().map()", renamed: "enumerated().map()") **
+
+     Projects each element of an observable sequence into a new form by incorporating the element's index.
+
+     - seealso: [map operator on reactivex.io](http://reactivex.io/documentation/operators/map.html)
+
+     - parameter selector: A transform function to apply to each source element; the second parameter of the function represents the index of the source element.
+     - returns: An observable sequence whose elements are the result of invoking the transform function on each element of source.
+     */
+    public func mapWithIndex<R>(_ selector: @escaping (E, Int) throws -> R)
+        -> Observable<R> {
+        return enumerated().map { try selector($0.element, $0.index) }
+    }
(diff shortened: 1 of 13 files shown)


freak4pc:
I'm using mapWithIndex here and there. 
So your point is instead we will have

Instead of this:
```swift
stream
    .mapWithIndex { index, item in }
```

This?:
```swift
stream
    .enumerated()
    .map { index, item in }
```

RxSwift/Deprecated.swift
```diff
@@ -35,6 +35,75 @@ extension Observable {
...
+     Projects each element of an observable sequence into a new form by incorporating the element's index.
+
+     - seealso: [map operator on reactivex.io](http://reactivex.io/documentation/operators/map.html)
+
+     - parameter selector: A transform function to apply to each source element; the second parameter of the function represents the index of the source element.
+     - returns: An observable sequence whose elements are the result of invoking the transform function on each element of source.
+     */
+    public func mapWithIndex<R>(_ selector: @escaping (E, Int) throws -> R)
+        -> Observable<R> {
+        return enumerated().map { try selector($0.element, $0.index) }
```
freak4pc: Do you think this should be flipped? index first, element second, like Swift's `enumerated()` method on Sequences ?

kzaher:
@freak4pc yes

RxSwift/Deprecated.swift
kzaher: I'm not sure what do you mean? Are you saying our `enumerated()` api is wrong?

https://github.com/ReactiveX/RxSwift/pull/1367/files#diff-caa8d1228e1aa3011315687e5dc8d4e2R19

freak4pc: No, it looks fine, I misread the implementation :)
Looks great!

PR #1368, opened 2017-08-07
Add rx.touchDown to UIButton

pietrobasso:
**Proposal**:
Currently, only `rx.tap` is implemented for `UIButton`. The _tap_ terminology is too vague. 
I think it's better to implement all UIControlEvents like `touchDown, touchUpInside, touchUpOutside` etc.

### DIFF
--- a/RxCocoa/iOS/UIButton+Rx.swift
+++ b/RxCocoa/iOS/UIButton+Rx.swift
@@ -15,9 +15,89 @@ import UIKit
 
 extension Reactive where Base: UIButton {
     
+    public var touchDown: ControlEvent<Void> {
+        return controlEvent(.touchDown)
+    }
+    
+    public var touchDownRepeat: ControlEvent<Void> {
+        return controlEvent(.touchDownRepeat)
+    }
+    
+    public var touchDragInside: ControlEvent<Void> {
+        return controlEvent(.touchDragInside)
+    }
+    
+    public var touchDragOutside: ControlEvent<Void> {
+        return controlEvent(.touchDragOutside)
+    }
+    
+    public var touchDragEnter: ControlEvent<Void> {
+        return controlEvent(.touchDragEnter)
+    }
+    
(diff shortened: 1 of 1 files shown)


kzaher:
Hm ... I agree that `rx.tap` is maybe vague but I don't think it is hard to figure out what it does.

If we were to include all of these events I would argue that they are added to wrong class. 
All of these events are available for any control since `rx.controlEvent` is also.

If we were to put this to all controls I don't think that `control.rx.touchUpInside` is more vague than `control.rx.controlEvent(.touchUpInside)`.

To be honest, I've rarely had to use `control.rx.controlEvent`. Do you use anything except `rx.tap` often and in what scenarios?

pietrobasso:
I wasn't aware of the existence of `rx.controlEvent`, I'm sorry for that. 

You're completely right to say that `control.rx.touchUpInside` is more vague than `control.rx.controlEvent(.touchUpInside)`, and I agree with you. 

The thing is that the `tap` terminology is, in fact, a combination of two UIControlEvents (touchDown + touchUpInside). That's the main reason why a UIControlEvent named `tap` doesn't exist, and that's why I consider it too vague.

My pull request can be closed but I think it's important to find a logical solution to the `tap` terminology.

freak4pc:
@pietrobasso You won't be able to deprecate `tap` as it's too heavily used. I would also personally be against changing what it currently does as there's also a used consensus among the community. 

But, to be honest, I have no issue with it being "tied" to `touchUpInside`. It does "the right thing" 99% of the time, and saves a lot of boilerplate. 

I would argue perhaps it needs to be `touchDown` and not `touchUpInside`, as the former is the actual tap event, and "touch up" is actually the release event. But that's sort of nitpicking :) 

I'm going to close this PR as discussed here, but if you want to keep discussing this and feel this is truly and issue, it would be good opening a new Issue and we can continue the discussion there. 

Thanks !

PR #1375, opened 2017-08-10
Fix two Xcode 9 beta 5 deprecation warnings

irace:


irace:
Hi!

OK, cool. Anything else I need to do?

kzaher:
I guess you can change urls in our travis file so unit tests on this PR are passing :))

PR #1379, opened 2017-08-15
Implement Generics parameter in DelegateProxy

tarunon:
Hello there, 
This proposal is first implementation of #1287, added generics parameter on DelegateProxy.

## Motivation
I wish avoid to use force casting when implement DelegateProxy subclasses.
```swift
// as is
class func setCurrentDelegate(_ delegate: AnyObject?, toObject object: AnyObject) {
  let scrollView: UIScrollView = castOrFatalError(object)
  scrollView.delegate = castOptionalOrFatalError(delegate)
}

// to be
override class func setCurrentDelegate(_ delegate: UIScrollViewDelegate?, toObject object: ParentObject) {
  object.delegate = delegate
}
```
Every DelegateProxy subclass had same force casting implementation, I have thought can move these force casting into DelegateProxy core implementation, and use Generics parameter instead.

### DelegateProxy
- The new `DelegateProxy` have 2 generics parameter that `ParentObject` and `Delegate`. 
- Made a workaround protocol `DelegateProxyBase` for protocol extension conformance.

### DelegateProxyType
- Moved some of method that affected by generics parameter to `DelegateProxyBase`.

### DelegateProxyFactory
- Added `static func sharedFactory(for proxyType: DelegateProxy.Type) -> DelegateProxyFactory` because generics class cannot define static stored property.
- Use `DelegateProxySubclass.prepareForDelegateProxy` instead of `DelegateProxyFactory.extended`.


RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -144,41 +182,42 @@ extension DelegateProxyType {
...
-        let currentDelegate: AnyObject? = Self.currentDelegateFor(object)
-
-        if currentDelegate !== proxy {
-            proxy.setForwardToDelegate(currentDelegate, retainDelegate: false)
-            assert(proxy.forwardToDelegate() === currentDelegate)
-            Self.setCurrentDelegate(proxy, toObject: object)
-            assert(Self.currentDelegateFor(object) === proxy)
-            assert(proxy.forwardToDelegate() === currentDelegate)
+        let currentDelegate = self.currentDelegateFor(object)
+        let delegateProxy = unsafeDowncast(proxy, to: self)
```
tarunon: I used `unsafeDowncast` instead of common force casting.
Let's think about `UITableView`.
We can get instance of `DelegateProxy` using `RxScrollViewDelegateProxy.proxyForObject(scrollView)` or `RxTableViewDelegateProxy.proxyForObject(tableView)` and if `scrollView` and `tableView` are same object, then DelegateProxy instances should be same object.
In this case, `proxyForObject` should cast two types that has different generics parameter.
Common force casting is failed, but `unsafeDowncast` is succeed.

RxCocoa/iOS/Proxies/RxScrollViewDelegateProxy.swift
```diff
@@ -14,24 +14,18 @@ import RxSwift
 import UIKit
 
 /// For more information take a look at `DelegateProxyType`.
-public class RxScrollViewDelegateProxy
-    : DelegateProxy
-    , UIScrollViewDelegate
-    , DelegateProxyType {
+open class RxScrollViewDelegateProxy<P: UIScrollView>
+    : DelegateProxy<P, UIScrollViewDelegate>
```
tarunon: I think the `Delegate` should fix in a type declared `DelegateProxyType`.
`Delegate` affect `DelegateProxy.installForwardDelegate` implementation, and of course we can call `DelegateProxy.installForwardDelegate` from upper casted parent object.
```swift
// e.g.
let tableView = UITableView()
RxScrollViewDelegateProxy.installForwardDelegate(myScrollViewDelegate, retainDelegate: false, onProxyForObject: tableView)
tableView.rx.delegate.forwardToDelegate() // -> It is myScrollViewDelegate.
```

We know same pattern in UIKit, it seems strange.
```swift
let tableView = UITableView()
(tableView as UIScrollView).delegate = myScrollViewDelegate
tableView.delegate // -> myScrollViewDelegate, but it behave as UITableViewDelegate
```

RxCocoa/Common/DelegateProxy.swift
```diff
@@ -21,18 +21,20 @@ var dataSourceAssociatedTag: UnsafeRawPointer = UnsafeRawPointer(UnsafeMutablePo
 /// Base class for `DelegateProxyType` protocol.
 ///
 /// This implementation is not thread safe and can be used only from one thread (Main thread).
-open class DelegateProxy : _RXDelegateProxy {
-
+open class DelegateProxy<P: AnyObject, D: NSObjectProtocol>: _RXDelegateProxy, DelegateProxyBase {
```
kzaher: Why `DelegateProxyBase`. Maybe `DelegateProxyProtocol` is more idiomatic to Swift?

RxCocoa/Common/DelegateProxy.swift
```diff
@@ -201,18 +203,45 @@ open class DelegateProxy : _RXDelegateProxy {
...
+    ///
+    /// Objects can have multiple delegate properties.
+    ///
+    /// Each delegate property needs to have it's own type implementing `DelegateProxyType`.
+    ///
+    /// It's abstract method.
+    ///
+    /// - parameter object: Object that has delegate property.
+    /// - returns: Value of delegate property.
+    open class func currentDelegateFor(_ object: ParentObject) -> Delegate? {
```
kzaher: Is there some reason we are conforming to `DelegateProxy` to `DelegateProxyBase`. It seems to me we aren't getting anything from that and using abstract methods in Swift is usually avoided.

Yes, we use them in a couple of places, but we don't need here IMHO.

RxCocoa/Common/DelegateProxy.swift
```diff
@@ -224,8 +253,8 @@ open class DelegateProxy : _RXDelegateProxy {
     /// through `self`.
     ///
     /// - returns: Value of reference if set or nil.
-    open func forwardToDelegate() -> AnyObject? {
-        return self._forwardToDelegate
+    open func forwardToDelegate() -> D? {
```
kzaher: Using associated type is probably more readable: `Delegate`.

RxCocoa/Common/DelegateProxy.swift
kzaher: I'm actually not sure we need both `DelegateProxyBase` and `DelegateProxyType`. Is there some reason we have two?

RxCocoa/Common/DelegateProxyFactory.swift
```diff
@@ -18,49 +18,71 @@ For example, in RxScrollViewDelegateProxy
...
+    private static var _sharedFactories: [ObjectIdentifier: DelegateProxyFactory] = [:]
+
+    /**
+     Shared instance of DelegateProxyFactory, if isn't exist shared instance, make DelegateProxyFactory instance for proxy type and extends.
+     DlegateProxyFactory have a shared instance per Delegate type.
+     - parameter proxyType: DelegateProxy type. Should use concrete DelegateProxy type, not generic.
+     - parameter extends: Extend DelegateProxyFactory if needs. See 'DelegateProxyType'.
+     - returns: DelegateProxyFactory shared instance.
+     */
+    public static func sharedFactory<DelegateProxy: DelegateProxyBase & DelegateProxyType>(for proxyType: DelegateProxy.Type, extends: (() -> Void)? = nil) -> DelegateProxyFactory {
```
kzaher: I think we should rename this. The name, intended usage and implementation are confusing to me.

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -86,56 +86,94 @@ and extend it
...
 }
 
-extension DelegateProxyType {
+extension DelegateProxyType where Self: DelegateProxyBase {
+
+    /// Store DelegateProxy subclass to factory.
+    /// When make 'RxXXXDelegateProxy' subclass, call 'RxXXXDelegateProxySubclass.prepareForFactory()' 1 time, or use it in DelegateProxyFactory
+    /// 'RxXXXDelegateProxy' can have one subclass implementation per concrete ParentObject type.
+    /// Should call it from concrete DelegateProxy type, not generic.
+    public static func prepareForFactory() {
```
kzaher: Maybe just call it `register`?

RxCocoa/Common/DelegateProxyFactory.swift
```diff
@@ -18,49 +18,71 @@ For example, in RxScrollViewDelegateProxy
...
+        MainScheduler.ensureExecutingOnScheduler()
+        if let factory = _sharedFactories[ObjectIdentifier(DelegateProxy.Delegate.self)] {
+            return factory
+        }
```
kzaher: This looks weird to me, if `factory` already exists, `extends` is ignored.

kzaher: I think we should rename `extends` to `knownImplementations`, make it a static method on `DelegateProxyType` and create a default implementation that has `void` implementation.

This API doesn't prevent anyone from calling `sharedFactory` twice with different `extends` parameter, which is not how this API show work IMHO.

RxCocoa/Common/DelegateProxyFactory.swift
```diff
@@ -18,49 +18,71 @@ For example, in RxScrollViewDelegateProxy
...
  
- 
+
  */
 public class DelegateProxyFactory {
+    private static var _sharedFactories: [ObjectIdentifier: DelegateProxyFactory] = [:]
+
+    /**
+     Shared instance of DelegateProxyFactory, if isn't exist shared instance, make DelegateProxyFactory instance for proxy type and extends.
+     DlegateProxyFactory have a shared instance per Delegate type.
```
kzaher: Typo `DlegateProxyFactory`.

tarunon:
Thank you, @kzaher 
I will improve this PR.

I think we can avoid using generics `ParentObject` in DelegateProxy subclass (e.g. RxTableViewDelegateProxy), I thought it for a few days after making this PR.
The biggest reason is swift not support covariant feature in generics parameter. 
I used some of workaround like `unsafeDownCast`, but I feel it's not good now.
Actually, I've made some of confusing API. 
I think next example is more clear than current implementation.
```swift
// Root DelegateProxy implementation
class RxScrollViewDelegateProxy
    : DelegateProxy<UIScrollView, UIScrollViewDelegate>, ... { ... }

// Subclass of Root DelegateProxy
class RxTableViewDelegateProxy: RxScrollViewDelegateProxy {
   require init(parentObject: UIScrollView) {
      self.tableView = castOrFatalError(parentObject)
      super.init(parentObject: parentObject)
   }
}

// Extends Root DelegateProxy's factory
RxTableViewDelegateProxy.register(for: UITableView)
```

RxCocoa/Common/DelegateProxy.swift
tarunon: I can merge it, thanks.

kzaher: It seems to me that we don't need this method defined here.

RxCocoa/Common/DelegateProxyFactory.swift
```diff
@@ -18,49 +18,71 @@ For example, in RxScrollViewDelegateProxy
...
 
     class RxScrollViewDelegateProxy: DelegateProxy {
-        static var factory = DelegateProxyFactory { (parentObject: UIScrollView) in
-                RxScrollViewDelegateProxy(parentObject: parentObject)
-            }
-            .extended { (parentObject: UITableView) in
-                RxTableViewDelegateProxy(parentObject: parentObject)
+        static var factory: DelegateProxyFactory {
+            return DelegateProxyFactory.sharedFactory(for: RxScrollViewDelegateProxy<UIScrollView>.self) {
+                RxTableViewDelegateProxy<UITableView>.register()
```
kzaher: It think we've changed this part.

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -86,56 +86,98 @@ and extend it
...
-    /// DelegateProxy factory
+    associatedtype ParentObject: AnyObject
+    associatedtype Delegate: AnyObject
+    
+    /// DelegateProxy factory. Use 'DelegateProxyFactory.sharedFactory'.
     static var factory: DelegateProxyFactory { get }
 
+    /// Creates new proxy for target object.
+    /// Should not call this function directory, use 'DelegateProxy.proxyForObject'
+    static func createProxy(for object: AnyObject) -> AnyObject
```
kzaher: I think we can remove this from `DelgateProxyType` since protocol extension is already implementing it depending on `factory`.

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -86,56 +86,98 @@ and extend it
...
+
+    /// Store DelegateProxy subclass to factory.
+    /// When make 'RxXXXDelegateProxy' subclass, call 'RxXXXDelegateProxySubclass.register()' 1 time.
+    /// 'RxXXXDelegateProxy' can have one subclass implementation per concrete ParentObject type.
+    static func register()
```
kzaher: It seems to me that `register` could be also removed from this interface.

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -144,41 +186,42 @@ extension DelegateProxyType {
...
+    ///            return RxSearchBarDelegateProxy<Base>.proxyForObject(base)
     ///         }
     ///
     ///         public var text: ControlProperty<String> {
     ///             let source: Observable<String> = self.delegate.observe(#selector(UISearchBarDelegate.searchBar(_:textDidChange:)))
     ///             ...
     ///         }
     ///     }
-    public static func proxyForObject(_ object: AnyObject) -> Self {
+    public static func proxyForObject(_ object: ParentObject) -> Self {
```
kzaher: We should make this `proxy(for:)`

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -144,41 +186,42 @@ extension DelegateProxyType {
     ///
     ///     extension Reactive where Base: UISearchBar {
     ///
-    ///         public var delegate: DelegateProxy {
-    ///            return RxSearchBarDelegateProxy.proxyForObject(base)
+    ///         public var delegate: DelegateProxy<Base, UISearchBarDelegate> {
+    ///            return RxSearchBarDelegateProxy<Base>.proxyForObject(base)
```
kzaher: I think that we should make this `RxSearchBarDelegateProxy<Base>.proxy(for: base)`

RxCocoa/Common/DelegateProxyType.swift
kzaher: I just want to understand, what did you force cast to here? To what type?

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -86,56 +86,98 @@ and extend it
...
     /// Objects can have multiple delegate properties.
     ///
     /// Each delegate property needs to have it's own type implementing `DelegateProxyType`.
     ///
+    /// It's abstract method.
+    ///
     /// - parameter object: Object that has delegate property.
     /// - returns: Value of delegate property.
-    static func currentDelegateFor(_ object: AnyObject) -> AnyObject?
+    static func currentDelegateFor(_ object: ParentObject) -> Delegate?
```
kzaher: I believe this should now be `func currentDelegate(for object: ParentObject) -> Delegate?`

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -86,56 +86,98 @@ and extend it
...
+
+    init(parentObject: ParentObject)
+
     /// Returns assigned proxy for object.
     ///
     /// - parameter object: Object that can have assigned delegate proxy.
     /// - returns: Assigned delegate proxy or `nil` if no delegate proxy is assigned.
-    static func assignedProxyFor(_ object: AnyObject) -> AnyObject?
-    
+    static func assignedProxyFor(_ object: ParentObject) -> Delegate?
```
kzaher: I believe this should now be `static func assignedProxy(for object: ParentObject) -> Delegate?`

RxCocoa/Common/DelegateProxyType.swift
kzaher: Shouldn't this now have the following signature?
`static func createProxy(for object: ParentObject) -> Self`

RxCocoa/iOS/Proxies/RxScrollViewDelegateProxy.swift
kzaher: Hi @tarunon ,

I'm sorry but I don't think I understand what did you want to say here 😅 ? Is there some problem?

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -86,56 +86,98 @@ and extend it
...
+    
+    /// DelegateProxy factory. Use 'DelegateProxyFactory.sharedFactory'.
     static var factory: DelegateProxyFactory { get }
 
+    /// Creates new proxy for target object.
+    /// Should not call this function directory, use 'DelegateProxy.proxyForObject'
+    static func createProxy(for object: AnyObject) -> AnyObject
+
+    /// Store DelegateProxy subclass to factory.
+    /// When make 'RxXXXDelegateProxy' subclass, call 'RxXXXDelegateProxySubclass.register()' 1 time.
```
kzaher: Maybe we can use `Rx*DelegateProxy` or some other wildcard instead of `'RxXXXDelegateProxy`. Seems a bit nicer 😄 .

I think the same would apply to other parts of docs.

RxCocoa/Common/DelegateProxyType.swift
tarunon: For example, in `RxTableViewDelegateProxy` case,

```swift
(tableView as UITableView).rx.delegate // DelegateProxy<UITableView, UIScrollViewDelegate>
(tableView as UIScrollView).rx.delegate //
DelegateProxy<UIScrollView, UIScrollViewDelegate>
```
In this case, actual type is `RxTableViewDelegateProxy<UITableView>` because it is checked on runtime in `DelegateProxyFactory`.
And in this case, I force cast from `RxTableViewDelegateProxy<UITableView>` to `RxScrollViewDelegateProxy<UIScrollView>`.

I think it is not good now...

RxCocoa/iOS/Proxies/RxScrollViewDelegateProxy.swift
tarunon: I think it is no problem here. 
I mean it is difficult making `RxTableViewDelegateProxy<UITableView, UITableViewDelegate>`.
I'm sorry confused you.

RxCocoa/Common/DelegateProxyType.swift
kzaher: Yeah, I think this is the biggest problem right now.

tarunon: Could I make commit that avoid `unsafeDowncast` way?
I think it will be better than current implemetation.

kzaher: We definitely need to figure this thing out, otherwise I'm not sure we can merge this :)

RxCocoa/Common/DelegateProxyType.swift
tarunon: I agree it, let's add it to `Deprecated.swift` also.

RxCocoa/Common/DelegateProxyType.swift
tarunon: If I use `static func createProxy(for object: ParentObject) -> Self` instead, 
I can see compile error like `Method 'createProxy(for:)' in non-final class 'RxScrollViewDelegate'`.

RxCocoa/Common/DelegateProxyType.swift
```diff
@@ -144,41 +181,42 @@ extension DelegateProxyType {
...
-        let currentDelegate: AnyObject? = Self.currentDelegateFor(object)
-
-        if currentDelegate !== proxy {
-            proxy.setForwardToDelegate(currentDelegate, retainDelegate: false)
-            assert(proxy.forwardToDelegate() === currentDelegate)
-            Self.setCurrentDelegate(proxy, toObject: object)
-            assert(Self.currentDelegateFor(object) === proxy)
-            assert(proxy.forwardToDelegate() === currentDelegate)
+        let currentDelegate = self.currentDelegate(for: object)
+        let delegateProxy = unsafeDowncast(proxy, to: self)
```
tarunon: The comment thread was closed by github, but still problem is here.

This is to solve cast between `RxTableViewDelegateProxy<UITableView>` and `RxScrollViewDelegateProxy<UIScrollView>`.
But I think it's better that quit using it. 
Remove `ParentObject` type generics parameter from `RxScrollViewDelegateProxy`, then we don't need to use `unsafeDowncast` anymore.

kzaher:
Hi @tarunon ,

are you still working on this? I think that we first need to figure out how does `UI{Table,Scroll}View` case work with this and what is the type of delegate proxy for those 2 classes.

tarunon:
Hello @kzaher ,

I'm still considering about what is the best way. 
It is my `UI{Table,Scroll}View` understanding,

Let's consider 3 cases when we use `DelegateProxy` in `UI{Table,Scroll}View`

|Run time|Compile time|
|---|---|
|`UIScrollView`|`UIScrollView`|
|`UITableView`|`UIScrollView`|
|`UITableView`|`UITableView`|

We declare `DelegateProxy` type based `ParentObject` type (Compile time), and we generate `DelegateProxy` instance uses `ParentObject`(Run time).
It will be...

|ParentObject|DelegateProxy|
|---|---|
|`UIScrollView`|`RxScrollViewDelegateProxy`|
|`UITableView`|`RxTableViewDelegateProxy`|

In master repo, `RxTableViewDelegateProxy` is subclass of `RxScrollViewDelegateProxy`, so there are no problem, but in this pr, `RxTableViewDelegateProxy<UITableView>` is not subclass of `RxScrollViewDelegateProxy<UIScrollView>`.
I use `unsafeDowncast` and it pass compile time and run time, but it's bad practice, maybe it is swift compiler bug(I think swift compiler don't check generics parameter when use `unsafeDowncast`).

I suggest avoid using `unsafeDowncast`, and I wish make 1 more commit that remove `unsafeDowncast`.
May I make commit?

kzaher:
Hi @tarunon , we definitely have to do this properly. Swift compiler doesn't support covariance as far as I can tell.

So I'm assuming `RxTableViewDelegateProxy` won't have generic arguments and will just inherit from `RxScrollViewDelegateProxy<UIScrollView>`? 

I think extension mechanism will also need to be modified a bit since it relies on `DelegateProxyType `.

tarunon:
> So I'm assuming RxTableViewDelegateProxy won't have generic arguments and will just inherit from RxScrollViewDelegateProxy<UIScrollView>?

I think so to, and it is better that implement `class RxScrollViewDelegateProxy: DelegateProxy<UIScrollView, UIScrollViewDelegate>`.
Let me create one more commit, plz wait few hours. 🙏

kzaher:
@tarunon do you want me to take a look now?

tarunon:
@kzaher Looks like test is going now, but maybe it will pass.
Current code is good for me, please check when you have time 🙇

kzaher:
Hi @tarunon ,

I think we could maybe improve this code. I'll try to prototype something on top of 679961e0e8496bb88bf7b1781f11c09535a43740 and show you if I succeed.

kzaher:
Hi @tarunon ,

I've merged your changes and simplified them a bit so we only need two `castOrFatalError` and have smaller public interface. You can take a look in swift4.0 branch, mostly at d1bc89b5ef942e8cfa1288b1914a47f26e74755f.

tarunon:
Hi @kzaher ,

I checked your refactoring. I like it, `DelegateProxyFactory` now be private, great! 
Thank you XD

Issue #1355, opened 2017-07-30
Extend RxBlocking to retrieve error from observable in unit tests


sgleadow:
**Short description of the issue**:

Feature request: Add a mechanism to retrieve an error from an observable, synchronously, via `RxBlocking`.

**Expected outcome**:

If this is a desirable feature, can we agree on a good interface?

I wrote an interface in #1354 (removed pending discussion here) that optionally returned the error if the observable terminated with error, or nil otherwise:

```
extension BlockingObservable {
    public func toError() -> Swift.Error? {  ... }
}
```

This is a different style interface to the existing `toArray()`, which will return the elements from the observable if they exist and throws an error otherwise.

Given this is for use in tests, I'd prefer all these to return optional values that we can assert on, and not to throw an error... the throwing seems to make the tests more verbose to eg. check that an error occurred. For example without throwing the style is more like:

```
        let error = Observable<Int>.error(testError).toBlocking().toError()
        XCTAssertNotNil(error)
```

As opposed to needing to catch functions if they throw:

```
        do {
            _ = try Observable<Int>.error(testError).toBlocking().single()
            XCTFail()
        }
        catch let e {
            XCTAssertErrorEqual(e, testError)
        }
```

**What actually happens**:

_Not implemented yet_

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

3.6.1

**Platform/Environment**

  - [x] iOS
  - [x] macOS
  - [x] tvOS
  - [x] watchOS
  - [x] playgrounds

**Xcode version**:

```
  Xcode version goes here
```

:warning: Fields below are optional for general issues or in case those questions aren't related to your issue, but filling them out will increase the chances of getting your issue resolved. :warning:

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [x] yes (which ones)
  - [ ] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [x] I have a small code base
  - [ ] I have a significant code base

kzaher:
Hi @sgleadow ,

I think this is a more general question "how to concisely test throwing method?".

I think you can do 

```swift
XCTAssertThrowsError(try xs.toBlocking().toArray())
```

or 

```swift
[...]

PR #1383, opened 2017-08-16
New materialize() operator over `BlockingObservable`

sgleadow:
This is based on a discussion in #1355 and on previous changes in #1354.

We introduce new extension to `BlockingObservable` called `materialize()` which returns a result type capturing the completed/failed status of the sequence, along with and elements and the error. The result type looks like this:

```
public enum SequenceMaterializeResult<T> {
    case completed(elements: [T])
    case failed(elements: [T], error: Error)
}
```

Internally, the rest of the `BlockingObservable+Operators` use `SequenceMaterializeResult`, and only `throw` at the last minute if an unexpected error occurs. This is mostly to maintain a backwards compatible interface for the existing operators, so it's a non-breaking change.

### DIFF
--- a/Documentation/UnitTests.md
+++ b/Documentation/UnitTests.md
@@ -86,7 +86,7 @@ It's easy to define `RxTests` extensions so you can write your tests in a readab
 
 It is also possible to write integration tests by using `RxBlocking` operators.
 
-Importing operators from the `RxBlocking` library will enable blocking the current thread and wait for sequence results.
+Using `RxBlocking`'s `toBlocking()` method, you can block the current thread and wait for the sequence to complete, allowing you to synchronously access its result.
 
 A simple way to test the result of your sequence is using the `toArray` method. It will return an array of all elements emitted once a sequence has completed successfully, or `throw` if an error caused the sequence to terminate.
 
@@ -98,7 +98,7 @@ let result = try fetchResource(location)
 XCTAssertEqual(result, expectedResult)
 ```
 
-Another option is to use the `materialize` operator which allows you to more granularly examine your sequence. It will return a `MaterializedSequenceResult` enumeration that could be either `.completed` along with the emitted elements if the sequence completed successfully, or `failed` if the sequence terminated with an error, along with the emitted error.
+Another option would be to use the `materialize` operator which lets you more granularly examine your sequence. It will return a `MaterializedSequenceResult` enumeration that could be either `.completed` along with the emitted elements if the sequence completed successfully, or `failed` if the sequence terminated with an error, along with the emitted error.
 
 ```swift
 let result = try fetchResource(location)
--- a/Documentation/UnitTests.md
+++ b/Documentation/UnitTests.md
@@ -88,10 +88,28 @@ It is also possible to write integration tests by using `RxBlocking` operators.
 
 Importing operators from `RxBlocking` library will enable blocking the current thread and wait for sequence results.
 
+If the squence succeeds, you can extract the array of results using `toArray()`, or `throw` if there's an error.
+
 ```swift
 let result = try fetchResource(location)
         .toBlocking()
         .toArray()
 
 XCTAssertEqual(result, expectedResult)
+```
+
+To extract a result representing the array of results _and_ an error if it occurred, the `materialize()` extension can be used.
+
+```swift
+let result = try fetchResource(location)
+        .toBlocking()
+        .materialize()
+
+switch result {
+        case .completed:
+            XCTFail("Expected result to be complete eith error, but result was successful.")
+        case .failed(let elements, let error):
+            XCTAssertEqual(elements, expectedResult)
+            XCTAssertErrorEqual(error, expectedError)
+        }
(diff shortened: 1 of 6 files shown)


RxBlocking/BlockingObservable+Operators.swift
```diff
@@ -10,14 +10,20 @@
     import RxSwift
 #endif
 
+public enum SequenceMaterializeResult<T> {
```
freak4pc: Not sure on the naming here, 
Would expect `MaterializedSequenceResult<T>` perhaps

Also wondering if it could be nice to go with `elements` and `errors` similar to RxSwiftExt's operators that extend materialize?

sgleadow: The rename totally make sense, I'll do that. In terms of `errors` do you mean in the enum case name or in the associated value parameter?

I'm not sure plural `errors` makes sense given it's only a single error that terminates the sequence? (I think, Rx newbie here).

1. In case name:

```
public enum MaterializedSequenceResult<T> {
    case completed(elements: [T])
    case errors(elements: [T], error: Error)
}
```

2. In associated value name

```
public enum MaterializedSequenceResult<T> {
    case completed(elements: [T])
    case failed(elements: [T], errors: [Error])
}
```

sgleadow:
Good point, examples added to UnitTests.md documentation

freak4pc:
@sgleadow Could you also real quickly add a CHANGELOG entry with something along the lines of "Added `materialize()` operator for RxBlocking's BlockingObservable`"

sgleadow:
Thanks for the help. I had no idea you could push to my fork!

I've added a line to the CHANGELOG referencing this PR.

freak4pc:
@sgleadow Yeah you can do it only through GitHub's UI apparently but its sometimes helpful for README stuff :) Thanks for all the hard work! @sgleadow 

Don't think there's anything blocking here, as soon as @kzaher finds the time

Documentation/UnitTests.md
```diff
@@ -95,3 +97,19 @@ let result = try fetchResource(location)
...
+Another option would be to use the `materialize` operator which lets you more granularly examine your sequence. It will return a `MaterializedSequenceResult` enumeration that could be either `.completed` along with the emitted elements if the sequence completed successfully, or `failed` if the sequence terminated with an error, along with the emitted error.
+
+```swift
+let result = try fetchResource(location)
+        .toBlocking()
+        .materialize()
+
+switch result {
+        case .completed:
+            XCTFail("Expected result to be complete eith error, but result was successful.")
```
kzaher: `eith` -> `with`?

RxBlocking/BlockingObservable+Operators.swift
```diff
@@ -10,14 +10,20 @@
     import RxSwift
 #endif
```
kzaher: I think we should document this, what `completed` means, what `failed` means and what is this `enum` for.

freak4pc: I actually added this in the documentation: https://github.com/ReactiveX/RxSwift/pull/1383/files/d416ff5e5c4f461c8afe262824f7d34f73d06e81#diff-6eb6c44aac5cc628cd1c254752e52e36R101

Did you want to see something besides that?

freak4pc: Or your mean inline documentation block for that enum ?

sgleadow: I've tweaked the `UnitTests.md` and added an example of the successful case, not just the error case.

Although if you mean inline docs like @freak4pc mentioned, I can add that.

sgleadow: I looked through other parts of the RxSwift codebase and thought an inline comment describing the `MaterializedSequenceResult` was also appropriate, so I've added that also.

RxBlocking/BlockingObservable+Operators.swift
```diff
@@ -10,14 +10,28 @@
...
 
+/// The `MaterializedSequenceResult` enum represents the materialized
+/// output of a BlockingObservable.
+///
+/// If the sequence terminates successfully, the result is represented
+/// by `.completed` with the array of elements.
+///
+/// If the sequene terminates with error, the result is represented
```
kzaher: Typo `sequene`

Tests/RxBlockingTests/Observable+BlockingTest.swift
```diff
@@ -337,3 +337,54 @@ extension ObservableBlockingTest {
...
+        }
+    }
+    
+    func testMaterialize_someData_fail() {
+        let sequence = Observable.concat(Observable.of(42, 43, 44, 45), Observable<Int>.error(testError))
+        let result = sequence.toBlocking().materialize()
+        
+        switch result {
+        case .completed:
+            XCTFail("Expected result to be complete eith error, but result was successful.")
```
kzaher: Typo `eith`.

PR #1384, opened 2017-08-17
Add a possibility to use XCTAssertEqual with elements of type [Record…

bryzinski:
The problem appears if a stream of events sends arrays. Due to swift 3 limitations it's not possible to write yet
```
extension Array: Equatable where Element: Equatable .
```
Assertion function is unavailable. Check the example below:

```
class FooTests: XCTestCase {
    
    let disposeBag = DisposeBag()
       
    func testBar() {
        let observable = Observable.from([[1], [2], [3], [4]])
        let expected = [next(0, [1]), next(0, [2]), next(0, [3]), next(0, [4]), completed(0)]
        let observer = scheduler.createObserver([Int].self)
        observable.subscribe(observer).addDisposableTo(disposeBag)
        scheduler.start()
        XCTAssertEqual(observer.events, expected)
    }
}

```
It rises:

> Cannot invoke 'XCTAssertEqual' with an argument list of type '([Recorded<Event<[Int]>>], [Recorded<Event<[Int]>>])'

### DIFF
--- a/RxTest/Event+Equatable.swift
+++ b/RxTest/Event+Equatable.swift
@@ -36,6 +36,29 @@ public func == <Element: Equatable>(lhs: Event<Element>, rhs: Event<Element>) ->
 ///
 /// In case `Error` events are being compared, they are equal in case their `NSError` representations are equal (domain and code)
 /// and their string representations are equal.
+public func == <Element: Equatable>(lhs: Event<[Element]>, rhs: Event<[Element]>) -> Bool {
+    switch (lhs, rhs) {
+    case (.completed, .completed): return true
+    case (.error(let e1), .error(let e2)):
+        #if os(Linux)
+            return  "\(e1)" == "\(e2)"
+        #else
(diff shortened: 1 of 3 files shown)


sergdort:
Hi, @bryzinski 
I see your point, and I have these extension as well :) Please see https://github.com/ReactiveX/RxSwift/pull/839 for more info. I don't think @kzaher opinion changed on it :)

freak4pc:
I agree on this not fitting the scope of Swift 3's capabilities, especially
since Swift 4 is right around the corner. Is there any improvement we can
make on the 4 branch around this ?

kzaher:
We can't possibly add all known types, so we need to make the cut somewhere.

The workaround is to map to some `EquatableXXX` adapter struct. For example we use `EquatableArray` internally.

bryzinski:
@kzaher thank you, I see. @sergdort Sorry, I missed previous discussion.

PR #1386, opened 2017-08-19
[WIP] Math Operators - min(), max(), sum()

freak4pc:
Hey All :)
I noticed we’re missing [`min()`](http://reactivex.io/documentation/operators/min.html), [`max()`](http://reactivex.io/documentation/operators/max.html) and [`sum()`](http://reactivex.io/documentation/operators/sum.html) so thought it would be cool to try and add these. 😄 

I've pushed this PR onto the Swift 4 branch so it would be part of the new release, and also to take advantage of the new Numeric protocol for `sum()`. 

This is my first attempt at making an operator PR so I’m sure there are some things that can be done better, which is exactly why I’m sharing this with the community. Would love the feedback of those more experienced 🤓 

The operators are based off of ReactiveX and also more or less the way they're implemented in RxJava: 

`min()` and `max()` work "out of the box" for all `Comparable`s. You can also [provide a `comparator` callback](https://github.com/ReactiveX/RxSwift/blob/feature/math/Tests/RxSwiftTests/Observable%2BMaxTests.swift#L92) to all non-`Comparable`s to "decide" which elements are lesser or greater. 

`sum()` works "out of the box" for all `Numeric`s. You can also [provide a `seed` and `accumulator`](https://github.com/ReactiveX/RxSwift/blob/feature/math/Tests/RxSwiftTests/Observable%2BSumTests.swift#L78) for non-`Numeric`s to provide custom summing of the items.

Looking forward to your feedback and hope this makes sense to all :)

🎉 🎉 🐰

### DIFF
--- a/RxSwift/Observables/Max.swift
+++ b/RxSwift/Observables/Max.swift
@@ -9,7 +9,7 @@
 extension ObservableType {
 
     /**
-     Emits the item from the source Observable that had the maximum value using a provided comparator closure.
+     Emits the element from the source Observable that had the maximum value using a provided comparator closure.
 
      - parameter comparator: A comparator method expected to return true if the first element has a greater value then the second element.
      - returns: An observable sequence containing the specified number of elements from the end of the source sequence.
@@ -23,7 +23,7 @@ extension ObservableType {
 extension ObservableType where E: Comparable {
 
     /**
-     Emits the item from the source Observable that had the maximum value.
+     Emits the element from the source Observable that had the maximum value.
 
      - returns: An observable sequence containing the specified number of elements from the end of the source sequence.
      */
@@ -35,13 +35,13 @@ extension ObservableType where E: Comparable {
 
 final fileprivate class MaxSink<O: ObserverType> : Sink<O>, ObserverType {
     typealias E = O.E
-    typealias Comparator = (E, E) -> Bool
+    typealias Parent = Max<E>
 
-    private let _comparator: Comparator
+    private let _parent: Parent
     private var _max: E?
 
-    init(comparator: @escaping Comparator, observer: O, cancel: Cancelable) {
-        _comparator = comparator
+    init(parent: Parent, observer: O, cancel: Cancelable) {
+        _parent = parent
         super.init(observer: observer, cancel: cancel)
     }
 
@@ -53,7 +53,7 @@ final fileprivate class MaxSink<O: ObserverType> : Sink<O>, ObserverType {
                 return
             }
 
-            _max = _comparator(max, value) ? _max : value
+            _max = _parent._comparator(max, value) ? _max : value
         case .error:
             forwardOn(event)
             dispose()
@@ -72,7 +72,7 @@ final fileprivate class Max<Element>: Producer<Element> {
     typealias Comparator = (E, E) -> Bool
 
     private let _source: Observable<Element>
-    private let _comparator: Comparator
+    fileprivate let _comparator: Comparator
 
     init(source: Observable<Element>, comparator: @escaping Comparator) {
         _source = source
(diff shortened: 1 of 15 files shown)


freak4pc:
I'll push a CHANGELOG entry and remove the WIP tag once this is reviewed

sergdort:
Hi, @freak4pc 

Looks nice 👍 
Not sure tho how common is use case of these operators

RxSwift/Observables/Max.swift
```diff
@@ -0,0 +1,87 @@
...
+
+extension ObservableType {
+
+    /**
+     Emits the element from the source Observable that had the maximum value using a provided comparator closure.
+
+     - parameter comparator: A comparator method expected to return true if the first element has a greater value then the second element.
+     - returns: An observable sequence containing the specified number of elements from the end of the source sequence.
+    */
+    public func max(_ comparator: @escaping (E, E) -> Bool)
```
sergdort: It seems like there is not difference between `Max` and `Min` producer or I missing something?

I feel like there could be only one implementation but with different comparison strategies.
E.g 
enum ComparisonStrategy {
    case min
    case max
}

So you just swap the order of comparison. Min: e2 < e1, Max: e1 < e2.

Am I right?

freak4pc:
Hey @sergdort , thanks for reviewing ! 
I agree this isn't really in the top of the list of "common operator", but for the sake of completeness I felt like it would be nice having them. I did find a couple of interesting use cases for them but definitely not "common"

RxSwift/Observables/Max.swift
freak4pc: In general, you're right. And I also sort of Contemplated if I could just have sort of a MinMaxSink with discrete implementations in the same file, but : 

1. I didn't know if that is the consensus in the RxSwift repo. It sort of seems most operators get their own file. 
2. Looking at Debounce and Throttle, they are very similar but they still get a file each. I guess the point is you could change implementations independently etc but TBH I really don't mind consolidating those if @kzaher will feel that's needed

kzaher:
Hi @freak4pc ,

I think that this is a dupe of #168 :) Sorry :)

I would really not want to pull some code in if we don't have good reasons why we would want to have it here :)

freak4pc:
@kzaher I have a couple points **for** this: 

1. I think we should aim to be as close as possible to ReactiveX, especially for people coming from other patterns. These are eventually missing operators ... IMO

2. You could also "easily" use `map` + `switchLatest` instead of a `flatMapLatest`... You could also perform arithmetic actions on numbers using just Bitwise operators instead of `+`, `-`, `/` etc, but it doesn't mean you should... But those doesn't mean you don't need `flatMapLatest` or arithmetic operators. If you have a language operator fitting a specific solution, it's easier to discover for a problem and is tailored specific to some problem. 

This is just me $0.02 but if you don't agree I'm 100% fine with closing this :) 

Up you 👍

sergdort:
I think there were two very important points from the previous discussion :

> * every added operator creates maintenance debt
> * it's hard to find important things in large codebases, and I'm afraid including all of these operators and specializations could bloat our codebase a lot, and thus increasing library size a lot. 

I feel like the last one is very important (hard for beginners to adopt, increase the build time and binary size)

freak4pc:
I understand , but also the fact this is under ReactiveX means in some way we should try and provide the operators defined by the organization. This isn't just a custom operator or anything crazy, it is on the ReactiveX manual and part of RxJava but not part of RxSwift... it kinda bothers me personally but as I said I'm not religious about it. Just think it makes sense IMO.

fpillet:
In the case these don't make it into core, don't hesitate to submit a PR for RxSwiftExt! Of course there's a rewrite of the operators themselves but at least the tests can stay the same...

kzaher:
Hi guys,

I'm sorry @sergdort for not merging this PR, but I really feel that not including them is the most optimal solution. Declining PRs doesn't come as a joy to me, quite the opposite.

> I think we should aim to be as close as possible to ReactiveX, especially for people coming from other implementation. These are eventually and practically missing operators ...

I agree that we should be as close as possible to ReactiveX, but I don't agree that means including and shipping all of the mentioned operators.

I think context also matters. I believe those operators were also only added for completeness with synchronous LINQ extensions in C# which we don't have.

> You could also "easily" use map + switchLatest instead of a flatMapLatest... You could also perform arithmetic actions on numbers using just Bitwise operators instead of +, -, / etc... But that doesn't mean you don't need flatMapLatest or arithmetic operators. If you have a language operator fitting a specific solution, it's easier to discover for a problem and is tailored specific to some problem.

Frequency of using `flatMapLatest` outweighs the proposed ones by far IMHO.

sergdort:
It was @freak4pc 's PR :)

Issue #1248, opened 2017-05-12
Transforming and Combining operators for PrimitiveSequence


hagmas:
**Short description of the issue**:

Some transforming and combining operators needs to be added to PrimitiveSequence, and some should be removed because they have no effects on some traits, such as `FlatMap` and `Map` for Completable.

**Related Issues**:

https://github.com/ReactiveX/RxSwift/issues/1206
https://github.com/ReactiveX/RxSwift/issues/1245

**Expected outcome**:

Here is a table that describes the relationship between some operators and PrimitiveSequence traits. ⭕️ means the operator has effect on the trait and ❌ means ineffective.

| Operators      | Single              | Maybe              | Completable  |
| ------------- |:--------------:|:--------------:|:--------------:|
|`AsMaybe`|⭕️|❌|⭕️|
|`AsSingle`|❌|❌|❌|
|`AsCompletable`|⭕️|❌|❌|
|`Combine`|⭕️(Same as `Zip`)|⭕️(Same as `Zip`)|⭕️(Same as `Zip`)|
|`Concat`|❌|⭕️(*1)|⭕️|
|`Map`|⭕️|⭕️|❌|
|`Merge`|❌|⭕️|⭕️|
|`Switch`,` FlatMap`|⭕️|⭕️|❌|
|`SwitchIfEmpty`|❌|⭕️|⭕️(Same as `Concat`)|
|`Zip`|⭕️|⭕️|⭕️(*2)|

*1: `Concat` for `Maybe` is effective only when either of the observables is completed.
*2: `Zip` for `Completable` is effective but a useless resultSelector needs to be defined. This can be replaced with `Merge` operator.

To implement these new transforming and combining operators for PrimitiveSequence, I would suggest to define them separately in three files "Single+Operators", "Maybe+Operators" and "Completable+Operators", e.g., defining `FlatMap` operator in "Single+Operators" and "Maybe+Operators", but not in "Completable+Operators". Operators valid for all the traits such as `Combine` or other creating/utility operators can be defined in the "PrimitiveSequence.swift" file as currently defined.

**What actually happens**:
The actual issue with the current version of code is that, `Map` and `FlatMap` can be used for `Completable` because they are defined in the extension of PrimitiveSequence. However, actually they have no effect and can be misleading.

```swift
Completable.empty().flatMap { _ in
    // the flow never comes here.
    return Completable.empty()
}.subscribe().disposed(by: disposeBag)
```

[...]

PR #1387, opened 2017-08-20
[WIP] Cleanup traits operators

sergdort:
This PR partly addresses #1248

### Description of change

* Moved Single, Maybe and Completable traits into separate files
* Remove some operators which does not make sense to have on particular traits (e.g `Map` on `Completable`)

### Questions

* There is [timer](https://github.com/ReactiveX/RxSwift/pull/1387/files#diff-5f830a64a3d55d39f04e65a487f45d52R247) operator on `PrimitiveSequenceType` do we even need it? It seems like it does not make sense to have
* What should we do with operators that was removed from the `Completable`? do we want to deprecate them?
* Do we need to add tests?


RxSwift/Traits/Completable.swift
```diff
@@ -0,0 +1,216 @@
...
+     
+     - parameter onNext: Action to invoke for each element in the observable sequence.
+     - parameter onError: Action to invoke upon errored termination of the observable sequence.
+     - parameter onCompleted: Action to invoke upon graceful termination of the observable sequence.
+     - parameter onSubscribe: Action to invoke before subscribing to source observable sequence.
+     - parameter onSubscribed: Action to invoke after subscribing to source observable sequence.
+     - parameter onDispose: Action to invoke after subscription to source observable has been disposed for any reason. It can be either because sequence terminates for some reason or observer subscription being disposed.
+     - returns: The source sequence with the side-effecting behavior applied.
+     */
+    public func `do`(onError: ((Swift.Error) throws -> Void)? = nil,
```
sergdort: I've removed `onNext` closure as it does not make sense to have it

RxSwift/Traits/Single.swift
```diff
@@ -0,0 +1,226 @@
...
+     
+     - parameter onNext: Action to invoke for each element in the observable sequence.
+     - parameter onError: Action to invoke upon errored termination of the observable sequence.
+     - parameter onSubscribe: Action to invoke before subscribing to source observable sequence.
+     - parameter onSubscribed: Action to invoke after subscribing to source observable sequence.
+     - parameter onDispose: Action to invoke after subscription to source observable has been disposed for any reason. It can be either because sequence terminates for some reason or observer subscription being disposed.
+     - returns: The source sequence with the side-effecting behavior applied.
+     */
+    public func `do`(onNext: ((ElementType) throws -> Void)? = nil,
+                     onError: ((Swift.Error) throws -> Void)? = nil,
```
sergdort: I've removed `onCompleted` for `Single`

RxSwift/Traits/Single.swift
```diff
@@ -0,0 +1,226 @@
...
+            )
+    }
+    
+    /**
+     Filters the elements of an observable sequence based on a predicate.
+     
+     - seealso: [filter operator on reactivex.io](http://reactivex.io/documentation/operators/filter.html)
+     
+     - parameter predicate: A function to test each source element for a condition.
+     - returns: An observable sequence that contains elements from the input sequence that satisfy the condition.
```
sergdort: This should probably say: returns `Maybe` even tho we filtering on the `Single`

sergdort:
At the moment CI is failing because I've removed `timeout` operator from the `Completable` do we even need it?

RxSwift/Traits/Single.swift
freak4pc: I think when I wrote the Docs when originally moving to Traits, @kzaher preferred leaving it as Observable Sequence instead of specifically naming Single, Maybe and Completable for the "returns" clause

kzaher: I think we can leave `observable sequence` terminology. It will also make our lives easier while we will try to keep documentation consistent, and `Maybe` is an observable sequence with additional guarantees IMHO.

kzaher:
I think timeout on `Completable` is useful to have IMHO.

sergdort:
I probably do not see bigger picture :) 
Anyway gonna put it back then

freak4pc:
@sergdort It would allow you to error out if a Completable doesn't complete in some expected time, which is actually useful if you expect something to be done in a timely manner

sergdort:
The thing is that, as far as I can tell `timeout` throws error if `next` event was not emitted, but `Completable` doesn't have `next` event

freak4pc:
@sergdort It would work for a completed event as well. 

Completed pre-timeout:
screen shot 2017-08-20 at 11 18 41 pm

Completed post-timeout:
screen shot 2017-08-20 at 11 19 00 pm

sergdort:
Yeah I guess it can be useful when you want to error out sequence if it was not completed before

sergdort:
Crazy question :) How important these traits to have them in the main repo?

I've never used them :)

freak4pc:
They are heavily used. Specifically Driver and Single.

RxMoya for example uses Single for all responses

-- 
Shai Mishali

kzaher:
In majority of cases I use `Single` and `Driver`.

kzaher:
@sergdort I've read similar opinions somewhere :) 

We are using Firebase and NetworkRequests in the same app. Replacing interfaces that return one element and terminate with persistent model observers (like FireBase or some other model layer) isn't realistic IMHO. 

Those interfaces are being consumed in completely different ways IMHO.

freak4pc:
The definition of Traits are variations of Observables that have different guarantees 

There is something nice about a stream that guarantees only a single element, or only a completion / error, or always on the main thread etc . It's good for visibility and exploration for other developers reading into your code base and it's easier to understand eventually.

kzaher:
Unit tests are failing ....

sergdort:
Weekends are coming :)

kzaher:
Hi @sergdort ,

just curious, are you still working on this?

> What should we do with operators that was removed from the Completable? do we want to deprecate them?

No, it was a bug that they existed in the first place, we only want to remove those kinds of operators.

> Do we need to add tests?

If some operator is just an alias, then I would just use sanity check unit test (like `Single` trait has). 
If some operator has completely new code base, then I would create detailed unit tests.
I think that should be a good trade-off.

kzaher:
You just need to call `./scripts/package-spm.swift` to fix unit tests.

sergdort:
Hi, @kzaher . The tests seems to be passing :)

Regarding your previous reply, I believe based on your answer there is nothing need to be added in this PR. But I may be mistaken =)

kzaher:
Hi @sergdort ,

I think `Completable` is missing: `deferred`, `delaySubscription`, `delay`, `observeOn`, `subscribeOn`, `catchError`, `retry`, `retryWhen`, `debug`, `using`, `timeout`.

Similar lists are for `Single` and `Maybe` :)

> If some operator is just an alias, then I would just use sanity check unit test (like Single trait has).
> If some operator has completely new code base, then I would create detailed unit tests.
> I think that should be a good trade-off.

I also don't see any code for sanity check unit tests. We didn't have them before, but I think it would be useful to have them so we don't run again into some weird public API issues like we have with some `Completable` APIs, and just to at least enumerate all of the operators we have.

It will probably be easier for me to finish this PR since there is a lot of work left anyway :)

sergdort:
@kzaher Sorry was on holidays :) Is there anything left to do?

kzaher:
@sergdort it's fine, I've done the boring part :)

PR #1388, opened 2017-08-25
Add RxSwift Chinese Documentation.

beeth0ven:
This documentation will help chinese to learn RxSwift much easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <img src="assets/Rx_Logo_M.png" alt="Miss Electric Eel 2016" width="36" height="36"> RxSwift: ReactiveX for Swift
 ======================================
 
-[![Travis CI](https://travis-ci.org/ReactiveX/RxSwift.svg?branch=master)](https://travis-ci.org/ReactiveX/RxSwift) ![platforms](https://img.shields.io/badge/platforms-iOS%20%7C%20macOS%20%7C%20tvOS%20%7C%20watchOS%20%7C%20Linux-333333.svg) ![pod](https://img.shields.io/cocoapods/v/RxSwift.svg) [![Carthage compatible](https://img.shields.io/badge/Carthage-compatible-4BC51D.svg?style=flat)](https://github.com/Carthage/Carthage) [![Swift Package Manager compatible](https://img.shields.io/badge/Swift%20Package%20Manager-compatible-brightgreen.svg)](https://github.com/apple/swift-package-manager)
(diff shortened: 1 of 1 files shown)


freak4pc:
This is actually very cool. 
My thought is though, that if we want to have multi-lingual documentation, we should make it part of the repo. And then we can possibly have a `Documentation/en` `Documentation/zh` etc etc.

fpillet:
Full documentation is a long way to go (and maintain!) but yeah, it's very cool to see translation efforts popping up 👍

beeth0ven:
@freak4pc @fpillet Thanks for all your encouragement and praising 😊. And  I am your Chinese fan @fpillet! Learn a lot from the conversation in Slack and the book.
@freak4pc Thanks for your advice. Because **RxSwift** is such a big topic, I publish a documentation  website using gitbook. It's not only include the documentation from RxSwift repo, but also cover some important topic such as: think reactive, architecture, rxswift ecosystem, reactiveX ecosystem etc.
So there is about 100 pages with 160 image asset files(continue to grow). I want to keep all this files separate from main repo which will make main repo lightweight. 

**This is the website address:** https://beeth0ven.github.io/RxSwift-Chinese-Documentation/

**Here are some screen shots:**



---


---

freak4pc:
@beeth0ven That's very cool!!
I'm just not 100% sure if it belongs in the main repo, or it could be more fitting in http://community.rxswift.org

@kzaher @fpillet what do you think ?

fpillet:
Yeah we had a discussion about this -- It's great to have stuff translated, although it also adds maintenance debt the project may not be able to sustain in the long term. Keeping these outside of the main repo may be the most reasonable option.

beeth0ven:
Yes, it will took a lot of time to keep documentation up to date. So I plan to update the documentation when there are some big changes.  e.g.  when RxSwift 3 update to RxSwift 4. That's more cost-effective. 

@fpillet I'll do my best to maintain the documentation. I think it will help a lot of people.

freak4pc:
@beeth0ven In this case I suggest submitting a PR here perhaps 
https://github.com/RxSwiftCommunity/rxswiftcommunity.github.io/blob/source/data/items.yml

beeth0ven:
@freak4pc @fpillet I'm sorry to trouble you. 

@freak4pc  Thanks for your advice. 

I see the https://github.com/RxSwiftCommunity/rxswiftcommunity.github.io/blob/source/data/items.yml file. But don't known which categories to choose. Could you give me some advises?

kzaher:
@freak4pc @beeth0ven @fpillet 

Yeah, I'm not sure how to best handle external translations.

On the other hand I can definitely see you've put a lot of effort into it, so thnx for that.

I agree with @freak4pc about http://community.rxswift.org suggestion.

beeth0ven:
@kzaher Thanks for provide such an amazing framework. It makes our life much easier. So I follow you to let more people benefit from it.

I will follow the advice about http://community.rxswift.org. @freak4pc @fpillet

freak4pc:
Closing this in favor of the Issue in RxSwiftCommunity.

PR #1389, opened 2017-08-25
Added reactive wrapper for UIStepper stepValue property

svdo:
Title says it all. 

**Question:** I actually need this commit on the `swift4.0` branch. How to arrange? Create separate pull request with same commit? Please advise...

### DIFF
--- a/RxCocoa/iOS/UIStepper+Rx.swift
+++ b/RxCocoa/iOS/UIStepper+Rx.swift
@@ -28,15 +28,10 @@ extension Reactive where Base: UIStepper {
     }
 
     /// Reactive wrapper for `stepValue` property.
-    public var stepValue: ControlProperty<Double> {
-        return UIControl.rx.value(
-            self.base,
-            getter: { stepper in
-                stepper.stepValue
-            }, setter: { stepper, stepValue in
-                stepper.stepValue = stepValue
-            }
-        )
+    public var stepValue: UIBindingObserver<Base, Double> {
+        return UIBindingObserver(UIElement: self.base) { stepper, value in
+            stepper.stepValue = value
+        }
     }
     
 }
--- a/RxCocoa/iOS/UIStepper+Rx.swift
+++ b/RxCocoa/iOS/UIStepper+Rx.swift
@@ -26,6 +26,18 @@ extension Reactive where Base: UIStepper {
             }
         )
     }
+
+    /// Reactive wrapper for `stepValue` property.
+    public var stepValue: ControlProperty<Double> {
+        return UIControl.rx.value(
+            self.base,
+            getter: { stepper in
+                stepper.stepValue
+            }, setter: { stepper, stepValue in
+                stepper.stepValue = stepValue
+            }
+        )
+    }
     
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ## Master
 
+* Adds Reactive wrapper for `UIStepper.stepValue` property. #1389
 * Adds `UIScrollView.willEndDragging` extension. #1365
 * Adds `materialize()` operator for RxBlocking's `BlockingObservable`. #1383
 * Adds `first` operator to ObservableType.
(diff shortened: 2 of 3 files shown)


devxoul:
Could you write a test for it? I think you might need to fork this repo to merge this pr in a Swift 4 branch.

svdo:
@devxoul Sure things, sorry for omitting the test. I will make do with my personal Swift 4 fork until the new beta release you mentioned.

kzaher:
Hi @svdo ,

I think we could add `stepValue` but it's type should be `UIBindingObserver`. It's write only property. Using  `UIControl.rx.value` only works for main data property.

svdo:
Hi @kzaher,

Thanks for your feedback. I don't quite understand what you mean though; `UIStepper.stepValue` is read+write (https://developer.apple.com/documentation/uikit/uistepper/1624083-stepvalue), and I don't feel there is a conceptual difference with the existing `value` property in `UIStepper+Rx.swift`, so I just copied that. Can you please clarify?

svdo:
Looking a bit further at some of the existing properties in other reactive wrappers, I am guessing this is what you mean:

```swift
    /// Reactive wrapper for `stepValue` property.
    public var stepValue: UIBindingObserver<Base, Double> {
        return UIBindingObserver(UIElement: self.base) { stepper, value in
            stepper.stepValue = value
        }
    }
```

Is that indeed what you meant? And if yes, for my understanding can you please explain (a) the difference with what I initially wrote, and (b) why the `value` is a `ControlProperty` and not a `UIBindingObserver`?

kzaher:
Hi @svdo ,

`UIControl.rx.value` uses target/action pattern for observing changes. Target/action only fires for user changes and not for programatic ones. As far as I can tell `stepValue` can only be changed programmatically, so using `UIControl.rx.value` is not the API that works out for this scenario.

svdo:
Ok I sort of get what you mean, even though it did work which I find confusing. 

So the solution above with `UIBindingObserver` is correct then? If so, I'll change the pull request.

kzaher:
Hi @svdo ,

what do you mean by 
> even though it did work which I find confusing.

> So the solution above with UIBindingObserver is correct then? If so, I'll change the pull request.

This is the correct version as far as I can tell.

```
    /// Reactive wrapper for `stepValue` property.
    public var stepValue: UIBindingObserver<Base, Double> {
        return UIBindingObserver(UIElement: self.base) { stepper, value in
            stepper.stepValue = value
        }
    }
```

svdo:
I have pulled in latest from develop and changed code as requested. Should be OK now.

kzaher:
@svdo I'm not sure what you've done but this doesn't look mergable, please take a look at the commit list and changes in this PR.

svdo:
Because the review took so long every time, develop has moved on quite a bit. I simply pulled it in. I think it should just merge, but if I have time next week I can see if I can do my changes again from scratch so that you'll have a more normal looking PR.

kzaher:
I would suggest always doing a rebase on top of `develop` branch. Git should handle these kinds of scenarios easily.

svdo:
Agreed, that would have been better.

svdo:
@kzaher I fixed the situation with the commits, my PR now only contains my own commits. It would be awesome if this could be merged before and be part of the next release, I noticed that it is in beta now. Thanks!

RxCocoa/iOS/UIStepper+Rx.swift
```diff
@@ -26,6 +26,13 @@ extension Reactive where Base: UIStepper {
             }
         )
     }
+
+    /// Reactive wrapper for `stepValue` property.
+    public var stepValue: UIBindingObserver<Base, Double> {
+        return UIBindingObserver(UIElement: self.base) { stepper, value in
```
devxoul: Don't we need to observe this value? I think we have to make this `ControlProperty` instead of `UIBindingObserver`.

svdo:
@kzaher @devxoul Any updates on this pull request?

svdo:
Hi guys, I completely understand that you are probably very busy and that this is probably something you have to squeeze in on the side, but I can't help being a bit disappointed by the lack of progress here. Is there anything I can do to move this forward? Thanks!

devxoul:
It looks fine to me. How do you think @kzaher ?

svdo:
Thx!

PR #1396, opened 2017-09-05
Add macOS NSPopUpButton reactive extensions

fpillet:
Reactive extensions for NSPopUpButton on macOS, with tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ## Master
 
+* Adds `NSPopUpButton` reactive extensions (macOS)
 * Adds `UIScrollView.willEndDragging` extension. #1365
 * Adds `materialize()` operator for RxBlocking's `BlockingObservable`. #1383
 * Adds `first` operator to ObservableType.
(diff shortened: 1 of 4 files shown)


freak4pc:
Just FYI you could've just changed the branch without a new PR :) 
0843def8-0ecd-4683-a0e2-c473134f54f7

fpillet:
That would have done it 👍  next time!

kzaher:
Hi guys,

I've tried to review this PR as best as I can, but it took a while since I don't have any experience with OSX.

The current behavior of `menuItems` now clears selection every time on bind. Yes, we could create some hacks how to not clear it, but it seems to me that conceptually setting `menuItems` should be coupled with setting `selectedItemIndex`. Those 2 values together make control state. Something like `public var menuItems: UIBindingObserver<Base, (selectedItemIndex: Int, titles: [String])>`. Otherwise you are depending on binding items and selected index in correct order which is prone to errors and current interface doesn't communicate this is a requirement.

I'm not sure is there any value in having `public var selectedItem: ControlProperty<NSMenuItem?>`.  `NSMenuItem` is not a value type so it's not an ideal sequence element and it seems to me that using a tag as unique identifier is far from ideal since it's an Int and not a generic identifier. As far as I can tell `selectedItemIndex` + access to data model of the items that are being bound to `NSPopUpButton` is the only consistent way to determine the typed selected item. I'm in favor of not including this API.

It seems to me that `public var selectedItemTitle: ControlProperty<String?>` API is depending on the title of the item, which is not a robust way to differentiate between two items, especially when drop down is populated with user generated content. I'm in favor of not including this API.

Reasoning for not including`public var selectedItemTag: ControlProperty<Int?>` is similar to `selectedItemTitle`. I don't think it's a generic way to differentiate between items and it seems to me that knowledge of `selectedItemIndex` + access to data model of the items that are being bound to `NSPopUpButton` is the only consistent way to determine the typed selected item.

It seems to me that `public func menuItems() -> UIBindingObserver<Base, [String]>` and `public func menuItemsWithTags() -> UIBindingObserver<Base, [(String,Int)]>` should be merged into something like `public var menuItems: UIBindingObserver<Base, (selectedItemIndex: Int, titles: [String])>` because of similar reasons. Tags aren't generic enough, Apple probably added them as a crotch.

Hope my explanations make sense.

fpillet:
Too much hassle, not sharing this anymore as I need it in my own projects.

PR #1407, opened 2017-09-14
Add wrappers for `observe` methods to use Swift 4 KeyPath

yusuke024:
This PR adds wrapper `observe` methods that accept new Swift 4 [`KeyPath`](https://github.com/apple/swift/blob/master/stdlib/public/core/KeyPath.swift) which includes both key path and type information.

So the below code:
```swift
self.rx.observe(CGRect.self, #keyPath(UIViewController.view.frame))
```
can be rewritten to:
```swift
self.rx.observe(\view.frame)
```

Let me know what you think about this solution, after that I will:
- [ ] Add some tests
- [ ] Update the document

### DIFF
--- a/RxCocoa/Foundation/NSObject+Rx.swift
+++ b/RxCocoa/Foundation/NSObject+Rx.swift
@@ -68,6 +68,27 @@ extension Reactive where Base: NSObject {
     public func observe<E>(_ type: E.Type, _ keyPath: String, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> {
         return KVOObservable(object: base, keyPath: keyPath, options: options, retainTarget: retainSelf).asObservable()
     }
+
+    /**
+     Observes values on `keyPath` starting from `self` with `options` and retains `self` if `retainSelf` is set.
+
+     `observe` is just a simple and performant wrapper around KVO mechanism.
+
+     * it can be used to observe paths starting from `self` or from ancestors in ownership graph (`retainSelf = false`)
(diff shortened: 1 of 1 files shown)


yusuke024:
Changed to `develop`.

kzaher:
Hi @yusuke024 ,

can you please add unit tests?

kzaher:
Ok, I guess I'll need to add tests to this and merge this :(

yusuke024:
Ah, I'm very sorry. I was having a little problem.

When I was writing the test I found out that if `E` (in `observe<E>`) is an optional the result observable would be double optional.

```swift
class MyClass: NSObject {
    var myVar: String? = nil
}

let m = MyClass()
let o1 = m.rx.observe(#keyPath(MyClass.myVar), String.self) // o1 is Observable<String?>
let o2 = m.rx.observe(\.myVar) // o2 is Observable<String??>
```

So, the type could be different for the same key path. And the test in `KVOObservableTests` can't be reused. So I want to fix this first and then add the test.

I'll continue working on this and add the test soon.

BTW, I messed up the forked repo a little bit, so I delete it and refork. But it seems like Github cannot remap to the new one. So I think I'll make a new PR later.

Again, I'm very sorry I didn't reply 😭

RowbotNZ:
Another option to get this behaviour could be to have `NSKeyValueObservation` conform to `Disposable`

```
extension NSKeyValueObservation: Disposable {
    public func dispose() {
        invalidate()
    }
}
```

Then you can use the Swift 4 API and just add the resulting observation directly to your dispose bag

```
view.observe(\.frame) { [unowned self] (view, change) in
    //Do stuff
}.disposed(by: dBag)
```

PR #1408, opened 2017-09-14
Added pretty init for AnyObserver with closures for each event

EvsenevDev:
**Added pretty init for AnyObserver with closures for each event.**
It's really simple for use:

```swift
let observer = AnyObserver<String>(onNext: { element in
    print(element)
}, onError: { error in
    print(error.localizedDescription)
}, onCompleted: {
    print("onCompleted")
})
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 ---
 
 ## Master
-
+* Adds `init` for `AnyObserver` with closures for each event. #1408
 * Adds `UIScrollView.willEndDragging` extension. #1365
 * Adds `materialize()` operator for RxBlocking's `BlockingObservable`. #1383
 * Adds `first` operator to ObservableType.
--- a/RxSwift/AnyObserver.swift
+++ b/RxSwift/AnyObserver.swift
@@ -32,6 +32,22 @@ public struct AnyObserver<Element> : ObserverType {
         self.observer = observer.on
     }
     
+    /// Construct an instance whose `on(event)` calls specific closure for events.
+    ///
+    /// - parameter observer: Observer that receives sequence events.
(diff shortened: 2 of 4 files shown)


kzaher:
Hi @EvsenevDev ,

could you please explain the context why you need this functionality? `subscribe` method has its own convenience overloads so you don't need to use `AnyObserver`.

EvsenevDev:
Hi @kzaher,
This like in **subscribe(onNext,onError...)**, but for `AnyObserver init`.
In our project we have principe: All public interface always returns **Observer** or **Observable**, and we often create `AnyObserver`. For create `AnyObserver` in each time i need write `switch` for events. My PR adds simplest `init`, which looks like `subscribe` func. For example i need `AnyObserver` and just `onNext` interested for me:

```Swift
class MyClass {
   var myObserver: AnyObserver<Void> {
       return AnyObserver<Void>(onNext:{ 
          print("Hello") 
       })
   }
}
```

Before this `init` we need to write:

```Swift
class MyClass {
   var myObserver: AnyObserver<Void> {
       return AnyObserver<Void> { event in
          switch event {
             case .next:
                 print("Hello")
             default: brek
       }
   }
}
```

And i think first sample more comfortable for use.

kzaher:
Hi @EvsenevDev ,

I think that we are now working towards making an observer just a simple lambda to unify general observable sequence observer and `Single`/`Maybe`/`Completable` "observers". 

`ObservableType` is just an adapter that would be used to easily construct that lambda.

I think that the interface shouldn't be designed in a way that allows accidentally ignoring errors like this one is:

```swift
var myObserver: AnyObserver<Void> {
       return AnyObserver<Void>(onNext:{ 
          print("Hello") 
       })
   }
```

`subscribe` method already has some default error handling logic and we also have `Binder` that will at least do some reasonable logging or fatalErroring in DEBUG.

I would feel much more comfortable leaving this code outside of the main repo for now since:
* one can easily implement it if needed
* I personally didn't ever have a need for it
* there is already a practical replacement called `Binder` that is specialized for those use cases

PR #1410, opened 2017-09-17
Introduce `Sink` using nested class

tarunon:
Hi there.

I proposed same topic when swift 3.1 released. #1157 
Maybe now is good, I propose it again.
I wish make same changes to every operator.

## Pros
* Can eliminate many duplicate type defines and use same generics type and `typealias` between `Sink` and their parent.
* Can increase `Sink`s confidentiality.

## Cons
* Make many code differences, several other improvements may affected by this proposal.
* In the past, `Sink` was always written before their parent, cannot do that anymore. (Can write at `extension`, we need to write after their parent.)
* If there are shared `Sink` class, this proposal might be a not good way in shared `Sink` class.

## Benchmark
This is my benchmark score.

|BenchmarkMethod|swift4.0|proposal/nested_sink|
|---|---|---|
|testMapFilterCreating|0.766sec|0.787sec|
|testMapFilterPumping|3.132sec|3.155sec|

### DIFF
--- a/RxSwift/Observables/Map.swift
+++ b/RxSwift/Observables/Map.swift
@@ -23,40 +23,6 @@ extension ObservableType {
     }
 }
 
-final fileprivate class MapSink<SourceType, O : ObserverType> : Sink<O>, ObserverType {
-    typealias Transform = (SourceType) throws -> ResultType
-
-    typealias ResultType = O.E
-    typealias Element = SourceType
-
-    private let _transform: Transform
-    
-    init(transform: @escaping Transform, observer: O, cancel: Cancelable) {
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @tarunon ,

I'm thinking that it might be worth exploring some potential changes that could improve our performance if we were to do some internal refactorings.

I'm assuming that maybe removing one generic parameter (`O: ObserverType`) and just passing lambdas through `run` and `subscribe` could improve performance, but I haven't tested it out yet.

tarunon:
Thanks, I understand. I try it later.

Issue #1414, opened 2017-09-20
Swift 4.0 support


swftvsn:
Would be awesome! Is there any work done on this?

nmdias:
I didn't had any issues with Swift 4 & Xcode 9 on my end. It's working great.

groue:
I *think* the swift4.0 branch has been superseded by beta releases such as [4.0.0-beta.0](https://github.com/ReactiveX/RxSwift/releases/tag/4.0.0-beta.0):

[tweet](https://twitter.com/KrunoslavZaher/status/909718322454900737):

> I've published 4.0.0-beta.0 this morning at ~ 2:00AM

[tweet](https://twitter.com/KrunoslavZaher/status/909716865886367746):

> We are planning to release 4.0 mid October, we are in beta right now.

But I'm not sure it's the correct way to go: the swift4.0 branch may have the same content as the 3.6.1 release + Swift 4 compatibility, when the 4.0.0-beta.0 tag may have breaking changes towards the future version 4.0.

@kzaher: it looks like the community could enjoy some "official" communication on Swift 4.

tomisacat:
But RxCocoa currently does not support Swift 4

kzaher:
I've updated the README.md file with 

* RxSwift 4.x / Swift 4.x can be found in [**rxswift4.0-swift4.0** branch](https://github.com/ReactiveX/RxSwift/tree/rxswift4.0-swift4.0).
* RxSwift 3.x / Swift 3.x can be found in [**master** branch](https://github.com/ReactiveX/RxSwift/tree/master).

Currently CHANGELOG.md represents official communication about releases.

```
## [4.0.0-alpha.0](https://github.com/ReactiveX/RxSwift/releases/tag/4.0.0-alpha.0)
* Swift 4.0 compatibility
* Changes delegate proxy to use plugin architecture. 
```

zhentan:
It seems like iTunes Connect does not like having "4.0.0-beta.0" as the CFBundleShortVersionString value. When trying to upload my app via fastlane pilot I get the following error:  `[Transporter Error Output]: ERROR ITMS-90060: "This bundle is invalid. The value for key CFBundleShortVersionString '4.0.0-beta.0' in the Info.plist file must be a period-separated list of at most three non-negative integers."
` Has anyone else had this problem?

Vkt0r:
@zhentan Same issue here trying to submit a new build to TestFlight using the pre-release

kzaher:
[...]

PR #1418, opened 2017-09-21
Update to PackageDescriptionV4

devxoul:
`swift4.0` branch seems to be disappeared 🤔


kzaher:
Hi @devxoul ,

we've renamed it to rxswift4.0-swift4.0 because we've received some complaints in https://github.com/ReactiveX/RxSwift/issues/1414 .

devxoul:
Ah, I see 👍

PR #1423, opened 2017-09-25
Disabled code coverage

tomquist:
When submitting a build to Testflight, we get the following error:

> **Invalid Bundle** - Disallowed LLVM instrumentation. Do not submit apps with LLVM profiling instrumentation or coverage collection enabled. Turn off LLVM profiling or code coverage, rebuild your app and resubmit the app.

For more information: https://github.com/Carthage/Carthage/issues/2056


kzaher:
Fun times.

PR #1424, opened 2017-09-25
Set CFBundleShortVersionString from 4.0.0-beta.0 to 4.0.0

tomquist:
The current frameworks can't be used to do a AppStore/Testflight release because of the following error:

> **ERROR ITMS-90060**: "This bundle is invalid. The value for key CFBundleShortVersionString '4.0.0-beta.0' in the Info.plist file must be a period-separated list of at most three non-negative integers."


kzaher:
Hi,

I think we'll need to fix unit tests to merge this.

PR #1443, opened 2017-10-09
Add `asDriver` and `asSharedSequence` methods to PublishRelay

yuzushioh:
PublishReplay can't fail with error or completed so it would not be strange if `PublishRelay` trait could be converted to `Driver` without any error handling. So I added `asDriver()` method to it! And I also added `asSharedSequence`

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -115,6 +115,10 @@
 		AEBE14181F8B52F3001FAE4C /* PublishRelay+Driver.swift in Sources */ = {isa = PBXBuildFile; fileRef = AEBE14151F8B52C7001FAE4C /* PublishRelay+Driver.swift */; };
 		AEBE14191F8B52F4001FAE4C /* PublishRelay+Driver.swift in Sources */ = {isa = PBXBuildFile; fileRef = AEBE14151F8B52C7001FAE4C /* PublishRelay+Driver.swift */; };
 		AEBE141A1F8B52F5001FAE4C /* PublishRelay+Driver.swift in Sources */ = {isa = PBXBuildFile; fileRef = AEBE14151F8B52C7001FAE4C /* PublishRelay+Driver.swift */; };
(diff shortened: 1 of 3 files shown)


tarunon:
[nits] 
I think it's great that if also support `asSharedSequence`, and `asSignal`.

yuzushioh:
@tarunon I will handle them in this PR! Thanks 👍

yuzushioh:
@tarunon `asSignal` for `PublishRelay` already exits so I just added `asSharedSequence`

kzaher:
Hi @yuzushioh ,

`PublishRelay` is meant to dispatch events, so only conversion to `Signal` (`share()`) makes sense.
`Driver` drives with state `share(replay: 1)`.

I think that maybe `ObservableConvertibleType+SharedSequence.swift` confused you. I think we need to deprecate this APIs.

```
    public func asSharedSequence<S>(sharingStrategy: S.Type = S.self, onErrorJustReturn: E) -> SharedSequence<S, E>
    
    public func asSharedSequence<S>(sharingStrategy: S.Type = S.self, onErrorDriveWith: SharedSequence<S, E>) -> SharedSequence<S, E> 

    public func asSharedSequence<S>(sharingStrategy: S.Type = S.self, onErrorRecover: @escaping (_ error: Swift.Error) -> SharedSequence<S, E>) -> SharedSequence<S, E>
```

Feel free to make a PR, or I can also deprecate them.

yuzushioh:
Hi
Ok! I am going to deprecate them. you mean making those methods unavailable by deprecating right? I will make a new PR.

And I just want to make sure if I understand correctly in that `PublishRelay` is only conversion to `Signal` trait and it doesn't make sense to drive with state `share(replay: 1)`. so the best way to use PublishRelay is to convert it to signal and use emit subscription in `Signal` trait. Is this correct?

kzaher:
> Ok! I am going to deprecate them. you mean making those methods unavailable by deprecating right? I will make a new PR.

Yes.

> And I just want to make sure if I understand correctly in that PublishRelay is only conversion to Signal trait and it doesn't make sense to drive with state share(replay: 1). so the best way to use PublishRelay is to convert it to signal and use emit subscription in Signal trait. Is this correct?

Yes.

> I also don't understand why BehaviorRelay can be converted to Driver but PublishRelay can't. Is this because BehaviorSubject already replays the last element and PublishSubject does not replay?

Yes, they have consistent replay strategies.

PR #1444, opened 2017-10-09
Add driver subscription to PublishRelay

yuzushioh:
I added driver subscription to PublishRelay

### DIFF
--- a/RxCocoa/Traits/Driver/Driver+Subscription.swift
+++ b/RxCocoa/Traits/Driver/Driver+Subscription.swift
@@ -69,6 +69,34 @@ extension SharedSequenceConvertibleType where SharingStrategy == DriverSharingSt
             relay.accept(e)
         })
     }
+    
+    /**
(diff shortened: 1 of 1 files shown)


yuzushioh:
oh there are already `emit` methods for `PublishRelay` 😅

kzaher:
Hi,

`Driver` represents state `share(replay: 1)`, so it shouldn't make sense to `drive` something that has `share()` strategy, like `PublishRelay`.

yuzushioh:
Hi @kzaher 
I see your point. how about `BehaviorRelay` having `drive` subscriptions? I might have misunderstood but I thought `BehaviorRelay` doesn't represent `share(replay: 1)`. I don't understand how it makes sense here.

kzaher:
Hi @yuzushioh ,

I would rather not add some public interface if we don't have compelling reasons do so. I doubt users will need to do `behaviorSubject.asDriver().drive` often, or ever :)

PR #1447, opened 2017-10-10
Delete unnecessary comment-outs

yuzushioh:
I found unnecessary comment-outs when I was going through the code base so I just cleaned them up.


kzaher:
Hi @yuzushioh ,

The reason why those comments are there is to document on what scope lock applies. We've just manually inlined those locks because Swift optimizer didn't know how to inline them without overhead.

PR #1449, opened 2017-10-11
Deprecate `asSharedSequence` in ObservableConvertibleType

yuzushioh:
ref #1443


yuzushioh:
I misunderstood something here I will resubmit

PR #1451, opened 2017-10-11
changed method visibility from public to open

pual:
Currently it's not possible to override the `setForwardToDelegate`method.
One use case to do that is described here http://stackoverflow.com/questions/35575305/transform-uiapplicationdelegate-methods-into-rxswift-observables

### DIFF
--- a/RxCocoa/Common/DelegateProxy.swift
+++ b/RxCocoa/Common/DelegateProxy.swift
@@ -199,7 +199,7 @@
         ///
         /// - parameter forwardToDelegate: Reference of delegate that receives all messages through `self`.
         /// - parameter retainDelegate: Should `self` retain `forwardToDelegate`.
-        open func setForwardToDelegate(_ delegate: Delegate?, retainDelegate: Bool) {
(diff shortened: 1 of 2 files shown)


devxoul:
How about adding a second parameter to `proxy(for:)`? For example:

```swift
public static func proxy(for object: ParentObject, retainDelegate: Bool = false) -> Self
```

pual:
Good idea!
Pull request is updated and I restored the previous method visibility.

devxoul:
I like this. :+1: Then I think we need a test for it. How do you think @kzaher?

kzaher:
Hi guys,

the reason why we have it on `public func setForwardToDelegate(_ delegate: Delegate?, retainDelegate: Bool)`  is because retaining delegate depends on usage context of delegate proxy and it's not inherent to its definition.

For example, you know do you want to retain delegate/datasource depending is it set to parent view controller (in which case probably no since it points to parent) or some standalone object (in which case probably yes since it's not pointing to parent).

kzaher:
I've cherry-picked ef436b03cbc56dd7c9e95965f215dff2513a2ab8 from this PR.

Thnx.

Issue #1416, opened 2017-09-20
ObservableType.subscribe(...) #if DEBUG causing problems with pre-compiled frameworks


asutula:
**Short description of the issue**:

I build a distribute a pre-compiled Swift framework that depends on RxSwift. That framework uses the `ObservableType.subscribe(...)` function, of which there are two versions wrapped in a `#if DEBUG #else #endif` (see the [RxSwift source here](https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/ObservableType%2BExtensions.swift#L24)). My framework is built using the Release configuration, so the compiled binary is expecting the release version of the `ObservableType.subscribe(...)` function from RxSwift. 

When using my framework in a project that builds other dependencies from source code (Cocoapods, for example), and running a Debug build, there is a runtime crash that happens when my framework calls `ObservableType.subscribe(...)` because it is the Debug version of `ObservableType.subscribe(...)` that is compiled and that doesn't match with the Release version my pre-compiled binary Framework is looking for.

If I use pre-compiled versions of RxSwift (using Carthage, or manually downloading a release from Github), the problem goes away because there is now just the Release version of `ObservableType.subscribe(...)` and it is consistent with what my framework was compiled with.

I see that the two different versions of the `ObservableType.subscribe(...)` function have existed since last summer, and this didn't seem to be a problem with older versions of Cocoapods (Cocoapods 1.2.0 was the last one that didn't seem associated with this issue). I'm not sure if this is strictly and issue with RxSwift, or strictly an issue with Cocoapods, or somewhere in between, but the underlying issue is clear.

**Expected outcome**:

 I expect my pre-compiled framework to be able to call `ObservableType.subscribe(...)` for both Debug and Release builds.

**What actually happens**:

There is a runtime crash where `ObservableType.subscribe(...)` is not found for Debug builds:
```
dyld: lazy symbol binding failed: Symbol not found: __TIFE7RxSwiftPS_14ObservableType9subscribeFT6onNextGSqFwx1ET__7onErrorGSqFPs5Error_T__11onCompletedGSqFT_T__10onDisposedGSqFT_T___PS_10Disposable_A1_
[...]

PR #1454, opened 2017-10-12
Fix remove method signature inconsistence

Vladlex:
Fix for the issue #1416 

The case is following:
1. You have a framework that uses RxSwift. Framework built as a static library in debug mode.
2. You use this framework in an app, and you want to build it in Release (or any non-debug) mode.

The problem is that there are two different methods declared in RxSwift: one for debug and one for release. So at compile-time, it's ok to call any method (because of default argument values they can be called the same way), but at runtime, there's no valid method signature can be found.

So the fix is to declare ONE method for both debug and non-debug schemes, but break method implementation into debug and non-debug parts.

This fix should not affect any working projects but give an opportunity to avoid a crash while debugging apps with release-build static libs which calls debug-built RxSwift methods.

### DIFF
--- a/RxSwift/ObservableType+Extensions.swift
+++ b/RxSwift/ObservableType+Extensions.swift
@@ -8,114 +8,103 @@
 
 extension ObservableType {
     /**
-    Subscribes an event handler to an observable sequence.
-
-    - parameter on: Action to invoke for each event in the observable sequence.
-    - returns: Subscription object used to unsubscribe from the observable sequence.
-    */
+     Subscribes an event handler to an observable sequence.
+     
+     - parameter on: Action to invoke for each event in the observable sequence.
+     - returns: Subscription object used to unsubscribe from the observable sequence.
+     */
     public func subscribe(_ on: @escaping (Event<E>) -> Void)
         -> Disposable {
-        let observer = AnonymousObserver { e in
-            on(e)
-        }
-        return self.subscribeSafe(observer)
+            let observer = AnonymousObserver { e in
+                on(e)
+            }
+            return self.subscribeSafe(observer)
     }
-
-    #if DEBUG
-        /**
-        Subscribes an element handler, an error handler, a completion handler and disposed handler to an observable sequence.
-
-        - parameter onNext: Action to invoke for each element in the observable sequence.
-        - parameter onError: Action to invoke upon errored termination of the observable sequence.
-        - parameter onCompleted: Action to invoke upon graceful termination of the observable sequence.
-        - parameter onDisposed: Action to invoke upon any type of termination of sequence (if the sequence has
-            gracefully completed, errored, or if the generation is canceled by disposing subscription).
-        - returns: Subscription object used to unsubscribe from the observable sequence.
-        */
-        public func subscribe(file: String = #file, line: UInt = #line, function: String = #function, onNext: ((E) -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil, onCompleted: (() -> Void)? = nil, onDisposed: (() -> Void)? = nil)
-            -> Disposable {
-
-            let disposable: Disposable
-
-            if let disposed = onDisposed {
-                disposable = Disposables.create(with: disposed)
-            }
-            else {
-                disposable = Disposables.create()
-            }
-
+    
+    
+    /**
+     Subscribes an element handler, an error handler, a completion handler and disposed handler to an observable sequence.
+     
+     - parameter onNext: Action to invoke for each element in the observable sequence.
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @Vladlex ,

mixing of debug and release configurations is not supported. 

E.g. we have
```
#if TRACE_RESOURCES
    struct Resources {
        static func incrementTotal() -> Int32 {
            return RxSwift.Resources.incrementTotal()
        }

        static func decrementTotal() -> Int32 {
            return RxSwift.Resources.decrementTotal()
        }

        static var numberOfSerialDispatchQueueObservables: Int32 {
            return RxSwift.Resources.numberOfSerialDispatchQueueObservables
        }

        static var numberOfMapOperators: Int32 {
            return RxSwift.Resources.numberOfMapOperators
        }

        static var total: Int32 {
            return RxSwift.Resources.total
        }
    }
#endif
```

Can you please explain what is the reason why you don't have separate builds for debug and release configurations?

kierangraham:
I'm currently experiencing the exact same issue as in #1416; we have a Swift framework which we distribute only the binary and it has RxSwift as a dependency. We build our framework in Release mode, linking against RxSwift also built in Release mode. Third party apps _always_ integrate the Release version of our pre-built framework.

When our framework is installed via CocoaPods and run in the simulator, built in Debug mode, we encounter the following runtime crash:

```
dyld: lazy symbol binding failed: Symbol not found: __T07RxSwift14ObservableTypePAAE9subscribeAA10Disposable_py1EQzcSg6onNext_ys5Error_pcSg0gI0yycSg0G9CompletedAM0G8DisposedtFfA1_
  Referenced from: /Application/83EC76A2-5F00-444D-B524-3C1EF370ED43/Example.app/Frameworks/MyFramework.framework/MyFramework
  Expected in: /Application/83EC76A2-5F00-444D-B524-3C1EF370ED43/Example.app/Frameworks/RxSwift.framework/RxSwift

dyld: Symbol not found: __T07RxSwift14ObservableTypePAAE9subscribeAA10Disposable_py1EQzcSg6onNext_ys5Error_pcSg0gI0yycSg0G9CompletedAM0G8DisposedtFfA1_
  Referenced from: /Application/83EC76A2-5F00-444D-B524-3C1EF370ED43/Example.app/Frameworks/MyFramework.framework/MyFramework
  Expected in: /Application/83EC76A2-5F00-444D-B524-3C1EF370ED43/Example.app/Frameworks/RxSwift.framework/RxSwift
```

kzaher:
Hi,

@Vladlex I've pasted wrong code. This is the code I should have pasted:
```
#if TRACE_RESOURCES
    fileprivate var resourceCount: AtomicInt = 0

    /// Resource utilization information
    public struct Resources {
        /// Counts internal Rx resource allocations (Observables, Observers, Disposables, etc.). This provides a simple way to detect leaks during development.
        public static var total: Int32 {
            return resourceCount.valueSnapshot()
        }

        /// Increments `Resources.total` resource count.
        ///
        /// - returns: New resource count
        public static func incrementTotal() -> Int32 {
            return AtomicIncrement(&resourceCount)
        }

        /// Decrements `Resources.total` resource count
        ///
        /// - returns: New resource count
        public static func decrementTotal() -> Int32 {
            return AtomicDecrement(&resourceCount)
        }
    }
#endif
```

I guess we can give this PR a shot, but please remove `file: String = #file, line: UInt = #line, function: String = #function` before we can merge this.

I'll try to think about alternative ways to handle this better.

Vladlex:
Ok, thank you very much!

kierangraham:
@kzaher firstly, thanks for approving this change and your work on RxSwift! 

Would you have plans to cherry-pick this into a `3.6.2` or `3.7.0` release? I guess what I am concerned about is breaking changes coming in `4.0.0` that would prevent me from upgrading to the next release candidate.

kzaher:
Hi @kierangraham ,

there are no significant breaking changes I'm aware of. I've managed to upgrade my apps pretty easily. If there are some, please let me know.

As far as for 3.6.2 and 3.7.0, we'll think about that after 4.0.0. If we detect that there is a need for it, then yes.

kierangraham:
@kzaher managed to make the upgrade to `4.0.0-rc.0` without any major issues 😄 – do you have an idea of when a `4.0.0-rc.1` will land which includes this fix? 🙇

HanniuCat:
Hi @kierangraham ,

HanniuCat:
Hi,@kierangraham，
    When our framework is installed via CocoaPods and run in the simulator, built in Debug mode, we encounter the following runtime crash:

PR #1457, opened 2017-10-14
Extended Sample operation to support a default value

aspitz:
Added the ability to set a default value for the sampler to return in the event that there were no new elements between sampler ticks


kzaher:
Let me think about this suggestion.

freak4pc:
This is kinda interesting actually. Since this is purely additive there’s no real issue with adding this.

The only issue I’m seeing here is that it’s diverging from ReactiveX and from the original usage of sample (turning into something like an interval withLatestFrom). 

Still could be an interesting addition 🤔🤔🤔 hard to say

aspitz-intrepid:
@kzaher any further thoughts?

freak4pc:
I still find this to be a good addition. More specifically, it would be quite hard accomplishing this behavior by combining other operators in RxSwift. 

@aspitz any chance you'd be willing to bring this up-to-date to the latest standards or the repo? Otherwise, I can find someone else to take over this code.

aspitz-intrepid:
I can fix this up :)

freak4pc:
Awesome :) @aspitz-intrepid

RxSwift/Observables/Sample.swift
```diff
@@ -13,16 +13,18 @@ extension ObservableType {
...
      Upon each sampling tick, the latest element (if any) in the source sequence during the last sampling interval is sent to the resulting sequence.
 
-     **In case there were no new elements between sampler ticks, no element is sent to the resulting sequence.**
+     **In case there were no new elements between sampler ticks, if a default value has been set then it is sent
+       to the resulting sequence otherwise no element is sent.**
 
      - seealso: [sample operator on reactivex.io](http://reactivex.io/documentation/operators/sample.html)
 
      - parameter sampler: Sampling tick sequence.
+                 defaultValue: a value to return if there are new new elements between sampler ticks
```
freak4pc:
```suggestion
     - parameter defaultValue: a value to return if there are no new elements between sampler ticks
```

RxSwift/Observables/Sample.swift
```diff
@@ -13,16 +13,18 @@ extension ObservableType {
 
      Upon each sampling tick, the latest element (if any) in the source sequence during the last sampling interval is sent to the resulting sequence.
 
-     **In case there were no new elements between sampler ticks, no element is sent to the resulting sequence.**
+     **In case there were no new elements between sampler ticks, if a default value has been set then it is sent
```
freak4pc:
```suggestion
     **In case there were no new elements between sampler ticks, you may provide a default value to be emitted, instead
```

RxSwift/Observables/Sample.swift
```diff
@@ -54,6 +56,8 @@ final private class SamplerSink<Observer: ObserverType, SampleType>
             if let element = parent.element {
```
freak4pc:
```suggestion
            if let element = self.parent.element ?? self.parent.defaultValue {
```

RxSwift/Observables/Sample.swift
```diff
@@ -54,6 +56,8 @@ final private class SamplerSink<Observer: ObserverType, SampleType>
...
                 self.parent.element = nil
                 self.parent.forwardOn(.next(element))
+            } else if let element = self.parent.parent.defaultValue {
```
freak4pc: remove this

RxSwift/Observables/Sample.swift
```diff
@@ -119,10 +123,12 @@ final private class SampleSequenceSink<Observer: ObserverType, SampleType>
 final private class Sample<Element, SampleType>: Producer<Element> {
     fileprivate let source: Observable<Element>
     fileprivate let sampler: Observable<SampleType>
-
-    init(source: Observable<Element>, sampler: Observable<SampleType>) {
+    fileprivate let defaultValue: Element?
```
freak4pc: add a `defaultValue` argument to `SampleSequenceSink` and pass it inside. Doing `parent.parent` doesn't make much sense in this case.

RxSwift/Observables/Sample.swift
```diff
@@ -51,7 +53,7 @@ final private class SamplerSink<Observer: ObserverType, SampleType>
     func synchronized_on(_ event: Event<Element>) {
         switch event {
         case .next, .completed:
-            if let element = parent.element {
+            if let element = parent.element ?? self.parent.defaultValue{
```
freak4pc:
```suggestion
            if let element = parent.element ?? self.parent.defaultValue {
```

freak4pc:
Thank you ! @aspitz

PR #1459, opened 2017-10-17
Add wrappers for `observe` methods to use Swift 4 KeyPath

yusuke024:
This PR adds wrapper `observe` methods that accept new Swift 4 [Smart KeyPaths](https://github.com/apple/swift-evolution/blob/master/proposals/0161-key-paths.md) which includes both key path and type information.

So the below code:
```swift
self.rx.observe(CGRect.self, #keyPath(UIViewController.view.frame))
```
can be rewritten to:
```swift
self.rx.observe(\view.frame)
```

Mostly the tests are copied from the old tests except for `property.property...` key paths and non-exists key paths which cannot be composed with the new style key path.

This is a revised PR for #1407

### DIFF
--- a/RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
+++ b/RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
@@ -45,7 +45,15 @@ extension Reactive where Base: NSObject {
         return observe(E.self, keyPathString, options: options, retainSelf: retainSelf)
     }
 
+    /**
+     Specialization of generic `observe` method.
 
+     This specialization first observes `KVORepresentable` value and then converts it to `RawRepresentable` value.
+
+     It is useful for observing bridged ObjC enum values.
+
+     For more information take a look at `observe` method.
+     */
     @available(swift 4.0)
     public func observe<E: RawRepresentable>(_ keyPath: KeyPath<Base, E?>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> where E.RawValue: KVORepresentable {
         guard let keyPathString = keyPath._kvcKeyPathString else {
--- a/RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
+++ b/RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
@@ -27,6 +27,32 @@ extension Reactive where Base: NSObject {
         return observe(E.RawValue.KVOType.self, keyPath, options: options, retainSelf: retainSelf)
             .map(E.init)
     }
+
+    /**
+     Specialization of generic `observe` method.
+
+     This specialization first observes `KVORepresentable` value and then converts it to `RawRepresentable` value.
+
+     It is useful for observing bridged ObjC enum values.
+
+     For more information take a look at `observe` method.
+     */
+    @available(swift 4.0)
+    public func observe<E: RawRepresentable>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> where E.RawValue: KVORepresentable {
+        guard let keyPathString = keyPath._kvcKeyPathString else {
+            rxFatalError("Key path cannot be observed. You may need to prefix it with @objc.")
+        }
+        return observe(E.self, keyPathString, options: options, retainSelf: retainSelf)
+    }
+
+
+    @available(swift 4.0)
+    public func observe<E: RawRepresentable>(_ keyPath: KeyPath<Base, E?>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> where E.RawValue: KVORepresentable {
+        guard let keyPathString = keyPath._kvcKeyPathString else {
+            rxFatalError("Key path cannot be observed. You may need to prefix it with @objc.")
+        }
+        return observe(E.self, keyPathString, options: options, retainSelf: retainSelf)
+    }
 }
 
 #if !DISABLE_SWIZZLING
@@ -46,6 +72,40 @@ extension Reactive where Base: NSObject {
         public func observeWeakly<E: RawRepresentable>(_ type: E.Type, _ keyPath: String, options: KeyValueObservingOptions = [.new, .initial]) -> Observable<E?> where E.RawValue: KVORepresentable {
             return observeWeakly(E.RawValue.KVOType.self, keyPath, options: options)
                 .map(E.init)
+        }
+
+        /**
+         Specialization of generic `observeWeakly` method.
+
+         This specialization first observes `KVORepresentable` value and then converts it to `RawRepresentable` value.
+
+         It is useful for observing bridged ObjC enum values.
+
+         For more information take a look at `observeWeakly` method.
+         */
+        @available(swift 4.0)
+        public func observeWeakly<E: RawRepresentable>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial]) -> Observable<E?> where E.RawValue: KVORepresentable {
+            guard let keyPathString = keyPath._kvcKeyPathString else {
+                rxFatalError("Key path cannot be observed. You may need to prefix it with @objc.")
+            }
+            return observeWeakly(E.self, keyPathString, options: options)
+        }
+
+        /**
+         Specialization of generic `observeWeakly` method.
+
+         This specialization first observes `KVORepresentable` value and then converts it to `RawRepresentable` value.
+
+         It is useful for observing bridged ObjC enum values.
+
+         For more information take a look at `observeWeakly` method.
+         */
+        @available(swift 4.0)
+        public func observeWeakly<E: RawRepresentable>(_ keyPath: KeyPath<Base, E?>, options: KeyValueObservingOptions = [.new, .initial]) -> Observable<E?> where E.RawValue: KVORepresentable {
+            guard let keyPathString = keyPath._kvcKeyPathString else {
+                rxFatalError("Key path cannot be observed. You may need to prefix it with @objc.")
+            }
+            return observeWeakly(E.self, keyPathString, options: options)
         }
     }
 #endif
--- a/RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
+++ b/RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
@@ -38,8 +38,17 @@ extension Reactive where Base: NSObject {
      For more information take a look at `observe` method.
      */
     @available(swift 4.0)
-    public func observe<E: RawRepresentable>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> where E.RawValue: KVORepresentable {
-        return observe(E.self, keyPath.kvcKeyPathString, options: options, retainSelf: retainSelf)
+    public func observe<E: RawRepresentable>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E> where E.RawValue: KVORepresentable {
+        return observe(E.self, keyPath.kvcKeyPathString, options: options, retainSelf: retainSelf).flatMap { value -> Observable<E> in
+            guard let value  = value else {
+                #if DEBUG
+                    rxFatalError("Something went wrong with KVO observing mechanism")
+                #else
+                    return Observable.empty()
+                #endif
+            }
+            return Observable.just(value)
+        }
     }
 
     /**
--- a/RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
+++ b/RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
@@ -39,10 +39,7 @@ extension Reactive where Base: NSObject {
      */
     @available(swift 4.0)
     public func observe<E: RawRepresentable>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> where E.RawValue: KVORepresentable {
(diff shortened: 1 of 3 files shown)


RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
```diff
@@ -27,6 +27,40 @@ extension Reactive where Base: NSObject {
...
+     Specialization of generic `observe` method.
+
+     This specialization first observes `KVORepresentable` value and then converts it to `RawRepresentable` value.
+
+     It is useful for observing bridged ObjC enum values.
+
+     For more information take a look at `observe` method.
+     */
+    @available(swift 4.0)
+    public func observe<E: RawRepresentable>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> where E.RawValue: KVORepresentable {
```
kzaher: It seems to me that the result should be `Observable<E?>`.

We could do something like.

```
return observe(E.self, keyPathString, options: options, retainSelf: retainSelf).flatMap { value in
    guard let value  = value else { 
         #if DEBUG
             rxFatalError("Something went wrong with KVO observing mechanism")
         #endif
         return Observable.empty()
    }
    return 
}
```

The reason why I wouldn't want to make this a fatal error in release mode is because ObjC <-> Swift interoperability is a bit flakey.

RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
```diff
@@ -27,6 +27,40 @@ extension Reactive where Base: NSObject {
...
+     Specialization of generic `observe` method.
+
+     This specialization first observes `KVORepresentable` value and then converts it to `RawRepresentable` value.
+
+     It is useful for observing bridged ObjC enum values.
+
+     For more information take a look at `observe` method.
+     */
+    @available(swift 4.0)
+    public func observe<E: RawRepresentable>(_ keyPath: KeyPath<Base, E?>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> where E.RawValue: KVORepresentable {
```
kzaher: The same comments as before.

RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
```diff
@@ -47,6 +81,40 @@ extension Reactive where Base: NSObject {
...
+         Specialization of generic `observeWeakly` method.
+
+         This specialization first observes `KVORepresentable` value and then converts it to `RawRepresentable` value.
+
+         It is useful for observing bridged ObjC enum values.
+
+         For more information take a look at `observeWeakly` method.
+         */
+        @available(swift 4.0)
+        public func observeWeakly<E: RawRepresentable>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial]) -> Observable<E?> where E.RawValue: KVORepresentable {
```
kzaher: I think this signature and other `observeWeakly` are correct because `nil` will be emitted if the object is `nil`.

I guess if somebody has `KeyPath<Base, E?>` they could easily do `observeWeakly.map { $0.flatMap { $0 }}` :)

RxCocoa/Foundation/NSObject+Rx.swift
```diff
@@ -68,6 +68,56 @@ extension Reactive where Base: NSObject {
...
+     If support for weak properties is needed or observing arbitrary or unknown relationships in the
+     ownership tree, `observeWeakly` is the preferred option.
+
+     - parameter keyPath: Key path of property names to observe.
+     - parameter options: KVO mechanism notification options.
+     - parameter retainSelf: Retains self during observation if set `true`.
+     - returns: Observable sequence of objects on `keyPath`.
+     */
+    @available(swift 4.0)
+    public func observe<E>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> {
```
kzaher: Should be `Observable<E>` here?

RxCocoa/Foundation/NSObject+Rx.swift
```diff
@@ -68,6 +68,56 @@ extension Reactive where Base: NSObject {
...
+     If support for weak properties is needed or observing arbitrary or unknown relationships in the
+     ownership tree, `observeWeakly` is the preferred option.
+
+     - parameter keyPath: Key path of property names to observe.
+     - parameter options: KVO mechanism notification options.
+     - parameter retainSelf: Retains self during observation if set `true`.
+     - returns: Observable sequence of objects on `keyPath`.
+     */
+    @available(swift 4.0)
+    public func observe<E>(_ keyPath: KeyPath<Base, E?>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> {
```
kzaher: Shouldn't this be
```
public func observe<E>(_ keyPath: KeyPath<Base, E?>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> 
```
?

RxCocoa/Foundation/NSObject+Rx.swift
```diff
@@ -95,6 +145,50 @@ extension Reactive where Base: NSObject {
...
+
+     - parameter keyPath: Key path of property names to observe.
+     - parameter options: KVO mechanism notification options.
+     - returns: Observable sequence of objects on `keyPath`.
+     */
+    @available(swift 4.0)
+    public func observeWeakly<E>(_ keyPath: KeyPath<Base, E?>, options: KeyValueObservingOptions = [.new, .initial]) -> Observable<E?> {
+        guard let keyPathString = keyPath._kvcKeyPathString else {
+            rxFatalError("Key path cannot be observed. You may need to prefix it with @objc.")
+        }
```
kzaher: Instead of repeating 
```
guard let keyPathString = keyPath._kvcKeyPathString else {
     rxFatalError("Key path cannot be observed. You may need to prefix it with @objc.")
}
```

It might be better to create extension on `KeyPath` so we can do something like:
```
return observeWeakly(E.self, keyPath.kvcKeyPathString, options: options)
```

```
extension KeyPath {
     var kvcKeyPathString: String {
           guard let keyPathString = keyPath._kvcKeyPathString else {
               rxFatalError("Key path cannot be observed. You may need to prefix it with @objc.")
           }
           return keyPathString
     } 
}
```

and thus avoid duplication of logic.

Tests/RxCocoaTests/KVOObservableTests.swift
```diff
@@ -16,14 +16,17 @@ import RxCocoa
     import Cocoa
 #endif
 
-final class KVOObservableTests : RxTest {
+final class KVOObservableStringBasedKVOTests : RxTest {
```
kzaher: I'll take a closer look at this file tomorrow, don't have the necessary concentration to go through it today.

RxCocoa/Foundation/NSObject+Rx.swift
yusuke024: Sorry, but I'm not sure what're the differences. Can you help point out them out?

RxCocoa/Foundation/NSObject+Rx.swift
yusuke024: I was merely trying to mimic how the old method work before:
```swift
observe<E>(_ type: E.Type, _ keyPath: String, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?>
```
of which returns `Observable<E?>`.

But if you want me to change the return type to `Observable<E>` please let me know.

Tests/RxCocoaTests/KVOObservableTests.swift
yusuke024: Thank you very much for reviewing my PR 🙇

RxCocoa/Foundation/NSObject+Rx+RawRepresentable.swift
```diff
@@ -27,6 +27,40 @@ extension Reactive where Base: NSObject {
...
+
+     This specialization first observes `KVORepresentable` value and then converts it to `RawRepresentable` value.
+
+     It is useful for observing bridged ObjC enum values.
+
+     For more information take a look at `observe` method.
+     */
+    @available(swift 4.0)
+    public func observe<E: RawRepresentable>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?> where E.RawValue: KVORepresentable {
+        guard let keyPathString = keyPath._kvcKeyPathString else {
```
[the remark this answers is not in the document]
yusuke024: Me too. But it's the only way to get the string for the new key path for now.

I think the long term solution would be to rewrite the entire thing to use the new [`observe(:)` API](https://github.com/apple/swift/blob/c5ff1f2cac8da6a14330f4b033b94c7c926d2126/stdlib/public/SDK/Foundation/NSObject.swift#L185). But I think it would be safer to go with wrappers for now.

Do you have any suggestion how to check or test?

kzaher:
Hi @yusuke024 ,

This is my reasoning, please let me know does this make sense.

* for `observe` set of methods the return type should have the same type as `E` from `KeyPath<Base, E>` because compiler is hinting following this path should return type `E` and if the result set isn't `E`, then there is an error. In `DEBUG` mode we would fatal in case we get `nil` and in release mode we would ignore `nil` because objc interoperability is flakey. 

I'm assuming we'll need to do something like this for `observe`, but I'll need to investigate are there any ambiguity issues.

```
public func observe<E>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E>
public func observe<E>(_ keyPath: KeyPath<Base, E?>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?>
```

* for `observeWeakly` set of methods the return type should have the type `E?`. I think we'll need to detect if key path is `KeyPath<Base, E?>` and make sure we also return `E?` and not `E??` as just direct forwarding of `E?` to observe method would return.

I'm assuming we'll need to do something like this for `observeWeakly`, but I'll need to investigate are there any ambiguity issues.

```
public func observeWeakly<E>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?>
public func observeWeakly<E>(_ keyPath: KeyPath<Base, E?>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E?>
```

> I was merely trying to mimic how the old method work before:

Got it.

The reason why the return type is optional in case you use string for keyPath is because there aren't any compile time guarantees that the value KVO returns is not `nil` and because `observeWeakly` also needs to return `nil` in case target object is deallocated.

If we are using `KeyPath<Base, E>` then we have some compiler hints what is the end result. I'm also curious will this work correctly for optionals, but need to test it.

yusuke024:
I didn't know the rationale behind the returned `Observable<E?>` type for the old method so I  tried to make it consistent. 

I agree with you if we're to have `Observable<E>` return type for `observe` methods.
This will make things simpler for `observe`, instead of having two set of methods we can only just keep:
```swift
public func observe<E>(_ keyPath: KeyPath<Base, E>, options: KeyValueObservingOptions = [.new, .initial], retainSelf: Bool = true) -> Observable<E>
```
And ditch out the optional version.

I'll update the code. Thank you.

kzaher:
@yusuke024 I'm not sure we'll be able to solve this with a single method because of the reasoning above.

yusuke024:
I've changed from `observe<E>(...) -> Observable<E?>` to `observe<E>(...) -> Observable<E>`.
Please see if it's okay.

kzaher:
I would like to merge this, but I'm still polishing this a bit, and I'm a bit uncertain about using `_kvcKeyPathString`.

To help clarify can we use `_kvcKeyPathString`, I've created this question for Swift team:
https://bugs.swift.org/browse/SR-6270

Let's see if they will respond.

kzaher:
I've been investigating some more and it seems to me that we can ditch all of our existing code and replace it with just a simple wrapper around `NSObject.observe`.

```
@objcMembers class B: NSObject {
    dynamic var text: String = ""

    init(text: String = "") {
        print("B init")
        self.text = text
    }

    deinit {
        print("B deinit")
    }
}

@objcMembers class A: NSObject {
    dynamic weak var b: B? = nil
    dynamic var c: String = ""
    dynamic var frame: CGRect = CGRect.zero
}

func a() {
    let a = A()

    let result = a.observe(\A.b?.text, options: [.new, .initial]) { selF, value in
        print("b = " + (value.newValue.flatMap { $0 } ?? "nil"))
    }

    let result2 = a.observe(\A.frame, options: [.new, .initial]) { selF, value in
        print("frame = " + (value.newValue.flatMap { $0 }?.debugDescription ?? "nil"))
    }

    a.frame = CGRect(x: 0.0, y: 0.1, width: 1.0, height: 2.0)

    print("before")
    autoreleasepool {
        let b = B()
        a.b = b
        b.text = "b"
        a.b = B(text: "next1")
        a.b = B(text: "next2")
    }
    print("after")
}
```

This looks wonderful, but it also means we can make this PR a lot shorter :)

kzaher:
There is one issue though ...

In case weak reference deallocates, it doesn't send any value.

```
        autoreleasepool {
            a.b = B(text: "122")
        }
```

```
b = nil
frame = (0.0, 0.0, 0.0, 0.0)
frame = (0.0, 0.1, 1.0, 2.0)
before
B init
b = 122
B deinit
after
```

kzaher:
I guess we could just define the new api as a wrapper for official `observe` API and be done with it, although the behavior is a bit inconsistent and weird.

I'll try to report a new bug for that behavior.

kittinunf:
This would immensely help us in our project too!

kzaher:
This is dependent on this issue I've reported.

https://bugs.swift.org/browse/SR-6280

freak4pc:
There is no progress or approval on that bug for the past two years, I'm closing this for now and hope to pursue adding wrapping `observe` itself later on.

Issue #1460, opened 2017-10-19
Implement DelegateProxyType for pure swift delegate


mobilinked:
**Short description of the issue**:

  "@objc" has been added to a pure swift delegate to implement DelegateProxyType

**Expected outcome**:

  Pass xcode compiling without adding "@objc"

**What actually happens**:

  We have to change the implementations of some DelegateProxyType when migrating macOS project to Swift 4 and using RxSwift 4.0.0
  By ref the implementation of RxTextFieldDelegateProxy at https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/macOS/NSTextField%2BRx.swift

  Compiling failed with two errors: 
  - "super.init" cannot be called outside of an initializer 
  - Type "..." does not conform to protocol 'DelegateProxyType" (because of not implementing forwardToDelegate ad setForwardToDelegate) 

  Our temp solution is to add "@objc" to the interface to pass compiling - our expectation is to use PURE swift. 
```swift
@objc protocol TestControlDelegate: AnyObject {
```
  
**Self contained code example that reproduces the issue**:

```swift
import Cocoa
import RxSwift
import RxCocoa

protocol TestControlDelegate: AnyObject {
    func delegateDoSomething()
}
class TestControl: NSView {
    weak var delegate: TestControlDelegate? = nil
}

extension TestControl: HasDelegate {
    public typealias Delegate = TestControlDelegate
}

class TestControlProxy
    : DelegateProxy<TestControl, TestControlDelegate>
    , DelegateProxyType
    , TestControlDelegate {
    
    fileprivate let _imageSubject = PublishSubject<Void>()
    fileprivate weak var _control: TestControl? = nil
    
    //#MARK: DelegateProxy
    init(parentObject: TestControl) {
        self._control = parentObject
        super.init(parentObject: parentObject, delegateProxy: TestControlProxy.self)
    }
    
    public static func registerKnownImplementations() {
        self.register { TestControlProxy(parentObject: $0) }
    }
    
    func delegateDoSomething() {
        _imageSubject.on(.next(()))
//        forwardToDelegate()?.delegateDoSomething()
    }
}
```
  
**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

  4.0.0

**Platform/Environment**

  - [ ] macOS

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [ ] easy, 100% repro
  

**Xcode version**:

```
  9.0.1
```

[...]

PR #1465, opened 2017-10-20
Support pure swift Delegate type in DelegateProxy

### LINKED ISSUE #1442: How to make a `DelegateProxy` for a pure swift delegate protocol using RxSwift 4.0?
When I upgrade to swift 4.0, I also upgrade `RxSwift` to `4.0.0-rc.0`.The `DelegateProxy` now need 2 type params, One for `ParentObject` type and one for `Delegate` type.But in my case some delegate type isn't conform to `NSObjectProtocol`, it's just a pure swift protocol,how can I use it as a `DelegateProxy`?

tarunon:
It will fix #1442, #1460 

I removed every Delegate type condition, and check it in runtime.

There are some of uncheckable corner case.
- Can set not class Delegate.
  - Can compile.
  - It will crash cannot pass assert. 
- Can use not NSObject Delegate as forwardToDelegate.
  - Can compile.
  - Method forwarding does not work, should implement own self.

### DIFF
--- a/RxCocoa/Common/DelegateProxy.swift
+++ b/RxCocoa/Common/DelegateProxy.swift
@@ -171,7 +171,7 @@
                 return
             }
 
-            guard ((self.forwardToDelegate() as? NSObject)?.responds(to: selector) ?? false) || voidDelegateMethodsContain(selector) else {
+            guard (forwardToDelegateResponds(to: selector) || voidDelegateMethodsContain(selector)) else {
                 rxFatalError("This class doesn't respond to selector \(selector)")
             }
         }
--- a/RxCocoa/Common/DelegateProxy.swift
+++ b/RxCocoa/Common/DelegateProxy.swift
@@ -18,7 +18,7 @@
     /// Base class for `DelegateProxyType` protocol.
     ///
     /// This implementation is not thread safe and can be used only from one thread (Main thread).
-    open class DelegateProxy<P: AnyObject, D: AnyObject>: _RXDelegateProxy {
+    open class DelegateProxy<P: AnyObject, D>: _RXDelegateProxy {
         public typealias ParentObject = P
         public typealias Delegate = D
 
@@ -28,8 +28,8 @@
         /// Parent object associated with delegate proxy.
         private weak private(set) var _parentObject: ParentObject?
 
-        fileprivate let _currentDelegateFor: (ParentObject) -> Delegate?
-        fileprivate let _setCurrentDelegateTo: (Delegate?, ParentObject) -> ()
+        fileprivate let _currentDelegateFor: (ParentObject) -> AnyObject?
+        fileprivate let _setCurrentDelegateTo: (AnyObject?, ParentObject) -> ()
 
         /// Initializes new instance.
         ///
@@ -37,8 +37,8 @@
         public init<Proxy: DelegateProxyType>(parentObject: ParentObject, delegateProxy: Proxy.Type)
             where Proxy: DelegateProxy<ParentObject, Delegate>, Proxy.ParentObject == ParentObject, Proxy.Delegate == Delegate {
             self._parentObject = parentObject
-            self._currentDelegateFor = delegateProxy.currentDelegate(for:)
-            self._setCurrentDelegateTo = delegateProxy.setCurrentDelegate(_:to:)
+            self._currentDelegateFor = delegateProxy._currentDelegate
+            self._setCurrentDelegateTo = delegateProxy._setCurrentDelegate
 
             MainScheduler.ensureExecutingOnScheduler()
             #if TRACE_RESOURCES
--- a/RxCocoa/Common/DelegateProxy.swift
+++ b/RxCocoa/Common/DelegateProxy.swift
@@ -171,7 +171,7 @@
                 return
             }
 
-            guard (forwardToDelegateResponds(to: selector) || voidDelegateMethodsContain(selector)) else {
+            guard ((self._forwardToDelegate?.responds(to: selector) ?? false) || voidDelegateMethodsContain(selector)) else {
                 rxFatalError("This class doesn't respond to selector \(selector)")
             }
         }
--- a/RxCocoa/Common/DelegateProxyType.swift
+++ b/RxCocoa/Common/DelegateProxyType.swift
@@ -74,7 +74,7 @@ Since RxCocoa needs to automagically create those Proxys and because views that 
 */
 public protocol DelegateProxyType: class {
     associatedtype ParentObject: AnyObject
-    associatedtype Delegate: AnyObject
+    associatedtype Delegate
     
     /// It is require that enumerate call `register` of the extended DelegateProxy subclasses here.
     static func registerKnownImplementations()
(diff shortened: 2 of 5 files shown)


kzaher:
@ufosky @RamwiseMatt could you please check out does this PR solve your issue? It seems to me it should.

kzaher:
@tarunon I think we could also investigate what would it take for us to introduce a separate class for Swift delegate proxies so we can investigate do we need to break some public API to do this.

Do you have time to try to do that, or shall I?

RamwiseMatt:
I can confirm that this fixes it. Many thanks!

kzaher:
I'll add some comments to this PR that explain why certain this were done this way, and then merge this.

tarunon:
@kzaher I was planning to separate classes for pure Swift Delegate past, but I investigated around DelegateProxy implementation and behavior, and now I think that the current implementation is fine.

We can organize Delegates that can be implemented in RxSwift users. These mean DelegateProxy's generics parameter.

1. `@objc protocol MyDelegate: NSObjectProtocol`
Should be everything is ok. `DelegateProxy<P: AnyObject, D: NSObjectProtocol>` can pass compile.

2. `protocol MyDelegate: NSObjectProtocol`
Should be everything is ok. `DelegateProxy<P: AnyObject, D>` can pass compile, but others cannot.

3. `@objc protocol MyDelegate: class`
Should be everything is ok if `forwardToDelegate` implementation is a NSObject.  `DelegateProxy<P: AnyObject, D: AnyObject>` can pass compile.
if `forwardToDelegate` is not subclass of NSObject, this line does not work. https://github.com/tarunon/RxSwift/blob/cbd82296d5d1b36a8babab9b10fce287ecd4b751/RxCocoa/Common/DelegateProxy.swift#L174
Regardless whether `forwardToDelegate` is subclass of NSObject, if `forwardToDelegate` has implementation of some of `optional func`, then seems the method forwarding work well. 

4. `protocol MyDelegate: class`
Should be everything is ok if `forwardToDelegate` implementation is a NSObject.  `DelegateProxy<P: AnyObject, D>` can pass compile, but others cannot.
There are same problem as case 3.

5. `protocol MyDelegate`
Should be everything is ok if `forwardToDelegate` implementation is a NSObject.  `DelegateProxy<P: AnyObject, D>` can pass compile.
There are same problem as case 3.
if `forwardToDelegate` is not class, maybe this assertion will be failure when install proxy.
https://github.com/tarunon/RxSwift/blob/cbd82296d5d1b36a8babab9b10fce287ecd4b751/RxCocoa/Common/DelegateProxyType.swift#L205-L208

6. `struct MyDelegate`
Should be everything is ng.
`DelegateProxy<P: AnyObject, D>` can pass compile, but it have no meaning.


So, most important thing is `forwardToDelegate` is NSObject or not, and problem is https://github.com/tarunon/RxSwift/blob/cbd82296d5d1b36a8babab9b10fce287ecd4b751/RxCocoa/Common/DelegateProxy.swift#L174 does not work well.

tarunon:
@kzaher 
I try to fix this line, https://github.com/tarunon/RxSwift/blob/cbd82296d5d1b36a8babab9b10fce287ecd4b751/RxCocoa/Common/DelegateProxy.swift#L174 
And add some of test cases for it. please give me 1 day.

tarunon:
It should be done, but... this workaround is super ugly. 😞

kzaher:
Hi @tarunon ,

I think we would ideally want to ship something that improves this behavior in a week or so.

The thing that worries me the most is that if we remove this conformance, what will exactly be in cases 4, 5 and 6. 

Here are the open questions:
* It looks to me that now we have 3 cases:
    * delegate isn't even `AnyObject` -> fatal error on `setForwardToDelegate` immediately 
    * delegate is `AnyObject` but not inheriting from `NSObject` -> in those cases `methodInvoked` and `sentMessage` won't work. 
    * delegate is `NSObject` -> all methods should work
* We also need to fatal error in `methodInvoked` and `sentMessage` in cases `delegate` is not `NSObject`. We need to document this behavior and add asserts because otherwise users will complain that the API doesn't work, then we'll need to explain to them that this API works only for `NSObject`.
* Adding those fatal errors is a bit problematic since `_forwardToDelegate` can be set after `methodInvoked`/`sentMessage` are being called or subscribed to, so we should also fatal error in case non `NSObject` is being set and there are some `methodInvoked`/`sentMessage` observers registered.

kzaher:
@tarunon 

> if forwardToDelegate is not subclass of NSObject, this line does not work. https://github.com/tarunon/RxSwift/blob/cbd82296d5d1b36a8babab9b10fce287ecd4b751/RxCocoa/Common/DelegateProxy.swift#L174

I'm a bit confused with this and 9c4b0c2. If it's non `NSObject`, then why would somebody be able to use `sentMessage` APIs?
Can you maybe explain that use case?

tarunon:
@kzaher 
Maybe from compatibility between Swfit and Objective-C.
We can check it in swift source. https://github.com/apple/swift/blob/f3bd55eb529d909b127790ad05fd1b896b24d9a0/stdlib/public/runtime/SwiftObject.h#L49

So... we cannot cast `Delegate(not NSObject)` to NSObject, but can use NSObject compatibility method from `AnyObject`. world... 

Anyway, if someone want to use `methodInvoked`/`sentMessage`, they need use selector.
Then they should add `@objc` keyword protocol or method.
Seems everything is fine.

kzaher:
Hi @tarunon ,

I'm not sure I understand why would we want for `respondsToSelector`, method forwarding and `messageSent`/`methodInvoked` to work for non objc classes?

Do you have some use case for it personally?

I got an impression that @RamwiseMatt and ufosky were advocating to just enable delegate proxy extension mechanism and support for manual observing/forwarding for Swift code.

tarunon:
Hi @kzaher ,

That is a matter of possibility. We cannot set type condition on Delegate, so I think it's necessary considering someone to use it.

And we have two implementation of calling `forwardToDelegate.responds(to:)` in master branch.

https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/Common/DelegateProxy.swift#L174
https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/Common/DelegateProxy.swift#L217

If I use forwardToDelegate as Not NSObject class, 1 is work, and 1 is not work.

We no longer have a condition in Delegate, so I think it's better to unify them to either. 
I used casting to AnyObject, because that work well on test case.

kzaher:
Hi @tarunon ,

> https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/Common/DelegateProxy.swift#L174
> https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/Common/DelegateProxy.swift#L217

I think you were right about these inconsistencies, we need to improve that.

I've cherry picked your first commit and made some changes here:
https://github.com/ReactiveX/RxSwift/tree/tarunon-fix/delegateProxy_without_objc_keyword

Please let me know do you see any issues with that implementation.

tarunon:
@kzaher It's fine for me, thank you, and we don't need this pr. I close it.

PR #1467, opened 2017-10-21
Add public keyword for workaround helper functions

ZombieWang:
I found that helper functions such as `castOrThrow<T>(_:_:) throws -> T` cannot not be accessed outside of the RxCocoa module.
Since they are global helper functions inside the RxCocoa module, I added public keyword for them.

By doing so, developers can access these helper functions when they are making their own delegate proxy classes.


kzaher:
Hi @ZombieWang ,

yeah, these functions are internal helpers and they aren't intended for `public` usage. Sorry.

They are here for cosmetics only and some are for workarounds around compiler bugs.

PR #1475, opened 2017-10-26
Add extension to bind Observable to keyPath to Variable's property

crazytonyli:
It can be helpful if there is an extension function that takes element from Observable's event sequence and set it to `Variable.E`'s property.


kzaher:
Hi @crazytonyli ,

I would rather not include this for now:
* it's easy to add this if somebody needs it
* we don't want to encourage people to use variables

PR #1483, opened 2017-11-06
Removing wrong type checking line in DelegateProxy

tarunon:
I'm sorry it was my mistaken, 
This line comes from Swift's bug 🐛 
`DelegateProxy.self` is not instance of `Delegate`.

I reported this behavior, https://bugs.swift.org/browse/SR-6304
And we'd remove this line from master branch.
Could you please review it? @kzaher

### DIFF
--- a/RxCocoa/Common/DelegateProxyType.swift
+++ b/RxCocoa/Common/DelegateProxyType.swift
@@ -388,7 +388,6 @@ extension DelegateProxyType where ParentObject: HasDataSource, Self.Delegate == 
         fileprivate func extend<DelegateProxy: DelegateProxyType, ParentObject>(make: @escaping (ParentObject) -> DelegateProxy) {
                 MainScheduler.ensureExecutingOnScheduler()
                 precondition(_identifier == DelegateProxy.identifier, "Delegate proxy has inconsistent identifier")
-                precondition((DelegateProxy.self as? DelegateProxy.Delegate) != nil, "DelegateProxy subclass should be as a Delegate")
                 guard _factories[ObjectIdentifier(ParentObject.self)] == nil else {
                     rxFatalError("The factory of \(ParentObject.self) is duplicated. DelegateProxy is not allowed of duplicated base object type.")
                 }


kzaher:
Hi @tarunon ,

could you maybe provide some example that demonstrates when the current solution fails? I'm a bit confused because I haven't seen that assertion firing.

I've tried to check out your bug report, but it's not clear to me:
* which failure is the one I should be looking at?
* why it worked so far and why haven't I seen any crashes?

tarunon:
This assertion condition is wrong. Correctly `DelegateProxy.Delegate.self is DelegateProxy.Delegate.Type`, but it was fail maybe swift bug.
If we use `@objc` protocol or class conditioning protocol, `DelegateProxy.Delegate.self is DelegateProxy.Delegate.Type` will be false, and instead `DelegateProxy.Delegate.self is DelegateProxy.Delegate` will be true.

> why it worked so far and why haven't I seen any crashes?

It's now fine because Delegate pattern almost use `@objc protocol` or `protocol *: class`. 

> which failure is the one I should be looking at?

If someone use common protocol as `Delegate`, then this assertion will be fail.
```swift
// Example
protocol MyDelegate {}
class MyClass {
  var proxy: MyDelegate?
}
class MyDelegateProxy: DelegateProxy<MyClass, MyDelegate>, DelegateProxyType, MyDelegate {
  ...
}
```

kzaher:
Hi @tarunon ,

Should we allow non `@objc protocol or protocol *: class.` though?
It seems to me that semantically it doesn't make sense and it will case runtime exception. 

Wouldn't it be better if we crashed sooner and provide some meaningful information instead of just removing this?

tarunon:
> Wouldn't it be better if we crashed sooner and provide some meaningful information instead of just removing this?

That make sense, but I'm not sure the way to checking the `Delegate` is `@objc protocol` or `protocol *: class`.
Now our solution uses swift bug, I think we'd avoid it and replace new solution.

How about checking it in `DelegateProxy.init` time? I think we can check same condition and avoid uses swift bug.

PR #1486, opened 2017-11-09
Add PublishRelay.asDriver()

AndrewSB:
This seems useful to me, since it's implied that a PublishRelay can't error. So a clean way to convert into a `Driver` was something I implemented for my own convenience

### DIFF
--- a/RxCocoa/Traits/PublishRelay.swift
+++ b/RxCocoa/Traits/PublishRelay.swift
@@ -37,4 +37,11 @@ public final class PublishRelay<Element>: ObservableType {
     public func asObservable() -> Observable<Element> {
         return _subject.asObservable()
     }
+
+    /// Warning: It is assumed this is called on the main thread; since Drivers require that.
+    public func asDriver() -> Driver<Element> {
+        return _subject.asDriver(onErrorRecover: { error in
+            rxFatalError("A PublishRelay shouldn't be able to emit an error: \(error)")
+        })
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @AndrewSB ,

yeah, `Driver` is `share(replay: 1)` vs `share()` that `PublishRelay` actually is.

It's better to use `PublishRelay` -> `Signal` since they have `share()` strategy.

orakaro:
@kzaher Is this the same reason that there is only [drive to BehaviourRelay](https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/Traits/Driver/Driver%2BSubscription.swift#L50) but no "drive to PublishRelay" method ?

kzaher:
@orakaro yes

AndrewSB:
Ahh, so if I want a Driver I should be using a `BehaviourRelay`. Cool. I'll amend this to be `toSignal()` instead

orakaro:
I think we are already have it [here](https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/Traits/Signal/PublishRelay+Signal.swift) 😄

AndrewSB:
Oh, perfect 😄

kzaher:
@gobetti Driver/drive represent state, and those extensions were made for binding state, so they make sense for drive. There aren't currently any scenarios where using emit makes sense with that form.

PR #1490, opened 2017-11-13
Use Signals for tapping events instead of Drivers

Kuniwak:
Driver does not fit to some UI events such as tapping. Because we do not expect to get past "tap" events.

But Driver is using `share(replay: 1)`. So it may replay accidentally the past event for some conditions. This replay is useful only for observables like Variables such as `rx.text`, but `rx.tap` is not Variable like.

BTW, Signal can fit the usecase. Signal is using `.share()`, so it does not replay past tapping events.


Kuniwak:
Oops I have a mistake about target branch...

Kuniwak:
I'm sorry, I opened the new PR that have a correct target branch at #1491...

PR #1493, opened 2017-11-15
Fix bad performance caused by stacktrace

sue71:
`Thread.callStackSymbols` takes too much time.
So, I want to call the method only when unhandled error is thrown.

### DIFF
--- a/RxSwift/ObservableType+Extensions.swift
+++ b/RxSwift/ObservableType+Extensions.swift
@@ -50,7 +50,7 @@ extension ObservableType {
                 
                 let synchronizationTracker = SynchronizationTracker()
 
-                let callStack = Thread.callStackSymbols
+                let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []
 
                 let observer = AnonymousObserver<E> { event in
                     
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @sue71 ,

that's a wrong call stack to capture. Can you maybe explain what troubles are you experiencing? 

There is also a release build that doesn't capture call stack.

devxoul:
@kzaher, I had a same problem: It takes too much time when the subscriptions are made in reusable cells.

screen shot 2017-11-16 at 5 28 13 am

Alternatively we could consider using low-level C API to bring call stacks as a `\n`-joined string rather than an array of string. It seems that a bridging C string array into a Swift string array takes long time:

screen shot 2017-11-16 at 5 38 27 am

I've been done with this at UINavigationItem-Margin (https://github.com/devxoul/UINavigationItem-Margin/commit/fa00009b2db12547277f189ec83c8a9b1a4c575a) but have no idea where to put C code. (Swift Package Manager requires a single language per target so it will be messed up 🤔_

kzaher:
Hi @devxoul @sue71 ,

maybe we can just have some low tech solution:
```
public enum Hooks {
    public static var recordCallStack: Bool = false
}
```

and add `let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []`.

Seems to me that could work.

devxoul:
@kzaher I like that. @sue71, can you edit this PR?

sue71:
Hi @kzaher, thx for the response. that's nice solution. I'll edit this PR.
@devxoul thx for the detailed explanation!

devxoul:
Thanks :+1:

PR #1500, opened 2017-11-22
Add distinct operator to ObservableType

SiarheiFedartsou:
Hi! I was surprised that there is still no `distinct` operator in RxSwift and decided to fill this gap.

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -199,6 +199,7 @@ custom_categories:
   - Delay
   - DelaySubscription
   - Dematerialize
+  - Distinct
   - DistinctUntilChanged
   - Do
   - ElementAt
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file.
 
 ## Master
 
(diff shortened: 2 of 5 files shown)


carlosypunto:
Hi, @SiarheiFedartsou, maybe you don't know about [RxSwiftExt: A collection of Rx operators & tools not found in the core RxSwift distribution](https://github.com/RxSwiftCommunity/RxSwiftExt), there you can found [distinct operator](https://github.com/RxSwiftCommunity/RxSwiftExt/blob/master/Source/RxSwift/distinct.swift)

carlosypunto:
@SiarheiFedartsou thanks for your effort. But, as you can see in the intention of that repository, the RxSwift core is being intentionally kept as compact as possible to avoid bloat. In any case, the big contributors and members, may they want to give you another answer.

SiarheiFedartsou:
@carlosypunto yes, I really didn’t know about RxSwiftExt, but almost all other ReactiveX implementations have such operator by default.

SiarheiFedartsou:
@kzaher what do you think?

kzaher:
@SiarheiFedartsou I agree with everything @carlosypunto said. 

We would need a really good reason why we would want to include some operator in this repo.

PR #1503, opened 2017-11-23
Added forwardingTarget method to be used as a fallback

zulkis:
Used standard forwarding method forwardingTargetForSelector: from NSObject in order to forward unrecognized messages first directly to the original delegate.

### DIFF
--- a/RxCocoa/Common/DelegateProxy.swift
+++ b/RxCocoa/Common/DelegateProxy.swift
@@ -229,6 +229,10 @@
                 || (self._forwardToDelegate?.responds(to: aSelector) ?? false)
                 || (self.voidDelegateMethodsContain(aSelector) && self.hasObservers(selector: aSelector))
         }
+        
+        override open func forwardingTarget(for aSelector: Selector!) -> Any? {
+            return self.forwardToDelegate()
+        }
 
         fileprivate func reset() {
             guard let parentObject = self._parentObject else { return }


zulkis:
Opened as an output from
https://github.com/pixeldock/RxAppState/issues/11

zulkis:
Interesting that tests are failing with such a changes...

kzaher:
Hi @zulkis ,

we already have forwarding implemented.

```
-(void)forwardInvocation:(NSInvocation *)anInvocation {
    BOOL isVoid = RX_is_method_signature_void(anInvocation.methodSignature);
    NSArray *arguments = nil;
    if (isVoid) {
        arguments = RX_extract_arguments(anInvocation);
        [self _sentMessage:anInvocation.selector withArguments:arguments];
    }
    
    if (self._forwardToDelegate && [self._forwardToDelegate respondsToSelector:anInvocation.selector]) {
        [anInvocation invokeWithTarget:self._forwardToDelegate];
    }

    if (isVoid) {
        [self _methodInvoked:anInvocation.selector withArguments:arguments];
    }
}
```

so I'm not sure why did you make that change.

Could you maybe provide a short example that demonstrates what issue are you trying to fix and how does this fix it?

freak4pc:
I'm closing this for the time being. I agree with @kzaher this is a bit of a  strange addition given we already provide forwarding capabilities. That added with no reply for over 2 years feels like a good time to close this. Thank you!

PR #1509, opened 2017-11-27
Set Content-Type header to application/json

ali-l:
So that the server recognizes the request body as JSON

Noticed this when sending a PUT request to a Rails backend. The server ignores the body when this header is not set.


ali-l:
The code is wrong, but what do you think about the idea of setting the header by default?

kzaher:
Hi @ali-l ,

I don't think we should be opinionated about types of calls we make. User can use `request` object to set the needed headers.

Issue #1508, opened 2017-11-27
[Poposal] Replace global func with static func to create Recorded Event in RxTest 


beeth0ven:
Hi all~~

We often use global functions `next`, `error`, `completed`  to create `Recorded<Event<T>>` when testing:

```swift
//  testElementAt_First
...
let xs = scheduler.createHotObservable([
    next(70, 6),  
    next(150, 4),
    next(210, 9),
    next(230, 13),
    next(270, 7),
    next(280, 1),
    next(300, -1),
    next(310, 3),
    next(340, 8),
    next(370, 11),
    completed(400)
    ])

let res = scheduler.start {
    xs.elementAt(0)
}

XCTAssertEqual(res.events, [
    next(210, 9),    
    completed(210)
    ])
...
```

Is it better to replace the following global functions:

```swift
public func next<T>(_ time: TestTime, _ element: T) -> Recorded<Event<T>> { ... }

public func completed<T>(_ time: TestTime, _ type: T.Type = T.self) -> Recorded<Event<T>> { ... }

public func error<T>(_ time: TestTime, _ error: Swift.Error, _ type: T.Type = T.self) -> Recorded<Event<T>> { ... }
```

 with `Recorded` static functions:

```swift
extension Recorded {

    public static func next<T>(_ time: TestTime, _ element: T) -> Recorded<Event<T>> 
        where Value == Event<T> { ... }

    public static func completed<T>(_ time: TestTime, _ type: T.Type = T.self) -> Recorded<Event<T>>
        where Value == Event<T> { ... }

    public static func error<T>(_ time: TestTime, _ error: Swift.Error, _ type: T.Type = T.self) -> Recorded<Event<T>>
        where Value == Event<T> { ... }
}
```

then the original test case becomes:

```swift
//  testElementAt_First
...
let xs = scheduler.createHotObservable([
    .next(70, 6),  
    .next(150, 4),
    .next(210, 9),
    .next(230, 13),
    .next(270, 7),
    .next(280, 1),
    .next(300, -1),
    .next(310, 3),
    .next(340, 8),
    .next(370, 11),
    .completed(400)
    ])

let res = scheduler.start {
    xs.elementAt(0)
}

XCTAssertEqual(res.events, [
    .next(210, 9),    
    .completed(210)
    ])
...
```

We just need to add `.` before each global function, this seems more Swifty.

kzaher:
Hi @beeth0ven ,

yes, I think it makes sense.

[...]

PR #1510, opened 2017-11-30
Replaces global functions next, error, completed with Recorded.next, Recorded.error, Recorded.completed in RxText

beeth0ven:
Hi all, 

This proposal is from #1508:

```diff
//  testElementAt_First
...
let xs = scheduler.createHotObservable([
-    next(70, 6),  
-    next(150, 4),
-    next(210, 9),
-    next(230, 13),
-    next(270, 7),
-    next(280, 1),
-    next(300, -1),
-    next(310, 3),
-    next(340, 8),
-    next(370, 11),
-    completed(400)
+    .next(70, 6),  
+    .next(150, 4),
+    .next(210, 9),
+    .next(230, 13),
+    .next(270, 7),
+    .next(280, 1),
+    .next(300, -1),
+    .next(310, 3),
+    .next(340, 8),
+    .next(370, 11),
+    .completed(400)
    ])

let res = scheduler.start {
    xs.elementAt(0)
}

XCTAssertEqual(res.events, [
-    next(210, 9),    
-    completed(210)
+    .next(210, 9),    
+    .completed(210)
    ])
...
```

Deprecated global functions `next`, `error`, `completed` in RxText:

```swift
public func next<T>(_ time: TestTime, _ element: T) -> Recorded<Event<T>> { ... }

public func completed<T>(_ time: TestTime, _ type: T.Type = T.self) -> Recorded<Event<T>> { ... }

public func error<T>(_ time: TestTime, _ error: Swift.Error, _ type: T.Type = T.self) -> Recorded<Event<T>> { ... }
```

in favor of `Recorded` static functions:

```swift
extension Recorded {

    public static func next<T>(_ time: TestTime, _ element: T) -> Recorded<Event<T>> 
        where Value == Event<T> { ... }

    public static func completed<T>(_ time: TestTime, _ type: T.Type = T.self) -> Recorded<Event<T>>
        where Value == Event<T> { ... }

    public static func error<T>(_ time: TestTime, _ error: Swift.Error, _ type: T.Type = T.self) -> Recorded<Event<T>>
        where Value == Event<T> { ... }
}
```

Much more swifty.


beeth0ven:
CI failed.

 I will try to fix it.

beeth0ven:
Ok, It's ready for review. Really a big PR 🤔.

RxTest/Deprecated.swift
```diff
@@ -19,3 +19,43 @@ extension TestScheduler {
...
+ */
+//extension XCTestCase {
+    /**
+     Factory method for an `.next` event recorded at a given time with a given value.
+ 
+     - parameter time: Recorded virtual time the `.next` event occurs.
+     - parameter element: Next sequence element.
+     - returns: Recorded event in time.
+     */
+    @available(*, deprecated, renamed: "Recorded.next(_:_:)")
```
kzaher: I don't think we should add `deprecated` attribute at this time because it will create a barrier for updating to next release of RxSwift since this is a pretty big change.

I would like to add this attribute near the end of 4.x lifecycle since this is mostly cosmetic change.

I'm ok with moving then to `Deprecated.swift`.

RxTest/Recorded.swift
```diff
@@ -32,6 +32,40 @@ extension Recorded {
...
+extension Recorded {
+    
+    /**
+     Factory method for an `.next` event recorded at a given time with a given value.
+     
+     - parameter time: Recorded virtual time the `.next` event occurs.
+     - parameter element: Next sequence element.
+     - returns: Recorded event in time.
+     */
+    public static func next<T>(_ time: TestTime, _ element: T) -> Recorded<Event<T>> where Value == Event<T> {
```
kzaher: Wouldn't it be more consistent with api guidelines if we used:
`public static func next<T>(time: TestTime, element: T) -> Recorded<Event<T>> where Value == Event<T>`
`public static func completed<T>(time: TestTime, type: T.Type = T.self) -> Recorded<Event<T>> where Value == Event<T>`
`public static func error<T>(time: TestTime, error: Swift.Error, type: T.Type = T.self) -> Recorded<Event<T>> where Value == Event<T> `

?

We can use regex to find replace all of the places.

beeth0ven: Yes, I totally agree. 👍

I've added the parameter label as suggested:

```swift
// This is call side

// Before:

let recordedEvents: [Recorded<Event<Int>>] = [
    next(205, 1),
    next(210, 2),
    next(240, 3),
    error(500, testError),
    completed(600)
]

// After:

let recordedEvents: [Recorded<Event<Int>>] = [
    .next(time: 205, element: 1),
    .next(time: 210, element: 2),
    .next(time: 240, element: 3),
    .error(time: 500, error: testError),
    .completed(time: 600)
]

``` 

Then i have a new idea if we can remove the `element`, `error` parameter label:

```swift

// Before:

let recordedEvents: [Recorded<Event<Int>>] = [
    .next(time: 205, element: 1),
    .next(time: 210, element: 2),
    .next(time: 240, element: 3),
    .error(time: 500, error: testError),
    .completed(time: 600)
]

// After:

let recordedEvents: [Recorded<Event<Int>>] = [
    .next(time: 205, 1),
    .next(time: 210, 2),
    .next(time: 240, 3),
    .error(time: 500, testError),
    .completed(time: 600)
]

```

This may add readability for test purpose. And will align the api when create `Event`:

```swift
let events: [Event<Int>] = [
    .next(1),
    .next(2),
    .next(3),
    .error(testError),
    .completed
]

let recordedEvents: [Recorded<Event<Int>>] = [
    .next(time: 205, 1),
    .next(time: 210, 2),
    .next(time: 240, 3),
    .error(time: 500, testError),
    .completed(time: 600)
]
```

Is this idea reasonable? 🤔

RxTest/Deprecated.swift
beeth0ven: Got it. 

It's better to treat this like `Variable` vs `BehaviorRelay` because a lot of people were using it.

RxTest/Recorded.swift
fpillet: Just my $.02: adding a named `time` parameter will just make it more tedious to write the hundreds of such lines that need to be written for tests. Writing tests is long enough, I don't think we need this useless burden added.

beeth0ven: Hi @fpillet Thanks for your valuable suggestions.

Now, There are three choices here:

1. function  with all named parameters:

    ```swift
    let recordedEvents: [Recorded<Event<Int>>] = [
        .next(time: 205, element: 1),
        .next(time: 210, element: 2),
        .next(time: 240, element: 3),
        .error(time: 500, error: testError),
        .completed(time: 600)
    ]
    ````

2. function  with just named `time` parameter:

    ```swift
    let recordedEvents: [Recorded<Event<Int>>] = [
        .next(time: 205, 1),
        .next(time: 210, 2),
        .next(time: 240, 3),
        .error(time: 500, testError),
        .completed(time: 600)
    ]
    ````

3. function  without any named parameter:

    ```swift
    let recordedEvents: [Recorded<Event<Int>>] = [
        .next(205, 1),
        .next(210, 2),
        .next(240, 3),
        .error(500, testError),
        .completed(600)
    ]
    ````

And option 3 seems to be light burden when writing test code.
@kzaher what do you think of it? 🦄

fpillet: I know the conventions mandates naming parameters, but the sheer volume of such calls that need to be written for tests make it somewhat impractical to follow convention in this case... thanks!

kzaher: I'm fine with option 3 since it's practical, although I'm not sure it's according to swift guidelines.

beeth0ven:
Thanks @kzaher 

I've removed deprecate attribute and kept option 3 selected.

RxTest/Recorded.swift
freak4pc: I think specifically for testing things option 3 makes more sense. 

The other options constitute so much boilerplate in this case where you have to repeatedly make use of this method. 

I think Guidelines, like any others, can be steered away from if there is a clearly good reason to, like this one :) just my 2 cents

freak4pc:
Really looking forward to a release including this ! 
The only annoying thing is we'll have to start adding type definition on groups of tests, but i'm all for getting rid of the global function. Thanks @beeth0ven !

kzaher:
@freak4pc 

I think we can create an extension that enables something as this:
```
Recorded(
    .next(200, 1),
    .next(201, 2),
    .next(202, 3),
)
```

freak4pc:
@kzaher true, that might clear this kind of code a bit. 
Might be better to add something around this before a new version? As it might add a lot of work for people otherwise (since this deprecates the global method which were the only way until now)

kzaher:
@freak4pc Agree.

I think we'll handle this something like variable. Since this was common publicly used API, we'll officially add deprecation warning in RxSwift 6.0, and then potentially remove it in RxSwift 7.0.

freak4pc:
@kzaher Sounds good. 

I would actually be in favor of deprecating officially in 5 and removing in 6.

2 release deprecation cycle is way more than a general LTS, it means people will have "deprecated" code for 2 Major Releases (which are generally 2 years) until a real deprecation is released. Maybe worth doing a yearly deprecation cycle? 

RxSwift 4 - Feature moved to Deprecated
RxSwift 5 - Feature marked with @deprecated 
RxSwift 6 - Feature removed completely 

Anyways I'll leave it up to you of course, it's just an opinion 😄

kzaher:
If we are changing an api that was here for 4 major versions just for cosmetics sake, I would like to avoid causing 7k warnings in people projects when they upgrade. 

We can maybe add the warning at the end of 5, but not start of 5 for sure.

I would like to add warnings in inline docs that this api will be changed in favor of new api.

beeth0ven:
Thanks for reviewing @freak4pc @fpillet @kzaher ☺️.

@freak4pc I'm glad you like this idea.

As @kzaher  suggested, we still have many improvements here：

1. Add warnings in inline docs to encourage ppl use new api: `.next(200, 20)`

2. Add `Recorded<Event<T>>` array factory method so that we don't need to define type:
 
    ```diff

    - let correctMessages: [Recorded<Event<Int>>] = [
    -     .next(210, 2),
    -     .next(220, 3),
    -     .next(230, 4),
    -     .next(240, 5),
    -     .completed(250)
    - ]

    + let correctMessages = Recorded.events(
    +     .next(210, 2),
    +     .next(220, 3),
    +     .next(230, 4),
    +     .next(240, 5),
    +     .completed(250),
    + )
    ```
3. Replace `Recored Event` [timeless global functions](https://github.com/ReactiveX/RxSwift/blob/master/Tests/Recorded%2BTimeless.swift) with static func in **Tests**:

    ```diff
    - func next<T>(_ value: T) -> Recorded<Event<T>> { ... }
    -
    - func completed<T>() -> Recorded<Event<T>> { ... }
    -
    - func error<T>(_ error: Swift.Error) -> Recorded<Event<T>> { ... }

    + extension Recorded {
    +
    +     public static func next<T>(_ element: T) -> Recorded<Event<T>>
    +         where Value == Event<T> { ... }
    +
    +     public static func completed<T>(_ time: TestTime, _ type: T.Type = T.self) -> Recorded<Event<T>>
    +         where Value == Event<T> { ... }
    +
    +     public static func error<T>(_ time: TestTime, _ error: Swift.Error, _ type: T.Type = T.self) -> Recorded<Event<T>>
    +         where Value == Event<T> { ... }
    + }
    ```

What do you think of these suggestions? 🙂

kzaher:
@beeth0ven currently yes for 2 and 3. 1 in future.

PR #1511, opened 2017-12-03
Avoid force cast

shoheiyokoyama:
Hi :sunglasses:
I used `castOrThrow` for force cast.


shoheiyokoyama:
@sergdort @kzaher 
Thanks!! :satisfied:

PR #1518, opened 2017-12-05
Search controller delegate

bobgodwinx:
### Review UISearchControllerDelegate
This #PR is related to a problem that I had while working with `UISearchController` all it's delegate methods where returning a `Void` type and this I couldn't do something like this: 
```swift 
viewController.searchController.rx.didPresent.bind(to: didPresentSearch).disposed(by: bag)
```
But when you actually look at the `UISearchControllerDelegate` it does return the `UISearchController` with all it's invocation. 
So I have decided to return it based on the `UIKit` implementation.  If someone still wants only the `Void` type from the events they can always use a map to `Void`. Let me know what you think. Thanks


kzaher:
Hi @bobgodwinx ,

The fact that it doesn't return `Observable<UISearchController>` is by design. You want to abstract away the source of events. Otherwise you haven't abstracted anything. If you need `UI*` that means that you are tight coupling your logic to UIKit.

PR #1528, opened 2017-12-16
Fix observeWeakly for property in category like UIView.frame

sora0077:
In GettingStarted, can observe UIView.frame, but actually occurs `RxCocoaError.invalidPropertyName`

```swift
view
  .rx.observeWeakly(CGRect.self, "frame")
  .subscribe(onNext: { frame in
    ...
  })
```

> https://github.com/ReactiveX/RxSwift/blob/master/Documentation/GettingStarted.md#kvo

### DIFF
--- a/RxCocoa/Foundation/NSObject+Rx.swift
+++ b/RxCocoa/Foundation/NSObject+Rx.swift
@@ -461,16 +461,27 @@ fileprivate final class KVOObservable<Element>
         weak var weakTarget: AnyObject? = target
 
         let propertyName = keyPathSections[0]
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @sora0077 ,

this seems like a really weird solution to me. The return type of method has different semantics than `property_getAttributes`.

Isn't this Apple bug?

freak4pc:
Any reply to @kzaher's comment here, @sora0077 ?

freak4pc:
Trying one more time ... @sora0077

PR #1529, opened 2017-12-18
[Feature] - OfType Operator

glwithu06:
- Implement `ofType` operator using type casting operator (`as`) in Swift
- Add `ofType` extension methods on `ObservableType`
- Create `OfType`, `OfTypeSink`class

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -206,6 +206,7 @@ custom_categories:
   - Enumerated
   - Error
   - Filter
+  - OfType
   - First
   - Generate
   - GroupBy
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -206,7 +206,6 @@ custom_categories:
   - Enumerated
   - Error
   - Filter
-  - OfType
   - First
   - Generate
   - GroupBy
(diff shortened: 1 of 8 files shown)


kzaher:
Hi @glwithu06 ,

did you need this functionality in some practical use case?

glwithu06:
Hi @kzaher 

Every ViewModel (I called it Binder) has Observable<UIEvent> as input and Observable<Display> as output in my architecture.
So, I need to handle every event by its type.
It is illustrated as follows.
screen shot 2017-12-21 at 10 43 29 am

Of course, it can be implemented with filter and map.
But I think this is better than combining two operators and `ofType` has been already implemented in others (RxKotlin, RxJava, RxGroovy, RxNet)

kzaher:
Hi @glwithu06 ,

Maybe a better place to put this functionality would be in RxSwiftExt. @fpillet Thoughts?

We try to minimize the number of operators inside this library.

glwithu06:
Yeah, I agree with that.
But I think the operator that is introduced in the webpage(http://reactivex.io/documentation/operators.html) should be implemented here.

kzaher:
Hi @glwithu06 ,

The union of all of the operators on http://reactivex.io/documentation/operators.html isn't in any way related with the number of operators that we plan to ship by default in this library.

Issue #1532, opened 2017-12-20
Extend global Hook to Traits


lucas34:
Hi,

Currently in RxSwift 4.0.0, the global Hook handle is only called for ObservableType.

The documentation for "Hook" is "/// Error handler called in case onError handler wasn't provided.".
But when I implement the following code 

```
Hooks.defaultErrorHandler = { callback, error in
    print("handled")
}

_ = Observable<String>.error(ApiError.unexpected).subscribe() // print
 _ = Single<String>.error(ApiError.unexpected).subscribe()
 _ = Maybe<String>.error(ApiError.unexpected).subscribe()
 _ = Completable.error(ApiError.unexpected).subscribe()
```

None of the observers implements onError. However, the Hook will only get called for the Observable. Why not forwarding error for Single, Maybe and Completable to Hook ? Otherwise, I think you should update the documentation for Hook to make it more explicit.

To workaround this I'm doing sequence.asObservable().subscribe()

kzaher:
Hi @lucas34 ,

yes, we should also add `Hooks.defaultErrorHandler` to `Single`, `Maybe`, `Completable`.

PR #1533, opened 2017-12-21
Implement error Hook for Traits

lucas34:
Link to issue #1532

Hi,

I tried to implement error Hook for Rx.Traits: Single, Maybe and Completable.
It's inspired by the implementation of error Hook for Observable. However, I'm not sure about the usage of SynchronizationTracker and if I should also add it. If it's not required I'll remove it.

Need to add
- [x] Changelog
- [x] Tests

### DIFF
--- a/RxSwift/Traits/Completable.swift
+++ b/RxSwift/Traits/Completable.swift
@@ -6,6 +6,10 @@
 //  Copyright © 2017 Krunoslav Zaher. All rights reserved.
 //
 
+#if DEBUG
+import Foundation
+#endif
+
 /// Sequence containing 0 elements
 public enum CompletableTrait { }
 /// Represents a push style sequence containing 0 elements.
@@ -75,10 +79,20 @@ public extension PrimitiveSequenceType where TraitType == CompletableTrait, Elem
      - returns: Subscription object used to unsubscribe from the observable sequence.
      */
     public func subscribe(onCompleted: (() -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil) -> Disposable {
+        #if DEBUG
+                let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []
+        #else
+                let callStack = [String]()
+        #endif
+
         return self.primitiveSequence.subscribe { event in
             switch event {
             case .error(let error):
-                onError?(error)
+                if let onError = onError {
+                    onError(error)
+                } else {
+                    Hooks.defaultErrorHandler(callStack, error)
+                }
             case .completed:
                 onCompleted?()
             }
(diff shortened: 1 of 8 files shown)


kzaher:
Hi @lucas34 ,

`Single`, `Maybe` and `Completable` can't have the same problems as general observable sequences since they only return one element, so you can't receive second element while processing the first one.

lucas34:
Alright I only added the Hook then. I also updated the changelog and added the tests. Let me know what do you think.

RxSwift/Traits/Completable.swift
```diff
@@ -75,14 +79,35 @@ public extension PrimitiveSequenceType where TraitType == CompletableTrait, Elem
...
      */
     public func subscribe(onCompleted: (() -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil) -> Disposable {
-        return self.primitiveSequence.subscribe { event in
-            switch event {
-            case .error(let error):
-                onError?(error)
-            case .completed:
-                onCompleted?()
+        #if DEBUG
+            let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []
```
kzaher: I believe we can just do
```swift
#if DEBUG
            let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []
#else
            let callStack = [String]()
#endif
```

and avoid duplication of logic.

RxSwift/Traits/Maybe.swift
```diff
@@ -84,16 +88,39 @@ public extension PrimitiveSequenceType where TraitType == MaybeTrait {
...
-        return self.primitiveSequence.subscribe { event in
-            switch event {
-            case .success(let element):
-                onSuccess?(element)
-            case .error(let error):
-                onError?(error)
-            case .completed:
-                onCompleted?()
+        #if DEBUG
+            let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []
```
kzaher: also here

RxSwift/Traits/Single.swift
```diff
@@ -77,14 +81,35 @@ extension PrimitiveSequenceType where TraitType == SingleTrait {
...
      */
     public func subscribe(onSuccess: ((ElementType) -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil) -> Disposable {
-        return self.primitiveSequence.subscribe { event in
-            switch event {
-            case .success(let element):
-                onSuccess?(element)
-            case .error(let error):
-                onError?(error)
+        #if DEBUG
+             let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []
```
kzaher: and here

lucas34:
No problem. I simplified.

PR #1538, opened 2017-12-28
Change `onNext` to `onSuccess` in `do` method in Single

SebastianOsinski:
This PR changes name of first parameter in `do` method in Single. Previously this parameter was named `onNext` which didn't match Single semantics and also was inconsistent with naming in `subscribe` method. I renamed this parameter to `onSuccess`. To provide backwards compatibility, previous version is still available but marked as deprecated.

### DIFF
--- a/RxSwift/Traits/Single.swift
+++ b/RxSwift/Traits/Single.swift
@@ -152,6 +152,34 @@ extension PrimitiveSequenceType where TraitType == SingleTrait {
 }
 
 extension PrimitiveSequenceType where TraitType == SingleTrait {
+
+    /**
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @SebastianOsinski ,

We've received some complaints about this in past so I guess we can merge this.

lucas34:
Having this two methods creates a conflict when you are not using all the callbacks
`.do(onError: { _ in })` will not compile anymore. Because the compiler don't know if you are actually referencing fo 
`.do(onNext, onError)` or `.do(onSuccess, onError)`
This is a breaking changes. (In my case)

PR #1539, opened 2017-12-28
Helping swift to infer type of 'complex' expression when used as a pod

kirillyakimovich:
Without this compiler is dying with "Ambiguous use of 'map'"

### DIFF
--- a/RxCocoa/Common/DelegateProxy.swift
+++ b/RxCocoa/Common/DelegateProxy.swift
@@ -208,8 +208,8 @@
             #endif
             self._setForwardToDelegate(delegate, retainDelegate: retainDelegate)
 
-            let sentSelectors: [Selector] = _sentMessageForSelector.values.filter { $0.hasObservers }.map { $0.selector }
-            let invokedSelectors: [Selector] = _methodInvokedForSelector.values.filter { $0.hasObservers }.map { $0.selector }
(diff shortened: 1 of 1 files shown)


kzaher:
Interesting, what compiler version are you using? This works just fine with Xcode Version 9.2 (9C40b)

RxCocoa/Common/DelegateProxy.swift
```diff
@@ -208,9 +208,9 @@
             #endif
             self._setForwardToDelegate(delegate, retainDelegate: retainDelegate)
 
-            let allUsedSelectors: [Selector] =
-                    self._sentMessageForSelector.values.filter { $0.hasObservers }.map { $0.selector } +
-                    self._methodInvokedForSelector.values.filter { $0.hasObservers }.map { $0.selector }
+            let sentSelectors: [Selector] = _sentMessageForSelector.values.filter { $0.hasObservers }.map { $0.selector }
```
kzaher: Please use `self._sentMessageForSelector` and `self._methodInvokedForSelector`.

kirillyakimovich:
Same Xcode as you: Version 9.2 (9C40b)
Trying to create minimal project when this will fail to provide some more info.

kirillyakimovich:
Ok, i found the reason my specified swift version was 3.2. Sorry for bothering. Closing

PR #1541, opened 2017-12-31
[Proposal] Class based Disposable

tarunon:
Hi, this PR is refactoring proposal making class based `Disposable`.
I refactor `Disposable` as class and remove `DisposeBase`, `Disposables`.
And I add `DisposableType` for 3rd party expansion and compatibility.

## Pros
* Improving performance. 
  Today we use protocol based `Disposable`. Seems class v-table method invocation is more efficient than protocol witness-table invocation, because todays protocol variable is type-existential.
  This is my benchmark result. `before` is master branch, `after` is this pr. `flatMap` has big impact.
  

* Don't need `Disposables` util class anymore.
  If `Disposable` is class, we can use `Disposable.create` functions. 

## Cons 
* Big PR, and breaking change.
* `DisposableType` may make bit confusing.


serges147:
I guess this will break those who:
- declare their own protocols inherited from `Disposable`
- iherit from some base class and implement Disposable as well, like:
`class X: Base, Disposable {`

serges147:
We do both in our project.

tarunon:
@serges147
Yes, I know it. And. I add `DisposableType` , I think we can use it instead. 
How about your use case?

serges147:
>>_How about your use case?_
Sorry, I'm not quite sure I get your the question.
Anyway, our use cases are the ones I described above, plus we use `SerialDisposable` & `CompositeDisposable` to dispose not only RxSwift stuff (like subscriptions and schedules) but also our own classes as well - it's very convenient.

tarunon:
@serges147 I added `DisposableType` for support making own `Disposable`.
`DisposableType` require `dispose` function, and have compatibility of `Disposable`.

```swift
class MyCustomDisposable: NSObject, DisposableType {
  func dispose() { ... }
}
```

Of course, we can use class that implementation `DisposableType` as `Disposable`,

```swift
let serialDisposable = SerialDisposable()
serialDisposable.disposable = MyCustomDisposable().asDisposable()
```

It certainly require calling `asDisposable()` but I think that's not a huge task.

kzaher:
Hi @tarunon ,

I've pushed some of the optimization attempts into https://github.com/ReactiveX/RxSwift/tree/optimizations2 branch.

I've managed to get execution of:
```swift
let subscription = Observable<Int>.createAnonymous { observer in
                    for _ in 0 ..< iterations * 10 {
                        observer.onNext(1)
                    }
                    return Disposables.create()
                    }
                    .flatMapLatest { x in Observable.just(x) }
                    .flatMapLatest { x in Observable.just(x) }
                    .flatMapLatest { x in Observable.just(x) }
                    .flatMapLatest { x in Observable.just(x) }
                    .flatMapLatest { x in Observable.just(x) }
                    .subscribe(onNext: { x in
                        sum += x
                    })

                subscription.dispose()
```

from ~`800`us (in develop) to ~`600`us (in optimizations2).

I'm open to playing around with codebase to improve performance.

I would rather if we did that into a separate branch and try to improve performance at least couple of X times. (X >= 2)

If we manage to do that in a separate branch where we don't need to worry about any backwards compatibility, then we can think about how can we port those improvements to main branch in next major version while considering backwards compatibility.

If we can't get at least couple of X improvement in optimizations2 branch without considering backwards compatibility, then I think the cost of breaking user code in next major version will outweigh the performance improvements and our efforts since nobody will probably notice them.

If you want to target this PR to optimizations2 branch or some other branch, I'm open to fresh ideas.

I've tried:
* defining observer as a `(Event<E>) -> ()` lambda
* defining observer as 3 lambdas (next, error, completed)

Haven't tried:
* making observable an enum
* making observer a class with 3 methods
* other ideas

I would suggest not doing cosmetic changes + performance updates in a single PR because it's hard to find the important changes.

E.g. Because of `Disposables.create()` -> ` Disposable.create()` it's really hard to find critical parts of code.

It is also beneficial to try minimizing the number of changes, so we can focus on the critical parts only.

The code in optimizations2 branch only includes the code of `flatMapLatest`, `create` and `just` operators, so it's a lot easier to experiment there. No need to propagate the changes across the entire project.

You can easily run performance tests I'm talking about by running `Microoptimizations` target inside `optimizations2` branch.

Hope this makes sense.

tarunon:
Hi @kzaher 
Sure, thanks information, I will rebase it to `optimizations2`, but seems `optimizations2` branch is broken. Could you please check it?

kzaher:
Hi @tarunon ,

yes, CI fails for optimizations2. This is intended only for experimentation.

Only performance tests and microoptimization targets work on optimizations2.

tarunon:
I mean about compile fail, I cannot play this sandbox.
screen shot 2018-01-05 at 13 08 50

I think it because there is no implementation of `func subscribe(_ observer: (Event<E>) -> ()) -> Disposable`.

kzaher:
@tarunon I think you should be able to run both configurations now.

tarunon:
Hmm, in `optimizations2` branch, it seems doesn't have impact in performance. I'll investigate, and if I can find performance improving, I'll make new PR. thanks

PR #1547, opened 2018-01-11
Fix ambiguous compile error Single.do

tarunon:
screen shot 2018-01-11 at 12 24 57

The latest update contains renaming argument label of `Single.do`, but it make ambiguous compile error now.
In `do(onDispose: xxx)` case, it match both `do(onNext:...` and `do(onSuccess:...`.
I remove default argument `onNext = nil` from `do(onNext:...`, it seems fine now.


freak4pc:
Yikes, happened to me as well :(

kzaher:
Hi @freak4pc ,

I think that a specific kind of unit test could test for a specific kind of ambiguity doesn't exist, but if you are adding that test, then you already know about ambiguity.

I don't think that tests can prove that there isn't any kind of ambiguity, but only that a specific kind of ambiguity that is being tested doesn't exist.

We have added some ambiguity tests in past, but I'm not sure how would we test this.

Issue #1550, opened 2018-01-15
Common `Bindable` protocol for any type that has `bind(to:`) sort of method


freak4pc:
This issue is a discussion and not an actual "Issue", per-se

Right now there is no good way to target all types that are Bindable - this includes Subjects, Relays and also in general there is no way to group Relays together. 

I noticed this problem in RxTestExt where we're forced to have separate definitions for `BehaviorRelay` and `PublishRelay`, for example: 

https://github.com/RxSwiftCommunity/RxTestExt/pull/15/files#diff-a4d1020eede0cd46ebf602cf4481baafR13

Would be really cool if we could have just a single method that applies to all `Bindable` types. 

Would love community feedback :)

freak4pc:
The thought is something like

```swift
protocol Bindable {
    func bind(to: Bindable)
}

extension PublishSubject: Bindable { ... } // In RxSwift
extension PublishRelay: Bindable { ... } // In RxCocoa
```

mosamer:
This would not solve it I think. May be something like
```swift
protocol Bindable {
   associatedtype E
}
extension ObserverType: Bindable {}

extension ObservableType {
   public func bind<B: Bindable>(to bindable: B) -> Disposable where B.E == E {
        ...
    }
}
```

freak4pc:
@mosamer I already removed my comment as I realized I was thinking backwards about it :)

kzaher:
Hi @mosamer @freak4pc  ,

I've posted my comments to https://github.com/ReactiveX/RxSwift/pull/1551.

freak4pc:
Closing this in favor of the PR discussion

PR #1551, opened 2018-01-15
Introduce `Bindable` protocol

mosamer:
This is a proof-of-concept implementation for `Bindable` protocol discussed in #1550 . 

It adds the following protocol
```swift
public protocol Bindable {
    associatedtype T
    func handle(_ event: Event<T>)
}
```
This moves responsibility of implementing `func bind(to:)` behavior  from `ObservableType` to `ObserverType`-like types.

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -1433,6 +1433,10 @@
 		CDDEF16B1D4FB40000CA8546 /* Disposables.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDDEF1691D4FB40000CA8546 /* Disposables.swift */; };
 		CDDEF16C1D4FB40000CA8546 /* Disposables.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDDEF1691D4FB40000CA8546 /* Disposables.swift */; };
 		CDDEF16D1D4FB40000CA8546 /* Disposables.swift in Sources */ = {isa = PBXBuildFile; fileRef = CDDEF1691D4FB40000CA8546 /* Disposables.swift */; };
+		CEA7688A200D2CCA002D4A96 /* Bindable.swift in Sources */ = {isa = PBXBuildFile; fileRef = CEA76889200D2CCA002D4A96 /* Bindable.swift */; };
+		CEA7688B200D2CCA002D4A96 /* Bindable.swift in Sources */ = {isa = PBXBuildFile; fileRef = CEA76889200D2CCA002D4A96 /* Bindable.swift */; };
+		CEA7688C200D2CCA002D4A96 /* Bindable.swift in Sources */ = {isa = PBXBuildFile; fileRef = CEA76889200D2CCA002D4A96 /* Bindable.swift */; };
+		CEA7688D200D2CCA002D4A96 /* Bindable.swift in Sources */ = {isa = PBXBuildFile; fileRef = CEA76889200D2CCA002D4A96 /* Bindable.swift */; };
 		D203C4F31BB9C4CA00D02D00 /* RxCollectionViewReactiveArrayDataSource.swift in Sources */ = {isa = PBXBuildFile; fileRef = C88253F11B8A752B00B02D69 /* RxCollectionViewReactiveArrayDataSource.swift */; };
 		D203C4F41BB9C52400D02D00 /* RxTableViewReactiveArrayDataSource.swift in Sources */ = {isa = PBXBuildFile; fileRef = C88253F21B8A752B00B02D69 /* RxTableViewReactiveArrayDataSource.swift */; };
 		D203C4F51BB9C52900D02D00 /* ItemEvents.swift in Sources */ = {isa = PBXBuildFile; fileRef = C88253F41B8A752B00B02D69 /* ItemEvents.swift */; };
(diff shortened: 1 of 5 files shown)


freak4pc:
Will hopefully find some time to more thoroughly poke through this but overall looks like a really nice change. Pretty much what I intended... 

One nit for now - I don't like the `handle` method name here that much ... it's so ... generic :p

mosamer:
@freak4pc I am totally open to change naming, documentation comments, placement of code etc :)

kzaher:
Hi @mosamer ,

I believe that would make the `Bindable` protocol equivalent to `ObserverType` protocol. 

The reason why we haven't made `*Relay`s conformant to `ObserverType` protocol is to improve type safety regarding `Error` and `Completed` events.

mosamer:
Hi @kzaher 

> I believe that would make the Bindable protocol equivalent to ObserverType protocol.

I think `Bindable` will be _similar_ to `ObserverType` the same way `func bind(to:)` is similar to `func susbscribe(_:)`. For `ObserverType` it's just an equivalent but for `*Relay`s will start to be a little bit different.

> The reason why we haven't made *Relays conformant to ObserverType protocol is to improve type safety regarding Error and Completed events.

IMO, this is still achieved after introducing `Bindable`. Same runtime checks for `Error` & `Completed` events can be made. However, instead of being defined as part of `ObservableType` extension implementation for `func bind(to:)`, this responsibility is moved to the types themselves. This might not be well-communicated with current code placement but we can move things :)

freak4pc:
ObservableType and Relay are similar besides the error/complete behavior, but what they *do* have in common is that you can Bind to both of them. The problem is there is no way defined to introduce what you _can_ bind to and what you _cannot_ bind to. This is where this PR comes into play eventually. It still leaves Relay and ObservableType separate, but giving them a unifying protocol to define that ability is something they both share (while we can still have separate type safety handling, contract, etc)

kzaher:
> I think Bindable will be similar to ObserverType the same way func bind(to:) is similar to func susbscribe(_:). For ObserverType it's just an equivalent but for *Relays will start to be a little bit different.

I'm not sure it is. `Relay`s are added to improve type safety. This is a step backward in regarding type safety.

> IMO, this is still achieved after introducing Bindable. Same runtime checks for Error & Completed events can be made. However, instead of being defined as part of ObservableType extension implementation for func bind(to:), this responsibility is moved to the types themselves. This might not be well-communicated with current code placement but we can move things :)

The same argument as before. `Relay`s are added to improve type safety. This is a step backward regarding type safety.

Do we have some realistic, practical, common examples why would somebody want to unify different relays outside of test environment?

freak4pc:
> Do we have some realistic, practical, common examples why would somebody want to unify different relays outside of test environment?

Any external library that would want to abstract an ability to accept something you can bind to. 

Right now you would have to have separate overrides for Relays and the rest of the ObserverType family. 

> This is a step backward regarding type safety.

I don't entirely understand this. How is Relay related to extra type-safety? Isn't relay just a Subject that can't error and complete ?

kzaher:
@freak4pc ,

> Any external library that would want to abstract an ability to accept something you can bind to.

Can you maybe name some specific cases you've encountered in past? The more the better. This isn't very specific.

> Are you just referring to sort of compile-time assurances? More than general type safety

Isn't compile-time assurance = type safety?

> I don't entirely understand this. How is Relay related to extra type-safety? Isn't relay just a Subject that can't error and complete ?

Right now you can only bind `Signal` -> `PublishRelay` and `Driver` to `BehaviorRelay` so you can't mix up semantics accidentally like you could if you could easily bind `Signal` to `BehaviorRelay` or `Driver` to `PublishRelay` which is conceptually invalid.

freak4pc:
> Right now you can only bind Signal -> PublishRelay and Driver to BehaviorRelay so you can't mix up semantics

This by itself is already a very good point - you can't really generalize something that has specific behavior, I sort of get what you're saying here.

@mosamer Since we only have the RxTest use-case here, I don't think I support this change anymore. It seems like an extensive change, with no _that many_ use cases, and we'd lose this "specificity" Kruno is talking about. 

What do you think?

nikolaykasyanov:
A very interesting discussion here 👍 

I'll just drop my use case as @kzaher asked for practical examples. I'd appreciate any suggestions how to deal with it without big API changes as proposed in this PR, too, as I definitely might be missing something.

Imagine having a coordinator with `AnyObserver` exposed to the view model as event handler of sorts. As long as we have 1:1 correspondence between coordinators and views, this is fine, as soon as view is deallocated, so does the view model and coordinator, usually.

```swift
protocol AwesomeCoordinatorType {
    var toggleItem: AnyObserver<Item>
}
```

It gets interesting when a coordinator is used by multiple views, though;

```swift
class AViewModel {
    let item: Item
    let coordinator: AwesomeCoordinatorType

    var toggle: AnyObserver<Void> {
        coordinator.toggleItem.mapObserver { [item] _ in item }
    }
}

// somewhre in `AView` imlpementation:
    radioButton.rx.tap.bind(to: viewModel.toggle).disposed(by: bag)
```

UIControl's event streams complete on control deallocation, so depending on the order in which `AView` properties deallocate, the button can deallocate _before_ the bag, sending `.completed` event to the observer. This event eventually reaches coordinator, and if coordinator's observer is backend by a regular subject, the subject will complete and be rendered unusable for _any other view that uses/will use_ this coordinator instance in the future.

If we had a protocol like this, though (note how it's more constrainted than the one proposed in this PR, as it doesn't allow `.completed` or `.error` events, only values, keeping the relay contract):

```
protocol RelaySinkType { // haven't give the name a lot of thought, sorry in advance
    associatedtype T
    accept(_ value: T)
}
```

then we could expose `AnyRelaySink` instead of `AnyObserver` in the coordinator and get static guarantees than no one can terminate such observers, ever.

nikolaykasyanov:
The bind method for such `RelaySinkType` can even be called `bindValues` to emphasize that completions are simply ignored (as well as error, perhaps, optionally with `rxDebugFatalError`).

freak4pc:
Hey, thanks for the discussion. I had a lengthy explanation to this here:
https://github.com/ReactiveX/RxSwift/pull/2039#issuecomment-513582247

My opinion is still that this is an over-abstraction, and that it doesn't allow full safety because you can bind replaying types to non-replaying types and vice-versa which is not desired.

nikolaykasyanov:
@freak4pc thanks a lot, should've checked this thread more thoroughly, my bad.

Good points, especially the one about replaying/non-replaying (in)compatibility, although I'd argue that in a lot of cases the actual type of a relay might be an implementation detail. I mean, If I have a behavior relay in my class just to have access to the latest state emitted by who-knows-what (publish subject, replaying subject etc.), this does indeed look like an implementation detail of this class. We don't always get to know what's behind an `ObserverType` in some public API, either.

As already mentioned by you though, the discussed API is pretty much achievable on the library user side by having a protocol and Relay extensions adopting it, so it's not a pressing matter by any means.

freak4pc:
> We don't always get to know what's behind an ObserverType in some public API, either.
Right, but that's exactly why the most basic component (Observer) makes _no_ guarantees, so it's easy to have type erasure for it this way. 

The type of Relay isn't really an implementation detail - saying it's an implementation detail is like saying "they are the same", but they are not. Publish* are meant to represent events while Behavior* are representing state, and mixing those together usually leads to bad things :)

nikolaykasyanov:
What if we forget about relays altogether, and consider having something akin to `ObserverType` but without completion and error handling? 🤔

freak4pc:
It would still not solve your problem because Relay isn't an ObserverType. (It only wraps one)

nikolaykasyanov:
Indeed, but it least it will be a library-level tool to have "value-only" consumers that don't care about completions and errors. But again, that's not necessarily a universally useful thing.

PR #1556, opened 2018-01-26
Use Shortened URL for RxSwift Slack

freak4pc:
Since http://swift.rxslack.org is down, the README has been using the long version of the URL.
I've made a shortened link to share in presentations etc that's easier to send so I've added it here as well: http://bit.ly/rxslack 

Feel free to close this if you prefer leaving the long URL or wait for the old subdomain to be fixed :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ KVO observing, async operations and streams are all unified under [abstraction o
 
 ###### ... interact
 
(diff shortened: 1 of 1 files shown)


freak4pc:
This isn't a code-change so I don't think develop branch is necessary - up to you:)

Nirma:
@freak4pc Stuff like this to the README is completely harmless, would it make sense to silence this warning from the bot when changes are made to non-critical areas like README or ChangeLog?

kzaher:
Hi @freak4pc ,

all changes need to go through `develop` branch. There should be no exceptions.

PR #1557, opened 2018-01-27
expand description of `zip` in tutorial.

Nirma:
The current explanation of  the `zip` operator in the tutorial I feel is incomplete. 
I felt that this explanation is a bit clearer.

### DIFF
--- a/Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
@@ -58,7 +58,8 @@ example("merge") {
 /*:
  ----
  ## `zip`
- Combines up to 8 source `Observable` sequences into a single new `Observable` sequence, and will emit from the combined `Observable` sequence the elements from each of the source `Observable` sequences at the corresponding index. [More info](http://reactivex.io/documentation/operators/zip.html)
+ Combines up to 8 source `Observable` sequences into a single new `Observable` sequence and will emit one element per full set of elements emitted by it's source sequences and not before receiving at least one element per each of its source sequences.
+     The elements emitted from the `zip` operator will be the combination of the elements from each of the source `Observable` sequences at the corresponding index. [More info](http://reactivex.io/documentation/operators/zip.html)
  ![](https://raw.githubusercontent.com/kzaher/rxswiftcontent/master/MarbleDiagrams/png/zip.png)
  */
 example("zip") {
--- a/Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
@@ -58,8 +58,9 @@ example("merge") {
 /*:
  ----
  ## `zip`
- Combines up to 8 source `Observable` sequences into a single new `Observable` sequence and will emit one element per full set of elements emitted by it's source sequences and not before receiving at least one element per each of its source sequences.
-     The elements emitted from the `zip` operator will be the combination of the elements from each of the source `Observable` sequences at the corresponding index. [More info](http://reactivex.io/documentation/operators/zip.html)
+ Combines up to 8 source `Observable` sequences into a single new `Observable` sequence.
+Emitted elements are the combination of elements from each of the source `Observable` sequences at the corresponding index.
+`zip` will emit an element every time all of the source `Observables` emit an element and not before. [More info](http://reactivex.io/documentation/operators/zip.html)
  ![](https://raw.githubusercontent.com/kzaher/rxswiftcontent/master/MarbleDiagrams/png/zip.png)
  */
 example("zip") {


Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
```diff
@@ -58,7 +58,8 @@ example("merge") {
 /*:
  ----
  ## `zip`
- Combines up to 8 source `Observable` sequences into a single new `Observable` sequence, and will emit from the combined `Observable` sequence the elements from each of the source `Observable` sequences at the corresponding index. [More info](http://reactivex.io/documentation/operators/zip.html)
+ Combines up to 8 source `Observable` sequences into a single new `Observable` sequence and will emit one element per full set of elements emitted by it's source sequences and not before receiving at least one element per each of its source sequences.
```
kzaher: Hi @Nirma ,

Maybe we could improve this initial description, but I'm not sure this is an improvement.

The suggested emphasizes first elements, which shouldn't be the case.

The meaning of "full set of elements" is also vague.

Nirma:
@kzaher Thanks for taking a look!

The written description of `zip` as it is now does not mention that `zip` will only emit an element after all of the source `Observables` have emitted an element and will not emit the next element until all of the source `Observables` have emitted an element again.

I pushed a new commit in an attempt to make the description clearer again. 
How does it look?

kzaher:
Hi @Nirma ,

> The written description of zip as it is now does not mention that zip will only emit an element after all of the source Observables have emitted an element and will not emit the next element until all of the source Observables have emitted an element again.

I believe that the former statement is a direct consequence of the following:

> elements from each of the source `Observable` sequences at the corresponding index.

isn't it?

Nirma:
@kzaher It may be implied but there is no explicit statement letting the reader know that, so adding a bit to the description hopefully will spell it out for the reader a little better. 

The text for `combineLatest` in the tutorial explains the need for all Source `Observables` to emit at least one element before an element is emitted from the `combineLatest` operator. 

So I thought making the description of `zip` symmetrical would help.

freak4pc:
I sort of understand the confusion, what do you guys think about something along the lines of: 

> Combines up to 8 source `Observable` sequences into a single new `Observable` sequence, and emits a combined `Observable` sequence with the elements from each of the source `Observable` sequences at the corresponding index, given all source observables emitted an element. For example, when using `zip` on A, B and C - a single, combined emission will occur, containing an element of each of the source Observables, but only when *all* three Observables have emitted an element. [More info](http://reactivex.io/documentation/operators/zip.html)

kzaher:
Hi guys,

I also understand the confusion, but I don't think emphasizing the first element

> given all source observables emitted an element.

improves the situation. The same is also valid for second element, and third one ...

freak4pc:
I'm not emphasizing an element - The description means "when all the observables emit something, zip is called." - it isn't limiting to the first one

kzaher:
@freak4pc after more careful reading, I think you might be right, but to me it still sounds less precise than the original docs.

I would just like to first like to understand the confusion.

I think the wording of the description could be first simplified a bit. Maybe separate along sentence into something like:

> Combines up to 8 source `Observable` sequences into a single new `Observable` sequence. Each resulting sequence element at some index with be generated from a combination of elements from each of the source `Observable` sequences at the same index.

One thing that is not clear from this description is how soon does the sequence generates some element. 

So for example. One could implement zip operator according to this description that doesn't generate any element until all sequences complete, and then generate all elements at once. It would be highly inefficient, but it would satisfy the description.

I believe this could be the only confusion point. One obviously can't generate the resulting element before it gets the corresponding elements, so the only vague part is the explanation how soon will the resulting element be generated.

From my point of view the thing that is missing could maybe be summarized as something like this:

> The resulting sequence will generate element at some index as soon as it receives all elements at corresponding index.

@Nirma in the case of `combineLatest` the behavior for first element is important to emphasize because that's the special case. The operator behaves differently before all sources emit at least one element and after.

`zip` on the other hand doesn't have that distinction.

freak4pc:
I think this: 

> Combines up to 8 source Observable sequences into a single new Observable sequence. Each resulting sequence element at some index with be generated from a combination of elements from each of the source Observable sequences at the same index.

Is much more confusing than what I wrote. I guess its a matter of taste - e.g. I feel like what I wrote is a bit more verbose but it is very explicit on how zip works, which I think is a nice touch. 

I'm good one way or the other :)

freak4pc:
I'm still up for this more verbose, but clearly explanation. Absolutely open to modifying/altering if any feedback is available: 

> Combines up to 8 source Observable sequences into a single new Observable sequence, and emits a combined Observable sequence with the elements from each of the source Observable sequences at the corresponding index, given all source observables emitted an element. For example, when using zip on A, B and C - a single, combined emission will occur, containing an element of each of the source Observables, but only when all three Observables have emitted an element.

kzaher:
Hi @freak4pc ,

I've read the explanation, but I find this part confusing
> For example, when using zip on A, B and C - a single, combined emission will occur, containing an element of each of the source Observables, but only when all three Observables have emitted an element.
... but it's just a rewording of the first sentence, but less precise.

Couldn't we just put ascii art marble diagram if people find the existing explanation confusing, wouldn't that explain this behavior best? It seems to me that you are trying to explain marble diagram anyway.

freak4pc:
@kzaher Not a huge fan of ASCII Marble diagrams (I find them of very little use) but its a personal preference. The reason I repeat it in a way is to provide a specific usage example with A, B, and C - instead of just a "general overview" kind of thing. I think it makes things clearer but I can rephrase it further if you don't find it fitting.

freak4pc:
It's hard to express with a marble diagram, this is the closest I could get with the description I mentioned above. Thoughts @Nirma @kzaher ? 

```
Combines up to 8 source `Observable` sequences into a single new `Observable` sequence, and emits a combined `Observable` sequence with the elements from each of the source `Observable` sequences at the corresponding index, given all source observables emitted an element. For example, when using `zip` on A, B and C - a single, combined emission will occur, containing an element of each of the source Observables, but only when **all** three Observables have emitted an element. [More info](http://reactivex.io/documentation/operators/zip.html)
 
 A +---1---2----3-----4--|
 B +-A------B-C--D-------|  
 C +----E--F---G---H---I-|

 zip(A, B, C):

 + +----1---2---3-----4--|

 1) 1,A,E
 2) 2,B,F
 3) 3,C,G
 4) 4,D,H
```

freak4pc:

Given we couldn't get to an agreement on better terminology for over 2 years, I think it's fine to close. Happy to discuss otherwise.

Nirma:
@freak4pc Sorry for falling off the radar with this one. Yeah for now lets just close it.

PR #1554, opened 2018-01-23
Add Zip<C: Collection> operator to Single trait

yuzushioh:
I added `fund zip<C: Collection>(_ collection: C, resultSelector: _)` operator to Single trait.

PR #1558, opened 2018-01-28
Add zip operator to Single

yuzushioh:
Resubmit #1554(because commits will be messy)
Added `zip<C: Collection>` operator to Single trait and fixed changes @kzaher mentioned in #1554

### DIFF
--- a/RxSwift/Traits/Single.swift
+++ b/RxSwift/Traits/Single.swift
@@ -258,7 +258,9 @@ extension PrimitiveSequenceType where TraitType == SingleTrait {
     public static func zip<C: Collection, R>(_ collection: C, _ resultSelector: @escaping ([ElementType]) throws -> R) -> PrimitiveSequence<TraitType, R> where C.Iterator.Element == PrimitiveSequence<TraitType, ElementType> {
         
         if collection.isEmpty {
-            return PrimitiveSequence<TraitType, R>(raw: .just(try! resultSelector([])))
+            return PrimitiveSequence<TraitType, R>.deferred {
+                return PrimitiveSequence<TraitType, R>(raw: .just(try resultSelector([])))
+            }
         }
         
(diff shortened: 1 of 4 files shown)


RxSwift/Traits/Single.swift
```diff
@@ -248,4 +248,35 @@ extension PrimitiveSequenceType where TraitType == SingleTrait {
...
+    /**
+     Merges the specified observable sequences into one observable sequence by using the selector function whenever all of the observable sequences have produced an element at a corresponding index.
+     
+     - parameter resultSelector: Function to invoke for each series of elements at corresponding indexes in the sources.
+     - returns: An observable sequence containing the result of combining elements of the sources using the specified result selector function.
+     */
+    public static func zip<C: Collection, R>(_ collection: C, _ resultSelector: @escaping ([ElementType]) throws -> R) -> PrimitiveSequence<TraitType, R> where C.Iterator.Element == PrimitiveSequence<TraitType, ElementType> {
+        
+        if collection.isEmpty {
+            return PrimitiveSequence<TraitType, R>(raw: .just(try! resultSelector([])))
```
kzaher: We can't use `try!`.

We should probably do something as this.

```
return PrimitiveSequence<TraitType, R>.defer {
     return PrimitiveSequence<TraitType, R>(raw: .just(try resultSelector([])))
}
```

and add a test case for that.

Defer is here to make sure `resultSelector ` is being called once per subscription and not per zip invocation. This shouldn't have much difference, but if somebody decides to do a dirty hack and perform side effects inside `resultSelector`, version with `defer` will work with `retry` operators nicely.

yuzushioh: gotcha! will fix it

yuzushioh: just one question. how should it be checked if the resultSelector is called not once per invocation but once per subscription? I'm not so used to tests in Rx.

PR #1559, opened 2018-01-29
Simplify by using forEach

ShenghaiWang:
Simplify code by using forEach

### DIFF
--- a/RxSwift/Extensions/Bag+Rx.swift
+++ b/RxSwift/Extensions/Bag+Rx.swift
@@ -11,52 +11,14 @@
 
 @inline(__always)
 func dispatch<E>(_ bag: Bag<(Event<E>) -> ()>, _ event: Event<E>) {
-    if bag._onlyFastPath {
-        bag._value0?(event)
-        return
-    }
-
-    let value0 = bag._value0
-    let dictionary = bag._dictionary
-
-    if let value0 = value0 {
-        value0(event)
-    }
-
-    let pairs = bag._pairs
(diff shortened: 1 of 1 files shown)


kzaher:
Hi @ShenghaiWang ,

The logic duplication there is due to performance reasons since bag is a commonly used data type. :)

ShenghaiWang:
Thanks for your reply and having created such a beautiful framework.

Yes, I noticed that @inline(__always) annotation. In order to avoid this duplication, could we somehow make that forEach method as an inline method? Or add a fileprivate method in Bag+Rx.swift file to hold the enumeration logic, which then will be used by these two global methods?

To me, the duplication here is just not in line with the overall beauty of the framework :)

kzaher:
@ShenghaiWang I agree that it's extremely ugly.

Unfortunately, there are various tradeoffs we've decided to make. This is just one of them.

We could try to do many different things to remedy this issue, but at the end of the day even though the solution could maybe work on a single compiler version (probably latest), optimizations could break without warning in any subsequent swift compiler version.

This is an ugly solution, but it's unit tested and at the end of day it's the least risky one IMHO.

I'm not worried that people would see this code and conclude that I couldn't see that this is duplicated logic and that there exists an obvious abstraction, even though this is probably constantly happening and people assume I'm clueless based on reading this code alone 😆😆😆.

ShenghaiWang:
Understood. You needn’t worry that at all. Thanks again for this beautiful framework.

PR #1563, opened 2018-02-06
[Proposal] Create binder with Smart Key Path

ra1028:
Hi✋ 

`Binder` is a good parts for binding` Observable` to object.
In current RxSwift, it's need to implement for every one of  property of object.
So, I propose to create `Binder` with Swift 4' smart key path.
Extensions that create Binder of properties is no longer needed after this.

It maybe not match ReactiveX rule, because smar key path is the Swift's unique feature.
However, the Binder is a part of RxCocoa, so I think this is acceptable extension.

Example:
```swift
textObservable
    .bind(to: label.rx[\.text])
    .disposed(by: bag)
```

### DIFF
--- a/RxCocoa/Common/KeyPathBinder.swift
+++ b/RxCocoa/Common/KeyPathBinder.swift
@@ -3,7 +3,7 @@
 //  RxCocoa
 //
 //  Created by Ryo Aoyama on 2/7/18.
-//  Copyright © 2018 Ryo Aoyama. All rights reserved.
+//  Copyright © 2018 Krunoslav Zaher. All rights reserved.
 //
 
 #if swift(>=3.2)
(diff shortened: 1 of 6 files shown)


RxCocoa/Common/KeyPathBinder.swift
```diff
@@ -0,0 +1,36 @@
...
+                base[keyPath: keyPath] = value
+            }
+        }
+        
+        /// Bindable sink for arbitrary property using the given key path.
+        /// Binding runs on the specified scheduler.
+        ///
+        /// - parameter keyPath: Key path to write to the property.
+        /// - parameter scheduler: Scheduler to run bindings on.
+        public subscript<Value>(keyPath: ReferenceWritableKeyPath<Base, Value>, on scheduler: ImmediateSchedulerType) -> Binder<Value> {
```
kzaher: Is there any reason why we don't have a default value for scheduler and instead have two methods?

ra1028: That's because the subscript does not support default arguments 😢

kzaher: :(

PR #1565, opened 2018-02-07
Performance improvements in thread lock.

ra1028:
RxSwift uses Foundation `NSRecursiveLock`, but it includes an implementation that is no needed.
See [here](https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSLock.swift) about it.

In FRP, thread lock is very frequently used.
So, can be expect drastic performance improvements by implementing that minimal using pthread mutex.

This is before and after of my benchmark results.
In the tests with the most impact, it has improved by more than **1 sec**.
Have little effects in creation tests that don't involving much thread lock.

Incidentally, I've removed the typealias of SpinLick that no longer needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 * Lower macOS Deployment Target to 10.9
 * Add `zip<C: Collection>(_ collection: C)` to Single trait
 * Add Smart Key Path subscripting to create a binder for property of object.
+* Performance improvement in thread locking.
 
 #### Anomalies
 
--- a/Platform/RecursiveLock.swift
+++ b/Platform/RecursiveLock.swift
@@ -22,7 +22,7 @@ final class RecursiveLock {
         let attr = UnsafeMutablePointer<pthread_mutexattr_t>.allocate(capacity: 1)
         attr.initialize(to: pthread_mutexattr_t())
         pthread_mutexattr_init(attr)
-        pthread_mutexattr_settype(attr, PTHREAD_MUTEX_RECURSIVE)
+        pthread_mutexattr_settype(attr, Int32(PTHREAD_MUTEX_RECURSIVE))
         
         pthread_mutex_init(mutex, attr)
         
--- a/Platform/RecursiveLock.swift
+++ b/Platform/RecursiveLock.swift
@@ -6,29 +6,55 @@
 //  Copyright © 2016 Krunoslav Zaher. All rights reserved.
 //
 
-import class Foundation.NSRecursiveLock
+import Foundation
 
-#if TRACE_RESOURCES
-    class RecursiveLock: NSRecursiveLock {
-        override init() {
+final class RecursiveLock {
+    
+    private let mutex = UnsafeMutablePointer<pthread_mutex_t>.allocate(capacity: 1)
+    
+    init() {
+        #if TRACE_RESOURCES
             _ = Resources.incrementTotal()
-            super.init()
-        }
-
-        override func lock() {
-            super.lock()
-            _ = Resources.incrementTotal()
-        }
-
-        override func unlock() {
-            super.unlock()
-            _ = Resources.decrementTotal()
-        }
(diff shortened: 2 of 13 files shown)


kzaher:
Hi @ra1028 ,

Interesting.

These were the results I got. The right side contains results with custom implemented RecursiveLock.

screen shot 2018-02-07 at 23 26 39

I've played with this code in the past :)

If you implement `RecursiveLock` in this way. 
```swift
struct RecursiveLock {
    func lock() {
    }

    func unlock() {
    }
}
```

It does shave off some time, but doesn't seem like locking is the bottleneck.

screen shot 2018-02-07 at 23 34 58

The biggest bottlenecks I've found were just related with Swift runtime and compiler. E.g.:
* value copying
* metadata lookups
* reference counting
* pattern matching (switch (event) { .next, .error, .completed })

I think that if we threw that out, we would get ~2x improvement :) 

But here's the thing. These benchmarks are synthetic. They assume integers are being passed and  lambdas don't perform any work at all. Which is not realistic use case IMHO.

I'm definitely open to discussions on what can we do to further improve performance. It's always interesting to know did we miss something obvious, better understand how the compiler works, or just race :)

ra1028:
@kzaher 
Hmmm, Didn't have performance changes?

I tested again with release build, then the performance was improved.
These were the results.
result

There are nearly 2x faster tests.

kzaher:
Hi @ra1028 ,

On what platform and configuration did you run those tests?

I've ran them on:
* 10.13.3 (17D47)
* MacBook Pro
* 2.5 GHz Intel Core i7
* 16 GB 1600 MHz DDR3

Yes, performance was more or less the same on my test, except for `testFlatMapLatestPumping`.  🤔

freak4pc:
To @kzaher's request, I've ran the performance tests locally as well. 

Specs: 
macOS 10.13.1
MBP 15" Mid-2015 - 2.5 GHz Intel Core i7
16 GB 1600 MHz DDR3

I do see some performance improvement but, in my eyes, it is relatively negligible - I'm not sure why the difference is more extreme in your case specifically @ra1028 

test

ra1028:
@kzaher @freak4pc 
Thank for the information.

My Specs:
macOS 10.12.6
MBP 15 2016 - 2.6 GHz Intel Core i7
16 GB 2133 MHz LPDDR3

I benchmarked again, 5 times for each on `develop` and 'pthread-lock' branch.
These are the results.
develop:
screen shot 2018-02-12 at 20 10 17

pthread-lock:
screen shot 2018-02-12 at 20 18 55

It seems that the results I saw before was extreme cases.
However, the results have improved slightly.
If this results is negligible difference for RxSwift, You can close this Pull-Request.

kzaher:
@ra1028 I think @tarunon was also interested in performance optimizations at one time.

I've played around with some optimization ideas in this branch https://github.com/ReactiveX/RxSwift/tree/optimizations2.

If something is a small change, then I guess we can change it mid release.

If something is a more risky change (like this), then it might make sense to wait for next major release so we don't accidentally break someone's code. I know why this looks like a small change on surface (especially since it's using part of Swift standard library source code), but messing around with threading, c structs (pthread_t) and Swift value types bridging is quite dangerous IMHO because of Swift language semantics.

I would rather want to explore what are all of the performance optimizations that we can make, and make them in the following major release if we can get a couple of X more performance.

I believe getting 2x improvements should be possible with some changes. I'm not so sure about 3x.

This is part of the straightforward changes we can potentially do in the following major release.

ra1028:
Got it. Thanks.
Should I closing this Pull-Request? Or set the base branch to `optimizations 2`?

kzaher:
@ra1028 If you can squash everything into a single commit, we can put it in optimizations2 branch so we don't forget about this.

ra1028:
Thanks @kzaher 
I made a new pull-request [here](https://github.com/ReactiveX/RxSwift/pull/1574).
https://github.com/ReactiveX/RxSwift/pull/1574

PR #1567, opened 2018-02-07
Lower macOS Deployment Target to 10.9

### LINKED ISSUE #1566: Why doesn't RxSwift support macOS 10.9?
I have a macOS app that I'm rebuilding in swift and I wanted to use RxSwift. I actually got a bunch of stuff done on top of RxSwift before I realized that it only supports back to 10.10. I cloned the repo, changed the deployment target for macOS to 10.9 and it builds fine.

Is there a specific reason 10.9 isn't supported or was it just an oversight when creating the project?

freak4pc:
Fixes #1566


freak4pc:
Funny that the warnings don't mind RxTest's podspec changing :)

PR #1570, opened 2018-02-08
Original Slack URL brought back to life 🧟‍♂️💀

freak4pc:
Happy to say that with immensely kind help from @bontoJR, the RxSwift.org domain is now back, so I’ve set up the `slack` subdomain forwarding back. Woohoo!


freak4pc:
Failure seems unrelated. Kicking off another build.

PR #1573, opened 2018-02-14
[PrimitiveSequence] Implement `catchErrorJustReturn`

ikesyo:
Just implemented `catchErrorJustReturn` for `PrimitiveSequence` (especially for `Single`).


ikesyo:
Should this be enclosed in `extension PrimitiveSequenceType where TraitType == SingleTrait`?

kzaher:
Hi @ikesyo ,

I think that this operator makes sense in case of `Single` or `Maybe`. The way it's currently implemented would add it also to `Completable`, even though one can't create element. It would work, but maybe it would make more sense to just add it to `Single` or `Maybe` for now.

We would also need to add those sanity check unit tests for this operator.

ikesyo:
Hi @kzaher, thanks for the feedback and I've addressed them. Could you take a second look?

PR #1575, opened 2018-02-14
[MaterializedSequenceResult] Add some useful getters for assertions

ikesyo:
Add four computed properties to `MaterializedSequenceResult`:

- `isCompleted`
- `isFailed`
- `elements`
- `error`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 * Lower macOS Deployment Target to 10.9
 * Add `zip<C: Collection>(_ collection: C)` to Single trait
 * Add Smart Key Path subscripting to create a binder for property of object.
+* Add some useful getters to `MaterializedSequenceResult` for assertions
 
 #### Anomalies
 
--- a/RxBlocking/BlockingObservable+Operators.swift
+++ b/RxBlocking/BlockingObservable+Operators.swift
@@ -21,6 +21,40 @@ public enum MaterializedSequenceResult<T> {
     case failed(elements: [T], error: Error)
 }
 
+extension MaterializedSequenceResult {
+    /// - returns: true if the result is `.completed`, otherwise false.
+    public var isCompleted: Bool {
+        switch self {
+        case .completed: return true
+        case .failed: return false
+        }
+    }
+
+    /// - returns: true if the result is `.failed`, otherwise false.
+    public var isFailed: Bool {
+        switch self {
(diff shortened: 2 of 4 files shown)


ikesyo:
> Please correct me if I'm wrong, but it seems to me that the reason for adding these convenience methods is to enable patterns as this:

That is not what I assume. The use-case is as follows:

```swift
let fooCompletable: Completable = ...
let result = fooCompletable.toBlocking().materialize()

// Check if the sequence successfully completed
// (and elements are not needed at all for `Completable`)
XCTAssertTrue(result.isCompleted)

// or

// Check if the sequence terminated with an error
XCTAssertTrue(result.isFailed)

// or
if let error = result.error {
    XCTFail("\(error)")
}
// The above is much simpler than
//
// if case .failed(let error) = result {
//     XCTFail("\(error)")
// }
// or
//
// switch result {
// case .completed: break
// case .failed(let error): XCTFail("\(error)")
// }

let barObservable: Observable<Int> = ...
let result = barObsrvable.toBlocking().materialize()
XCTAssertEqual(result.elements, [1, 3,5]) // Check the elements regardless of the sequence's termination event
```

kzaher:
Hi @ikesyo ,

Because of generality Rx blocking works with any observable sequence, so completable type is erased when using `toBlocking`. Otherwise we would need to create tedious overloads for each type, which wouldn't be much better.

I understand that this API could be used to test completable, but even then it seems to me that there is no downside of doing `XCTAssertTrue(result.completed, [])`.

In case of `XCTAssertTrue(result.isFailed)`, I believe that test could be improved by writing in in the following way: `XCTAssertTrue(result.error, [elements received], expectedError)`. 
Testing only for `XCTAssertTrue(result.isFailed)` seems more fragile to me.

In case of:
```swift
let barObservable: Observable<Int> = ...
let result = barObsrvable.toBlocking().materialize()
XCTAssertEqual(result.elements, [1, 3, 5])
```

I believe that it is extremely important to test how the sequence terminates. Testing that sequence has produced only a certain sequence of elements and not caring about successful termination or erroring seems fragile to me.

ikesyo:
Hmm I'm not interested in `public var completed: [Element]?` and `public var failed: (elements: [Element], error: Error)?` so I'll withdraw this.

diogot:
What if it just adds the helpers:
```Swift
public var elements: [T]
public var error: Error?
```
It would avoid some boilerplate in tests.

PR #1582, opened 2018-02-22
Add operator to convert Single to Maybe

### LINKED ISSUE #1580: Add operators on Single to convert to Maybe
I was looking at RxJava Single, and there is two operators that I need and I would like to add on RxSwift.
* `flatMapMaybe` - Same as flatMap but returns a `Maybe`;
* `toMaybe` - Convert Single to Maybe.

Do you think it’s worth to adding these operators?
If so, I am able to do this and open a pull request.

Thanks!

pietrocaselani:
This pull request add two operators on Single trait
* `Single.flatMapMaybe`: Same as `Single.flatMap` but returns a Maybe trait;
* `Single.asMaybe`: Converts the Single trait to Maybe trait.

Closes #1580


kzaher:
Hi @pietrocaselani ,

I would like to merge this, but there is just one thing. You are using tabs, and we are using 4 spaces instead. If you could change this, I would be happy to merge this.

PR #1592, opened 2018-03-12
UISegmentedControl set title and image for segment wrappers.

LucianoPAlmeida:
Hey guys :)) 
This PR adds just two convenience wrappers to RxCocoa UISegmented Control.

### DIFF
--- a/RxCocoa/Deprecated.swift
+++ b/RxCocoa/Deprecated.swift
@@ -367,6 +367,13 @@ extension Reactive where Base: UIImageView {
         }
     }
 }
+    
+extension Reactive where Base: UISegmentedControl {
+    @available(*, deprecated, renamed: "enabledForSegment(at:)")
+    public func enabled(forSegmentAt segmentAt: Int) -> Binder<Bool> {
+        return enabledForSegment(at: segmentAt)
+    }
+}
 #endif
 
 #if os(macOS)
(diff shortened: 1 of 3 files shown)


kzaher:
Hi @LucianoPAlmeida ,

what is the reason to use
```swift
segmentedControl.rx.title(forSegmentAt: 0)
```
vs 
```swift
segmentedControl.titleForSegment(at: 0)
```

LucianoPAlmeida:
Hey @kzaher :))
The use case for this one was more to set the text according to a model event. 
Something like:
```swift
// Where the textChangeEvent is a BehaviourRelay<String>
viewModel.textChangeEvent.asDriver().drive(cell.segmentedControl.rx.title(forSegmentAt: 0)).disposed(by: disposeBag)

```
The idea was just to provide a reactive way to change the text according to model changes. :))

kzaher:
But wouldn’t a mote appropriate form be titleForSegment(at:).

LucianoPAlmeida:
Ahh I see, sorry I miss understand the question before :)) 
I just follow the same naming pattern of the existing`enabled(forSegmentAt segmentAt: Int)`. 
But I agree that  `segmentedControl.titleForSegment(at: 0)` and sure can change it  :)

kzaher:
Hi @LucianoPAlmeida ,

I think we've maybe made an error with `enabled(forSegmentAt segmentAt: Int)` :(.

LucianoPAlmeida:
Hey @kzaher  
Thank you for the feedback :)) 
What do you think about this changes?

PR #1597, opened 2018-03-14
[Documentation] Clarify retry documentation

namolnad:
Attempt to reduce confusion caused by documentation stating a source sequence would be "repeated" when it might be run only once.

### DIFF
--- a/RxSwift/Traits/PrimitiveSequence.swift
+++ b/RxSwift/Traits/PrimitiveSequence.swift
@@ -148,13 +148,11 @@ extension PrimitiveSequence {
     }
 
     /**
-     Repeats the source observable sequence the specified number of times in case of an error or until it successfully terminates.
-
-     If you encounter an error and want it to retry once, then you must use `retry(2)`
+     If the initial subscription to the observable sequence emits an error event, try repeating it up to the specified number of attempts (inclusive of the initial attempt) or until is succeeds. For example, if you want to retry a sequence once upon failure, you should use retry(2) (once for the initial attempt, and once for the retry).
 
      - seealso: [retry operator on reactivex.io](http://reactivex.io/documentation/operators/retry.html)
 
-     - parameter maxAttemptCount: Maximum number of times to repeat the sequence.
+     - parameter maxAttemptCount: Maximum number of times to attempt the sequence subscription.
      - returns: An observable sequence producing the elements of the given sequence repeatedly until it terminates successfully.
      */
     public func retry(_ maxAttemptCount: Int)


freak4pc:
Hey @namolnad, first of all I really want to thank you for your contribution and time. It's highly appreciated ! 

I think kruno can look at this later but IMHO this change doesn't add clarity unfortunately. 

"Run" is a very arbitrary and overloaded term to use, and its also slightly "wrong" in this context. I think the terminology of Repeat is fine, especially when combined with the "condition", e.g. - 

> ... in case of an error or until it successfully terminates.

namolnad:
Thanks @freak4pc, I'll think about this one a little further as I imagine we aren't the first ones to have used this operator incorrectly even after scanning the docs.

It confused us over at Instacart for a bit because it says it `repeats the specified number of times` which to us felt like it had to have already occurred once before the second occurrence counts as a *repeat*. Given that `retry(1)` occurs a max of 1 time, rather than after a single *repeat*, we were obviously wrong. Without this following line, we likely wouldn't have noticed for quite a while: `If you encounter an error and want it to retry once, then you must use "retry(2)"`

I think some slight rewording or reordering can help de-muddy those waters. Will get back to you with one more go at this.

Thanks!

freak4pc:
@namolnad Oh I see where the confusion lies at ! 

I think the issue is something like this is missing: 

> Repeats the source observable sequence the specified number of times, **on top of the initial attempt**, in case of an error or until it successfully terminates.

namolnad:
@freak4pc Thanks for the quick reply! I think I like where you're headed, except that I don't think that's actually the behavior here. Perhaps something like:

> Repeats the source observable sequence in case of an error until it successfully terminates. The total number of times the sequence *executes/runs/rx_fancy_word* won't exceed the `maxAttemptCount` value included.

Does that make sense?

kzaher:
@namolnad @freak4pc I think I understand the confusion.

Wouldn't it make more sense to say `concatenates` or `subscribes`.
```
Concatenates the source observable sequence at most the specified number of times in case of an error or until it successfully terminates.
```

```
Subscribes to the source observable sequence at most the specified number of times in case of an error or until it successfully terminates.
```

freak4pc:
I don't think any of these are really a good fit. I think "repeats" or "retries" is essential here since its the essence of the operator. 

The confusion @namolnad stated with all explanations are parts like this: 
> at most the specified number of times

It is not 100% true because 
thing
 .retry(1)

will actually have 2 tries (initial and a single retry) , so the terminology of "at most the specified number" might be confusing. (even if correct) 

I still think my explanation with a bit of tightening can work well but up to you guys :)

RxSwift/Traits/PrimitiveSequence.swift
```diff
@@ -148,13 +148,13 @@ extension PrimitiveSequence {
     }
 
     /**
-     Repeats the source observable sequence the specified number of times in case of an error or until it successfully terminates.
+     Runs the source observable sequence the specified number of times in case of an error or until it successfully terminates.
 
      If you encounter an error and want it to retry once, then you must use `retry(2)`
```
namolnad: @freak4pc If your last comment is true, then this line of documentation is very confusing.

> It is not 100% true because
> ```
> thing
>  .retry(1)
> ```
> will actually have 2 tries (initial and a single retry) , so the terminology of "at most the specified number" might be confusing. (even if correct)

freak4pc: So maybe I'm wrong? That could very well be true haha :) Let me confirm.

freak4pc: Whoops, I totally am. Let me remove my last comments to not confuse anyone reading later.

namolnad: @freak4pc does my confusion make sense now? :) Do you have any thoughts on improved wording?

freak4pc:
@namolnad @kzaher. Gave this some more thought. How would you guys feel about:

> Repeats the subscription to the source observable sequence in case of an error, up to the specified number of attempts (inclusive of the initial attempt), or until a successful termination of the observable sequence.

RxSwift/Traits/PrimitiveSequence.swift
freak4pc: I was sure I replied to this a few days ago. Ha! Too much work 😅

namolnad:
@freak4pc that wording seems __*much*__ better to me :)

I'll update this PR with that suggested wording, if that sounds good to you.

kzaher:
Hi,

I think we need to understand the issue here.

The description says

> Attempt to reduce confusion caused by documentation stating a source sequence would be "repeated" when it might be run only once.

According to Google

re·peat
rəˈpēt/

verb
1. say again something one has already said.
2. do (something) again, either once or a number of times.

The problem is that people assume the first meaning, and not the second, like in:
`repeat this excercise 3 times`.

Our docs actually say.

> Repeats the source observable sequence the specified number of times in case of an error or until it successfully terminates.
>
> If you encounter an error and want it to retry once, then you must use `retry(2)`.

I guess the second sentence **If you encounter an error and want it to retry once, then you must use `retry(2)`** solves that ambiguity, 

Honestly, I think that most people don't ever see this other sentence, so maybe the simplest thing would be to just remove the space between them and say.

> Repeats the source observable sequence the specified number of times in case of an error or until it successfully terminates. If you encounter an error and want it to retry once, then you must use `retry(2)`

If we merge this PR then this critical second sentence that resolves the potential ambiguity of interpretation by a practical example (which is optimal from my point of view) is even less visible.

namolnad:
@kzaher I don't disagree that the practical example is clarifying in and of itself, but I don't think I agree with your interpretation of the definition provided by Google.

> According to Google
> 
> re·peat
> rəˈpēt/
> 
> verb
> 
> 1.  say again something one has already said.
> 2.  do (something) again, either once or a number of times.

For clarity's sake, if you remove `(something)` here, then the definition reads:
> 2.  do again, either once or a number of times.

The "_again_" here is critical to the definition of repeating something. If you were telling someone to do an exercise and they hadn't yet done the exercise once, you would not say `repeat this exercise once`, you would say `do this exercise once`.

Perhaps, to keep it shortest, you could combine the suggested approaches here:

> Repeats the source observable sequence the specified number of times (inclusive of the initial attempt) in case of an error or until it successfully terminates. If you encounter an error and want it to retry once, then you must use `retry(2)`.

Ultimately not my call, but I think this first sentence needs a bit of help beyond just putting the practical example closer as, in it's current form, I think it's slightly contradictory to the example.

kzaher:
@namolnad If repeat is the wrong word to use here, why would we put repeat as the first word, and then explain how it's not actually repeat?

freak4pc:
I'm with @namolnad here. The dictionary definition isn't relevant in my eyes. There are many terms we use in technical documentation that isn't necessarily directly on par with the dictionary term, and that's fine. I think a repeat accurately conveys the meaning of this operator. 

I'm also cool with just using `Retry the source observable...` etc. I'm not usually a fan of using a word in its own definition but python and Google for example are doing just that: 
https://pypi.python.org/pypi/retrying
http://google-cloud-python.readthedocs.io/en/latest/core/retry.html

kzaher:
@freak4pc @namolnad

I agree that `retry` is a confusing, and `repeat` could be also considered confusing.

I would either want to find some word that conveys intuitively on high level what this operator does, or just leave the word `repeat` with a practical example afterwords.

There are two interpretations:
* either word repeat conveys the correct meaning, but it's ambiguous, and we need to disambiguate it with an example.
* word `repeat` always conveys the wrong meaning, in which case we shouldn't try to redefine the meaning of the word `repeat`

The word has latin origins:
late Middle English: from Old French repeter, from Latin repetere, from re- ‘back’ + petere ‘seek.’

... which would imply some kind of going back and trying again, but it's still ambiguous as far as I can tell. 

In the context of learning/training, is it a valid use case for trainer to explain the exercise (without actually performing it) and then say, `repeat this exercise X times`?

That seems to me like the same use case here. We define an operation and can say repeat this operation max X times.

I can agree that the current situation is a bit clumsy and would like to improve it, but I'm not convinced that using a word and then redefining its meaning in later part of sentence is a better approach.

freak4pc:
I think you're missing a third interpretation which is the one I'm standing by : 

- Repeat in the context of an observable sequence makes sense as an alias for a retry operation. 

I don't think it requires extra explanation (personal opinion), and I also thing that the bottom line is: 

Do we want an explanation that people understand, or one that's dictionary-correct? 

Because the current explanation is neither and is indeed relatively confusing, while the later suggestions at least accurately represent what this operator does with no guesses on the amount of times it does it.

kzaher:
Hi @freak4pc ,

I don't think there can be logically a third interpretation besides:
* either word repeat conveys the correct meaning, but it's ambiguous, and we need to disambiguate it with an example.
* word repeat always conveys the wrong meaning, in which case we shouldn't try to redefine the meaning of the word repeat

> Repeat in the context of an observable sequence makes sense as an alias for a retry operation.

is in the first bullet

> * either word repeat conveys the correct meaning, but it's ambiguous, and we need to disambiguate it with an example.

The thing that you don't agree is how to disambiguate. With an example or explaining the actual word.

We've discussed about `retry` operator a bunch of times previously and yes, the naming does have issues, so I would like to wrap this once and for all.

> Do we want an explanation that people understand, or one that's dictionary-correct?

Ideally both, otherwise we are creating our own language, been there, don't want to ever do that again :)

The problem with the current suggestion is that it probably won't solve anything once again.
Why do I have this opinion?
Because I have a suspicion people only read the start of docs, so they will read
> **Repeats the subscription to the source observable sequence in case of an error**, bla bla bla

I have no other explanation why this is still causing confusion when we've provided actual example how to use it in most common use case and what argument to pass.

> If you encounter an error and want it to retry once, then you must use `retry(2)`

freak4pc:
I think what we don't agree about is the fact this needs disambiguation at all. I think it's clear from the context of the sequence itself, but let's put that aside.

I agree people usually would read only the first sentence of a documentation as it's supposed to be the summary of the method. Moving that second sentence in-line with the first would also be better than nothing. 

The main problem is that re-try means "Try again" , but retry 2 doesn't mean try again twice, it means try again once on top of the original attempt - and that's where the clarification is needed TBH.

kzaher:
> The main problem is that re-try means "Try again" , but retry 2 doesn't mean try again twice, it means try again once on top of the original attempt - and that's where the clarification is needed TBH.

Yes, the main problem is that there is a collision between regular meaning and its meaning in the context of Rx (that's what I meant by inventing our own language).

The reason why it says `retry` is brevity and lack of a better alternative.

What is your explanation why our first attempt at solving this by adding
>  If you encounter an error and want it to retry once, then you must use `retry(2)`

failed?

freak4pc:
> If the initial subscription to the observable sequence emits an error event, try repeating it up to the specified number of attempts (inclusive of the initial attempt). For example, if you want to retry a sequence once upon failure, you should use `retry(2)` (once for the initial attempt, and once for the retry). 

Here is a more "explicit" attempt. We can also get rid of the parentheses at the end if you think its too much. Is this any better in your opinion ? @kzaher

freak4pc:
Any thoughts here ?

namolnad:
@freak4pc Thanks for following up here — this had fallen off of my radar. FWIW, I think your explicit attempt clearly conveys the actual behavior here and better sets expectations vs current documentation.

kzaher:
Hi @freak4pc ,

right now we have

> Repeats the source observable sequence the specified number of times in case of an error or until it successfully terminates.
> If you encounter an error and want it to retry once, then you must use `retry(2)`

This explanation
> If the initial subscription to the observable sequence emits an error event, try repeating it up to the specified number of attempts (inclusive of the initial attempt). For example, if you want to retry a sequence once upon failure, you should use retry(2) (once for the initial attempt, and once for the retry).

would make it seem like if it errors out first time, it always repeats the sequence up to the specified number of attempts (no matter have those additional attempts errored out or no).

The original explanation doesn't have that problem.

kzaher:
If we want to explain how the operator works, maybe we should just do something like this.

> This operator forwards next and completed events and creates an internal counter that starts with 0. Every time an error event is received, counter is incremented. If the counter is less then the maximal number of attempts, then new subscription is made, otherwise the error event is forwarded.

:)

freak4pc:
@kzaher I don't think that is a good piece of documentation. It explains how it works, not what it does. Implementation details aren't interesting details for a user of an operator IMO :) 

The issue with mine isn't that critical IMO because "retry" is implicit in the fact something failed. But it can easily be resolved by adding 3-4 words: 

> If the initial subscription to the observable sequence emits an error event, try repeating it up to the specified number of attempts (inclusive of the initial attempt) **or until is succeeds**. For example, if you want to retry a sequence once upon failure, you should use retry(2) (once for the initial attempt, and once for the retry).

namolnad:
@freak4pc I adjusted the PR to use the wording you suggested (and tweaked the parameter description.)

freak4pc:
@namolnad Thanks! Squashed all your commits into one to keep things clean. 

@kzaher Thanks for the openness. Regarding your feedback:

> it separates the initial attempt for the subsequent ones, so when reading it my first impression is that we treat the first attempt somehow differently, which is not the case

It doesn't exactly separate it - but the first failure is the thing that causes the "Retry sequence", in a way. So I think its just easier to explain that way, and I don't think it necessarily conveys anything IRT the initial attempt being any different.

> It doesn't explain what happens with sequence elements and how is the resulting sequence constructed based on the original one.

Can you clarify further on what specific parts of this are you referring to ? Just the fact the emissions are "held" until retries are exhausted?

kzaher:
Ok guys, if you want to merge this, I'm fine with it.

Issue #1598, opened 2018-03-16
CurrentThreadScheduler.swift is incompatible with Swift 4.1


alanzeino:
Luckily, this is the only visible warning out of using master with Swift 4.1. I'll put up a PR for this.

```
RxSwift/RxSwift/Schedulers/CurrentThreadScheduler.swift:54:17: warning: 'deallocate(capacity:)' is deprecated: Swift currently only supports freeing entire heap blocks, use deallocate() instead
            key.deallocate(capacity: 1)
                ^
```

freak4pc:
The main project itself seems to have a few more warnings around 4.1, but library-wise this seems like the only issue. It's pretty much just removing the `capacity:` piece I imagine ?

kzaher:
I think we can close this.

PR #1599, opened 2018-03-16
Update CurrentThreadScheduler.swift for Swift 4.1

alanzeino:
As seen in #1598


RxSwift/Schedulers/CurrentThreadScheduler.swift
```diff
@@ -51,7 +51,11 @@ public class CurrentThreadScheduler : ImmediateSchedulerType {
     private static var isScheduleRequiredKey: pthread_key_t = { () -> pthread_key_t in
         let key = UnsafeMutablePointer<pthread_key_t>.allocate(capacity: 1)
         defer {
+#if swift(>=4.1)
             key.deallocate(capacity: 1)
```
freak4pc: These Swift branches currently do the same, I think you meant `key.deallocate()` in this case?

freak4pc:
Extra reference: This addresses changes introduced in [SE-0184](https://github.com/apple/swift-evolution/blob/master/proposals/0184-unsafe-pointers-add-missing.md).

alanzeino:
Updated the commit.

kzaher:
Hi @alanzeino @freak4pc ,

let's wait for official Xcode 9.3 before merging this. Iron is still hot 😆

keith:
Now that Xcode 9.3 is out, can we merge this change?

freak4pc:
Seems straightforward, I'll merge this and if kruno has issues he can revert. Thanks @alanzeino !

Issue #1485, opened 2017-11-07
Add support for static library frameworks


piercifani:
**Short description of the issue**:

I want to build RxSwift with the `static_framework` flag when using Cocoapods 1.4.0 beta in order to have a faster launch time

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [x] yes (which ones) 8.3.3 and 9.1
  - [ ] no

kzaher:
Hi @piercifani ,

any hints or PRs that describe what we need to do to support this in backwards compatible way?

piercifani:
Just so you know, I'm using RxSwift to try out this feature on Cocoapods, you can follow the thread [here](https://github.com/CocoaPods/CocoaPods/issues/7213)

kzaher:
@piercifani appreciate the help.

sergdort:
Seems like Cocoapods added support for [static frameworks](https://github.com/CocoaPods/Core/pull/386) and in order to support this we just need to put [`"static_framework": true`](https://github.com/CocoaPods/Specs/search?p=12&q=static_framework&type=&utf8=✓) in the podspec.
Will try to find some time tomorrow to look into it

sergdort:
So, I've managed to make it work locally for `RxSwift`, `RxBlocking` and `RxTest`. But it does not work for `RxCocoa` because it contains some `ObjC` files, not sure what is the issue TBH ¯\_(ツ)_/¯

sergdort:
I think it should work out of the box right now with CocoaPods 1.5.0 http://blog.cocoapods.org/CocoaPods-1.5.0/

kzaher:
Ok, let's close this then.

PR #1601, opened 2018-03-18
Add static libs support

sergdort:
Addresses #1485 

### Description of change

- Added `s.static_framework = true` flag to `RxSwift`, `RxBlocking`, `RxTest` podspecs
- Everything seems to work fine for `RxSwift`, `RxBlocking`, `RxTest`, I was able to build them in the test project.

screen shot 2018-03-18 at 11 54 00

### Issues

- It seems like adding `s.static_framework = true` is not enough for `RxCocoa`  as it is mixed `Objc` and `Swift` library.  [Example](https://drive.google.com/open?id=1rmT7ORe6W_tXrgHXZUIwWpIi6J8WDTU7)
- Also, It seems that we can not leave `RxCocoa` to be dynamic at the moment as well, as it can't link statically compiled `RxSwift` ¯\_(ツ)_/¯. The error I'm getting:

screen shot 2018-03-18 at 13 02 23

### DIFF
--- a/RxBlocking.podspec
+++ b/RxBlocking.podspec
@@ -16,6 +16,7 @@ Waiting for observable sequence to complete before exiting command line applicat
   s.source           = { :git => "https://github.com/ReactiveX/RxSwift.git", :tag => s.version.to_s }
 
   s.requires_arc          = true
+  s.static_framework = true
 
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.9'
--- a/RxSwift.podspec
+++ b/RxSwift.podspec
@@ -26,6 +26,7 @@ gitDiff().grep("bug").less          // sequences of swift objects
   s.source           = { :git => "https://github.com/ReactiveX/RxSwift.git", :tag => s.version.to_s }
 
   s.requires_arc          = true
+  s.static_framework = true
 
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.9'
--- a/RxTest.podspec
+++ b/RxTest.podspec
@@ -46,6 +46,7 @@ func testMap() {
   s.source           = { :git => "https://github.com/ReactiveX/RxSwift.git", :tag => s.version.to_s }
 
   s.requires_arc          = true
+  s.static_framework = true
 
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.9'


DenHeadless:
Does RxSwift *need* to be compiled as static? 

CocoaPods 1.5.0 will introduce a way to make any pod dependencies static without any code changes. 
https://github.com/CocoaPods/CocoaPods/issues/6899

Also mixed targets are also being looked into 
https://github.com/CocoaPods/CocoaPods/pull/7519

This way any developer can use RxSwift as static framework without RxSwift podspec actually declaring it as static on podspec level, thus denying any possibility to use framework as dynamic(if that's something developer is interested in).

sergdort:
Hm, I'm a little bit confused, I think that [this](https://github.com/CocoaPods/Core/pull/386#issuecomment-309282234) says that  vendor of the library is in charge how they want to ship a library.

DenHeadless:
Yeah, you absolutely can make this decision, but does RxSwift have a reason to be always static, and not dynamic?

Actually, from that artsy/eigen PR you mentioned, no podspec changes were required to make all libraries static.

I think `static_framework` will have the most value for libraries which currently are prepackaged as a static library inside dynamic one, for example Crashlytics, Fabric, TwitterKit etc. Previously there were no way to compile a static library, so those use workarounds to be compatible with CocoaPods and Swift.

DenHeadless:
@kzaher Basically there are two kinds of integration, one is podspec property "build me as a static framework", and it's released in CocoaPods 1.4.0 - [Blog post](http://blog.cocoapods.org/CocoaPods-1.4.0/).

Second integration is Podfile integration that allows you to say "Build all dependencies as static", similar to "Build all dependencies as dynamic", which all Swift CocoaPods projects use now(`use_frameworks!`). And second integration is currently in the works, and scheduled to be released in [1.5.0 release of CocoaPods](https://github.com/CocoaPods/CocoaPods/pulls?utf8=✓&q=milestone%3A1.5.0+)

I'm not sure what changes will be required for second integration to work(if any), but it looks more important than the first one, it's best to let client select how frameworks should be integrated, and not make this decision subjectively.

kzaher:
@sergdort @DenHeadless 
So if my understanding is correct, shouldn't we just rely on second method?
> Second integration is Podfile integration that allows you to say "Build all dependencies as static", similar to "Build all dependencies as dynamic", which all Swift CocoaPods projects use now(use_frameworks!). And second integration is currently in the works, and scheduled to be released in 1.5.0 release of CocoaPods

and avoid the first one completely?

I can't see any benefits for us doing the first type of integration.

freak4pc:
I'm worried about this - I think we should rely on letting people define the framework as static in their own framework. if we dictate this, we prevent people who prefer a dynamic library any options pretty much and I'm not sure its the best idea 🤔

sergdort:
TBH, I'm not sure how it worked for artsy as `use_frameworks!(false)` is the same as without `use_frameworks!`

freak4pc:
@sergdort Yeah that's what I understand as well - but IMO removing `use_frameworks` (or using the `false` option which I actually like since its explicit), is done on the `Podfile` on the user end, so there's no action needed on our end. What do you think ?

mackoj:
I understand that you might not want to change your Podfile but please consider testing your SDK in this mode.

PR #1605, opened 2018-04-01
reactive wrapper for UIPopoverPresentationController

vldalx:
The pull request adds a reactive wrapper for UIPopoverPresentationController.
Extensions:
- `prepareForPresentation`
- `willReposition`
- `didDismiss`

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -85,6 +85,7 @@ custom_categories:
   - UINavigationItem+Rx
   - UIPageControl+Rx
   - UIPickerView+Rx
+  - UIPopoverPresentationController+Rx
   - UIProgressView+Rx
   - UIRefreshControl+Rx
   - UIScrollView+Rx
@@ -116,6 +117,7 @@ custom_categories:
   - RxNavigationControllerDelegateProxy
   - RxPickerViewDataSourceProxy
   - RxPickerViewDelegateProxy
+  - RxPopoverPresentationControllerProxy
   - RxScrollViewDelegateProxy
   - RxSearchBarDelegateProxy
   - RxSearchControllerDelegateProxy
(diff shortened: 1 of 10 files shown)


freak4pc:
Thanks for the contribution @vldalx. I think there is an attempt to keep RxSwift core as light as possible and reduce the amount of code we add here. 

I feel like a better place for this extension will be at RxSwiftCommunity/RxSwiftExt. What do you think?

vldalx:
According to [Human Interface Guidelines](https://developer.apple.com/ios/human-interface-guidelines/views/popovers/), generally, popovers should be reserved for use in iPad apps. So, it means a small percentage of all the iOS apps use popovers. From this perspective, the best place for the extension is `RxSwiftExt`.

On the other hand, `UIPopoverPresentationController` is a part of `UIKit` and `RxCocoa` is a framework that helps make Cocoa APIs easier to use with reactive techniques.

freak4pc:
@vldalx RxCocoa Core doesn't include _any_ Cocoa/UIKit APIs. 
Only extremely popular ones, and usually code isn't added on this repo in general unless its an extremely requested feature.

I'd suggest perhaps proposing this as an issue in RxSwiftExt, but in general I think this might be too much of a "niche" to make an extension for in a public repo, that's just my opinion though :)

freak4pc:
@vldalx Are you OK with closing and moving this to RxSwiftExt possibly?

vldalx:
@freak4pc I completely forgot about this RP. I'll close and move it in two weeks.

freak4pc:
Cool. Closing this for now, the branch still lives so you may transfer in your free time. 

Thanks!

vldalx:
@freak4pc 
I'm moving this PR to `RxSwiftExt`. However, there is a problem with it.
In order to move tests it's necessary to have [RxTest](https://github.com/ReactiveX/RxSwift/blob/master/Tests/RxTest.swift) class in `RxSwiftExt`. Should I copy that class?

freak4pc:
RxSwiftExt already uses RxTest for tests: https://github.com/RxSwiftCommunity/RxSwiftExt/blob/master/Tests/RxSwift/MapManyTests.swift#L11

vldalx:
My question wasn't about `RxTest` framework
https://github.com/ReactiveX/RxSwift/blob/e27cded029c8c0df61d4f4778705921015f5741b/Tests/RxTest.swift#L11
It was about `RxTest` class
https://github.com/ReactiveX/RxSwift/blob/e27cded029c8c0df61d4f4778705921015f5741b/Tests/RxTest.swift#L47-L48
It's the base class for tests instead of `XCTestCase`.

freak4pc:
@vldalx That's fine, you'll just have to refactor your test to not use `RxTest` as its a helper class of `RxSwift` itself. I don't see why you'd forcefully need it just to write a test for this.

PR #1607, opened 2018-04-01
Update to Swift4.1.

RockerHX:
Remove warning.


alanzeino:
Is the deinitialize() required here? I fixed this in a different way in f3305e92e5725f0fb6815f9e5366d29b2dc1b290

freak4pc:
@RockerHX This is already fixed as @alanzeino suggested. I don't think the deinitialize is required.

freak4pc:
This was fixed in #1599 AFAIK and this can be closed

PR #1633, opened 2018-04-26
Added `timeInterval` operator

aspitz:
Added the `timeInterval` operator to `ObservableType`. Added a unit test to verify that the `timeInterval` operator is working as expected


aspitz:
Hi  @kzaher,

`timeInterval` is defined as one of the twelve ReactiveX observable utility operators. Of those twelve, seven exist today in RxSwift main repo so it seems appropriate for `timeInterval` to also exist in the main repo.

freak4pc:
Hey Ayal :) @aspitz 

I was also under that impression at the beginning but have grown to realize RxSwift core is already a pretty massive repo. Any new operator added to the library (even ones in ReactiveX) need to have good reasoning, and a wide demand. 

There is no good reason to add an operator only a few people use, as it creates additional Technical Debt etc. 

With that in mind, I think this operator could potentially be highly useful for measurements, profiling, calculations, etc. But it would have to be decided by @kzaher if this is proper to merge.

Thanks for all of your work, regardless! :)

kzaher:
Hi @aspitz ,

I agree with @freak4pc . Right now they only function of ReactiveX web site is a brief overview of the concept and a glossary of operators that one can find across different ReactiveX implementations.

We don't plan to implement each and every operator described there.

PR #1639, opened 2018-05-01
Changing BehaviorRelay to have a setter.

Herbal7ea:
This is to make working with BehaviorRelays more convenient and to have Variable syntax and BehaviorRelay syntax similar.
BehaviorRelay is the suggested successor for Variable (when it becomes deprecated), and with the value setter, this will make transition more intuitive and less cumbersome.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ All notable changes to this project will be documented in this file.
 
 #### Anomalies
 
+## [4.1.3](https://github.com/ReactiveX/RxSwift/releases/tag/4.1.3)
+
+* Adds setter to `BehaviorRelay`'s `value` field
+
 ## [4.1.2](https://github.com/ReactiveX/RxSwift/releases/tag/4.1.2)
 
 * Adds deprecation warner.
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -128,6 +128,9 @@
 		A5CD038A1F1660F40005A376 /* RxPickerViewAdapter.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5CD03891F1660F40005A376 /* RxPickerViewAdapter.swift */; };
 		A5CD038C1F1660F40005A376 /* RxPickerViewAdapter.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5CD03891F1660F40005A376 /* RxPickerViewAdapter.swift */; };
 		A5CD038D1F1660F40005A376 /* RxPickerViewAdapter.swift in Sources */ = {isa = PBXBuildFile; fileRef = A5CD03891F1660F40005A376 /* RxPickerViewAdapter.swift */; };
+		A91A139E2098FC1500B8E4B2 /* BehaviorRelayTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A91A139D2098FC1500B8E4B2 /* BehaviorRelayTests.swift */; };
+		A91A139F209900E100B8E4B2 /* BehaviorRelayTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A91A139D2098FC1500B8E4B2 /* BehaviorRelayTests.swift */; };
+		A91A13A0209900E100B8E4B2 /* BehaviorRelayTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A91A139D2098FC1500B8E4B2 /* BehaviorRelayTests.swift */; };
 		AAE623761C82475700FC7801 /* UIProgressView+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = AAE623751C82475700FC7801 /* UIProgressView+Rx.swift */; };
 		AAE623771C82475700FC7801 /* UIProgressView+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = AAE623751C82475700FC7801 /* UIProgressView+Rx.swift */; };
 		B44D73EC1EE6D4A300EBFBE8 /* UIViewController+RxTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 271A97421CFC99FE00D64125 /* UIViewController+RxTests.swift */; };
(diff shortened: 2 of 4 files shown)


freak4pc:
Hey @Herbal7ea - First of all I wanna thank you for the PR. 
I, personally, don't agree this would be a good change for two reasons.

1. Relays are modeled after RxJava's RxRelay - where `.accept()` is used as well
2. Part of the reasoning to get rid of `Variable` is to remove confusion and abuse of `Variable`-like implementation. The act of setting a "Value" imperatively, causes people to think an observable as a value, when it is indeed a set of values over the axis of time. In that sense I (personally) find it adds little value in the Rx-sense to add a value setter here.

Herbal7ea:
I think I understand where you are going with this - you goal is to to have people use Rx as it was intended (values of time).  

Whether the syntactical sugar of a setter exists, accept & value can still  be used imperatively.  But the bigger point is that it isn't immediately clear how to "set" value through `accept` for people coming from `Variable`.  A setter would be more of a convenient way of using accept, and it doesn't take away RxJava's equivalent of `.accept()`.  

If `BehaviorRelay` is to be `Variable`'s successor, people are hoping for the same simplicity and straight forwardness of setting the value.  

What I am hoping to argue for is that a setter doesn't take away from the current model, but it allows for simpler syntax to work with, without requiring a larger rewrite for people adopting RxSwift's next version.

freak4pc:
> accept & value can still be used imperatively. 

Yeah, but practically having a `value` setter makes the imperative abusing of this even worse and more common. 

`Variable` probably shouldn't have been part of RxSwift to begin with, given the fact its not part of _any_ other Rx implementation. 

I understand "syntax simplicity", but if something is simple and causes many people (usually new-comers) to make wrong stream-modeling choices or wrong design choices, that might be a good sign it isn't necessarily the best solution. 

I believe there was a discussion to eventually have a Variable alias to BehaviorRelay which might have the `value` setter back but I think that 90% out of the time, using a `.value` setter in a Rx unit just introduces the imperativeness of the design and breaks out of the Rx paradigm.

Herbal7ea:
In following that line of thought, it seems that `.value` should be removed completely if the desire is to not have people use RxSwift imperatively.  

There is little to no difference between `.accept()` and a `setter` other than to add confusion.  Both can be abused.  

For convenience I think a `setter` should be used.  But I think I have my view, and you have yours, and we both have good points.  Is there input from others or is this rejected?

freak4pc:
I'll repeat my point again. 

Having a setter creates the illusion you're changing the Observable's "value", when in reality they have no value. They only accept more next events on top of them. using a method is more explicit of that process, when a standard setter creates (and we've seen this in many questions and messages) the wrong impression on what Rx is and how it works. 

I'm not the authority to reject, that'd be @kzaher.

Herbal7ea:
But doesn't having the property even as a getter give the same idea?  Because it's a BehaviorRelay, and not something like a Replay or Publish, we only get/care about the last value set in the relay.  Subscribers will only get the most current value and `.accept()` really just sets the most current value, exactly like a setter would.  

We don't make everyone drive 20mph on the freeway because some people go 100mph.  If our concern is abuse, the value property should be removed completely.  If not, then why not have a convenience setter?

freak4pc:
As I mentioned:

- the original Relay implementation uses the `accept()` convention
- If one Relay has `.value` I'd argue they all should (and here lies the issue of "value").
- A getter `.value` isn't harmful conceptually, but I agree it's a tad strange. I guess this related to point 1 here. 

> We don't make everyone drive 20mph on the freeway because some people go 100mph.

I would go the other way - we shouldn't force everyone to use training wheels just because some people don't like riding on a bicycle as it was designed :)

Herbal7ea:
Haha! 😂  But really, people aren't forced to use the setter, they can continue to use `.accept()`.  And using the setter isn't the problem, right?  It's using RxSwift imperatively that's that problem.  `.accept()` can be misused just as easily as a setter. I.E. not having a setter doesn't protect us any more against misuse.  

It's a type of setter whether we call it one or not.  They are french fries, whether people call them freedom fries or not.  It's the bad patterns that have been taught and shared that are the problem not a convenience method.  

I'm just saying: if a) we have a getter (`.value`), and b) we have a way of setting the new value which the getter uses (`.accept()`) doesn't it make sense then to c) make a setter that is easier to use?  I mean `.accept()` is acting as a setter, just not a convenient one.

freak4pc:
I feel like we're running in circles here :) You're not going to convince me otherwise unfortunately, but this isn't my call, I'm just stating my opinion.

It's definitely much easier to abuse and use Rx-imperatively with a raw setter and also get the wrong impression as to what Rx does under the hood with that "value". `accept` is much more explicit. 

Regarding your last paragraph - no, I don't feel you should add a convenience to something that doesn't necessarily need to be used that way.

DenHeadless:
The question is, how do you model `Variable` s behavior when you do need a variable, that will change over time, and ability to check it's current value?

Rx has enough read-only objects like Observable, Driver, ControlEvent and others. But there are situations, where read-write object is more convenient than, for example, writing a bunch of CombineLatest, or Zips, or withLatestFrom.

Unfortunately, I'm not really familiar with implementations in other languages, but maybe you have some concrete examples, how this behavior can be achieved, or, how this behavior can be replaced with more "right" approach, while still being simple and usable.

DenHeadless:
Sorry about my previous, now deleted comment, I completely misunderstood the problem and jumped to wrong conclusions) For some reason I thought BehaviorRelay does not have *getter*, but it has, only setter is missing. I guess it’s fine then. First of all, alignment with other implementations is a good thing, second, ‘accept’ method already exists on PublishRelay, which makes them similar in use, as they should be.

Herbal7ea:
I would still like to pursue this behavior. I think it does add value if used correctly.

freak4pc:
@luispadron Thanks for your feedback. 

I think your kind of perspective is exactly why I am, personally, against leaving `Variable`-style behavior or adding a `value` setter for any relays. It puts beginners on the wrong path to make everything `Variable` "just to get started", and eventually end up with this more-imperative-than-declarative mutant that does not reflect what RxSwift is trying to solve. 

So the gist is: It doesn't exist in any other Rx implementation, It encourages bad architecture and behavior, it encourages wrongful thinking of how Rx internal work.

With all that, my opinion on this hasn't changed. I think this would be quite a bad change.

luispadron:
Understandable, are you all thinking of removing the getter in general? Personally, I'd prefer if we're going to discourage imperative (which I think we should) that we don't allow using it at all.

freak4pc:
@luispadron I think there are some realistic use cases to using a getter. I usually prefer to not go down that path but we should allow the option IMO since we also allow manually adding a value onto the stream.

I think Relays offer the "recommended" way of binding reactive streams, while also being flexible to provide imperative usage via accept() and `.value` setter, which is also the way its done in RxJava's [RxRelay](https://github.com/JakeWharton/RxRelay) by Jake Wharton, where these ideas were brought to RxSwift from.

freak4pc:
@Herbal7ea I know this is very old, but - 

I found a very disturbing example of way a `.value` setter is a really bad idea compared to `.accept()`. The entire premise of this discussion was "you can do the same with both", but that isn't the case. 

Note the following case (asked yesterday on the Slack channel): 

```swift
let itemsVar = Variable<[Item]>([])
let itemsRelay = BehaviorRelay<[Item]>(value: [])

// If you want to emit a new value, they act the same
itemsVar.value = [new, items]
itemsRelay.accept([new, items])

// BUT, Variable lets you abuse modifying sub-parts of collections such as Arrays.
// Modifying a member of the collection won't cause an emission, breaking the reactive chains depending on the data source.
itemsVar.value[0] = newItem // Doesn't cause an emission

// You can't abuse Relays the same way
itemsRelay.value[0] // can't mutate read-only property
itemsRelay.accept([newItem] + itemsRelay.value) // You are forced to be explicit about your mutation and provide an "entire" new `next` value
```

I had enough "theoretical" reasons on way a setter is a really bad idea, but now that I have a practical one as well, I absolutely see no way this would be a good idea for the community. 

This specific example has set back a developer a day or two into development since his mutations to a Variable didn't cause his UI to change at all. If `Variable` doesn't exist, this situation can't happen, and we don't have to worry about providing constructs that miscommunicate the intents of Reactive Programming, or misguide consumers of the framework.

freak4pc:
This has been open long enough (2 major releases) with enough community opposition that I feel comfortable closing this. Thanks for the suggestion and work, regardless!

PR #1649, opened 2018-05-11
Modified completed condition of zip

FuYouFang:
If there Three Observables like this:

|  | E1 | E2 | E3 | E4 | E5 | E6 |
| -- | -- | -- | -- | -- | -- | -- |
| O1 | 🍎| 🍎| 🍎| 🍎| 🍎| completed |
| O2 | 🍎| 🍎| 🍎| completed | | |
| O3 | 🍎| completed | | | | |

Should it end at E2 or should it end at E4 ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ All notable changes to this project will be documented in this file.
 #### Anomalies
 
 * Fixes ambiguity issue with  `Single.do(onNext:onError:onSubscribe:onSubscribed:onDispose:)` and `Single.do(onSuccess:onError:onSubscribe:onSubscribed:onDispose:)`.
+* Modified `completed` condition of `zip`.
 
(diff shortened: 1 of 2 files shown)


kzaher:
Hi @FuYouFang ,

I believe it should actually end at E6. The current behavior is confusing and we need to change this for sure. 

We are tracking this under https://github.com/ReactiveX/RxSwift/issues/1666

This is intended for RxSwift 5.0. If you like to do the changes, you are welcome.

MortyMerr:
Hey @kzaher what's the final solution on that? Conclusion in issue and in this PR seems to be different :(

kzaher:
We've made the changes to end at `E6` and released them in 5.0.

c0diq:
This changes has impacted our tests. It appears that a TestScheduler.start() method will now never return in some cases (deadlock) when that scheduler is used in an Observable that never completes. I will file a ticket separately.

PR #1651, opened 2018-05-12
Add operator asCompletable on Single trait

pietrocaselani:
This PR adds the operator `asCompletable` on Single, related to what we discussed on #1580

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 * Add `Single.flatMapMaybe(_:)` and `Single.asMaybe()`
 * Add `UICollectionView.rx.prefetchItems`, `UICollectionView.rx.cancelPrefetchingForItems`,  `UITableView.rx.prefetchRows`, and `UITableView.rx.cancelPrefetchingForRows`.
 * Add `Single.flatMapCompletable()`
+* ADD `Single.asCompletable()`
 
 #### Anomalies
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ All notable changes to this project will be documented in this file.
 * Add `Single.flatMapMaybe(_:)` and `Single.asMaybe()`
 * Add `UICollectionView.rx.prefetchItems`, `UICollectionView.rx.cancelPrefetchingForItems`,  `UITableView.rx.prefetchRows`, and `UITableView.rx.cancelPrefetchingForRows`.
 * Add `Single.flatMapCompletable()`
-* ADD `Single.asCompletable()`
+* Add `Single.asCompletable()`
 
 #### Anomalies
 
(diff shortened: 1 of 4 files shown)


pietrocaselani:
Looks like iOS tests failed with this message:
```
Error Domain=IDETestOperationsObserverErrorDomain 
Code=6 "Early unexpected exit, operation never finished bootstrapping - no restart will be attempted"
UserInfo={NSLocalizedDescription=Early unexpected exit, operation never finished bootstrapping - no restart will be attempted}
```
I closed and reopened the PR to force CI run again.
Do you know why this happened?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 * Add `Single.flatMapMaybe(_:)` and `Single.asMaybe()`
 * Add `UICollectionView.rx.prefetchItems`, `UICollectionView.rx.cancelPrefetchingForItems`,  `UITableView.rx.prefetchRows`, and `UITableView.rx.cancelPrefetchingForRows`.
 * Add `Single.flatMapCompletable()`
+* ADD `Single.asCompletable()`
```
freak4pc: ADD -> Add

RxSwift/Traits/Single.swift
```diff
@@ -327,4 +327,17 @@ extension PrimitiveSequenceType where TraitType == SingleTrait {
     public func asMaybe() -> Maybe<ElementType> {
         return Maybe(raw: primitiveSequence.source)
     }
+
+    /// Converts `self` to `Completable` trait.
+    ///
+    /// - returns: Completable trait that represents `self`.
+    public func asCompletable() -> Completable {
+        return Completable.create { completable -> Disposable in
```
freak4pc: Wouldn't it be better to just do `asObservable().ignoreElements()` ? 

I don't love the idea of subscribing inside a creation-block and using a side effect-ish kind of approach to feeding the completable.

Tests/RxSwiftTests/SingleTest.swift
```diff
@@ -585,6 +585,30 @@ extension SingleTest {
             .completed(200)
             ])
     }
+
+    func test_asCompletable() {
+        let scheduler = TestScheduler(initialClock: 0)
+
+        let res = scheduler.start {
+            (Single<Int>.just(5).asCompletable() as Completable).asObservable()
```
freak4pc: I think the multiple casting and hinting here make the test non-deterministic.
e.g. making the Completable an Observable. Not sure if there's a nicer way around this but that's my personal opinion at least.

RxSwift/Traits/Single.swift
pietrocaselani: I didn't know about `ignoreElements`. It's really better this way!

Tests/RxSwiftTests/SingleTest.swift
pietrocaselani: The problem is that the closure from `scheduler.start` returns an `Observable`. 
I could call `subscribe` on my own, store the events on an array and check the received events, but this sounds so bad 😝 
All tests on traits end with this `asObservable` call. I really don't how we can make this better.

freak4pc: Gotta. That's fair I think. If Kruno dislikes this he'll review as well.

PR #1653, opened 2018-05-13
use typealiased name for consistency

taichino:
Slightly improves readability.

### DIFF
--- a/RxExample/RxExample/Examples/TableViewPartialUpdates/PartialUpdatesViewController.swift
+++ b/RxExample/RxExample/Examples/TableViewPartialUpdates/PartialUpdatesViewController.swift
@@ -25,7 +25,7 @@ class PartialUpdatesViewController : ViewController {
 
     var timer: Foundation.Timer? = nil
 
-    static let initialValue: [AnimatableSectionModel<String, Int>] = [
+    static let initialValue: [NumberSection] = [
         NumberSection(model: "section 1", items: [1, 2, 3]),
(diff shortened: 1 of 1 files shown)


freak4pc:
I would prefer leaving as is, as its an Example - making it easier for people to understand what are the type they need to use. Type aliases are great for hiding implementation details in a code base you know, but for educational purposes it is less than optimal as you need to "dive" into the definition in this case. 

This is just my own opinion :)

taichino:
@freak4pc I see your point. But shouldn't it be consistent?

    static let initialValue: [AnimatableSectionModel<String, Int>] = [
         NumberSection(model: "section 1", items: [1, 2, 3]),
         NumberSection(model: "section 2", items: [4, 5, 6]),
         ...
     ]

^ this was little confusing to me. I think it's better to use `NumberSection` for everything or not to use it at all in the example.

freak4pc:
@taichino I see what you're saying. I don't mind either way, but I actually think the "raw type" is there to make it obvious the type alias is pointing to that type, and makes it clear what "NumberSection" is. 

We could possibly change `NumberSection` to just `.init(model:items:)` but not sure if this adds a lot for clarity's sake

taichino:
@freak4pc 

> the "raw type" is there to make it obvious the type alias is pointing to that type

Ummmm. I'm not sure if this actually helps and I still think using `NumberSection` is easier to understand (at least for me). But it's not big deal, I'll close this. Thanks for your input!

Issue #1472, opened 2017-10-24
Crash when convert Double to Int


dangthaison91:
**Short description of the issue**:

**Expected outcome**:

**What actually happens**:

Stacktrace point to: `DispatchQueueConfiguration.swift`

```
interval	TimeInterval	30875491.331709981
fatal error: Double value cannot be converted to Int because the result would be greater than Int.max

DispatchQueueConfiguration.swift - Line 21: return DispatchTimeInterval.milliseconds(Int(interval * 1000.0))
```
```swift
 var shouldExpireBefore: TimeInterval { //... }
 var willExpireAt: TimeInterval {
        return expiryDate?.timeIntervalSinceNow ?? 0
    }
    
    var expiresSoon: Observable<Void> {
        let expiresSoonAt = willExpireAt - shouldExpireBefore
        
        guard expiresSoonAt > 0 else {
            return .just(()) // fire now if it expiresSoonAlready
        }
        
        return Observable<Int>.timer(expiresSoonAt, scheduler: MainScheduler.instance).map { _ in }
    }
  ```

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

  3.6.1

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  Xcode 9.0
```

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [ ] yes (which ones)
  - [ ] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [ ] I have a small code base
  - [x] I have a significant code base

minuscorp:
I think this is just an issue that you're excceding Int value, which depending of the device's architecture, it can be of 32 or 64 bit length. You can use a safe version of the crashing line using `Int.init(exactly:)` and either `round` or `floor` to avoid losing the conversion due to the decimal part of the numer.

```swift
[...]

PR #1658, opened 2018-05-23
Rename `take(_:scheduler:)` to `take(duration:scheduler:)` to disambiguate overloads

### LINKED ISSUE #1625: `take(_ duration:scheduler:)` and `take(_ count:)` usage ambiguity
There is a great confusion between `take(_ duration:scheduler:)` and `take(_ count:)` in the sense they both don't require a label for the first argument.

### take(_ duration:scheduler:) : 
https://github.com/ReactiveX/RxSwift/blob/f1e215a3fe1303568e415095cb00a6ca2a27b706/RxSwift/Observables/Take.swift#L41

### take(_ count:) :

https://github.com/ReactiveX/RxSwift/blob/f1e215a3fe1303568e415095cb00a6ca2a27b706/RxSwift/Observables/Take.swift#L19

I would suggest dropping the `_` from the duration variation, making it `take(duration:scheduler:)` to get rid of the ambiguity. 

This was noticed when @irace asked a question with an example that uses both of them in the same sequence, with code similar to this: 

```swift
observable
  .take(5)
  .take(8, scheduler: MainScheduler.instance)
```

There is no real way to know what the latter does. The "count" one makes sense to be label-less.

This can be done via either
1. Soft deprecating in a minor patch
2. Hard deprecating in next major
3. A combination of both (stepped deprecation)

Let me know your thoughts.

freak4pc:
Fixes #1625. 

I took a minor decision to already mark the old version with an `@available` clause given the fact this is a relatively low-impact method being rarely used, and given the code change extremely low-effort (just adding `duration:` to the call-site). 

Using something like `DeprecationWarner` here seems like an overkill but I can go that way if you feel you prefer it.


freak4pc:
Would you also mind adding this before cutting the new release? @kzaher

kzaher:
@freak4pc Sorry, didn't notice the PR early enough.

kzaher:
Hi @freak4pc ,

I've checked out the PR. I think that because of this issue https://github.com/ReactiveX/RxSwift/issues/1472 we should immediately make the type of parameter name `duration` to `DispatchTimeInterval`.

freak4pc:
Hey @kzaher 
I still think those are two entirely different changes. 

Even if we fix the signature to support DispatchTimeInterval, the rest of the methods won't. That would be a much more vast change.

freak4pc:
Yeah so @kzaher I quickly went through it - There is no easy "clean" way to make this. 

You would need to have all SchedulerType(s) provide a `scheduleRelative` and `schedulePeriodic` that take a `DispatchTimeInterval` and also decide if we completely want to drop the `RxTimeInterval` alias to `TimeInterval` or support both in a backward compatible way. 

IMO this a **much more widespread** change that is really out of scope for this tiny fix. 

I'd love to do a separate PR for introducing `DispatchTimeInterval` but it needs more discussion and **much** more work.

kzaher:
Hi @freak4pc ,

> You would need to have all SchedulerType(s) provide a scheduleRelative and schedulePeriodic that take a DispatchTimeInterval and also decide if we completely want to drop the RxTimeInterval alias to TimeInterval or support both in a backward compatible way.

Yes, we would need to add two new overloads and a default implementation for new methods so we don't break custom schedulers.

We would also need to reroute methods that use RxTimeInterval to use DispatchTimeInterval in current scheduler implementations.

Third step is adding operators that just use DispatchTimeInterval and deprecate RxTimeInterval.

 > IMO this a much more widespread change that is really out of scope for this tiny fix.

I would agree. But it doesn't seem to me that it makes sense to deprecate the same API twice. Especially since we are already have #1472 reported.

freak4pc:
@kzaher I see. I just think landing a PR of that vast of a change would take much more work, as well as reviewing that sort of change. Would definitely look into it though. 

I would be OK with a double deprecation, as changing all schedulers will take much longer. But If you feel strongly about this, I prefer closing this PR for now as it will not be relevant.

kzaher:
Hi @freak4pc ,

I can also make the scheduler change. It doesn't seem to me like that big of a change at the moment, but I might be wrong. I can check it out in a couple of days.

freak4pc:
Thanks @kzaher 
I tired playing with it a bit but TBH feel like I'm dropping right into the rabbit hole :) 
Schedulers are so deeply rooted and even though I know most of the system pretty well, I don't know Scheduler implementation that well. 

Also these kind of notes leave me worried of touching this: 😂 
```
// TODO:
// This looks horrible, and yes, it is.
```

freak4pc:
@kzaher I vote for closing this for now until you have time to look into redoing Schedulers. I don't think me doing them would be good since I really don't 100% understand the underlying implementation and its a really low-level concept to RxSwift.

PR #1659, opened 2018-05-23
Fix Project Warnings & Deprecations

freak4pc:
No CHANGELOG entry since this is #trivial. 

- Update project to resolve warnings

- Fix test using the only `shareReplay, shareReplayLatestWhileConnected` methods

- Use `targetEnvironment` to correctly target simulator

- Fix Wikipedia example to use compactMap vs. flatMap


Tests/RxSwiftTests/Anomalies.swift
```diff
@@ -59,10 +59,9 @@ extension AnomaliesTest {
         }
 
         for op in [
-                { $0.shareReplay(1) },
+                { $0.share(replay: 1) },
                 { $0.replay(1).refCount() },
-                { $0.publish().refCount() },
-                { $0.shareReplayLatestWhileConnected() }
```
freak4pc: I removed `shareReplayLatestWhileConnected` since its the same as `share(replay: 1)` with default scope

kzaher:
Hi @freak4pc ,

Build unfortunately server is failing, and it doesn't look like a transient error to me.

freak4pc:
@kzaher Seems clean now. Let me know if this works for you. Had to bump CI to Xcode 9.3.

PR #1661, opened 2018-05-28
Reduce Bag dispatch implementation code size

hmlongco:
Performance enhancement reduces Bag dispatch inlined code size by 12%.

I ran the Benchmarks on this change and the original version about a dozen times each, and depending upon the position of the sun, moon, and stars the revised code is faster / slower / exactly the same speed as the original. I even wrote another test case to check multiple subscription dispatches.

All of which leads me to believe that any variation in execution speed depends solely on whatever else my system happens to be doing at the same time.

The generated inline code is, however, about 12% smaller, the size benefit primarily obtained by removing the need to handle the optional _value0 twice.

Since dispatch is inlined and used in quite a few places, I felt that the change, while small, would be significant in an app that used RxSwift extensively.


kzaher:
Hi @hmlongco ,

When you say
> I ran the Benchmarks on this change and the original version about a dozen times each, and depending upon the position of the sun, moon, and stars the revised code is faster / slower / exactly the same speed as the original. I even wrote another test case to check multiple subscription dispatches.

I don't think that neither of the current benchmark tests, not the one you've added tests this scenario.

Fast path is optimization intended for the case when there is only a single observer attached. I'll try to add a benchmark that tests for that case, or you can also try if you like.

Issue #1673, opened 2018-06-14
RxCocoa error in Xcode 10 (`Cycle inside RxCocoa; building could produce unreliable results.`)


jdisho:
I get the following error, when I try to build a project that has RxCocoa as a dependency.  🤔

<img width="454" alt="screenshot 2018-06-14 at 14 03 57" src="https://user-images.githubusercontent.com/8223362/41411087-e82f970c-6fdb-11e8-8d71-38cff6d94295.png">

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

```
  4.0
```

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  Xcode 10
```
**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [ ] I have a small code base
  - [x] I have a significant code base

kzaher:
Hi @jdisho ,

have no idea, PRs are welcome. I've created a new branch for [Xcode10](https://github.com/ReactiveX/RxSwift/tree/xcode10).

carlosypunto:
Appkit and UIKit has API changes in Swift 4.2. @kzaher, new branch [Xcode10](https://github.com/ReactiveX/RxSwift/tree/xcode10) will be for Swift 4.2?
@jdisho, has you try change swift version from 4.2 to 4.0 _RxCocoa_ target in Pod project?

kzaher:
@carlosypunto I think we should first figure out how to fix it, and if it's possible to make compatible changes with Xcode 9, then we'll use the same branch, otherwise we'll think about it more.

carlosypunto:
Ok @kzaher. I took a little while and did the update, basically the problem is with RxCocoa and RxExample, since RxSwift is Swift 4.2 code compatible. I could not pass the tests because (like in #1677 issue) runnin `./scripts/package-spm.swift` twice don't make the same content of `Sources/AllTestz/main.swift` so it will always modify the git working copy stoping the tests. I hope I have clarified something.

hmlongco:
[...]

PR #1676, opened 2018-06-16
Add insert(...) to DisposeBag

hmlongco:
DisposeBags are a necessary evil in Swift's ARC-based environment. Who hasn't written the following code?

```
    viewModel.name.bind(to: nameLabel).disposed(by: disposeBag)
    viewModel.address.bind(to: addressLabel).disposed(by: disposeBag)
    viewModel.city.bind(to: cityLabel).disposed(by: disposeBag)
    viewModel.state.bind(to: stateLabel).disposed(by: disposeBag)
    viewModel.zip.bind(to: zipLabel).disposed(by: disposeBag)
```

Every element is properly disposed of with its own `.disposed(by: disposeBag)`, but we're writing a lot of boilerplate code in the process.

Worse, almost 40% of our code is devoted to housekeeping and memory management, obscuring our actual intent.

This [PR](https://github.com/ReactiveX/RxSwift/pull/1676) adds a simple addition to RxSwift's DisposeBag: Bag. 

With it, the above code can be written as follows:

```
    disposeBag.bag(
        viewModel.name.bind(to: nameLabel),
        viewModel.address.bind(to: addressLabel),
        viewModel.city.bind(to: cityLabel),
        viewModel.state.bind(to: stateLabel),
        viewModel.zip.bind(to: zipLabel)
    )
```

The disposeBag.bag() function simply gathers up (bags) all of the disposables into a Swift variadic function and inserts all of them into the DisposeBag as a single group.

Even better, our code is a lot cleaner and our actual intent is much more obvious.

DisposeBag.bag can be written as a simple extension...

```
extension DisposeBag {
    public func bag(_ disposables: Disposable...) {
        disposables.forEach { self.insert($0) }
    }
}
```

But baking it directly into DisposeBag makes the functionality much more performant, requiring only a single lock/unlock pair and a single memory reallocation to the _disposables array. 

```
extension DisposeBag {
    public func bag(_ disposables: Disposable...) {
        if _isDisposed {
            disposables.forEach { $0.dispose() }
        } else {
            _disposables += disposables
        }
    }
}
```

One should note that in our first code sample each individual `.bind(to:).disposed(by: disposeBag)` is going to require its own individual lock/insert/unlock as well. DisposeBag.bag() avoids this.

So. Less code. Clearer code. Faster code. What's not to like?

### DIFF
--- a/RxSwift/Disposables/DisposeBag.swift
+++ b/RxSwift/Disposables/DisposeBag.swift
@@ -86,12 +86,7 @@ public final class DisposeBag: DisposeBase {
 extension DisposeBag {
     /// Convenience function allows a list of disposables to be gathered for disposal.
     public func insert(_ disposables: Disposable...) {
-        _lock.lock(); defer { _lock.unlock() }
-        if _isDisposed {
-            disposables.forEach { $0.dispose() }
-        } else {
-            _disposables += disposables
-        }
+        insert(disposables)
     }
 
     /// Convenience function allows an array of disposables to be gathered for disposal.
--- a/RxSwift/Disposables/DisposeBag.swift
+++ b/RxSwift/Disposables/DisposeBag.swift
@@ -32,14 +32,14 @@ public final class DisposeBag: DisposeBase {
     private var _lock = SpinLock()
     
     // state
-    private var _disposables = [Disposable]()
-    private var _isDisposed = false
+    fileprivate var _disposables = [Disposable]()
+    fileprivate var _isDisposed = false
     
     /// Constructs new empty dispose bag.
     public override init() {
         super.init()
     }
-    
+
     /// Adds `disposable` to be disposed when dispose bag is being deinited.
     ///
     /// - parameter disposable: Disposable to add.
@@ -80,5 +80,17 @@ public final class DisposeBag: DisposeBase {
     
     deinit {
         dispose()
+    }
+}
+
+extension DisposeBag {
+    /// Convenience function allows a list of disposables to be gathered for disposal.
+    public func insert(_ disposables: Disposable...) {
+        _lock.lock(); defer { _lock.unlock() }
+        if _isDisposed {
+            disposables.forEach { $0.dispose() }
+        } else {
+            _disposables += disposables
+        }
     }
 }
--- a/RxSwift/Disposables/DisposeBag.swift
+++ b/RxSwift/Disposables/DisposeBag.swift
@@ -93,4 +93,14 @@ extension DisposeBag {
             _disposables += disposables
         }
     }
+
+    /// Convenience function allows an array of disposables to be gathered for disposal.
+    public func insert(_ disposables: [Disposable]) {
+        _lock.lock(); defer { _lock.unlock() }
+        if _isDisposed {
+            disposables.forEach { $0.dispose() }
+        } else {
+            _disposables += disposables
+        }
+    }
 }
--- a/RxSwift/Disposables/DisposeBag.swift
+++ b/RxSwift/Disposables/DisposeBag.swift
@@ -84,6 +84,19 @@ public final class DisposeBag: DisposeBase {
 }
 
 extension DisposeBag {
+
+    /// Convenience init allows a list of disposables to be gathered for disposal.
+    public convenience init(disposing disposables: Disposable...) {
+        self.init()
+        _disposables += disposables
+    }
+
+    /// Convenience init allows an array of disposables to be gathered for disposal.
+    public convenience init(disposing disposables: [Disposable]) {
+        self.init()
+        _disposables += disposables
+    }
+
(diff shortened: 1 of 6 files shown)


hmlongco:
Running tests locally goes into loop.

`Package for Swift package manager isn't updated, please run ./scripts/package-spm.swift and commit the changes`

I do so, commit, run it again and get the same error.

hmlongco:
Fails with following issue: https://github.com/ReactiveX/RxSwift/issues/1673

Running all tests locally succeeds.

kzaher:
Hi @hmlongco ,

I think we could add something like this, but it would probably make sense to add variadic version of `insert` for consistency sake.

RxSwift/Disposables/DisposeBag.swift
```diff
@@ -82,3 +82,31 @@ public final class DisposeBag: DisposeBase {
         dispose()
     }
 }
+
+extension DisposeBag {
+    /// Convenience function allows a list of disposables to be gathered for disposal.
+    ///
+    /// With disposeBag.bag:
```
kzaher: I would remove this explanation and just leave
```
Convenience function allows a list of disposables to be gathered for disposal.

/// - parameter disposables: List of disposables to add.
```

This is a pretty simple API.

hmlongco:
@kzaher I updated the PR, renaming bag to insert (more prosaic) and updating the tests. However, when I go a git status I see the following:
``` 
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	typechange: Sources/AllTestz/DisposeBagTest.swift
	typechange: Sources/RxSwift/DisposeBag.swift
	typechange: Sources/RxSwift/ObservableType+Extensions.swift
	typechange: Sources/RxSwift/Single.swift
```
I can't get rid of the type changes, so it's listing conflicts. (git status | grep typechange | awk '{print $2}' | xargs git checkout, which usually works, doesn't)

kzaher:
Hi @hmlongco ,

I believe that rebasing on top of `develop` should do the trick.

freak4pc:
Hey @hmlongco 
I fixed the AllTestz conflict and squashed your commits - should be good to go as soon as we have a green. 

TBH I'm not entirely sure `DisposeBag.insert(...)` is necessarily the best name but don't have a better suggestion at this point. 

I see insert used for `Set`s in general and `Array` for specific indices, but I don't know if it entirely conveyes what this helper does. In any way its very helpful to have.

RxSwift/Disposables/DisposeBag.swift
```diff
@@ -93,4 +93,14 @@ extension DisposeBag {
             _disposables += disposables
         }
     }
+
+    /// Convenience function allows an array of disposables to be gathered for disposal.
+    public func insert(_ disposables: [Disposable]) {
```
freak4pc: DRY - No reason to "rewrite" this method twice. You can use the splatted/variadic splatted version, make it an array and pass it down to this method IMO

hmlongco: Didn't occur to me that the veriadic array was an array I could pass into the second version. Thanks.

hmlongco:
I still prefer the disposeBag.bag( .... ) syntax, but kzaher thought it better to use insert since that's the singular function name too.

freak4pc:
`bag` is more confusing IMO (since its a verb and noun). Between the two, `insert` is better. It's just overloaded with different connotations from `Set` vs. `Array`.

freak4pc:
By the way, just as an after-thought, what would you think of an (additional) initializer for `DisposeBag`: 

```swift
self.disposeBag = DisposeBag(disposing: [ 
  lblName.rx.text.bind(to: viewModel.name), 
  // etc etc...
])
```

The up-side of this variation is that, for non-storyboard classes, it could have a finite DisposeBag that is potentially immutable, and also helps in making the subscriptions more "finite" in their feel. 

It is an addition to solve the seem "readability" issue, that just came to mind. Would it be something you'd be interested in?

hmlongco:
Could work. Also tend to like that name better than insert...
```
disposeBag.disposing(
     viewModel.name.bind(to: nameLabel)
     viewModel.address.bind(to: addressLabel)
     viewModel.city.bind(to: cityLabel)
     viewModel.state.bind(to: stateLabel)
     viewModel.zip.bind(to: zipLabel)
    /// etc etc...
)
```

hmlongco:
Sigh. I think the variadic to array conversion fails on earlier versions of swift.

freak4pc:
@hmlongco I think `disposing` on an instance of `DisposeBag` isn't good, because it conveyes the meaning it replaces the current disposables instead of adding or inserting.

freak4pc:
@hmlongco I think you just missed the SPM tests. This should work now.

Issue #501, opened 2016-02-21
doOnSubscribe and doOnUnsubscribe?


JSchaenzle:
Is there an equivalent of doOnSubscribe and doOnUnsubscribe that can be tacked on to a signal?

kzaher:
Hi @JSchaenzle ,

This is similar to https://github.com/ReactiveX/RxSwift/issues/398

I'm just curios of the context you need `doOnSubscribe` and `doOnUnsubscribe`?

People usually need it to tie lifetime of some external hidden resource with observable sequence lifetime, or debugging.

In case you need it for external resource lifetime, maybe `using` operator is better.
In case you need it for debugging, we do have `debug` operator.

Trying to figure out do we really need to add this convenience functionality.

JSchaenzle:
@kzaher,
Thanks for the response. I am used to ReactiveCocoa and RxJava where I use operators like `doOnSubscribe` and `doOnUnsubscribe` for debugging complex signals. I'm just getting started with RxSwift (and really liking it!) and I didn't realize that the debug operator was a thing. I think that will be exactly what I need. Thanks again for pointing that out to me.

lbatteau:
I like to use `doOnSubscribe` for indicating network activity. When creating an Observable that performs a network request, the subscribe event is the actual start of the request. Hope that makes sense.

sergdort:
Hi, @lbatteau I guess you can use `using` operator or [ActivityIndicator](https://github.com/ReactiveX/RxSwift/blob/master/RxExample/RxExample/Services/ActivityIndicator.swift)

Issue #398, opened 2016-01-09
Request: Add doOnDisposed operator


thanegill:
As seen here: http://www.introtorx.com/Content/v1.0.10621.0/11_AdvancedErrorHandling.html#Finally

DanielAsher:
I recently needed this operator. I was advised to make use of the new `using` operator, but `finally` seems like a much more elegant solution.

+1

kzaher:
We currently have `using` operator. It looks like a boring solution, but I believe people really do need it more then `finally` :)

It looks to me that `finally` is one of those side effect operators, and could be maybe added to `do` operator.

http://reactivex.io/documentation/operators/do.html

I'm not sure that naming makes much sense in the context of Swift. 

In C# you have `try`,`catch`,`finally`.
I believe the same is valid for Java.

I think the closest thing in Swift would be `defer`, but that is a keyword, so not sure how would I add it as an operator.

It seems to me that simplest and most consistent way to implement finally would be to add `onDisposed` to `do` operator.

Disposed is executed in all cases finally is.

DanielAsher:
@kzaher a `doOn(onDisposed:` would work nicely.

I feel a convenience function named `finally` conveys well its purpose.

DanielAsher:
I would add that I believe a `finally` combinator is lighter-weight than the `using` factory function.

thanegill:
An `doOn(Disposed: )` operator would work for me. And while we're at it an `doOn(Subscribed: )` would also be nice. 

Is there an example of `using` with the equivalent functionality of `doOn(Disposed: )` and `doOn(Subscribed: )`?

thanegill:
Below is what I figured out for a `doOn(Disposed: )` operator, is this the right way to do this?

```swift
extension Observable {
    public func doOnDisposed(onDisposed: () -> ()) -> Observable<Element> {
        return Observable<Element>.using({
                return AnonymousDisposable(onDisposed)
        }, observableFactory: { _ -> Observable<Element> in
            return self
        })
    }
}
```

kzaher:
@thanegill I believe the code you've written should work.

```
extension Observable {
    public func doOnMyExtensions(onSubscribed: () ->(), onDisposed: () -> ()) -> Observable<Element> {
        return Observable<Element>.using({
                onSubscribed()
                return AnonymousDisposable(onDisposed)
        }, observableFactory: { _ -> Observable<Element> in
            return self
        })
    }
}
```

[...]

PR #1681, opened 2018-06-19
Add subscribe(...,onDisposed:) to Single

sirlantis:
I was surprised that this parameter from `Observable.subscribe` was not available on `Single.subscribe`.

I'm not sure whether this qualifies as a [new operator](https://github.com/ReactiveX/RxSwift/blob/master/CONTRIBUTING.md#new-operators). It sounds like there was some resistance to #501 and #398 recommending to do `using` instead. Just let me know if you feel this PR belongs in the same bucket.

---

In my specific instance the `Single` is a promise/future, so I'm looking for a *finally*-like solution. This could be very concise like this:

```
single
  .subscribe(
    onSuccess: { ... }
    onError: { ... }
    onDisposed: alwaysDoThis
  )
```

Currently, I would have to use `do`:

```
single
  .do(
    onSuccess: { ... }
    onError: { ... }
    onDispose: alwaysDoThis
  )
  .subscribe()
```

or switch to Observable:

```
single
  .asObservable()
  .subscribe(
    onNext: { ... }
    onError: { ... }
    onDisposed: alwaysDoThis
  )
```

or repeat code (at the risk that one forgets calling it in one branch):

```
let alwaysDoThis = { ... }

single
  .subscribe(
    onSuccess: { ...; alwaysDoThis() }
    onError: { ...; alwaysDoThis() }
  )
```

or rely that `onDispose` will be called after onSuccess/onError (which feels odd and breaks the second someone inserts `observeOn` between the two instead of in front of them):

```
single
  .do(onDispose: { ... })
  .subscribe(
    onSuccess: { ... }
    onError: { ... }
  )
```

### DIFF
--- a/RxSwift/Traits/Single.swift
+++ b/RxSwift/Traits/Single.swift
@@ -78,16 +78,18 @@ extension PrimitiveSequenceType where TraitType == SingleTrait {
      
      - parameter onSuccess: Action to invoke for each element in the observable sequence.
      - parameter onError: Action to invoke upon errored termination of the observable sequence.
+     - parameter onDisposed: Action to invoke upon any type of termination of sequence (if the sequence has
+     gracefully completed, errored, or if the generation is canceled by disposing subscription).
      - returns: Subscription object used to unsubscribe from the observable sequence.
      */
-    public func subscribe(onSuccess: ((ElementType) -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil) -> Disposable {
+    public func subscribe(onSuccess: ((ElementType) -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil, onDisposed: (() -> Void)? = nil) -> Disposable {
         #if DEBUG
-             let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []
+            let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []
         #else
             let callStack = [String]()
         #endif
-    
-        return self.primitiveSequence.subscribe { event in
+
+        let subscription = self.primitiveSequence.subscribe { event in
             switch event {
             case .success(let element):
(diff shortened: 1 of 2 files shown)


RxSwift/Traits/Single.swift
```diff
@@ -99,6 +101,15 @@ extension PrimitiveSequenceType where TraitType == SingleTrait {
                 }
             }
         }
+
+        if let disposed = onDisposed {
```
hmlongco: I don't think this is gong to do what you want. As just one case, consider:
```
Just(1).asSingle().subscribe(onSuccess: { _ in }).disposed(by: bag)
```
The call to self.primitiveSequence.subscribe is going to fire the subscription to Just, return the value, complete and dispose all before this line is even hit. As a result, the subscription is already disposed, but your onDispose handler isn't going to be fired until the disposeBag is deallocated and emptied.

If you're going to do this, you're going to need to hook into `self.primitiveSequence.subscribe` much in the same way as it's done in the subscribe method in `ObservableType+Extensions.swift`.

That would probably also require that you add onDisposed to the other traits as well.

sirlantis: Oh, good point. I'll wait for some feedback on whether I should continue with the PR or scrap it altogether.

kzaher:
Hi @sirlantis ,

I guess we'll need to add this sooner or later because people will want to use it. I'm fine with pulling this, just please fix the code so it behaves correctly. Please take a look at observable extensions and how we've implemented it there.

freak4pc:
Hey @sirlantis - Are you still interested in pursuing this PR?

sirlantis:
Hey @freak4pc - I'd like to but I'm swamped right now and can't say when I'll find the time to follow-up. 😞

freak4pc:
Work being done in #1977

PR #1692, opened 2018-07-12
Missing emit to BehaviorRelay methods for Signal

sinarionn:
I guess you guys just missed it? On the other hand - for Driver you missed methods for PublishRelay, maybe it's not a coincidence?


kzaher:
Hi @sinarionn ,

yeah, that was on purpose. `Signal` is event like, and `BehaviorRelay` is state like. It would be weird to do this.

sinarionn:
@kzaher it's just "where do we want to forward event from sequence". is there any difference between abstract observer or specialized PublishRelay/BehaiviourRelay/Binder? 

We are just getting events from our sequences (observables/drivers/signals they all just sequences with some different logic layers) and forwarding them to someone who can handle it (observer/BehaviourRelay/PublishRelay/Binder and so on - they all can accept those values). 

That "purpose" looks strange for me, because it looks like i cant use Signal's with Binder's/BehaviorRelay "just because". You are saying about "event like" Signal nature, but what about Driver? You think it's more "state like" just because it replay's latest value, and so i can drive BehaviorRelay and  can't PublishRelay? If so - you can close this PR, and i will be forced to write a bunch of additional extensions on my side )

freak4pc:
> "I will be forced to write a bunch of additional extensions on my side"
I would not recommend writing extensions just to go around best-practices :) 

Signal, to me, models Events just like Kruno stated. For example, a tap on a button should not have a replay, since only future-taps are relevant. 

`Signal` doesn't replay while `Driver` does replay - those are the "end" results in a way. 

Whatever you "convert" into these types (be it `Behavior*` or `Publish*`) are "clamped" by whatever behavior the traits define, and not by the original emissions. 

For example, when using a PublishRelay and making it a Completable, you wouldn't expect it to emit `next` events - because `Completable`'s promise is to "not emit next events", exactly like `Signal`'s is to not replay on subscription. 

Hope this makes some sense :)

sinarionn:
Not really :) You are trying to convince me that there is some "Bad practice" to use Signal and BehaviourRelay together, but they all are just components that should be used in a proper place in a proper time.

Here is my example: i'm using Signal's for routing events from viewModel's, from all available observables - signal is best for me (sharing suvscription and main thread is great, but no need in replay, so Signal > Driver). My another component Router gathers all routing events from all currently active vm's and handles them by building/presenting new ui. At some moment i want to write additional logic layer based on current app state, so i want to map all routing event's into proper enum state and emit results into Behaviour* (it's the best component in this case, as it behaves exactly like i need - it stores latest active state) and i cant do it directly without emit methods i was asking about. 

Long story short - you think that if for some reason the fact that at some place of my code event-like Signal's was the best to use - it dictates me not to use their events with state-like Behaviour* in another parts of my code when i need exactly this behaviour.

freak4pc:
I’m not saying anything. You can do whatever you want :) 

I’m just saying that if you use Signal - there will be no replay, doesn’t matter if the signal is originating from a BehaviorSubject or PublishSubject. You can do whatever you want with this information :)

sinarionn:
Wait a second, are you sure i understand me right?

sinarionn:
We are discussing why Signal's doesn't have method `emit(to: BehaviourRelay)`
I'm not asking Signal's to start replying - we have Driver's for this.

freak4pc:
Yikes, sorry I totally got blacked out. Just re-read the code and realized what you're asking for. 

Still not 100% sure this is a good choice to do (bridging Signal to a "replaying" subject), but on the other hand - even if it is "weird", I think there should be a way to make that connection if a user has a good use case for it.

kzaher:
Hi @sinarinonn,

Can you please write some pseudo code with that example you’ve explained?

And also the point at which the transformation occurs. 

Just the method signatures for thoses apis would be enough. 

I’ve never seen example like that, so I’m curios did I miss something.

sinarionn:
Hi @kzaher . I don't know why do you need that code, it's just Signal that emits it's value into BehaviorRelay. It was just explanation about example case when it can be needed.
Here i wrote fast example: [https://pastebin.com/S23hnfc0](https://pastebin.com/S23hnfc0)
And it's just one from many possible examples (i have few more in my project). I'm seeing some problem (IMHO) in your logic, you think that if some feature or code produces event-oriented result, this result can't be used directly with state-oriented code. I think that this is contrived restrictions.

sinarionn:
In such cases i think: Observables/Signals/Drivers - outputs, Publish/Behavior/Binder - inputs, and this should be developer choice what to use and when to use it.

kzaher:
Hi @sinarionn ,

> And it's just one from many possible examples (i have few more in my project). I'm seeing some problem (IMHO) in your logic, you think that if some feature or code produces event-oriented result, this result can't be used directly with state-oriented code. 

I guess one part why I'm doing this project is to educate myself. If there is something obvious that I've missed, I would like to understand what that is. 

The APIs that we've built into this project should support the most general use case.

The example that you've provided is a degenerated reducer. I've experienced similar situations myself once or twice. You have a situation where the state is so fine grained that the new state is equal to the actual event.

Reducer is in the form
```
func reduce<S, S>(state: S, event: S) -> S {
     return event
}
```

People who are using MVVM are probably encountering this quite often. E.g. if there is a `BehaviorSubject` for storing a single string, and textbox changes can be represented by a single string.

> I think that this is contrived restrictions.

It's a matter of perspective.

> In such cases i think: Observables/Signals/Drivers - outputs, Publish/Behavior/Binder - inputs, and this should be developer choice what to use and when to use it.

I'm also a developer :)

Fine, if you want to use such fined grained state, be my guest. I won't let my dislike for MVVM stop this PR.

Just please add unit tests before we can merge this.

kzaher:
> On the other hand - for Driver you missed methods for PublishRelay, maybe it's not a coincidence?

I would need some convincing why would Driver -> PublishRelay be a good idea. Until then, we won't have that pair.

sinarionn:
@kzaher,  sorry, I didn't want to appear rude. i guess i need to work on my english language skills )
Anyway - i wrote some tests (copied them from Driver actually). 

> I would need some convincing why would Driver -> PublishRelay be a good idea.

Well, for example i don't see any other way to convert Driver into event-like Signal or simple observable, and remove `replay` logic. Of course you'll say about "general use case" and maybe even will be right (in this case) :)

kzaher:
Both `Driver` and `Signal` have `asObservable` so you can always bind what ever concepts you like.

sinarionn:
Yep, but resulting observable will still replay latest value if it was converted from Driver.

PR #1698, opened 2018-07-18
Minor cleanup for Documentation

freak4pc:
After some questions about Driver sharing strategy, I linked to the documentation and while reading noticed there are some issues with it. 

This is in no way a “complete fix up”, but I cleaned it up and added some missing info to the best of my limited time :) 

If there are any suggestions or feedback I’d love to address them.

The main points addressed are:

- Replacing instances of `shareReplay`, `shareReplayLatestWhileConnected` with `share(replay:scope:)`.

- Add a few words on Relays.

- Fix `Event` case names.

- Removed references to `Variable` as well as an entire section dedicated to it, as it’s deprecated. 

- Removed outdated Migration docs (They refer to `RxSwift 1.9 -> 2.0`)

- Add basic explanation of `Signal`. 

- General cleanup and formatting.


LightD:
Didn't even know the real difference between `Signal` and `Driver`. Thanks for that and for removing all `Variable` references.

kzaher:
Perfect 👍

PR #1699, opened 2018-07-18
Fixes unused parameter in UIControl+Rx

luispadron:
Simple fix I came across while debugging some other stuff.
Seems like the idea was to have a parameter with a default value but that parameter isn't actually being used, this PR fixes that

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ All notable changes to this project will be documented in this file.
 * Add  `Single.catchErrorJustReturn(_:)` operator.
 * Add  `Single.asMaybe(_:)` operator.
 * Add  `Single.asCompletable(_:)` operator.
+* Fixes unused parameter in `UIControl+Rx`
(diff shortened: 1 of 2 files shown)


freak4pc:
Nice catch @luispadron !
Looks good to me but I'll give @kzaher the option to look over it.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ All notable changes to this project will be documented in this file.
 * Add  `Single.catchErrorJustReturn(_:)` operator.
 * Add  `Single.asMaybe(_:)` operator.
 * Add  `Single.asCompletable(_:)` operator.
+* Fixes unused parameter in `UIControl+Rx`
```
freak4pc: I would change this to "Use `editingEvents` argument in `controlPropertyWithDefaultEvents`"

luispadron: Will update soon

PR #1700, opened 2018-07-19
disposed:by: instead of disaposable:with:

yonivav:
disaposable:with: is depracated


freak4pc:
Hey @yonivav, thank you very much for your contribution but this already addressed in #1698 :) 

Appreciate your help.

Shai.

PR #1702, opened 2018-07-20
Allow binding to multiple observers

freak4pc:
This is based on a PR in RxSwiftExt: https://github.com/RxSwiftCommunity/RxSwiftExt/pull/166

As well as a discussion here: https://github.com/RxSwiftCommunity/RxSwiftExt/issues/168

I wondered if you'd be open to adding something like this to the main repo as it seems useful, and would also save us from a few external overloads, possibly. 

It's also possible having separate overloads for `bind(to: Observer)`, `bind(to: Observer...)`, `bind(to: [Observer])` but I don't think there would be a significant performance difference in that case. (**edit**: Running the `Benchmarks` tests confirms that)

If this makes sense to you, I'll add everything needed including tests etc. This maintains backward compatibility.

### DIFF
--- a/RxCocoa/Common/Observable+Bind.swift
+++ b/RxCocoa/Common/Observable+Bind.swift
@@ -6,40 +6,47 @@
 //  Copyright © 2015 Krunoslav Zaher. All rights reserved.
 //
 
-
 import RxSwift
 
 extension ObservableType {
-    
     /**
-    Creates new subscription and sends elements to observer.
-    
-    In this form it's equivalent to `subscribe` method, but it communicates intent better, and enables
-    writing more consistent binding code.
-    
-    - parameter to: Observer that receives events.
-    - returns: Disposable object that can be used to unsubscribe the observer.
-    */
-    public func bind<O: ObserverType>(to observer: O) -> Disposable where O.E == E {
-        return self.subscribe(observer)
-    }
-
-    /**
-     Creates new subscription and sends elements to observer.
-
-     In this form it's equivalent to `subscribe` method, but it communicates intent better, and enables
+     Creates new subscription and sends elements to observer(s).
+     In this form, it's equivalent to the `subscribe` method, but it better conveys intent, and enables
      writing more consistent binding code.
-
-     - parameter to: Observer that receives events.
-     - returns: Disposable object that can be used to unsubscribe the observer.
+     - parameter to: Observers to receives events.
+     - returns: Disposable object that can be used to unsubscribe the observers.
      */
-    public func bind<O: ObserverType>(to observer: O) -> Disposable where O.E == E? {
-        return self.map { $0 }.subscribe(observer)
+    public func bind<O: ObserverType>(to observers: O...) -> Disposable where O.E == E {
+        return self.bind(to: observers)
     }
-    
+
+    /**
+     Creates new subscription and sends elements to observer(s).
+     In this form, it's equivalent to the `subscribe` method, but it better conveys intent, and enables
+     writing more consistent binding code.
+     - parameter to: Observers to receives events.
+     - returns: Disposable object that can be used to unsubscribe the observers.
+     */
+    public func bind<O: ObserverType>(to observers: O...) -> Disposable where O.E == E? {
+        return self.map { $0 as E? }.bind(to: observers)
+    }
+
+    /**
+     Creates new subscription and sends elements to observer(s).
+     In this form, it's equivalent to the `subscribe` method, but it better conveys intent, and enables
+     writing more consistent binding code.
+     - parameter to: Observers to receives events.
+     - returns: Disposable object that can be used to unsubscribe the observers.
+     */
+    private func bind<O: ObserverType>(to observers: [O]) -> Disposable where O.E == E {
+        return self.subscribe { event in
+            observers.forEach { $0.on(event) }
+        }
+    }
+
     /**
     Subscribes to observable sequence using custom binder function.
-    
+
     - parameter to: Function used to bind elements from `self`.
     - returns: Object representing subscription.
     */
@@ -50,11 +57,11 @@ extension ObservableType {
     /**
     Subscribes to observable sequence using custom binder function and final parameter passed to binder function
     after `self` is passed.
-    
+
         public func bind<R1, R2>(to binder: Self -> R1 -> R2, curriedArgument: R1) -> R2 {
             return binder(self)(curriedArgument)
         }
-    
+
     - parameter to: Function used to bind elements from `self`.
     - parameter curriedArgument: Final argument passed to `binder` to finish binding process.
     - returns: Object representing subscription.
@@ -62,14 +69,13 @@ extension ObservableType {
     public func bind<R1, R2>(to binder: (Self) -> (R1) -> R2, curriedArgument: R1) -> R2 {
          return binder(self)(curriedArgument)
     }
-    
-    
-    /**
-    Subscribes an element handler to an observable sequence. 
 
+
+    /**
+    Subscribes an element handler to an observable sequence.
     In case error occurs in debug mode, `fatalError` will be raised.
     In case error occurs in release mode, `error` will be logged.
-    
+
     - parameter onNext: Action to invoke for each element in the observable sequence.
     - returns: Subscription object used to unsubscribe from the observable sequence.
     */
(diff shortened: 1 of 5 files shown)


pedrommcarrasco:
I'm in for variadic parameters and if the benchmarks prove that the overload impact isn't significant I guess it's fine :)
👍 and ❤️ for this initiative!

kzaher:
Hi @freak4pc ,

I think that adding these extensions to generic observer is dangerous. The name of this thing is `bind(to:`, and it doesn't mention anything about creating additional `share` intermediate step.

This unfortunately can't cover all of the cases properly and creates a lot of confusion.

E.g. If you use `share`, this will fail in case you are binding stateful entity (that emits first value immediately) because other observers won't receive the initial value.

If you  `shareReplay` then it will probably work, but it's weird, and I'm sure there will be some other unwanted side effect.

I think it might sense to have this for `Driver` or `Signal`. 

It might make sense to hold these kinds of changes from single to variadic until next major version since it is public interface breakage.

freak4pc:
> E.g. If you use share, this will fail in case you are binding stateful entity (that emits first value immediately) because other observers won't receive the initial value.

If that is the case, there should be a test case to cover it IMO - My impression was this is fine since the entire test suite passed. 

Anyways I think binding to many has the underlying meaning of using a shared resource. I don't mind having an array-only variation, without variadic, and I think that works well since it fits both scenarios. 

Regarding stateful, yeah you're right this would probably need a `share(replay: 1)` instead.

kzaher:
> If that is the case, there should be a test case to cover it IMO - My impression was this is fine since the entire test suite passed.

The fact that test case passes isn't the evidence that the code works, it only means that there is the lack of evidence that it doesn't work :)

Ok, I've checked out the code again. If you change the implementation to be

```
public func bind<O: ObserverType>(to observers: O...) -> Disposable where O.E == E {
    return self.subscribe { event in
        observers.map { $0(event) }
    }
}
```
and ditch the array version then you don't have any multiple subscription issues, we don't add any additional methods and just improve the existing one.

It still breaks the public interface, so if we want to pull this in, just to be safe, maybe we should wait until RxSwift 5.0.

freak4pc:
@kzaher 

You mean `$0.on(event)`, not `$0(event)` - correct? Also we'll need to add a `_ =` before the map since the result isn't actually used, in which case, perhaps better to use `forEach` ?

freak4pc:
@kzaher Made the changes required, let me know if makes sense for you. 

I had to add array overloads, since Arrays don't support splatting yet (so it wouldn't be able to reuse the Optional vs. non-Optional uses). I made them private since you wanted to avoid Array-style bindings, but TBH I think it would be better to make them public and have both options open. LMK your thoughts on this. 

I also extended this to Relays, but I just noticed - could it be there are no tests for any kind of relays at this point?

kzaher:
> I also extended this to Relays, but I just noticed - could it be there are no tests for any kind of relays at this point?

I believe they are below your code.

freak4pc:
Oh, sorry, duh. Blacked out there :)

RxCocoa/Common/Observable+Bind.swift
```diff
@@ -10,82 +10,140 @@
...
+     Creates new subscription and sends elements to behavior relay(s).
      
      In case error occurs in debug mode, `fatalError` will be raised.
      In case error occurs in release mode, `error` will be logged.
      
      - parameter to: Target behavior relay for sequence elements.
      - returns: Disposable object that can be used to unsubscribe the observer.
      */
-    public func bind(to relay: BehaviorRelay<E>) -> Disposable {
+    public func bind(to relays: BehaviorRelay<E?>...) -> Disposable {
```
kzaher: We are missing tests for binds to behavior and publish relays.

freak4pc:
@kzaher Added the tests, ready for you

freak4pc:
Push passed but PR didn't - not sure if related 🤔 
Thoughts @kzaher ?

RxCocoa/Common/Observable+Bind.swift
```diff
@@ -10,82 +10,140 @@
...
      In case error occurs in debug mode, `fatalError` will be raised.
      In case error occurs in release mode, `error` will be logged.
      
-     - parameter to: Target publish relay for sequence elements.
+     - parameter to: Target publish relays for sequence elements.
      - returns: Disposable object that can be used to unsubscribe the observer.
      */
-    public func bind(to relay: PublishRelay<E>) -> Disposable {
-        return self.subscribe { e in
+    public func bind(to relays: PublishRelay<E?>) -> Disposable {
```
kzaher: This doesn't look right to me. Not variadic.

Tests/RxCocoaTests/Observable+BindTests.swift
```diff
@@ -112,13 +162,39 @@ extension ObservableBindTest {
...
+            ])
+    }
     
     func testBindToOptionalPublishRelay() {
         var events: [Recorded<Event<Int?>>] = []
         
         let relay = PublishRelay<Int?>()
         
-        _ = relay.subscribe{ event in
+        _ = relay.subscribe { event in
```
kzaher: This is not testing the variadic case, is there a need for this test and the non variadic ones?

Tests/RxCocoaTests/Observable+BindTests.swift
```diff
@@ -30,7 +30,32 @@ extension ObservableBindTest {
...
+
+        XCTAssertEqual(events1, [
+            .next(1),
+            .completed()
+        ])
+
+        XCTAssertEqual(events2, [
+            .next(1),
+            .completed()
+        ])
```
kzaher: It seems to me that we should remove all tests with plural, like `testBindToOptionalObserver`. Don't see a need for them.

Tests/RxCocoaTests/Observable+BindTests.swift
```diff
@@ -129,12 +205,38 @@ extension ObservableBindTest {
...
+
+        _ = relay1.subscribe { event in
+            events1.append(Recorded(time: 0, value: event))
+        }
+
+        _ = relay2.subscribe { event in
+            events2.append(Recorded(time: 0, value: event))
+        }
+
+        _ = Observable.just(1).bind(to: relay1, relay2)
```
kzaher: This should have
```
(Observable.just(1) as Observable<Int>)
```

Tests/RxCocoaTests/Observable+BindTests.swift
```diff
@@ -165,6 +278,16 @@ extension ObservableBindTest {
         XCTAssertEqual(relay.value, 1)
     }
 
+    func testBindToOptionalBehaviorRelays() {
+        let relay1 = BehaviorRelay<Int?>(value: 0)
+        let relay2 = BehaviorRelay<Int?>(value: 0)
+
+        _ = Observable.just(1).bind(to: relay1, relay2)
```
kzaher: This should have 
```
(Observable.just(1) as Observable<Int>)
```

Tests/RxCocoaTests/Observable+BindTests.swift
freak4pc: How so ? We need to always validate that doing `bind(to: relay)` and `bind(to: relay1, relay2)` works in our tests.

Tests/RxCocoaTests/Observable+BindTests.swift
freak4pc: As written below, in my though process, yes - 
I believe our tests should always test that the existing interface and the old interface works (e.g. binding to a single one vs. binding to a variadic list).

RxCocoa/Common/Observable+Bind.swift
freak4pc: Yup, my bad :)

freak4pc:
I've addressed your feedback @kzaher. 

I believe we should test the single and variadic tests separately to ensure reliability and backward-compatibility. Let me know if you'd like this to be done otherwise.

freak4pc:
BTW let's merge #1924 first, would make it easier to rebase this.

freak4pc:
As expected this is conflicted. Taking care of this RN.

freak4pc:
Ok, I redid the work on top of the RxRelay work. This should be good to go as well.

freak4pc:
@kzaher CI is green whenever you're ready to give this another review.

PR #1707, opened 2018-07-25
Support iOS 11 for Geolocation example 

tedgonzalez:
Requesting Always authorization requires NSLocationAlwaysAndWhenInUseUsageDescription for iOS 11.
Without this, iOS will not display the authorization popup.


freak4pc:
Thanks @tedgonzalez !

PR #1711, opened 2018-08-01
Add missing doc/comment for `leeway` in scheduler

Cee:
Also using the latest Swift preferred comment format. :)

### DIFF
--- a/RxSwift/Schedulers/ConcurrentDispatchQueueScheduler.swift
+++ b/RxSwift/Schedulers/ConcurrentDispatchQueueScheduler.swift
@@ -23,7 +23,7 @@ public class ConcurrentDispatchQueueScheduler: SchedulerType {
 
     let configuration: DispatchQueueConfiguration
     
-    /// 
+    /// Constructs new `ConcurrentDispatchQueueScheduler` that wraps `queue`.
     ///
     /// - parameter queue: Target dispatch queue.
     /// - parameter leeway: The amount of time, in nanoseconds, that the system will defer the timer.
--- a/RxSwift/Schedulers/ConcurrentDispatchQueueScheduler.swift
+++ b/RxSwift/Schedulers/ConcurrentDispatchQueueScheduler.swift
@@ -23,9 +23,10 @@ public class ConcurrentDispatchQueueScheduler: SchedulerType {
 
     let configuration: DispatchQueueConfiguration
     
-    /// Constructs new `ConcurrentDispatchQueueScheduler` that wraps `queue`.
+    /// 
     ///
     /// - parameter queue: Target dispatch queue.
+    /// - parameter leeway: The amount of time, in nanoseconds, that the system will defer the timer.
     public init(queue: DispatchQueue, leeway: DispatchTimeInterval = DispatchTimeInterval.nanoseconds(0)) {
         configuration = DispatchQueueConfiguration(queue: queue, leeway: leeway)
     }
@@ -33,6 +34,7 @@ public class ConcurrentDispatchQueueScheduler: SchedulerType {
     /// Convenience init for scheduler that wraps one of the global concurrent dispatch queues.
     ///
     /// - parameter qos: Target global dispatch queue, by quality of service class.
+    /// - parameter leeway: The amount of time, in nanoseconds, that the system will defer the timer.
     @available(iOS 8, OSX 10.10, *)
     public convenience init(qos: DispatchQoS, leeway: DispatchTimeInterval = DispatchTimeInterval.nanoseconds(0)) {
         self.init(queue: DispatchQueue(
--- a/RxSwift/Schedulers/SerialDispatchQueueScheduler.swift
+++ b/RxSwift/Schedulers/SerialDispatchQueueScheduler.swift
@@ -38,6 +38,11 @@ public class SerialDispatchQueueScheduler : SchedulerType {
 
     let configuration: DispatchQueueConfiguration
     
+    /// Constructs new `SerialDispatchQueueScheduler` that wraps `queue`.
+    ///
+    /// - Parameters:
+    ///   - serialQueue: Target dispatch queue.
+    ///   - leeway: The amount of time, in nanoseconds, that the system will defer the timer.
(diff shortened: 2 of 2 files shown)


RxSwift/Schedulers/ConcurrentDispatchQueueScheduler.swift
```diff
@@ -16,23 +16,28 @@ import Dispatch
...
 
     let configuration: DispatchQueueConfiguration
-    
+
     /// Constructs new `ConcurrentDispatchQueueScheduler` that wraps `queue`.
     ///
-    /// - parameter queue: Target dispatch queue.
+    /// - Parameters:
+    ///   - queue: Target dispatch queue.
+    ///   - leeway: The amount of time, in nanoseconds, that the system can defer the timer.
```
freak4pc: is it "can" defer, or "will" defer ?

RxSwift/Schedulers/ConcurrentDispatchQueueScheduler.swift
```diff
@@ -44,38 +49,35 @@ public class ConcurrentDispatchQueueScheduler: SchedulerType {
...
     }
 
-    /**
-    Schedules an action to be executed immediately.
-    
-    - parameter state: State passed to the action to be executed.
-    - parameter action: Action to be executed.
-    - returns: The disposable object used to cancel the scheduled action (best effort).
-    */
+    /// Schedules an action to be executed immediately.
```
freak4pc: Any reason you changed all comment blocks from 
```
/**

*/
```
To
``` 
///
///
///
```

Please leave the existing style as that is the style used in all source files in the framework.

RxSwift/Schedulers/SerialDispatchQueueScheduler.swift
```diff
@@ -27,41 +27,46 @@ In case some customization need to be made on it before usage,
 internal serial queue can be customized using `serialQueueConfiguration`
 callback.
 */
-public class SerialDispatchQueueScheduler : SchedulerType {
+public class SerialDispatchQueueScheduler: SchedulerType {
     public typealias TimeInterval = Foundation.TimeInterval
     public typealias Time = Date
-    
-    /// - returns: Current time.
```
freak4pc: Please revert this change. `returns: Something` is the correct syntax Xcode detects.

RxSwift/Schedulers/SerialDispatchQueueScheduler.swift
```diff
@@ -27,41 +27,46 @@ In case some customization need to be made on it before usage,
...
     }
 
     let configuration: DispatchQueueConfiguration
-    
+
+    /// Constructs new `SerialDispatchQueueScheduler` that wraps `queue`.
+    ///
+    /// - Parameters:
+    ///   - serialQueue: Target dispatch queue.
+    ///   - leeway:
```
freak4pc: Missing explanation for leeway

freak4pc:
PS There is no "preferred" documentation format from any official entity such as Apple etc. So I don't see a reason making this change. Both options are valid, but better to keep the style already across the framework.

Cee:
@freak4pc Okay, let's revert that.

Cee:
@freak4pc When you click "Editor -> Structure -> Add Documentation" or press <kbd>Command</kbd>+<kbd>Option</kbd>+<kbd>/</kbd>, Xcode would generate `///` one rather than the old comment `/* ... */` block.

RxSwift/Schedulers/SerialDispatchQueueScheduler.swift
```diff
@@ -38,6 +38,11 @@ public class SerialDispatchQueueScheduler : SchedulerType {
 
     let configuration: DispatchQueueConfiguration
     
+    /// Constructs new `SerialDispatchQueueScheduler` that wraps `queue`.
+    ///
+    /// - Parameters:
+    ///   - serialQueue: Target dispatch queue.
```
freak4pc: This has to be the same format as others: 

- parameter serialQueue: description ...
- parameter leeway: text...

freak4pc:
About the comment-style, it's very interesting. Some files inside the project use different kinds of comment format which is super strange. 

Perhaps @kzaher can have his opinion. It's relatively minor but better to have everything aligned to one style.

Cee:
Comparsion: 
https://github.com/Cee/RxSwift/blob/9daf8e437136c1486b2ed6391bb91e86b8b0840a/RxSwift/Schedulers/ConcurrentDispatchQueueScheduler.swift#L26-L29
and
https://github.com/Cee/RxSwift/blob/9daf8e437136c1486b2ed6391bb91e86b8b0840a/RxSwift/Schedulers/ConcurrentDispatchQueueScheduler.swift#L49-L55

IDK currently we are strictly following one of these formats. I'd like to use Xcode generated one, which is `///`.

RxSwift/Schedulers/SerialDispatchQueueScheduler.swift
```diff
@@ -38,6 +38,12 @@ public class SerialDispatchQueueScheduler : SchedulerType {
 
     let configuration: DispatchQueueConfiguration
     
+    /**
+    Constructs new `SerialDispatchQueueScheduler` that wraps `queue`.
```
freak4pc: One last nit-pick, it should be "that wraps `serialQueue`". I think I'm good beyond that.

Cee: @freak4pc What about https://github.com/Cee/RxSwift/blob/9daf8e437136c1486b2ed6391bb91e86b8b0840a/RxSwift/Schedulers/ConcurrentDispatchQueueScheduler.swift#L26

freak4pc: The argument is `queue` in that link. Here its `serialQueue` that's why I mentioned it.

Cee: oh okay then

freak4pc:
Thanks for your patience @Cee !

Cee:
@freak4pc Thank you for your nice catches!

PR #1713, opened 2018-08-06
Add `string` to `NSTextView`

Cee:
**TL;DR**: Add an Rx `string` property to `NSTextView`.

Mainly did several things to this pr:
- Added an extension to `NSTextView` called `NSTextView+Rx.swift`.
- Wrote a test for `NSTextView+Rx.swift`.
- Made `<->` two-way binding operator exposed to both iOS and macOS.
- Completed the macOS example of synchronizing text between two different `NSTextView`s.

untitled

### DIFF
--- a/RxCocoa/macOS/NSTextView+Rx.swift
+++ b/RxCocoa/macOS/NSTextView+Rx.swift
@@ -1,6 +1,6 @@
 //
 //  NSTextView+Rx.swift
-//  RxCocoa-macOS
+//  RxCocoa
 //
 //  Created by Cee on 8/5/18.
 //  Copyright © 2018 Krunoslav Zaher. All rights reserved.
--- a/RxCocoa/macOS/NSTextView+Rx.swift
+++ b/RxCocoa/macOS/NSTextView+Rx.swift
@@ -66,7 +66,7 @@ extension Reactive where Base: NSTextView {
     }
 
     /// Reactive wrapper for `string` property.
-    public var string: ControlProperty<String?> {
+    public var string: ControlProperty<String> {
         let delegate = RxTextViewDelegateProxy.proxy(for: base)
 
         let source = Observable.deferred { [weak textView = self.base] in
--- a/RxCocoa/macOS/NSTextView+Rx.swift
+++ b/RxCocoa/macOS/NSTextView+Rx.swift
@@ -31,7 +31,7 @@ open class RxTextViewDelegateProxy: DelegateProxy<NSTextView, NSTextViewDelegate
         self.register { RxTextViewDelegateProxy(textView: $0) }
     }
 
-    fileprivate let textSubject = PublishSubject<String?>()
+    fileprivate let textSubject = PublishSubject<String>()
 
     // MARK: Delegate methods
 
@@ -70,11 +70,11 @@ extension Reactive where Base: NSTextView {
         let delegate = RxTextViewDelegateProxy.proxy(for: base)
 
         let source = Observable.deferred { [weak textView = self.base] in
-            delegate.textSubject.startWith(textView?.string)
+            delegate.textSubject.startWith(textView?.string ?? "")
         }.takeUntil(deallocated)
 
-        let observer = Binder(base) { (control, value: String?) in
-            control.string = value ?? ""
+        let observer = Binder(base) { (control, value) in
+            control.string = value
         }
 
         return ControlProperty(values: source, valueSink: observer.asObserver())
--- a/RxCocoa/macOS/NSTextView+Rx.swift
+++ b/RxCocoa/macOS/NSTextView+Rx.swift
@@ -73,7 +73,7 @@ extension Reactive where Base: NSTextView {
             delegate.textSubject.startWith(textView?.string)
         }.takeUntil(deallocated)
 
-        let observer = Binder(base) { (control: Base, value: String?) in
+        let observer = Binder(base) { (control, value: String?) in
             control.string = value ?? ""
         }
 
--- a/RxCocoa/macOS/NSTextView+Rx.swift
+++ b/RxCocoa/macOS/NSTextView+Rx.swift
@@ -1,0 +1,85 @@
+//
+//  NSTextView+Rx.swift
+//  RxCocoa-macOS
+//
+//  Created by Cee on 8/5/18.
+//  Copyright © 2018 Krunoslav Zaher. All rights reserved.
+//
+
+#if os(macOS)
+
+import Cocoa
+import RxSwift
+
+/// Delegate proxy for `NSTextView`.
+///
+/// For more information take a look at `DelegateProxyType`.
+open class RxTextViewDelegateProxy: DelegateProxy<NSTextView, NSTextViewDelegate>, DelegateProxyType, NSTextViewDelegate {
+
+    /// Typed parent object.
+    public weak private(set) var textView: NSTextView?
+
+    /// Initializes `RxTextViewDelegateProxy`
+    ///
+    /// - parameter textView: Parent object for delegate proxy.
+    init(textView: NSTextView) {
+        self.textView = textView
+        super.init(parentObject: textView, delegateProxy: RxTextViewDelegateProxy.self)
+    }
+
+    public static func registerKnownImplementations() {
+        self.register { RxTextViewDelegateProxy(textView: $0) }
+    }
+
+    fileprivate let textSubject = PublishSubject<String?>()
+
+    // MARK: Delegate methods
+
+    open func textDidChange(_ notification: Notification) {
+        let textView: NSTextView = castOrFatalError(notification.object)
+        let nextValue = textView.string
+        self.textSubject.on(.next(nextValue))
+        _forwardToDelegate?.textDidChange?(notification)
+    }
+
+    // MARK: Delegate proxy methods
+
+    /// For more information take a look at `DelegateProxyType`.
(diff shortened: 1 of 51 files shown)


freak4pc:
Thanks for your contribution! 
I think the usual consensus is that we avoid adding new Cocoa Reactive Extensions to core, since otherwise this repo will be very bloated. Though, this specifically mirrors our existing extension on `UITextView` so I feel like it should be OK to add this. 

I'll have to defer to @kzaher on this; might take some time since he's very busy these days.

Tests/RxCocoaTests/NSTextView+RxTests.swift
```diff
@@ -49,7 +49,9 @@ extension NSTextViewTests {
                 name: NSText.didChangeNotification,
                 object: textView,
                 userInfo: nil)
-            textView.delegate?.textDidChange?(notification)
+            if let delegate = textView.delegate {
+                delegate.textDidChange?(notification)
```
freak4pc: Do you really need to be explicit about this? Optional chaining should work 🤔

Cee: Oh I pushed a wrong one... lemme fix this.

Tests/RxCocoaTests/RxTest+Controls.swift
```diff
@@ -62,6 +62,68 @@ extension RxTest {
...
+
+        // this code is here to flush any events that were scheduled to
+        // run on main loop
+        DispatchQueue.main.async {
+            let runLoop = CFRunLoopGetCurrent()
+            CFRunLoopStop(runLoop)
+        }
+        let runLoop = CFRunLoopGetCurrent()
+        CFRunLoopWakeUp(runLoop)
+        CFRunLoopRun()
```
kzaher: Is this code really needed since `wait(for:` is running the RunLoop?

RxCocoa/macOS/NSTextView+Rx.swift
```diff
@@ -0,0 +1,85 @@
...
+    /// Reactive wrapper for `string` property.
+    public var string: ControlProperty<String?> {
+        let delegate = RxTextViewDelegateProxy.proxy(for: base)
+
+        let source = Observable.deferred { [weak textView = self.base] in
+            delegate.textSubject.startWith(textView?.string)
+        }.takeUntil(deallocated)
+
+        let observer = Binder(base) { (control, value: String?) in
+            control.string = value ?? ""
```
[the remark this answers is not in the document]
Cee: While `NSTextView` is a subclass of the `NSText` class, the `string` property requires a nonoptional `String`.

Ref: https://developer.apple.com/documentation/appkit/nstext/1528601-string

Tests/RxCocoaTests/RxTest+Controls.swift
Cee: Good catch! Delete it!

Tests/RxCocoaTests/NSTextView+RxTests.swift
```diff
@@ -0,0 +1,71 @@
...
+        let completeExpectation = XCTestExpectation(description: "NSTextView completion")
+
+        autoreleasepool {
+            let textView = NSTextView()
+            let delegate = TextViewDelegate()
+            textView.delegate = delegate
+            var rxDidChange = false
+
+            _ = textView.rx.string
+                .skip(1) // Initial value
```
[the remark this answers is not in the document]
Cee: We're currently using `XCTAssertFalse(rxDidChange)` to test initial value?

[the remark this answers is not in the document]
Cee: https://github.com/ReactiveX/RxSwift/blob/master/Tests/RxCocoaTests/NSTextField+RxTests.swift#L35

kzaher:
Hi @Cee ,

I think we can just record all of the events and use normal `XCTAssertEqual` extensions to test the recorded events.

```
public func XCTAssertEqual<T: Equatable>(_ lhs: [Event<T>], _ rhs: [Event<T>], file: StaticString = #file, line: UInt = #line) {
```

Cee:
Hi @kzaher 

I removed `XCTAssertFalse` since we are now testing initial value.

Cee:
Hi @kzaher , are you available to take a look?

Tests/RxCocoaTests/RxTest+Controls.swift
```diff
@@ -62,6 +62,57 @@ extension RxTest {
...
+
+
+            _ = (control as NSObject).rx.deallocated.subscribe(onNext: { _ in
+                deallocateExpectation.fulfill()
+            })
+
+            control = nil
+        }
+
+        wait(for: [completeExpectation, deallocateExpectation], timeout: seconds, enforceOrder: true)
```
kzaher: I think that we can modify original implementation of `func ensurePropertyDeallocated` and add `timeout seconds: TimeInterval = 0.0` and the following line to it.

` wait(for: [completeExpectation, deallocateExpectation], timeout: seconds, enforceOrder: true)`

Tests/RxCocoaTests/NSTextView+RxTests.swift
```diff
@@ -0,0 +1,77 @@
...
+import XCTest
+
+final class NSTextViewTests: RxTest {
+    /// NSTextView is much more complicated than other NS-prefix views or controls,
+    /// which takes a longer time to go through the `onCompleted` block. Here, we are
+    /// using `wait(for:timeout:)` to check dealloc/complete operations.
+    static let timeout: TimeInterval = 0.5
+
+    static let notificationForwardTestStrings: [String] = ["T", "Te", "Tes", "Test"]
+    static let notificationForwardTestAssertStrings: [String] = [""] + NSTextViewTests.notificationForwardTestStrings
```
kzaher: This should be moved to `testTextView_TextDidChange_ForwardsToDelegates` and removed `AssertStrings` and `Strings` part from the identifier name since they are redundant.

Tests/RxCocoaTests/NSTextView+RxTests.swift
```diff
@@ -0,0 +1,77 @@
...
+
+        autoreleasepool {
+            let textView = NSTextView()
+            let delegate = TextViewDelegate()
+            textView.delegate = delegate
+            var rxDidChange = false
+
+            var index = 0
+
+            _ = textView.rx.string
```
kzaher: This part should be something like `textView.rx.string.subscribe { recorded.append($0 }`.

Tests/RxCocoaTests/NSTextView+RxTests.swift
```diff
@@ -0,0 +1,77 @@
...
+
+    func testTextView_TextDidChange_ForwardsToDelegates() {
+
+        let completeExpectation = XCTestExpectation(description: "NSTextView completion")
+
+        autoreleasepool {
+            let textView = NSTextView()
+            let delegate = TextViewDelegate()
+            textView.delegate = delegate
+            var rxDidChange = false
```
kzaher: No need for this.

Tests/RxCocoaTests/NSTextView+RxTests.swift
```diff
@@ -0,0 +1,77 @@
...
+
+            var index = 0
```
kzaher: And this.

Tests/RxCocoaTests/NSTextView+RxTests.swift
```diff
@@ -0,0 +1,77 @@
...
+            for testString in NSTextViewTests.notificationForwardTestStrings {
+                textView.string = testString
+                let notification = Notification(
+                    name: NSText.didChangeNotification,
+                    object: textView,
+                    userInfo: nil)
+                (textView.delegate as NSTextDelegate?)?.textDidChange?(notification)
+            }
+
+            XCTAssertTrue(rxDidChange)
```
kzaher: There should be only one assert after ` wait(for: [completeExpectation], timeout: NSTextViewTests.timeout)`. Something like:
```
XCTAssertEqual(recorded, [
   next("T"),
   next("Te"),
   next("Tes"),
    next("Test"),
....
])
```

Tests/RxCocoaTests/NSTextView+RxTests.swift
```diff
@@ -0,0 +1,77 @@
...
+            XCTAssertTrue(delegate.didChange)
```
kzaher: This should be moved after `(textView.delegate as NSTextDelegate?)?.textDidChange?(notification)`. In front of the `for testString in NSTextViewTests.notificationForwardTestStrings {` should be.
`XCTAssertFalse(delegate.didChange)` to assert first call to `(textView.delegate as NSTextDelegate?)?.textDidChange?(notification)` causes the switch.

Tests/RxCocoaTests/RxTest+Controls.swift
Cee: Seems that `wait(for:timeout:enforceOrder:)` function doesn't have `file` and `line` parameters.

kzaher: > Seems that wait(for:timeout:enforceOrder:) function doesn't have file and line parameters.

Maybe I'm overlooking something, why is this important?

kzaher:
I've merged the changes and merged this manually.

https://github.com/ReactiveX/RxSwift/commit/05306ee6760da852c2e5bb44448fed86eb8f8e7b

PR #1715, opened 2018-08-08
Correct method name for Getting Started in Documentation.

vishalwaka:
Correct method name to 'myJust' from 'just'.

### DIFF
--- a/Documentation/GettingStarted.md
+++ b/Documentation/GettingStarted.md
@@ -286,7 +286,7 @@ let cancel = searchForMe
 
 There are a lot of ways to create your own `Observable` sequence. The easiest way is probably to use the `create` function.
 
-Let's write a function that creates a sequence which returns one element upon subscription. That function is called 'just'.
+Let's write a function that creates a sequence which returns one element upon subscription. That function is called 'myJust'.
 
 *This is the actual implementation*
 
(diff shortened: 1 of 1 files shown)


Documentation/GettingStarted.md
```diff
@@ -286,7 +286,7 @@ let cancel = searchForMe
 
 There are a lot of ways to create your own `Observable` sequence. The easiest way is probably to use the `create` function.
 
-Let's write a function that creates a sequence which returns one element upon subscription. That function is called 'just'.
```
kzaher: Oh, now I understand the confusion. I've meant to say here that we already have a standard Rx operator called `just` in the library and that the plan of the demo is to write it from scratch.

Maybe something like `There is already a built-in operator that performs this function named 'just', but let us reimplement it from scratch.`

freak4pc: Wanted to comment on this as well. I don't' think this fix is needed; if you look at the `from` example, it mentions the exact same format. `myJust` is just your own implementation, and it prefixes that by saying "just is the name of that operator [inside RxSwift]". I think it's pretty clear as is, TBH.

kzaher: I'm fine with what ever you decide. Don't have strong feelings about this.

vishalwaka: Yes, @freak4pc. I agree with your suggestion to reframe the sentence this way. Since it informs about the method available in RxSwift and then it asks to provide the implementation for the same.

freak4pc: Sounds good. Want to alter this PR ?

PR #1716, opened 2018-08-11
Add `record` and `bind` to TestScheduler

freak4pc:
The `record` method is part of the `RxExample` project. I (and many others I know) find myself copying this over to every project I work on, mainly since it cuts out a lot of the boilerplate, and highly simplifies the mental load by just “recording” a stream of events. I think it highly fits being part of the core repo.

The `bind` method is from RxTestExt by @mosamer - I think it fits with the same family of test helpers. It’s useful mainly for driving fake inputs to ViewModels or any other kind of Observer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 * Adds `queuePriority` parameter (defaults to `.normal`) to `OperationQueueScheduler`.
 * Performance enhancement reduces Bag dispatch inline code size by 12%.
 * Adds `customCaptureSubscriptionCallstack` hook to allow custom subscription callstacks to be generated.
+* Add `record(source:duration:`) and `bind(_:to:duration:)` to `TestScheduler`.
 
 #### Anomalies
 
--- a/RxTest/Schedulers/TestScheduler.swift
+++ b/RxTest/Schedulers/TestScheduler.swift
@@ -164,6 +164,21 @@ public class TestScheduler : VirtualTimeScheduler<TestSchedulerVirtualTimeConver
 
         return observer
     }
+
+    /**
+     Builds a hot observable with predefined events, binds it's result to given observer and sets up disposal.
+
+     - parameter events: Array of recorded events to emit over the scheduled observable.
+     - parameter target: Observer to bind to.
+    */
+    public func bind<O: ObserverType>(_ events: [Recorded<Event<O.E>>], to target: O, duration: TestTime = 100000) {
+        let driver = self.createHotObservable(events)
+        let disposable = driver.asObservable().subscribe(target)
+
+        self.scheduleAt(duration) {
+            disposable.dispose()
+        }
+    }
 }
 
 
--- a/RxTest/Schedulers/TestScheduler.swift
+++ b/RxTest/Schedulers/TestScheduler.swift
@@ -147,6 +147,23 @@ public class TestScheduler : VirtualTimeScheduler<TestSchedulerVirtualTimeConver
     public func start<Element>(_ create: @escaping () -> Observable<Element>) -> TestableObserver<Element> {
         return start(created: Defaults.created, subscribed: Defaults.subscribed, disposed: Defaults.disposed, create: create)
     }
+
+    /**
+     Builds testable observer for a specific observable sequence, binds it's results and sets up disposal.
+
+     - parameter source: Observable sequence to observe.
+     - returns: Observer that records all events for observable sequence.
+     */
+    func record<O: ObservableConvertibleType>(source: O, duration: TestTime = 100000) -> TestableObserver<O.E> {
+        let observer = self.createObserver(O.E.self)
+        let disposable = source.asObservable().subscribe(observer)
+
+        self.scheduleAt(duration) {
+            disposable.dispose()
+        }
+
+        return observer
+    }
 }
 
 


kzaher:
Hi @freak4pc ,

I've written those examples some time ago when I was still trying to figure out how to best test it.

At that time I've been checking RxJS guys, but as it turns out, I haven't actually used anything like that in concrete apps.

RxJS guys were doing all of these ASCII art sequences. I can understand that might work for some projects.

I've actually always used exactly the same style for testing operators as we are using in tests. It's not fancy and it's extremely explicit.

The downside of all of these convenience extensions is that it adds layer upon layer upon layer of abstractions and conveniences and hides subscription calls. Even current `start` methods that we have might be too much, but what is done is done.

Right now I prefer just the dumb, simple, straightforward format that we use for tests and that other implementations use.

I think that less code we have, less opinionated we are and less code we need to maintain. For the example app I didn't care much about API design since it was just something I was hacking with, but I'm not sure it would be a good idea to add more convenience methods inside the test library itself.

I know that you were talking something about some tutorials, if you are going to do them, you will need to explain what these methods do anyway, so you won't gain a lot by adding them here.

freak4pc:
Hey @kzaher, Thanks for your comments. 

I think there is a room to explain how a specific helper method works if its added, but it doesn't negate the fact that helper methods are ... well, helpful :) 

My thought when it comes to Helper methods is, if there is a common use case that is commonly repeated, there is no real good reason to not reduce it into a single operation. 

For example, If I need to test a single output of my ViewModel, or whichever other variation:

```swift
let results1 = scheduler.createObserver(Int.self)
let disposable1 = viewModel.output.subscribe(results) // or (`bind(to: results)` for clarity).

scheduler.scheduleAt(1000) {
    disposable1.dispose()
}

XCTAssertEqual(result1.events, [...])
```

That's a bit annoying, but fine given the fact I'm only testing a single output. But what happens if I'm testing 5?

```swift
let results1 = scheduler.createObserver(Int.self)
let disposable1 = viewModel.output1.subscribe(results) // or (`bind(to: results)` for clarity).

let results2 = scheduler.createObserver(String.self)
let disposable2 = viewModel.output2.subscribe(results) // or (`bind(to: results)` for clarity).

let results3 = scheduler.createObserver(Double.self)
let disposable3 = viewModel.output3.subscribe(results) // or (`bind(to: results)` for clarity).

let results4 = scheduler.createObserver(String.self)
let disposable4 = viewModel.output4.subscribe(results) // or (`bind(to: results)` for clarity).

let results5 = scheduler.createObserver(Int.self)
let disposable5 = viewModel.output5.subscribe(results) // or (`bind(to: results)` for clarity).

scheduler.scheduleAt(1000) {
    disposable1.dispose()
    disposable2.dispose()
    disposable3.dispose()
    disposable4.dispose()
    disposable5.dispose()
}

scheduler.start()

XCTAssertEqual(results1.events, [...])
XCTAssertEqual(results2.events, [...])
XCTAssertEqual(results3.events, [...])
XCTAssertEqual(results4.events, [...])
XCTAssertEqual(results5.events, [...])
```

This is more of a mess, since I need two lines for every output whose events I want to "record" and inspect later on. I cannot see a downside to abstract these specific two lines to a singular `record` method since it:

1. Makes things cleaner.
2. Conveys intent much better than "subscribe".
3. Encourages writing proper, readable tests.

Here is that same code with the `record` helper, and I don't think there's any question about the fact it is much more readable, and much cleaner to look at.

```swift
let results1 = scheduler.record(source: output1)
let results2 = scheduler.record(source: output2)
let results3 = scheduler.record(source: output3)
let results4 = scheduler.record(source: output4)
let results5 = scheduler.record(source: output5)

scheduler.start()

XCTAssertEqual(results1.events, [...])
XCTAssertEqual(results2.events, [...])
XCTAssertEqual(results3.events, [...])
XCTAssertEqual(results4.events, [...])
XCTAssertEqual(results5.events, [...])
```

You say you prefer the "dumb, straightforward solution", and this one is exactly that (IMO). Not too many people write tests unfortunately, and making the testing methods accessible and discoverable is a huge win. Since `record` is easier to understand conceptually then a simple `subscribe`, even if that's what's happening under the hood.

I don't care much for ASCII marble diagrams, that wasn't the purpose of this :) 

Hope you reconsider, and would love your thoughts. 

Thanks!

freak4pc:
While writing the tutorial you mentioned, I've already found several examples where the record/bind syntax makes things much clearer and "cleaner", in its basis. Just much easier to read. 

compare

freak4pc:
Not sure if it would be trivial (or even possible) to add the same support for Relays since the Unix/SPM targets don't take RxCocoa as a dependency for tests (which I imagine it shouldn't).

kzaher:
Hi @freak4pc ,

I'm not questioning are the methods helpful. I'm arguing that this might not be the right place to put them.

It seems to me that this is more of a MVVM testing extensions thing. The primary purpose of RxTest is to provide minimal and sufficient API for operator testing.

If somebody wants something fancier, then they can just build the fancy logic on top of the basic testing primitives.

I don't feel comfortable adding MVVM related convenience extensions to something called RxTest. RxSwift doesn't even have a concept of `bind`ing.

freak4pc:
Hey:) as usual, thanks for your thorough feedback.

I personally get to disagree but I’m leaving the final call to you of course since this is your baby :) 

I believe that:

1. This is beyond “helpful”, but basically the best way to convey actual meaning of what it does. Recording events is infinitely more understandable and readable compared to manually comparing a TestableObserver of a specific type and subscribing events to it. The record method also saves having to manually define the static type due to type inference.

2. This has nothing to do with MVVM. This is entirely architecture agnostic in my book. It would make it easier to just record any other stream from any other architecture or piece of the app.

2. It’s not entirely correct binding isn’t a thing in RxSwify; we literally have a bind(to:) alias for subscribe, so the terminology is very clear IMO.

This is the best I’ve got 🤓 
If it doesn’t convince you it’s fine, I do understand your stand of this even though I disagree mostly. Adding a small helper to reduce testing boilerplate is a win in my book :) 

Feel free to close otherwise

Thanks !

freak4pc:
Should I close this as of now ?

kzaher:
@freak4pc Hey, yeah, let's close this for now.

PR #1720, opened 2018-08-14
Add `elements` to `TestableObserver`

freak4pc:
This adds `elements` to `TestableObserver` (RxTest) to let you easily assert emitted elements or their count (similar to `assertValues` / `assertValuesCount` in RxJava).


kzaher:
Hi @freak4pc ,

I don't think that when testing events one should use anything else then `events` directly. If we are doing that somewhere, then we need to improve those parts.

E.g. one could test that events have some value, but ignoring `complete` or `error` is a fragile approach IMHO.

freak4pc:
Thanks for the feedback @kzaher :) 

I don't think this is a fragile approach, but it is definitely a matter of taste. When dealing with UI and UI-related streams, non-terminating sequences are very common. 

This wasn't something I was specifically using, but discussion in the #reactive channel in one of the iOS Slacks had some good comments and discussions with @mbrandonw @stephencelis, mentioning that: 

1. This is already part of the RxJava equivalent: http://reactivex.io/RxJava/javadoc/io/reactivex/observers/BaseTestConsumer.html#assertValues-T...-

2. Asserting specific values are being emitted from a stream is a viable option for many; I don't feel like we should be that opinionated about "what conveys good tests vs. non-good tests". As I mentioned, it's quite common to test non-terminating sequences. Even though I wouldn't necessarily test things this way, I found that many people would only test for `.next` events in their sequences quite often, or try solutions such as `RxBlocking`.

3. Speaking of `RxBlocking` - it is unfortunately not a viable alternative, since it requires the sequence to complete to be inspected. So if you just want to inspect the elements/values emitted from the stream, you don't have a choice for doing so. 

I'm not entirely religious to this, but I think it's a fairly common use case that _could_ be useful and doesn't provide a huge surface area code-wise.

kzaher:
Hi @freak4pc ,

I think that this is similar to this PR.

https://github.com/ReactiveX/RxSwift/pull/1575/files

I'm aware of the certain compromises that were made to simplify the usage of this library. Each and every time I've made those compromises it was because the alternative was either extremely more verbose, or there was a lot of uncertainty, or it was just ship it now or spent the next 2 years planning, testing and prototyping.

What I'm trying to say, I'm fine with making compromises when the cost of not making them in some form is high.

Then there are times like this. I can see why people would want a silly shortcut that is not correct, makes their code fragile in the long run, but it kind of looks plausible on code review.

If the primary purpose of this interface is to do testing this way:

```swift
XCTAssertEqual(result.elements, [1, 2, 3])
```

vs 

```swift
XCTAssertRecordedValues(result.events, [1, 2, 3])
```

then yes, I'm against this.

Version 1 is fragile.
Version 2 is not fragile.

Version 1 ignores parts of tested set.
Version 2 performs proper test.

There are various way how to design helpers that create events, so I would rather leave that to the actual users.

If you want to add something like `XCTAssertRecordedValues`, that would probably make more sense IMHO.

freak4pc:
I’m 100% fine with the compromise of making it an XCT* function. Sounds good by me. Let me revise this PR and get your opinion prior. 

Thanks for your feedback :)

kzaher:
If want to add that, then lets name it `XCTAssertRecordedValuesOnly` or something that makes sure that in case there isn't a value in there, it will cause a failure.

freak4pc:
I think only is a bit excessive -
What do you think of keeping the naming familiar and name it XCTAssertRecordedElements ?

kzaher:
If it fails for other events, I guess I can see how the name might fit. In the worst case it will fail and cause a surprise. Ok, lets go with `XCTAssertRecordedElements`.

freak4pc:
Closed in favor of #1726  cc/ @kzaher

PR #1722, opened 2018-08-15
Remove usage of `Variable` from Playground, Example projects and Tests.

freak4pc:
This PR Removes all usages of `Variables` from:

- Playground pages.
- Example projects.
- Test.

It only leaves tests actually pertaining to ensuring `Variable` itself functions properly.


RxExample/RxExample/Operators.swift
```diff
@@ -51,43 +51,42 @@ func <-> <Base>(textInput: TextInput<Base>, variable: Variable<String>) -> Dispo
...
-            if let nonMarkedTextValue = nonMarkedTextValue, nonMarkedTextValue != variable.value {
-                variable.value = nonMarkedTextValue
+            if let nonMarkedTextValue = nonMarkedTextValue, nonMarkedTextValue != relay.value {
+                relay.accept(nonMarkedTextValue)
             }
         }, onCompleted:  {
             bindToUIDisposable.dispose()
         })
 
-    return Disposables.create(bindToUIDisposable, bindToVariable)
```
kzaher: We would probably want to expose `Disposables.create`.

RxExample/RxExample/Operators.swift
```diff
@@ -51,43 +51,42 @@ func <-> <Base>(textInput: TextInput<Base>, variable: Variable<String>) -> Dispo
...
+    let bindToUIDisposable = relay.bind(to: property)
+    let bindToRelay = property
         .subscribe(onNext: { n in
-            variable.value = n
+            relay.accept(n)
         }, onCompleted:  {
             bindToUIDisposable.dispose()
         })
 
-    return Disposables.create(bindToUIDisposable, bindToVariable)
```
kzaher: The same.

freak4pc: Updated both. 

Not sure what's the difference though? Seems `CompositeDisposable(a, b)` == `Disposables.create(a,b)`, no ?

PR #1726, opened 2018-08-22
Add `XCTAssertRecordedElements` to `XCTest+Rx`

freak4pc:
Add `XCTAssertRecordedElements` to `XCTest+Rx` 

Supersedes #1720.

### DIFF
--- a/RxTest/XCTest+Rx.swift
+++ b/RxTest/XCTest+Rx.swift
@@ -170,6 +170,47 @@ public func XCTAssertEqual<T: Equatable>(_ lhs: [Recorded<Event<T?>>], _ rhs: [R
     printSequenceDifferences(lhs, rhs, ==)
 }
 
+/**
+ Assert a list of Recorded events has emitted the provided elements.
+ This method does not take event times into consideration.
+
+ This method will assert a failure if any stop events have been emitted (e.g. `completed` or `error`).
+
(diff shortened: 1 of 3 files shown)


RxTest/XCTest+Rx.swift
```diff
@@ -170,6 +170,47 @@ public func XCTAssertEqual<T: Equatable>(_ lhs: [Recorded<Event<T?>>], _ rhs: [R
...
+    if let stopEvent = stream.first(where: { $0.value.isStopEvent }) {
+        #if os(Linux)
+        XCTFail("A non-next stop event has been emitted: \(stopEvent)")
+        #else
+        XCTFail("A non-next stop event has been emitted: \(stopEvent)", file: file, line: line)
+        #endif
+        return
+    }
+
+    let streamElements = stream.map { event -> T in
```
freak4pc: I started with `stream.map { $0.event.value }` but this potentially returns `nil`. 
I thought of doing a `compactMap`, but if the event itself would be `nil` that would cause issues IMO.

freak4pc:
Here are the use cases. I thought of actually adding these tests but I'm not sure if you want tests for test helpers: 

screen shot 2018-08-22 at 10 52 09 am

freak4pc:
Anything prevents merging this, given the addition was approved in a separate issued discussion

PR #1728, opened 2018-08-24
Update UITableView+Rx.swift comment

cozzin:
Edit comment because configureCell was changed to required init
https://github.com/RxSwiftCommunity/RxDataSources/issues/253

### DIFF
--- a/RxCocoa/iOS/UITableView+Rx.swift
+++ b/RxCocoa/iOS/UITableView+Rx.swift
@@ -110,7 +110,12 @@ extension Reactive where Base: UITableView {
      
      Example 
 
-        let dataSource = RxTableViewSectionedReloadDataSource<SectionModel<String, Double>>()
+        let dataSource = RxTableViewSectionedReloadDataSource<SectionModel<String, Double>>(
+            configureCell: { dataSource, tableView, indexPath, element in
+                let cell = tv.dequeueReusableCell(withIdentifier: "Cell")!
+                cell.textLabel?.text = "\(element) @ row \(indexPath.row)"
+                return cell
+        })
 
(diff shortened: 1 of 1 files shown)


RxCocoa/iOS/UITableView+Rx.swift
```diff
@@ -110,7 +110,12 @@ extension Reactive where Base: UITableView {
      
      Example 
 
-        let dataSource = RxTableViewSectionedReloadDataSource<SectionModel<String, Double>>()
+        let dataSource = RxTableViewSectionedReloadDataSource<SectionModel<String, Double>>(
```
freak4pc: Does `RxTableViewSectionedReloadDataSource` actually exist in this scenario? This isn't RxDataSources.

cozzin: @freak4pc It contained a comment for RxTableViewSectionedReloadDataSource. However, to ensure the purity of RxCocoa, it seems appropriate to remove the part of RxDataSources.

freak4pc: I think we should entirely remove everything related to RxDataSources here - the code isn't part of this framework.

kzaher: Let's remove this example completely then.

kzaher: Somebody wants to remove this then?

freak4pc: I'll take care of it, 2 minutes.

freak4pc:
@kzaher Let me know if that looks OK.

PR #1729, opened 2018-08-26
Conditionally conforming Recorded, Event, SingleEvent, MaybeEvent to equatable.

LucianoPAlmeida:
Only changes to take advantage of conditional conformance to remove some == operators of [Event+Equatable](https://github.com/ReactiveX/RxSwift/blob/master/RxTest/Event%2BEquatable.swift) and [Recorded](https://github.com/ReactiveX/RxSwift/blob/master/RxTest/Recorded.swift) on RXTest.


RxTest/Event+Equatable.swift
```diff
@@ -13,111 +13,97 @@ import class Foundation.NSError
...
+        switch (lhs, rhs) {
+        case (.error(let e1), .error(let e2)):
+            #if os(Linux)
             return  "\(e1)" == "\(e2)"
-        #else
+            #else
             let error1 = e1 as NSError
             let error2 = e2 as NSError
-
+
```
kzaher: Can you please remove these spaces and set Xcode to trim trailing whitespace.

RxTest/Event+Equatable.swift
```diff
@@ -13,111 +13,97 @@ import class Foundation.NSError
...
+        case (.completed, .completed): return true
+        case (.error(let e1), .error(let e2)):
+            #if os(Linux)
             return  "\(e1)" == "\(e2)"
-        #else
+            #else
             let error1 = e1 as NSError
             let error2 = e2 as NSError
-
+
```
kzaher: Can you please remove these spaces and set Xcode to trim trailing whitespace.

RxTest/Event+Equatable.swift
```diff
@@ -13,111 +13,97 @@ import class Foundation.NSError
...
```
kzaher: Can you please remove these spaces and set Xcode to trim trailing whitespace.

LucianoPAlmeida:
Hi @kzaher :)
Does it also need a changelog entry?

kzaher:
It would be nice.

CHANGELOG.md
```diff
@@ -13,6 +13,8 @@ All notable changes to this project will be documented in this file.
 
 ## [4.X.X](https://github.com/ReactiveX/RxSwift/releases/tag/4.X.X)
 
+* Xcode 9.3 / Swift 4.1 compatibility
```
freak4pc: I would change this line to "Increase minimum requirement to Swift 4.1". (and we should also make sure that's updated in the README)

LucianoPAlmeida:
About the minimum requirement, is possible to maintain it on the 4.0 by using 
`#if swift(>=4.1)` adding this new implementations under this condition but also maintaining the old implementation on a else clause. 
Maybe it can be an option because it not requires people to upgrade to 4.1, but people who does can take advantage of this equatable conformance :)) 
What you think? @kzaher @freak4pc

freak4pc:
I feel comfortable with just support 4.1 and up given 4.2 was already released and 5 is around the corner ...

freak4pc:
My thought was we want to support Xcode 9.3.1 and Xcode 10, which means Swift 4.1 and Swift 4.2 - both support conditional conformance, and older users can simply target an older version. I don't think it's fair to expect of us to maintain the newest version supporting Xcode < 9.3 or Xcode 8. 

We can definitely add support for both but I think it's a bit moot.

LucianoPAlmeida:
Hey @freak4pc and @kzaher :))

@freak4pc I agree 👍 
But I think conditional checks are a fair option, mostly because It's not a big change since most of the code are reusable and I believe it's good enough since users who use >=4.1 can take advantage of this feature :)
 
It's a very useful feature to me for example, to pass recorded events to a Nimble equal that takes an equatable parameter like 

```swift 
let correctValues: [Recorded<Event<Bool>>] = [
      next(0, true)
]
expect(observer.events).to(equal(correctValues))
```

PR #1746, opened 2018-09-20
Update Rx.xcworkspace

kylinroc:
- Add all documentations
- Remove duplicated playground ref


kzaher:
Hi @kylinroc ,

where is the second reference? I can't find it.

Issue #1552, opened 2018-01-22
Binary operator '==' cannot be applied to two '[Recorded<Event<Void>>]' operands


Arcovv:
**Short description of the issue**:

Binary operator '==' cannot be applied to two '[Recorded<Event<Void>>]' operands. `Void` isn't implement Equatable.

**What actually happens**:

Hi,

Recently I tried to use RxTest with ViewModel architecture in my project. But I got some wrong from Xcode, that "Binary operator '==' cannot be applied to two '[Recorded<Event<Void>>]' operands" error log.

There is my MVVM code:

```swift
protocol ProfileViewModelInputs {
  func didTapSettingButton()
}

protocol ProfileViewModelOutputs {
  var showSetting: Signal<Void> { get }
}

final class ProfileViewModel {

  let showSetting: Signal<Void>

  init() {
    showSetting = tappingSettingButtonSubject.asSignal()
  }

  fileprivate let tappingSettingButtonSubject = PublishRelay<Void>()
  func didTapSettingButton() {
    tappingSettingButtonSubject.accept(())
  }
}
```

And my test code:

```swift
class ProfileViewModelTests: XCTestCase {
  
  private var viewModel: ProfileViewModel!
  private var disposeBag: DisposeBag!
  private var schedule: TestScheduler!
  
  override func setUp() {
    super.setUp()
    viewModel = ProfileViewModel()
    disposeBag = DisposeBag()
    schedule = TestScheduler(initialClock: 0)
  }
  
  override func tearDown() {
    super.tearDown()
    viewModel = nil
    disposeBag = nil
    schedule = nil
  }
  
  func test_tapping_setting_button() {
    let tapObserver = schedule.createObserver(Void.self)
    viewModel.outputs.showSetting
      .debug(#function)
      .emit(to: tapObserver)
      .disposed(by: disposeBag)
    schedule.start()
    
    viewModel.inputs.didTapSettingButton()
    
    let expectedEvents = [
      next(0, ())
    ]
    let tapEvents = tapObserver.events

    XCTAssert("\(tapObserver.events)" == "\(expectedEvents)")
  }
}
```

In this code, I can't use `tapObserver.events == expectedEvents` because Void is not implemented Equatable. Only I can transform it to string. How to solve this problem?

Thanks for you help.

kzaher:
Hi,

this doesn't look like an issue with the project itself but more of a "how do I do XXX" question. 
[...]

PR #1747, opened 2018-09-21
Add XCTAssertEqual for [Recorded<Event<Void>>]

kouki-dan:
The observer of Void is not testable on RxSwift.
This is a test case for this.
```
        let observer = scheduler.createObserver(Void.self)

        scheduler.scheduleAt(100, action: {
            observer.onNext(())
        })

        scheduler.start()

        XCTAssertEqual(observer.events, [
            .next(100, ())
        ])
```

This code fails to compile on latest RxSwift code because XCTAssertEqual is not defined for `Recorded<Event<Void>>>`.

This PR adds XCTAssertEqual method and equatable for related them.

This PR also may fix the issue's problem: https://github.com/ReactiveX/RxSwift/issues/1552

### DIFF
--- a/RxTest/Event+Equatable.swift
+++ b/RxTest/Event+Equatable.swift
@@ -64,14 +64,14 @@ public func == (lhs: Event<Void>, rhs: Event<Void>) -> Bool {
     case (.completed, .completed): return true
     case (.error(let e1), .error(let e2)):
         #if os(Linux)
-        return  "\(e1)" == "\(e2)"
+            return  "\(e1)" == "\(e2)"
         #else
-        let error1 = e1 as NSError
-        let error2 = e2 as NSError
+            let error1 = e1 as NSError
+            let error2 = e2 as NSError
 
-        return error1.domain == error2.domain
-            && error1.code == error2.code
-            && "\(e1)" == "\(e2)"
+            return error1.domain == error2.domain
+                && error1.code == error2.code
+                && "\(e1)" == "\(e2)"
         #endif
     case (.next, .next): return true
     default: return false
--- a/RxTest/Event+Equatable.swift
+++ b/RxTest/Event+Equatable.swift
@@ -55,6 +55,30 @@ public func == <Element: Equatable>(lhs: Event<Element?>, rhs: Event<Element?>) 
     }
 }
 
+/// Compares two events with Void elements. They are equal always.
+///
+/// In case `Error` events are being compared, they are equal in case their `NSError` representations are equal (domain and code)
+/// and their string representations are equal.
+public func == (lhs: Event<Void>, rhs: Event<Void>) -> Bool {
+    switch (lhs, rhs) {
+    case (.completed, .completed): return true
+    case (.error(let e1), .error(let e2)):
+        #if os(Linux)
+        return  "\(e1)" == "\(e2)"
+        #else
+        let error1 = e1 as NSError
+        let error2 = e2 as NSError
+
+        return error1.domain == error2.domain
+            && error1.code == error2.code
+            && "\(e1)" == "\(e2)"
+        #endif
+    case (.next, .next): return true
+    default: return false
+    }
+}
+
+
 /// Compares two events. They are equal if they are both the same member of `Event` enumeration.
 ///
 /// In case `Error` events are being compared, they are equal in case their `NSError` representations are equal (domain and code)
--- a/RxTest/Recorded+Event.swift
+++ b/RxTest/Recorded+Event.swift
@@ -108,3 +108,14 @@ extension Recorded {
     }
 }
 
+extension Recorded where Value == Void {
+    /**
+     Factory method for a `.next` event recorded at a given time.
+
+     - parameter time: Recorded virtual time the `.next` event occurs.
+     - returns: Recorded event in time.
+     */
+    public static func next(_ time: TestTime) -> Recorded<Event<Void>> {
+        return Recorded<Event<Void>>(time: time, value: .next(()))
+    }
+}
--- a/RxTest/Recorded.swift
+++ b/RxTest/Recorded.swift
@@ -43,3 +43,7 @@ public func == <T: Equatable>(lhs: Recorded<Event<T>>, rhs: Recorded<Event<T>>) 
 public func == <T: Equatable>(lhs: Recorded<Event<T?>>, rhs: Recorded<Event<T?>>) -> Bool {
     return lhs.time == rhs.time && lhs.value == rhs.value
 }
+
+public func == (lhs: Recorded<Event<Void>>, rhs: Recorded<Event<Void>>) -> Bool {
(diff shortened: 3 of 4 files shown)


freak4pc:
If we’re doing a specific overload for void, might as well make it time only, since the value is irrelevant and constant

freak4pc:
Pulling your branch, I'm still getting issues. Did you test this with real code?

freak4pc:
I added an overload to accept `time` only for Void events, I think this would prove very useful so you vould just do

```swift
XCTAssertEqual(voidStream.events, [
    .next(5),
    .next(20),
    .next(55),
    .next(70)
])
```

What do you think ?

kouki-dan:
I also think this factory method for Void events is very useful!

kouki-dan:
Thank you for your review and fantastic idea!

freak4pc:
@kzaher Have a quick minute for this? pretty quick change, but remedies an issue some users have. What do you think?

kzaher:
Hi guys,

I can understand why this is annoying. I'm also annoyed by it. But as far as I can tell this is Swift compiler bug/inadequacy.
In case this is fixed in the compiler layer and it is possible to make `()` conform to `Equatable` or they make `()` conform to `Equatable` directly we will have ambiguity errors.

I would prefer if we stayed away from trying to introduce workarounds for language/compiler inadequacies.

```swift
 1> () == ()

$R0: Bool = true
  2>  
  3> func a<T: Equatable>(l: T, r: T) -> Bool { return l == r }
  4> a(l: (), r: ())
error: repl.swift:4:13: error: argument type '()' does not conform to expected type 'Equatable'
a(l: (), r: ())
```

If this was some serious issue, then sure, we would pull it in because hell, world isn't perfect. But it seems to me that this is just an annoyance.

If somebody needs to assert a list of voids, just transform it to one of your custom `Empty: Equatable` types in the unit tests and temporarily test these. It should be equivalend. This is what the compiler should do anyway.

freak4pc:
@kzaher I disagree with this assumption. 

We can only program for the language we have, not for the language we _might_ have. There is no intention from the Swift core team to Fix this, it seems to me at least. And even if that happens, we can always have it as part of a new release, like we always do. 

The problem is substantial, as right now the only way to assert equality of `Recorded<Event<Void>>` is to map it to something else (like `voidOutput.map { "x" }.bind(to: stringTestableObserver)`), since otherwise there would be no possible way to assert the validity of these events. 

I urge you to reconsider this as its a huge pain in the butt for asserting anything void (which is substantial, as mentioned).

kzaher:
Hi @freak4pc ,

what evidence do you have that needing to add one additional `map` is a substantial problem? It seems to me that this library exists for 4 years, so if this was a substantial issue, why is this PR made now and not sooner?

I would say that we need to be consistent and only provide `Equality` extensions. We can't provide tuple specific overloads for every single API where we have `Equatable`. This is not a realistic expectation.

If we accept this PR, how are you going to decide should we also add overload for `(E) where E: Equatable`, or `(E1, E2) where E1: Equatable, E2: Equatable`? At what number should we stop?

Should we also add these overloads for `distinctUntilChanged` operator?

And even if we did all that, that still wouldn't solve any of my problems since there is myriad of other APIs I have that accept `Equatable`, but that don't work with tuples.

> I urge you to reconsider this as its a huge pain in the butt for asserting anything void (which is substantial, as mentioned).

It is. I'm looking at this every day. My code is littered with structs that are only there because tuples aren't `Equatable`.

freak4pc:
I don't think the fact users were able to hack around this for 4 years makes this a non-considerable change. The fact is - Void events are not going anywhere, and they are very very common. Testing against those is also very very common. 

Tuples are out of the question for this specific case IMO - since there are so many variations on tuples, but Voids are just a singular thing. 

`distinctUntilChanged` will cause a stream to be stuck since `Void` cannot change - again this isn't the best example as Void isn't a value that has "transitions", it is a singular static value that is basically just a "Signal" something has happened, so only the time piece is relevant to a stream of Void. Which is why, IMO it deserves this special consideration (even if it was `Equatable`, the overload asserting only times is helpful since values have no meaning in this case).

freak4pc:
Any advanced thoughts of this or should we close? 
I still bump into this a lot, we've moved these extensions to many projects but think it is a really nice and senseful piece of code to have in the core project.

kzaher:
@freak4pc I have some different ideas how to tackle this. There is a chance that we'll write a swift evolution proposal that would indirectly solve this issue on the compiler level, where it should be solved. Last I've heard was that there was some support in the compiler community, but I think that we are still waiting for approval.

I would really want to solve this in a more correct way.

freak4pc:
@kzaher I actually thank that "API-wise" this is the much cleaner path. 

Even if in the compiler you'd make `Void` conform to Observable, you would still need to manually repeat a value that never changes. 

e.g. 

```
.next(10, ())
.next(20, ())
.next(25, ())
.completed
```

This API, even though is more manual, provides us with a much cleaner usage, in this specific case where value has no meaning:

```
.next(10)
.next(20)
.next(25)
.completed
```

I still think that the road to fix this in the compiler is very _very_ far, so we should have a normal solution in the meantime that doesn't rely on mapping to a string just to perform a simple test. That's just my personal opinion :)

freak4pc:
Hey @kzaher - would appreciate if you would give this another thought.
Especially my last comment, even if Void would be Equatable, it would still not provide an "easy" syntax to only provide team, which provides a lot of boilerplate (e.g. repeating `()` even though it's a singular entity)

YiYiZheng:
Does this pull request merge into any branch?

YiYiZheng:
@freak4pc 
As the pull request has not merge, how can I fix this problem now?

freak4pc:
Manually mapping void streams to some string and testing against that

kouki-dan:
Thank you for your discussion and sorry for absent. I understood you are saying.

When I'm facing this problem, I thought why Tuple does not conform Equatable. If this is resolved by compiler, I think that is better.

@YiYiZheng I always map to true of Bool in my testing file when I want to test Void stream. Like this.
```
let observer = scheduler.createObserver(Bool.self)
anyVoidOutputs
    .map { true }
    .bind(to: observer)
XCTAssertEqual(observer.events, [
    .next(time, true),
])
```

Any other ideas?
I thought using own enum instead of Void which has only one case and the enum conforms Equatable may fix this issues. But I'm not sure that is the right way to use Rx, so I always use map to true in my testing.
I mean define own enum like this and use it instead of Void in my code(not test code).
```
enum Event: Equatable {
    case publish
}
```

lordzsolt:
Dear diary,

It is March, 2021. This PR was opened 2.5 years ago, but it's still as relevant today.
Nothing has changed.
I am sad.

freak4pc:
Dear diary, 
Apple apparently still doesn't consider Void equal to Void or equatable at all for that matter :)

PR #1748, opened 2018-09-21
Remove compiler conditionals that support swift 3.0

carlosypunto:
Diven that RxSwift do not support swift 3.x version, that repetition of code is unnecessary. Removing benefits reading.


carlosypunto:
No CHANGELOG changes made because, I think, it is indifferent for the final user.

PR #1749, opened 2018-09-22
Add @discardableResult for Resources

kylinroc:
Indent for conditional compilation block in project seems inconsistent, so I chosen the Xcode default way (`Control + I`).


kylinroc:
I think this change not affect users, the old `_ =` way still works. Not sure should I change the CHANGLOG?

Issue #1752, opened 2018-09-25
Update WATCHOS_DEPLOYMENT_TARGET to 3.0 for Submission to Apple


dlewanda:
**Short description of the issue**:
With the final release of Xcode 10 and watchOS 5, Apple is rejecting submission of watchOS applications with a `WATCHOS_DEPLOYMENT_TARGET` below **3.0**

**Expected outcome**:

Apple accepts the application via upload through the normal means.

**What actually happens**:

The application upload is rejected with the following message:
```
[02:22:23]: [iTMSTransporter] [2018-09-25 02:22:23 MDT] <main> ERROR: ERROR ITMS-90081: "This bundle is invalid. Applications built for more than one architecture require an iOS Deployment Target of 3.0 or later."
```
While the error message suggests the iOS Deployment target is wrong, it appears this is a misnomer, as our iOS deployment target is 9.0. Others are seeing this issue as well:
https://forums.developer.apple.com/thread/108853

We did have to update the watchOS deployment target from 2.0 (the default) on other frameworks we use. We had been relying on the precompiled version of RxSwift to expedite our build times, but it appears that the current minimum version compiled in is 2.0 as shown here:
https://github.com/ReactiveX/RxSwift/blob/master/Rx.xcodeproj/project.pbxproj#L6428

**Self contained code example that reproduces the issue**:

_The issue isn't specific to any code, but how it's built via the project setting linked above_

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**
https://github.com/ReactiveX/RxSwift/releases/tag/4.3.1

**Platform/Environment**

  - [ ] iOS
  - [ ] macOS
  - [ ] tvOS
  - [x] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  10.0 (10A255)
```

:warning: Fields below are optional for general issues or in case those questions aren't related to your issue, but filling them out will increase the chances of getting your issue resolved. :warning:

**Installation method**:
  - [ ] CocoaPods
  - [x] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
[...]

PR #1753, opened 2018-09-25
Update watchOS deployment target for submission to Apple

dlewanda:
It appears that, while unpublished, Apple is requiring all watchOS deployment targets to be updated to 3.0

https://forums.developer.apple.com/thread/108853 for reference

Addresses https://github.com/ReactiveX/RxSwift/issues/1752

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -6425,7 +6425,7 @@
 				VALIDATE_PRODUCT = YES;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
-				WATCHOS_DEPLOYMENT_TARGET = 2.0;
+				WATCHOS_DEPLOYMENT_TARGET = 3.0;
 			};
 			name = "Release-Tests";
 		};
@@ -6841,7 +6841,7 @@
 				TVOS_DEPLOYMENT_TARGET = 9.0;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
-				WATCHOS_DEPLOYMENT_TARGET = 2.0;
+				WATCHOS_DEPLOYMENT_TARGET = 3.0;
 			};
 			name = Debug;
 		};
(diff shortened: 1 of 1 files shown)


dlewanda:
My apologies here for the premature posting, I didn't see the warnings turn into errors locally. I'm still working through the necessary changes to get this to pass, but if anyone else has input here, I'm all ears!

kzaher:
Hi @dlewanda ,

we've reimplemented those atomic operations for Linux, but as you can see, it's not the most performant solution out there :(

We are still waiting for `stdatomic` extensions to become available. But I guess they do work.

dlewanda:
Hi @kzaher,

Thanks for your response. Given that the deprecation message from Apple is only a warning, would you consider allowing those warnings in for at least this release until there's a better option? I don't believe they're entirely broken, and seems like the least disruptive option given Apple's requirement of WatchOS 3.0 as a minimum deployment target as of a couple weeks ago.

dlewanda:
Hi @freak4pc - Any chance you've had an opportunity to look at the situation here? I'd love to get your thoughts on how we can update to watchOS 3.0 as a deployment target.

kzaher:
Does anyone have some new information on this? 

Is there a way for us to implement this warning?
```
'OSAtomicXor32OrigBarrier' was deprecated in watchOS 3.0: Use atomic_fetch_xor() from <stdatomic.h> instead
```

Last time I've checked this was broken and I couldn't update the atomic operations.

dlewanda:
I'm trying to figure out how to avoid these warnings, and there don't appear to be any [project settings](https://stackoverflow.com/a/32861678) that can do it. When I try to edit the file to insert any solution specific to that file, I get the following error:
Xcode Error

Any suggestions as to how I can edit that file?

freak4pc:
@dlewanda, I'm closing this in favor of #1772. I came back to this with a clean head and think I figured it out, I also took your commit there.

kzaher:
@dlewanda Using any editor outside of Xcode. Vim, Visual Studio Code, Atom .... You name it.

freak4pc:
Yup - Xcode is crappy with symlinks.

PR #1760, opened 2018-10-06
Add CombineLatest operator to Single trait

yuzushioh:
Added `combineLatest` operator to Single trait

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -307,6 +307,7 @@ custom_categories:
   - Completable
   - Maybe
   - ObservableType+PrimitiveSequence
+  - PrimitiveSequence+CombineLatest
   - PrimitiveSequence+Zip+arity
   - PrimitiveSequence
   - Single
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 ---
 ## Master
 
+* Add `combineLatest` operator to Single trait
+
 ## [4.3.1](https://github.com/ReactiveX/RxSwift/releases/tag/4.3.1)
 
 #### Anomalies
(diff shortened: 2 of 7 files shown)


kzaher:
Hi @yuzushioh,

Can you please explain how is this different then the zip operator?

yuzushioh:
Hi @kzaher 

some view model in my project fetches three different elements via api. It requires all of 3 elements at first call but it needs UI to be updated if any of them emits another element after.  If using Zip operator, all of three observables needs to emit elements to update UI. I believe this can not be implemented by zip or merge operator unless to do something tricky by using merge operator and Enums.

Something like this for example but more complex in my project
```swift
let someTrigger1 = PublishSubject<Void>()
let someTrigger2 = PublishSubject<Void>()
let someTrigger3 = PublishSubject<Void>()

let firstElement = someTrigger1.asSingle()
    .flatMap { repository.getUser() // Single<User> }

let secondElement = someTrigger2.asSingle()
    .flatMap { repository.getItem() // Single<Item> }

let thirdElement = someTrigger3.asSingle()
    .flatMap { repository.getStatus() // Single<Status> }

let dataSource = Single.combineLatest(firstElement, secondElement, thirdElement) {
     return DataSource($0, $1, $2)
}
```

ifabijanovic:
It seems like you are using the `Single` sequence incorrectly here.

http://reactivex.io/documentation/single.html
`A Single is something like an Observable, but instead of emitting a series of values — anywhere from none at all to an infinite number — it always either emits one value or an error notification.`

In your example if any of the `PublishSubjects` emitted a second value you would get an error: `Sequence contains more than one element`. It seems you're using `Single` just because your repository API returns a `Single`. Instead you should do something like this:

```
let firstElement = someTrigger1
  .flatMap(Latest) { repository.getUser().asObservable() }
...
...
let dataSource = Observable.combineLatest(firstElement, ...) { ... }
```

PR #1765, opened 2018-10-12
Adding RxCocoa reactive extension badgeColor on UITabBarItem 

LucianoPAlmeida:
Adding RxCocoa reactive extension badgeColor on UITabBarItem  available for iOS 10+ and tvOS 10+.

### DIFF
--- a/Tests/RxCocoaTests/UITabBarItem+RxTests.swift
+++ b/Tests/RxCocoaTests/UITabBarItem+RxTests.swift
@@ -15,17 +15,36 @@ final class UITabBarItemTests : RxTest {
 }
 
 extension UITabBarItemTests {
-    func testBadgetValue_Text() {
+    func testBadgeValue_Text() {
         let subject = UITabBarItem(tabBarSystemItem: .more, tag: 0)
         Observable.just("5").subscribe(subject.rx.badgeValue).dispose()
         
         XCTAssertTrue(subject.badgeValue == "5")
     }
     
-    func testBadgetValue_Empty() {
+    func testBadgeValue_Empty() {
         let subject = UITabBarItem(tabBarSystemItem: .more, tag: 0)
         Observable.just(nil).subscribe(subject.rx.badgeValue).dispose()
         
-        XCTAssertTrue(subject.badgeValue == nil)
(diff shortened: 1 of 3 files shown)


freak4pc:
Hey @LucianoPAlmeida ! Thanks for the addition and willingness to help. 🙏 

In general, if you'll look at other similar PRs - a decision was made to not add additional RxCocoa extensions that are trivial to create (e.g. simple Binders) because it creates a heavier code base that needs more maintenance and provides value only to a small portion of users. 

When the addition is not trivial (e.g. some fancy DelegateProxy of a popular delegate that's hard to get right), it might be worth adding it to RxCocoa core, but otherwise - these simpler ones might be better suited as a PR to RxCocoa. 

Let me know what you think :)
Shai.

LucianoPAlmeida:
Hey @freak4pc :)
Is very common to create specific custom binders, this one thought could be useful for other users. 
Thank you for the feedback and I totally understand the point 👍 
Feel free to close this one :))

LucianoPAlmeida:
> When the addition is not trivial (e.g. some fancy DelegateProxy of a popular delegate that's hard to get right), it might be worth adding it to RxCocoa core, but otherwise - these simpler ones might be better suited as a PR to RxCocoa.

Sorry to ask but I'm a bit confused 🙃. RxCocoa Core is different from RxCocoa? 😆

freak4pc:
@LucianoPAlmeida - Sorry! I was distracted, I meant RxSwiftExt could be a good place for it :) 
Thanks for your understanding!

Tests/RxCocoaTests/UITabBarItem+RxTests.swift
```diff
@@ -15,17 +15,36 @@ final class UITabBarItemTests : RxTest {
 }
 
 extension UITabBarItemTests {
-    func testBadgetValue_Text() {
```
freak4pc: If you have a second to make a PR just for this fix that would be awesome - typos in code are a bummer 😂

LucianoPAlmeida: For sure 😄

PR #1766, opened 2018-10-12
Update deprecated code examples in UnitTests.md

pontus-andersson:
The `next()` global function is planned for future deprecation. Now using `Recorded.next()` instead.

### DIFF
--- a/Documentation/UnitTests.md
+++ b/Documentation/UnitTests.md
@@ -9,50 +9,50 @@ This is an example of a typical `RxSwift` operator unit test:
 
 ```swift
 func testMap_Range() {
-        // Initializes test scheduler.
-        // Test scheduler implements virtual time that is
-        // detached from local machine clock.
-        // This enables running the simulation as fast as possible
-        // and proving that all events have been handled.
-        let scheduler = TestScheduler(initialClock: 0)
-
-        // Creates a mock hot observable sequence.
-        // The sequence will emit events at designated
-        // times, no matter if there are observers subscribed or not.
-        // (that's what hot means).
-        // This observable sequence will also record all subscriptions
-        // made during its lifetime (`subscriptions` property).
-        let xs = scheduler.createHotObservable([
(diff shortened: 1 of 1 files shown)


pontus-andersson:
@freak4pc Yes, that's much better! Thanks for pointing it out. Give me a minute to fix it.

pontus-andersson:
@freak4pc Right (I'm still a learner of RxSwift. That's why I read these docs from the start).

I'm running the code in my unit tests and they pass (even without the `completed(...)` call though).

Documentation/UnitTests.md
```diff
@@ -9,50 +9,51 @@ This is an example of a typical `RxSwift` operator unit test:
...
+    // This observable sequence will also record all subscriptions
+    // made during its lifetime (`subscriptions` property).
+    let xs = scheduler.createHotObservable(
+        Recorded.events(
+            .next(150, 1),  // first argument is virtual time, second argument is element value
+            .next(210, 0),
+            .next(220, 1),
+            .next(230, 2),
+            .next(240, 4),
             completed(300) // virtual time when completed is sent
```
freak4pc: this needs to have the `.` before. The reason it works without is that the `completed` one refers to the global free function and not the case. e.g.: 

```swift
Recorded.events(
            .next(150, 1),  // first argument is virtual time, second argument is element value
            .next(210, 0),
            .next(220, 1),
            .next(230, 2),
            .next(240, 4),
            .completed(300) // virtual time when completed is sent
           )
```

Same for the second example.

freak4pc:
Yup, I didn't say otherwise :)

pontus-andersson:
@freak4pc so much work for so little documentation. Well, I guess it was just sloppy from my end. Thanks for approving!

freak4pc:
Ha, really don't worry about it :) It's a good exercise, your next PR will definitely require less fiddling. Always keep learning :]

PR #1767, opened 2018-10-12
Extension to sequence of Disposable elements

pontus-andersson:
Convenient extension to sequence used to add several Disposable elements to a dispose bag.


pontus-andersson:
@freak4pc did not know. Now I know. Thanks!

PR #1769, opened 2018-10-15
Add `PublishRelay.asDriver()`

dmcyk:
This PR adds `PublishRelay.asDriver()` method. 
Seems like `PublishRelay` is the only trait that's missing such conversion. All `BehaviorRelay`, `ControlEvent` and `ControlProperty` already implement `asDriver`.

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -45,6 +45,7 @@ custom_categories:
   - Driver+Subscription
   - Driver
   - ObservableConvertibleType+Driver
+  - PublishRelay+Driver
 - name: RxCocoa/Traits/SharedSequence
   children:
   - ObservableConvertibleType+SharedSequence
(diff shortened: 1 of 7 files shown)


freak4pc:
Hey @dmcyk - Thanks for your contribution. 
The fact Publish* cannot be turned into a `Driver` is not coincidental -

BehaviorRelay/BehaviorSubject models a State (hence it replays its latest value) and so does a Driver (models State) 

PublishRelay/PublishSubject model Events (hence it does *not* replay latest value), but `Driver` does not model event as it requires replaying its latest event. The correct `SharingStrategy` for `Publish*` is actually a `Signal` - which is why `asSignal` is available for these units. 

Hope this makes sense :)

Thanks again for the thought and assistance.
Shai.

dmcyk:
oh I see.  I wasn't aware of the separate `Signal` type Thanks for the explanation!

PR #1772, opened 2018-10-15
Fix watchOS minimum deployment and deprecation of OSAtomicIncrement32Barrier

### LINKED ISSUE #1752: Update WATCHOS_DEPLOYMENT_TARGET to 3.0 for Submission to Apple

freak4pc:
Supersedes #1753.
Fixes #1752. 

This PR:

* Takes @dlewanda's changes for the watchOS 3.0 minimum deployment.
* Fixes the deprecation of `OSAtomicIncrement32Barrier` by switching to `atomic_fetch_add` from `stdatomic.h`. 

As a reminder - we should merge this ASAP as it causes a critical issue when submitting to the App Store with a watchOS target.

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -6425,7 +6425,7 @@
 				VALIDATE_PRODUCT = YES;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
-				WATCHOS_DEPLOYMENT_TARGET = 2.0;
+				WATCHOS_DEPLOYMENT_TARGET = 3.0;
 			};
 			name = "Release-Tests";
 		};
@@ -6841,7 +6841,7 @@
 				TVOS_DEPLOYMENT_TARGET = 9.0;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
-				WATCHOS_DEPLOYMENT_TARGET = 2.0;
+				WATCHOS_DEPLOYMENT_TARGET = 3.0;
 			};
 			name = Debug;
 		};
@@ -6902,7 +6902,7 @@
 				VALIDATE_PRODUCT = YES;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
-				WATCHOS_DEPLOYMENT_TARGET = 2.0;
+				WATCHOS_DEPLOYMENT_TARGET = 3.0;
 			};
 			name = Release;
 		};
--- a/RxCocoa/Runtime/_RXObjCRuntime.m
+++ b/RxCocoa/Runtime/_RXObjCRuntime.m
@@ -11,6 +11,7 @@
 #import <objc/runtime.h>
 #import <objc/message.h>
 #import <libkern/OSAtomic.h>
+#import <stdatomic.h>
 
 #import "include/_RX.h"
 #import "include/_RXObjCRuntime.h"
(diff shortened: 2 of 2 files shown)


freak4pc:
<p align="center"></p>

freak4pc:
@kzaher Let me know if this makes sense for you

freak4pc:
Hey @kzaher - Just want to clarify a few things:

1. This PR moves all watchOS targets to 3.0, not just the RxCocoa one. 
2. The reason the fix was applied to the RxCocoa repo is because that methods is specifically hard deprecated (with an error, not a warning) in Objective C. 
3. In the Swift-related code this only causes a warning at the moment, not a build error (even for watchOS 3).

This PR should be sufficient to release a patch to resolve the critical issue at hand (can't build for watchOS 3, can't submit to App Store with watchOS 3) for now in my opinion, I think we should wait to see if we're getting a proper public API for stdatomic before we use our own implementation for everything, but that's just my personal opinions. 

tl;dr - 
RxCocoa is a build error (since its Objective-C) - so it was fixed 
RxSwift is only a warning - so I left it as is, for the time being. 

Appreciate your thoughts on this - I'm OK with whatever you decide, clearly.

Shai.

kzaher:
Hi @freak4pc ,

If we are publishing fixes, we should fix these warnings IMHO otherwise we might need another patch release for these warnings.

freak4pc:
I don't super mind but we had many releases with warning for a while and it
was never a issue- since we can't control random depreciation without
fallback from Apple.

I can switch to your implementation but I hope we won't see a performance
hit with that.

-- 
Shai Mishali

freak4pc:
Hey @kzaher - Removing the two build warnings is not enough to make this code work on watchOS because its importing Glibs and XCTest, but even when resolving this it seems that 

`currentThread.threadDictionary = threadDictionary`

doesn't work for the watchOS target (only for the Linux one). Any thoughts?

freak4pc:
Re-ping @kzaher - Without this fix (or forking the repo and applying it), people with watchOS targets can't submit to the App Store (reject).

kzaher:
> Hey @kzaher - Removing the two build warnings is not enough to make this code work on watchOS because its importing Glibs and XCTest, but even when resolving this it seems that

> currentThread.threadDictionary = threadDictionary

> doesn't work for the watchOS target (only for the Linux one). Any thoughts?

Ok, I will fix the RxSwift then.

PR #1777, opened 2018-10-20
Replace Queue's count check Logic

syjdev:
Hello, I want to contribute RxSwift Community.
So I try to replace Queue's count check logic.

- AS-IS
*.count > 0

- TO-BE
!*.isEmpty


kzaher:
Hi @syjdev ,

if you are doing this, then you should also fix the template files.

https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Observables/Zip%2Barity.tt

`Zip` files are generated.

PR #1781, opened 2018-10-27
Make `completable.andThen` an alias to `completable.concat`

freak4pc:
Was asked yesterday what's the difference between `completable.andThen(completable)` and `completable.concat(completable)`. In my head I imagined its simply an alias, but in practice it uses a separate sink which seems weird and increases the surface area for tests. 

These two operators variations should use the same implementation to prevent any weird side-effects if only one of the implementations change. 

The separate Sink is still useful for `completable.andThen(single/maybe/observable)`. 

Let me know if that makes sense to you :)


kzaher:
` This operator has tail recursive optimizations that will prevent stack overflow.`

You can check out the source code to get a better idea of the differences in complexity between them.

PR #1782, opened 2018-10-27
update RxSwift version into latest version

isfaaghyth:


freak4pc:
Thanks @isfaaghyth - I'm not sure this change is entirely needed. For README & SPM it's still useful to indicate the latest major release, so I'm personally in favour of leaving as is. Depending on other people's opinions :)

mendesbarreto:
Yeap! I agree with @freak4pc. Indeed if you run the swift package manager or the carthage update with the current file setup, it will get the last version anyway 😉.

In my opinion is better leaving as is, because when we change the RxSwift version adding new features or fix some bugs, we don't need to worry about change this part of Readme file.

isfaaghyth:
I see, thankyou. closed! 😄

PR #1783, opened 2018-10-27
Add takeUntil(_ predicate:) and takeUntil(_ element:)

freak4pc:
The first variation - `takeUntil(_ predicate:)` - is the inverse of `takeWhile(_ predicate:)`. 

In RxJava, and other implementations, `takeUntil(predicate:)` is one of the most popular usages of this operator (compared to the `takeUntil(secondObservable)` variation) and I feel I always have to "flip around terms" (or use different terminology compared to android co-workers). I usually wouldn't bother adding operators just for the sake of alignment but in this case I think its quite valuable. 

The second version - `takeUntil(_ element:)` is simply syntactic sugar but I think it's quite helpful. I'm not religious to having it but I think it's an extremely common use case. 

Appreciate any opinions / thoughts :)

### DIFF
--- a/RxSwift/Observables/TakeUntil.swift
+++ b/RxSwift/Observables/TakeUntil.swift
@@ -26,14 +26,26 @@ extension ObservableType {
 
      - seealso: [takeUntil operator on reactivex.io](http://reactivex.io/documentation/operators/takeuntil.html)
 
+     - parameter behavior: Whether or not to include the last element matching the predicate.
      - parameter predicate: A function to test each element for a condition.
      - returns: An observable sequence that contains the elements from the input sequence that occur before the element at which the test passes.
      */
-    public func takeUntil(_ predicate: @escaping (E) throws -> Bool)
+    public func takeUntil(_ behavior: TakeUntilBehavior,
+                          predicate: @escaping (E) throws -> Bool)
         -> Observable<E> {
-        return TakeUntilPredicate(source: asObservable(), predicate: predicate)
+        return TakeUntilPredicate(source: asObservable(),
+                                  behavior: behavior,
+                                  predicate: predicate)
     }
 }
+
+/// Behaviors for the `takeUntil(_ behavior:predicate:)` operator.
+public enum TakeUntilBehavior {
+    /// Include the last element matching the predicate.
+    case inclusive
+
+    /// Exclude the last element matching the predicate.
+    case exclusive
 }
 
 // MARK: - TakeUntil Observable
@@ -177,6 +189,10 @@ final fileprivate class TakeUntilPredicateSink<O: ObserverType>
             if _running {
                 forwardOn(.next(value))
             } else {
+                if _parent._behavior == .inclusive {
+                    forwardOn(.next(value))
+                }
+
                 forwardOn(.completed)
                 dispose()
             }
@@ -193,9 +209,13 @@ final fileprivate class TakeUntilPredicate<Element>: Producer<Element> {
 
     fileprivate let _source: Observable<Element>
     fileprivate let _predicate: Predicate
+    fileprivate let _behavior: TakeUntilBehavior
 
-    init(source: Observable<Element>, predicate: @escaping Predicate) {
+    init(source: Observable<Element>,
+         behavior: TakeUntilBehavior,
+         predicate: @escaping Predicate) {
         _source = source
+        _behavior = behavior
         _predicate = predicate
     }
 
--- a/RxSwift/Observables/TakeUntil.swift
+++ b/RxSwift/Observables/TakeUntil.swift
@@ -20,8 +20,23 @@ extension ObservableType {
         -> Observable<E> {
         return TakeUntil(source: self.asObservable(), other: other.asObservable())
     }
+
+    /**
+     Returns elements from an observable sequence until the specified condition is true.
+
+     - seealso: [takeUntil operator on reactivex.io](http://reactivex.io/documentation/operators/takeuntil.html)
+
+     - parameter predicate: A function to test each element for a condition.
+     - returns: An observable sequence that contains the elements from the input sequence that occur before the element at which the test passes.
+     */
+    public func takeUntil(_ predicate: @escaping (E) throws -> Bool)
+        -> Observable<E> {
+        return TakeUntilPredicate(source: asObservable(), predicate: predicate)
+    }
+}
 }
 
+// MARK: - TakeUntil Observable
 final fileprivate class TakeUntilSinkOther<Other, O: ObserverType>
     : ObserverType
     , LockOwnerType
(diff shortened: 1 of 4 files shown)


kzaher:
Hi @freak4pc ,

it seems to me that we could just add an alias and a `!` transformation to `takeWhile`, right? I don't think that we need to implement the entire operator?

yikuo123:
> Hi @freak4pc ,
> 
> it seems to me that we could just add an alias and a `!` transformation to `takeWhile`, right? I don't think that we need to implement the entire operator?

it seems different between takeWhile and takeUntil. takeUntil includes the last element witch matches the condition, but takeWhile doesn't.

freak4pc:
Hey @kzaher - Thanks for taking a look.

@yikuo123 is correct. My initial thought was using an inverted version of `takeWhile` - but that one is exclusive of the last element, and this one is inclusive which is extremely important. 

Let me know your thoughts :)

freak4pc:
Also, rebased on top of develop.

yikuo123:
@kzaher  RxJava describes it as following:


tim 20181227120655

freak4pc:
My original point was that I have to flip terms in my head when working with takeWhile 

If a co worker does takeUntil { $0 == 3 } 
I have to do takeWhile { $0 < 4 }. That’s what I meant with it being truly it confusing. They mean different things. 

Until doesn’t have anything in it, linguistic-wise that enforces inclusivity or exclusivity. If your boss says “your deadline is until next Monday” - it could mean including Monday or excluding (usually the former).

DevAndArtist:
I would love to see the inclusion of this operator. I just had a case where `takeWhile` was not enough because it rejected the last element of the sequence and had to workaround it with a `materialize / dematerialize` dance. `takeUntil` operator with a predicate closure would have solved my issue more elegantly.

kzaher:
Hi @freak4pc ,

I can understand why this operator would be useful, but I think that we would need to improve the interface to pull this in.

IMHO the way it's currently defined in RxJava is extremely confusing and ambiguous. 

If we named it something like: 
* `takeUntil(.includeXXX) { $0 == green }`
* `takeUntil(.excludeXXX) { $0 == green }`

It might be easy to figure out from the API what it does. I have no idea what XXXX should be.

freak4pc:
@kzaher I don't think this is a good idea TBH
I think if we take an operator that has a certain behavior in RxJava, Rx.NET, etc - we should retain the official behavior to not confuse people switching around different (supposedly) platform-agnostic implementations. 

We can possibly do 

`takeUntil(inclusive: Bool = true, predicate:)` (Where `inclusive` is . `true` by default).

TBH I personally am against that because `takeUntil` is always `inclusive` among implementations and I don't find it super platform-agnostic to ReactiveX to make "our own" interpretation of it.

kzaher:
Hi @freak4pc ,

in general I agree, just please note that there is no such thing as official behavior. I have learned a lot from other implementations, but that doesn't mean that we shouldn't be critical and trying to improve APIs.

If we see something wrong or confusing with other APIs we shouldn't just blindly copy bad API. I don't consider this project to be a religion :)

Other APIs aren't perfect. RxJava 2.0 had a complete rewrite from hacked up 1.x version. RxJS in 6.0 broke the universe when they went from `Observable` extensions to `.pipe(map())` versions of the API. Rx.NET has mistaken `throttle` for `debounce`. They probably still do.

http://reactivex.io site just uses RxJava docs and current RxJava behavior. What ever is there doesn't make it official in any way. It is just a showcase of all of the **current** versions of RxJava operators.

> TBH I personally am against that because takeUntil is always inclusive among implementations and I don't find it super platform-agnostic to ReactiveX to make "our own" interpretation of it.

To the best of my knowledge Rx.NET doesn't even have that operator because it's too confusing https://github.com/dotnet/reactive/search?p=2&q=takeuntil&unscoped_q=takeuntil The version they have only accepts other `Observable` or time which makes it a really clear API.

freak4pc:
@kzaher Thanks for the feedback. 

Overall, totally agree with you. When I say "official behavior" I mean that the marble chart that appears on ReactiveX should be reflected in the operator behavior that we create, in my opinion. 

As long as we preserve that behavior, even if we add some API flag, I think we're good overall.

Would you be OK with pulling this PR if I change the sig to `takeUntil(inclusive:predicate:)` ?

kzaher:
I think I would be ok pulling in
```
enum TakeUntilBehavior {
    // or some better name 
    case inclusive
    // or some better name
    case exclusive
}
public func takeUntil(_ behavior: TakeUntilBehavior, _ predicate: @escaping (E) throws -> Bool) -> Observable<E>
```

I would remove this API because I don't think I have seen enough evidence that we should bloat our API with this version.
```
    public func takeUntil(_ element: E) -> Observable<E> {
        return TakeUntilPredicate(source: asObservable(), predicate: { $0 == element })
    }
```

freak4pc:
Is there any reason you want to introduce another enum for that? Given it's a binary - yes/no option ? Or do you expect additional behaviors to be added here ?

DevAndArtist:
@freak4pc having a binary state doesn't mean it must be a `Bool`. I think such an enum is more descriptive / readable than a pain boolean.

freak4pc:
@DevAndArtist I am personally not a fan of adding additional abstraction layers just for the sake of abstraction. If the options are `inclusive` or `exclusive`, Bool `inclusive` means exactly the same to me. If it's a preference of both, I'm not religious to it but it feels moot.

kzaher:
It seems to me that the API would be shorter and more readable in this form.

```swift
takeUntil(.inclusive) { $0 == green  }
takeUntil(.exclusive) { $0 == green  }
```
vs 
```swift
takeUntil(inclusive: true) { $0 == green  }
takeUntil(inclusive: false) { $0 == green  }
```

It also provides us with an escape hatch in case we need to extend the behavior in the future.

freak4pc:
I disagree on the readability but I don't mind enough to argue for no good reason here, so I'll go with it :) Commit coming up.

DevAndArtist:
Just `.inclusive` be the default value or should we always type out the behavior manually?

freak4pc:
I'm going with `.inclusive` default to keep in line with the ReactiveX behavior while still allowing additional flexibility.

kzaher:
> I disagree on the readability but I don't mind enough to argue for no good reason here,

🙇 

> Just .inclusive be the default value or should we always type out the behavior manually?

I would not make it default. It is ambiguous when excluded. You can create your own convenience if you dislike it much :)

freak4pc:
Ugh, I really think this is not the way to go :) 
`throttle` has `latest` that is defaulted to true - we don't force a user to read documentation just to get the "standard" behavior. We just let them choose the non-standard behavior via a flag in this case. 

I'll appreciate if, specifically for the default var, you'll reconsider here, as I think it really dirties up the callsite for the most common use case of this operator.

kzaher:
@freak4pc If we don't make it explicit that API will be annoying to understand.

This is what you've written on the top of this thread.

> The first variation - takeUntil(_ predicate:) - is the inverse of takeWhile(_ predicate:).

> The second version - takeUntil(_ element:) is simply syntactic sugar but I think it's quite helpful. I'm not religious to having it but I think it's an extremely common use case.

Which is not true. This means that you were also confused how this operator really works. 

If you just wrote the implementation and were confused and if I've seen the name and was also assuming it worked in a different way, I don't think it's realistic to assume that arbitrary person will be able to intuitively figure out how does this operator behave.

Having a more verbose but clear API is better option IMHO.

> throttle has latest that is defaulted to true - we don't force a user to read documentation just to get the "standard" behavior. 

I should have probably designed that interface better. It was my fault. I just needed something that worked and was too tired to spend much time thinking about the correct naming. In the case of throttle it was really hard to compress  `(2.0, .latestElementReceivedInADueTimeWideTimeWindowSinceLastElementEmissionBeEmitted)`. 

Other implementations were (and still are) also implementing `throttle` horribly, so in the end my conclusion was, if somebody complains, he is free to make a PR and improve the API.

RxSwift/Observables/TakeUntil.swift
```diff
@@ -129,3 +156,72 @@ final fileprivate class TakeUntil<Element, Other>: Producer<Element> {
...
+
+    init(parent: Parent, observer: O, cancel: Cancelable) {
+        _parent = parent
+        super.init(observer: observer, cancel: cancel)
+    }
+
+    func on(_ event: Event<Element>) {
+        switch event {
+        case .next(let value):
+            if !_running {
```
kzaher: I would probably put `self` everywhere. I know that we weren't consistent in the past, but looks good enough to pull in.

freak4pc: I’ll  take care of this in a separate PR :) thanks !

RamblinWreck77:
Oh wow, I just ran into this problem and could really use this! Excited to see it's already been merged. Any estimates on when this will ship?

kzaher:
We need to fix https://github.com/ReactiveX/RxSwift/issues/1853 and then we'll probably do a new release.

PR #1784, opened 2018-10-28
Fix deprecated atomic APIs.

### LINKED ISSUE #1730: OSAtomicXX functions deprecated in iOS 10
*If you still want to report issue, please delete above statements before submitting an issue.*

iOS11 deprecates functions OSAtomicXX which are used by RxSwift under Platform.Darwin.swift. 

**Expected outcome**:

Rewrite Atomic methods using alternative solutions.

**What actually happens**:

Xcode warns you of deprecated functions.

4.2.0

**Platform/Environment**

  - [X] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

  - [X] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

9.4.1

**Installation method**:
  - [ ] CocoaPods
  - [ ] Carthage
  - [X] Git submodules

  - [ ] yes (which ones)
  - [X] no

  - [ ] just starting
  - [X] I have a significant code base

kzaher:
This PR should:
* Fix #1730.
* Improve Linux performance.
* Cleanup our implementation.

Unfortunately I've had to add another library. It would be ideal to not add it, but I'm not sure that is possible.

@freak4pc could you maybe help me test this?

### DIFF
--- a/RxAtomic/include/RxAtomic.h
+++ b/RxAtomic/include/RxAtomic.h
@@ -1,0 +1,48 @@
+//
+//  RxAtomic.h
+//  RxAtomic
+//
+//  Created by Krunoslav Zaher on 10/28/18.
+//  Copyright © 2018 Krunoslav Zaher. All rights reserved.
+//
+
+#ifndef RxAtomic_h
+#define RxAtomic_h
+
+#include <stdatomic.h>
+
+#define SWIFT_NAME(_name) __attribute__((swift_name(#_name)))
+
+#define Atomic(swift_type, llvm_type) \
+    typedef struct { volatile atomic_##llvm_type atom; } Atomic##swift_type;\
+    static __inline__ __attribute__((__always_inline__)) SWIFT_NAME(Atomic##swift_type.initialize(self:_:)) \
+    void Atomic##swift_type##_Initialize(Atomic##swift_type * _Nonnull self, llvm_type value) { \
+        atomic_init(&self->atom, value);\
+    }\
+    \
+    static __inline__ __attribute__((__always_inline__)) SWIFT_NAME(Atomic##swift_type.load(self:)) \
+    llvm_type Atomic##swift_type##_Load(Atomic##swift_type * _Nonnull self) { \
+        return atomic_load(&self->atom);\
+    }\
+    \
+    static __inline__ __attribute__((__always_inline__)) SWIFT_NAME(Atomic##swift_type.fetchOr(self:_:)) \
+    llvm_type Atomic##swift_type##_FetchOr(Atomic##swift_type * _Nonnull self, llvm_type mask) { \
+        return atomic_fetch_or(&self->atom, mask);\
+    }\
+    \
+    static __inline__ __attribute__((__always_inline__)) SWIFT_NAME(Atomic##swift_type.add(self:_:)) \
+    llvm_type Atomic##swift_type##_Add(Atomic##swift_type * _Nonnull self, llvm_type value) { \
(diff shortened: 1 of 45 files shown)


freak4pc:
Sure, I'll take a look tomorrow. Agreed another library is a bit jarring but if there's no choice, there's no choice. Can't this live in RxSwift itself ?

kzaher:
I think that Swift package manager doesn't allow mixing of C and Swift inside a single module. If you can make it to work inside RxSwift that would be awesome.

RxAtomic/include/RxAtomic.h
```diff
@@ -0,0 +1,48 @@
...
+
+#define Atomic(swift_type, llvm_type) \
+    typedef struct { volatile atomic_##llvm_type atom; } Atomic##swift_type;\
+    static __inline__ __attribute__((__always_inline__)) SWIFT_NAME(Atomic##swift_type.initialize(self:_:)) \
+    void Atomic##swift_type##_Initialize(Atomic##swift_type * _Nonnull self, llvm_type value) { \
+        atomic_init(&self->atom, value);\
+    }\
+    \
+    static __inline__ __attribute__((__always_inline__)) SWIFT_NAME(Atomic##swift_type.load(self:)) \
+    llvm_type Atomic##swift_type##_Load(Atomic##swift_type * _Nonnull self) { \
```
freak4pc: Oh boy... this is a bit of a crazy hack in a way - I guess there's no other way using stdatomic from Swift code that you know of ?

freak4pc:
> I think that Swift package manager doesn't allow mixing of C and Swift inside a single module. If you can make it to work inside RxSwift that would be awesome.

I didn't notice it's using underlying C code - this smells a bit weird to me but I imagine there's no other good way to do this. Was the pure-Swift linux impl. you wrote not performant enough ?

kzaher:
@freak4pc Apple was exposing their own wrappers for atomic operations. Since they've deprecated their own implementations we don't have any official Apple API to rely on. This wraps llvm intrinsics.

Since llvm intrinsics aren't directly available in Swift, this is a workaround.

There is no pure Swift implementation. Swift uses NSRecursiveLocks or OSAtomic***, both of which aren't pure Swift.

freak4pc:
OK, Makes sense - and second question is - you mentioned SPM can't bundle C files, but it does seem there is a SPM target for the separate `RxAtomic` framework - putting these same files in the main RxSwift target won't help in this specific case? I won't have time to look into this but wondering what are your thoughts. 

In any way I'll run my unit tests on this later and see if there are any regressions.

kzaher:
> OK, Makes sense - and second question is - you mentioned SPM can't bundle C files, but it does seem there is a SPM target for the separate RxAtomic framework - putting these same files in the main RxSwift target won't help in this specific case?

There is a separate SPM target.

freak4pc:
Ran 460 View Model unit tests on my side and I'm not seeing any issues.

freak4pc:
FYI: https://bugs.swift.org/browse/SR-9144

Would help if you can join the convo @kzaher

kzaher:
@freak4pc cool. Thnx.

RxAtomic/include/RxAtomic.h
kzaher: The other was was `OSAtomic*` :)

PR #1786, opened 2018-10-29
Add operator take in SharedSequence

### LINKED ISSUE #1761: Add operator take to SharedSequenceConvertibleType
Hi.
I want to use `take` operator in Driver.
In order to do it, it need to implement `take` operator to `SharedSequenceConvertibleType`.
I think that we can add this operator to `SharedSequenceConvertibleType`, because other operators also known as `delay`, `startWith`, `skip` are already implemented.

Do you think it’s worth to adding this operators?
If so, I am able to do this and open a pull request.

Thanks!

kzumu:
close #1761 
Added `take` operator  in SharedSequence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Master
 
 ## [4.3.1](https://github.com/ReactiveX/RxSwift/releases/tag/4.3.1)
+* Add `take` operator in SharedSequenceConvertible.
 
 #### Anomalies
 
--- a/RxCocoa/Traits/SharedSequence/SharedSequence+Operators.swift
+++ b/RxCocoa/Traits/SharedSequence/SharedSequence+Operators.swift
@@ -512,4 +512,12 @@ extension SharedSequenceConvertibleType {
 
         return SharedSequence(source)
     }
+    
+    public func take(_ count: Int)
+        -> SharedSequence<SharingStrategy, E> {
+        let source = self.asObservable()
+            .take(count)
+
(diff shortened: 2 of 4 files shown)


kzaher:
Hi @kzumu @freak4pc ,

Here is the rationale why we have `skip` and not `take`. `Driver` and `Signal` should ideally be non terminating, otherwise there could be unexpected work performed in subscription.

This might be ok if the `Driver` `Signal` is finite and const (`just`, `empty`) and no work is performed on subscription, but I don't think we should support adding operators that transform potentially non terminating observables sharing subscription to terminating.

Please bear in mind that the whole reason why `Driver` and `Signal` are shared in the first place is to reuse subscription work.

`take` will make any `Driver` and `Signal` terminating, which is something we probably shouldn't encourage.

Alexey-Matjuk:
@kzaher why then `Driver` even has completed event then?
```swift
/// - parameter onCompleted: Action to invoke upon graceful termination of the observable sequence.
public func drive(onNext: ((Self.E) -> Void)? = nil, onCompleted: (() -> Void)? = nil, onDisposed: (() -> Void)? = nil) -> RxSwift.Disposable
/// - parameter onCompleted: Action to invoke upon graceful termination of the observable sequence.
public func `do`(onNext: ((Self.E) -> Void)? = nil, onCompleted: (() -> Void)? = nil, onSubscribe: (() -> Void)? = nil, onSubscribed: (() -> Void)? = nil, onDispose: (() -> Void)? = nil) -> RxCocoa.SharedSequence<Self.SharingStrategy, Self.E>
```

PR #1794, opened 2018-11-05
Remove type checking performance debugging flags

tonyarnold:
This PR disables the following Swift flags:

- `-debug-time-function-bodies`
- `-warn-long-expression-type-checking=100`
- `-warn-long-function-bodies=100`

They are appearing in our build logs during debugging and testing. I think you could probably turn these on and off as you actually need them, rather than forcing their use in all projects that build RxSwift?

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -4285,7 +4285,7 @@
 				MACOSX_DEPLOYMENT_TARGET = 10.9;
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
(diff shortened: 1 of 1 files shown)


freak4pc:
I think these are basically there so people notice these issues during development. I don't think removing them entirely would be the best. Which dependency manager are you using for importing it? 

If CocoaPods, What I personally usually do is `inhibit_all_warnings!` for 3rd parties, since those aren't usually related to my app. But YMMV.

kzaher:
Does anybody else have issues with those debugging flags?

freak4pc:
The "issue" is that - if you take in the dependency as a submodule / subproject / CocoaPods (without 'inhibit_all_warnings'), you will definitely get these invoked. 

I basically always use `inhibit_all_warnings` for my 3rd parties (since I don't feel responsible for my dependecies' warnings but only my own project), so I don't find it bothering, but I can see why it would bother some users.

kzaher:
Ok, I'll move this to `Release-Tests`. We are running that on CI anyway.

PR #1795, opened 2018-11-05
Refactor self usage

cruisediary:
Refactor 72 self usage cases
```swift
// as is
with(self.asObservable())(curriedArgument)
return BlockingObservable(timeout: timeout, source: self.asObservable())
return Observable.concat([self.asObservable(), second.asObservable()])

// to be
with(asObservable())(curriedArgument)
return BlockingObservable(timeout: timeout, source: asObservable())
return Observable.concat([asObservable(), second.asObservable()])

and so on
```

- Always using self. to access instance members and functions. 
- Using self. only when the compiler needs it. 
- All Unit test passed

#### Swift evolution
https://github.com/apple/swift-evolution/blob/master/proposals/0009-require-self-for-accessing-instance-members.md#counter-argument

#### Clean code
http://thebugcode.github.io/when-to-self-in-swift-and-when-not-to-2/

### DIFF
--- a/RxSwift/Observables/Amb.swift
+++ b/RxSwift/Observables/Amb.swift
@@ -35,7 +35,7 @@ extension ObservableType {
     public func amb<O2: ObservableType>
         (_ right: O2)
         -> Observable<E> where O2.E == E {
-        return Amb(left: asObservable(), right: right.asObservable())
+        return Amb(left: self.asObservable(), right: right.asObservable())
     }
 }
 
--- a/RxSwift/Observables/Catch.swift
+++ b/RxSwift/Observables/Catch.swift
@@ -18,7 +18,7 @@ extension ObservableType {
      */
     public func catchError(_ handler: @escaping (Swift.Error) throws -> Observable<E>)
         -> Observable<E> {
-        return Catch(source: asObservable(), handler: handler)
+        return Catch(source: self.asObservable(), handler: handler)
     }
 
     /**
@@ -31,7 +31,7 @@ extension ObservableType {
      */
     public func catchErrorJustReturn(_ element: E)
         -> Observable<E> {
-        return Catch(source: asObservable(), handler: { _ in Observable.just(element) })
+        return Catch(source: self.asObservable(), handler: { _ in Observable.just(element) })
     }
     
 }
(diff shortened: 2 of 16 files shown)


freak4pc:
I'm OK with this PR. Prefer getting rid of self if it's not needed, as well.

cruisediary:
Already some operators are not using self like `self.asObservable()` 👻 
- Amb
- Catch
- ElementAt
- Filter
- Merge
- SingleAsync
- Skip
- SkipUntil
- SkipWhile
- Switch
- SwitchIfEmpty
- Take
- TakeLast
- TakeUntil
- TakeWhile
- WithLatestFrom

kzaher:
Hi @cruisediary ,

I'm not sure I agree with the https://github.com/apple/swift-evolution/blob/master/proposals/0009-require-self-for-accessing-instance-members.md#counter-argument

"I'm not sure how many Swift users this effects, but I'm colorblind and I really struggle with the local vs properties syntax coloring." -- Tyler Cloutier, cloutiertyler@aol.com

Even if one is not colorblind and it not using an IDE when reading the code (like during PRs), referencing `self.` helps.

It is actually our current team policy to always use `self` when referencing instance variables. We've used some combination of `_` and `self` in this project, and yes, it's not ideal because Swift doesn't enforce using `self`.

cruisediary:
Hi @kzaher 
If the team's policy is to always use `self`
How about using `self` for unification. because some operators use `self` or NOT 👻

kzaher:
Yeah, we should put self where it doesn’t exist for consistency.

cruisediary:
I updated putting `self` where it doesn't exist like above list (Amb ~ WithLatestFrom) 👻 FYI @kzaher

freak4pc:
@cruisediary - I think this is a relatively partial fix right? It seems like very few updates :) I don't mind merging partial improvements but let me know if you'd like to include a few more additions here

kzaher:
I guess I'm fine merging this. @freak4pc  Are we waiting for some changes?

freak4pc:
Note to my knowledge. I mean, there are many other locations where this should be fixed, but this is also fine as is partially.

PR #1805, opened 2018-11-18
Fix DISABLE_SWIZZLING compilation

### LINKED ISSUE #1804: Can't compile RxCocoa withDISABLE_SWIZZLING flag
Compiling `RxCocoa` using `-DDISABLE_SWIZZLING` is broken on current master. _RxDelegateProxy is calling methods that are guarded by the flag. For example [this one](https://github.com/ReactiveX/RxSwift/blob/master/RxCocoa/Runtime/_RXDelegateProxy.m#L118).

## General Information

* RxSwift Version: master
* Swift Version: 4.2

Reflejo:
Some methods under the compiler flag are being used by RxCocoa's proxy delegate. So this commit is moving these outside the conditional.

Fixes #1804

### DIFF
--- a/RxCocoa/Runtime/include/_RXObjCRuntime.h
+++ b/RxCocoa/Runtime/include/_RXObjCRuntime.h
@@ -87,6 +87,8 @@ void * __nonnull RX_reference_from_selector(SEL __nonnull selector);
 /// Ensures interceptor is installed on target object.
 IMP __nullable RX_ensure_observing(id __nonnull target, SEL __nonnull selector, NSError *__autoreleasing __nullable * __nullable error);
 
+#endif
+
 /// Extracts arguments for `invocation`.
 NSArray * __nonnull RX_extract_arguments(NSInvocation * __nonnull invocation);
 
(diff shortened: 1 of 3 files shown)


RxCocoa/Runtime/include/_RXObjCRuntime.h
```diff
@@ -98,5 +100,3 @@ BOOL RX_is_method_signature_void(NSMethodSignature * __nonnull methodSignature);
 
 /// Default value for `RXInterceptionObserver.targetImplementation`.
 IMP __nonnull RX_default_target_implementation(void);
```
kzaher: I don't think this needs to be included, right?

kzaher: I think that we just need to expose 

```
/// Returns `YES` in case method has `void` return type.
BOOL RX_is_method_with_description_void(struct objc_method_description method);
 /// Extracts arguments for `invocation`.
NSArray * __nonnull RX_extract_arguments(NSInvocation * __nonnull invocation);
 /// Returns `YES` in case methodSignature has `void` return type.
BOOL RX_is_method_signature_void(NSMethodSignature * __nonnull methodSignature);
```

PR #1809, opened 2018-11-22
Make bind debuggable

muukii:
In development, It occurs fatalError if `bind` receives.
But now, it's hard for us to find where the error occurred.
So, I would like to make it more debuggable.

This PR has my first idea.
Let me know what do you think about it.

### DIFF
--- a/Documentation/GettingStarted.md
+++ b/Documentation/GettingStarted.md
@@ -935,7 +935,7 @@ self.rx.observe(CGRect.self, "view.frame", retainSelf: false)
 
 ### `rx.observeWeakly`
 
-`rx.observeWeakly` has somewhat slower than `rx.observe` because it has to handle object deallocation in case of weak references.
+`rx.observeWeakly` is somewhat slower than `rx.observe` because it has to handle object deallocation in case of weak references.
 
 It can be used in all cases where `rx.observe` can be used and additionally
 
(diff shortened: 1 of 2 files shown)


freak4pc:
I think that’s quite weird. You’d have to go around and change subscribe, drive, emit, etc.

It seems like an OK hack to track something in your system (eg add for your own use) but personally don’t think it’d be a good addition for the library. Especially since rxFatalError would usually happens as you implement something so it would be caught quite early. 

Even if this would’ve been a proper solution it should be DEBUG-only, but I still think it doesn’t really fit here

muukii:
Yes, we have to change to other observer instead of `bind` in that case.
But, the problem is even If we are stopped there with` fatalError`, we can not find `bind` where are written.
We also can't find from stack-trace.
I need file and line to fix this.

kzaher:
Hi @muukii ,

I understand why you would want something like that. We have a similar situation with `subscribe`. We first had a debug fatal error and release error logging. 

Different signatures of the methods were causing linking issues.

That's why we've introduced call stack recorder. 
https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/ObservableType%2BExtensions.swift#L54

It seems to me that there is a similar situation here. We might reuse that call stack recorder.

muukii:
@kzaher 

Thank you for your understanding.
And, I didn't know what features of Hooks.
Definitely, it may help for debugging where it is subscribed.
Because `bind` call `subscribe` inside.
So, I think this PR can be closed.
Thanks!

PR #155, opened 2015-09-22
Use guard for simplification.

yoshinorisano:
When I was looking into the RxExample code, I thought it could be improved by using `guard` statements introduced in Swift 2.
Though it is not essential change for RxExample, if it is good for you, please consider merge it.
Thanks :)

PR #1813, opened 2018-11-26
[Pitch] Creating type abstracting relay types common duplicated logic

LucianoPAlmeida:
On Signal+Subscriptions there were two methods with the same implementation for two different types of Relays (`Publish` and `Behavior`). So maybe it's worth creating a protocol level of abstraction over relay and remove this duplication using the generic type. It's a very little duplication and maybe doesn't worth the abstraction. 
@kzaher and @freak4pc let me know your thoughts on this one :)

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -33,6 +33,7 @@ custom_categories:
   - DispatchQueue+Extensions
 - name: RxCocoa/Traits
   children:
+  - ObservableRelayType
   - BehaviorRelay
   - ControlEvent
   - ControlProperty
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -33,10 +33,10 @@ custom_categories:
   - DispatchQueue+Extensions
 - name: RxCocoa/Traits
   children:
-  - ObservableRelayType
   - BehaviorRelay
   - ControlEvent
   - ControlProperty
+  - ObservableRelayType
   - PublishRelay
 - name: RxCocoa/Traits/Driver
   children:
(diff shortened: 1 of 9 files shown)


freak4pc:
I think there was already a discussion about unifying binding related operations that was rejected 
https://github.com/ReactiveX/RxSwift/pull/155

In any way even if we want to unify Relays as a single thing it needs to happen for all types, not just Signal. I’d suggest leaving this open to see if Krunoslav could take a peek.

LucianoPAlmeida:
Now I see the point, thank you :)) 
I think this one is specific to Relay types, more conceptual abstraction around `accept()`. And different form 1551 it doesn't affect ObserverType and by doing that being an abstraction only affecting RxCocoa scope and Relay types `Publish` and `Behavior`. 
This is more a pitch to bring the discussion and feel  free to close if this doesn't fit :))

LucianoPAlmeida:
Just updating the pitch :))
Also, this is something that can be done with a class `Relay<Element>` with an "abstract" method `accept` where `PublishRelay` and `BehaviorRelay` inherits from it. 
But now I'm thinking about it better and maybe this whole pitch seems like an over-abstraction, not sure 😆

kzaher:
Hi @LucianoPAlmeida ,

The purpose of the particular overloads was to only allow particular types of binds.

> But now I'm thinking about it better and maybe this whole pitch seems like an over-abstraction, not sure 😆

Yeah, this seems to me like an over-abstraction.

LucianoPAlmeida:
I see the point 👍 
Closing this then, thank's for the comments @freak4pc @kzaher :))

PR #1822, opened 2018-12-10
Remove Code Signing

freak4pc:
Some targets still have code signing for some reason. Did a quick cleanup.


LucianoPAlmeida:
Hey @freak4pc :)
Just notice that these redundant builds happen here too.  
When you working from a branch it always triggers two builds 😭 
It started to bother a bit on another project and actually seems to be a very simple solution. 
You can check it [the issue and PR](https://github.com/SwifterSwift/SwifterSwift/issues/613) 
Maybe it could be applied here too :))

freak4pc:
Great note ! Would definitely want to add that.

PR #1830, opened 2018-12-21
Improve doc comment for asSingle and asMaybe

hiragram:
Updated `asSingle()` and `asMaybe()` 's document comments.


hiragram:
@GeekTree0101 Thank you for your review. I pushed a new commit. What do you think of it?

freak4pc:
I think this accurately describes a little-unknown edge case of `asSingle()`. Any thoughts here @kzaher ?

kzaher:
@freak4pc 
Yep, looks good.

PR #1834, opened 2018-12-23
Removing` _stopped` in PublishSubject

mykoma:
The ` _stoppedEvent` can be represent the `_stopped`，so it's better to remove the `_stopped`


kzaher:
Last time I've tested this there was a big difference in using `_stoppedEvent != nil` or a boolean.

Will need to recheck benchmarks.

PR #1837, opened 2018-12-26
Add swift_version to podspecs

timdonnelly:
`swift_version` is supported as of [Cocoapods 1.4](https://blog.cocoapods.org/CocoaPods-1.4.0/#swift-version-dsl). Adding this removes the requirement that the pod be integrated into a target which uses the same Swift version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,35 +41,19 @@ matrix:
     include:
       - osx_image: xcode10
         os: osx
-        env: BUILD="pod lib lint --swift-version=4 RxSwift.podspec"
+        env: BUILD="pod lib lint RxSwift.podspec"
 
       - osx_image: xcode10
         os: osx
-        env: BUILD="pod lib lint --swift-version=4 RxCocoa.podspec"
+        env: BUILD="pod lib lint RxCocoa.podspec"
 
       - osx_image: xcode10
         os: osx
-        env: BUILD="pod lib lint --swift-version=4 RxBlocking.podspec"
+        env: BUILD="pod lib lint RxBlocking.podspec"
 
       - osx_image: xcode10
         os: osx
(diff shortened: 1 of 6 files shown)


timdonnelly:
I also removed `--swift_version` from podspec linting in the Travis config. As per Cocoapods documentation:

> Note: The usage of the .swift-version file as well as the --swift-version parameter during lint is now deprecated. You may continue to use it but you will receive a warning to use the swift_version attribute instead.

kzaher:
Hi @timdonnelly ,

The changes you've made seem reasonable, but I don't think we can merge this because this would break backward compatibility.

We have a bunch of `#if swift( >= 4.1)` or `#if swift (>=4.2)` in our code base. Fixing the swift version to 4.0 could cause breakage.

freak4pc:
Agree with @kzaher - But I guess we can possibly add this to RxSwift 4.5.0 while removing all of these conditional build statements? Or would you want to wait for RxSwift 5 at this point?

bjtitus:
Can this issue be re-evalutated in light of the multiple `swift-versions` in CocoaPods 1.7.0?
http://blog.cocoapods.org/CocoaPods-1.7.0-beta/

freak4pc:
We can definitely do this when 1.7.0 will be releasing a RC / Final release. I don't think it would be a good idea to count on a beta of CocoaPods for anything like this.

freak4pc:
With the release of Swift 5 (and RxSwift 5 soon), i don't think this would matter much due to binary stability.

freak4pc:
Closing this in favour of the upcoming RxSwift 5 release. We'll only support Xcode 10.2 and Swift 5 (and above), so the Swift versions shouldn't matter anymore.

PR #1842, opened 2018-12-28
Make AnyRecursiveScheduler implement Disposable.

mykoma:
AnyRecursiveScheduler had implement the function `dispose`, so It's should be implement the protocol Disposable


freak4pc:
I don't think that's a correct assumption... Don't think you would want someone to add a Scheduler directly into a DisposeBag. Also, note that `dispose` and the other methods are internal to the framework and not public.

mykoma:
Yes, you are right.

PR #1845, opened 2019-01-03
Fix typo(?): `ValidationResult` -> `Availability`

dgoldman-pdx:
I'm new here, but I *think* that this was simply a typo.

### DIFF
--- a/Documentation/Examples.md
+++ b/Documentation/Examples.md
@@ -146,7 +146,7 @@ self.usernameOutlet.rx.text
         let loadingValue = Availability.pending(message: "Checking availability ...")
 
         // This will fire a server call to check if the username already exists.
-        // Its type is `Observable<ValidationResult>`
+        // Its type is `Observable<Availability>`
         return API.usernameAvailable(username)
           .map { available in
               if available {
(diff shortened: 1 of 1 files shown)


Documentation/Examples.md
```diff
@@ -146,7 +146,7 @@ self.usernameOutlet.rx.text
         let loadingValue = Availability.pending(message: "Checking availability ...")
 
         // This will fire a server call to check if the username already exists.
-        // Its type is `Observable<ValidationResult>`
+        // Its type is `Observable<Availability>`
```
freak4pc: The fix below is correct, but this one isn't - 
I believe its better to fix this to "Its type is `Observable<Bool>`" - because the `map` instantly askes `if available` at the entry point.

Also this example is a tad weird - because instead of `map` + `switchLatest` we usually use `flatMapLatest`... might need a tiny update

dgoldman-pdx: Oh, right. Thanks.

I'll change this one to `Observable<Bool>`. But I think I'd better leave the `flatMapLatest` update for you.

freak4pc: Yup, that's 💯% fine :) 
Thanks!

PR #1858, opened 2019-01-22
Add functions to Driver that sends elements to PublishRelay

lammertw:
There are already functions to `drive` to a `BehaviorRelay`. Sinds sending them to a `PublishRelay` is pretty much the same I think it's useful to have functions for that as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Master
 
 * Adds `takeUntil(_ behavior:predicate:)`.
+* Adds functions to `Driver` that sends elements to `PublishRelay`
 
 ## [4.4.0](https://github.com/ReactiveX/RxSwift/releases/tag/4.4.0)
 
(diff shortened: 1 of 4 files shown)


freak4pc:
Hey there! Thanks for your contribution :) 

The fact these methods don't exist isn't accidental - 
It doesn't make much sense driving a Driver into a PublishRelay - as the latter has no replay, and Driver does have a Replay.

In general Driver(s) represents State and PublishRelay(s) represents Events in a way, so based on previous PRs suggesting this, these two shouldnt really be interchangeable. 

If you use Signal instead (which has no replay), it can `emit(to)` a PublishRelay - a composition which makes more sense. 

Feel free to look here: 

https://github.com/ReactiveX/RxSwift/pull/1769#issuecomment-429733057

lammertw:
Thank you for your answer. I guess as you mentioned I has to use `.asSignal()` on my `ControlEvent` instead of `asDriver()`.

Just to make sure would you say the following is the correct way or is there a better way?

```
let isLoadingRelay = PublishRelay<Bool>()
let refreshControl = self.refreshControl
refreshControl.rx
    .controlEvent(.valueChanged)
    .asSignal()
    .map { refreshControl.isRefreshing }
    .emit(to: isLoadingRelay)
    .disposed(by: disposeBag)
```

freak4pc:
Hey, on specific usage information its better to come to our slack http://slack.rxswift.org 
I think you could actually just use a ControlProperty for the UIRefreshControl, instead of using a relay. 

Since the issue this was opened for is resolved, I'll close this for now, but let me know if you have more feedback on this specific issue.

PR #1859, opened 2019-01-24
Fix warnings from the Swift 5 compiler

keith:
This makes some backwards compatible warnings fixes for the Swift 5
compiler in Swift 4.2 mode. There are 4 warnings being fixed here.

1. Functions inside public extensions don't need to be marked public.
For this case I opted to remove the public from the extension instead so
that no method that is meant to be internal in one of these extensions
accidentally becomes public API.
2. Members inside of a fileprivate extension don't need to be marked
fileprivate. Here I removed it from the member so that new members will
also be fileprivate unless decided otherwise.
3. private(set) is redundant on a private property. I assume this was
left over from some changes.
4. Removed unreachable return statements that now produce warnings

### DIFF
--- a/RxCocoa/Common/DelegateProxy.swift
+++ b/RxCocoa/Common/DelegateProxy.swift
@@ -24,7 +24,7 @@
         private var _methodInvokedForSelector = [Selector: MessageDispatcher]()
 
         /// Parent object associated with delegate proxy.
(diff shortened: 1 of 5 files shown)


keith:
Thanks!

kzaher:
No prob.

PR #1862, opened 2019-01-26
remove RxExample Dependencies singleton keeping simplicity

carlosypunto:
I remember that a long time ago we opted to use a singleton to facilitate the understanding that the dependencies (especially those of the Schedulers) could be shared. I think that this new approach does away with the singleton, making an injection of dependencies more correct, and at the same time maintain the understanding that instances are shared.

### DIFF
--- a/RxExample/RxExample/Examples/Dependencies.swift
+++ b/RxExample/RxExample/Examples/Dependencies.swift
@@ -13,29 +13,22 @@ import class Foundation.OperationQueue
 import enum Foundation.QualityOfService
 
 class Dependencies {
-
-    // *****************************************************************************************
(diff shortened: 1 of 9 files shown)


carlosypunto:
I cannot understand CI error, I've not modify RxCocoa.podspec

kzaher:
Hi @carlosypunto ,

I'm not sure this change improves anything. The resulting code is neither more testable or clear. I guess it's even more confusing since it's not clear why would anyone construct a struct without parameters and then discard it and just pick a single member. 

The primary reason why using singleton is not ideal is because it doesn't isolate the components, but since this is example app, you've proposed that we shouldn't care.

carlosypunto:
I agree

PR #1877, opened 2019-02-13
Fixing check to see if queue is on main thread.

ryu1006:


ryu1006:
Fixing check to see if queue is on main thread.

PR #1881, opened 2019-02-19
Adding isIgnoringInteractionEvents in UIApplication+Rx

vkcldhkd:
Added isIgnoringInteractionEvents in UIApplication+Rx 
can use it with indicator.rx.isAnimating :)

### DIFF
--- a/RxCocoa/iOS/UIApplication+Rx.swift
+++ b/RxCocoa/iOS/UIApplication+Rx.swift
@@ -8,17 +8,24 @@
 
 #if os(iOS)
 
-    import UIKit
-    import RxSwift
+import UIKit
+import RxSwift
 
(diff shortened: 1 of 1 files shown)


freak4pc:
Thanks for your contribution @vkcldhkd ! 

Usually we only add these kind of things to RxSwift/RxCocoa core if either:

 1. They are non-trivial to implement yourself (e.g. more than a simple Binder, etc.); or,
 2. They are very often used by developers (like UIButton)

I believe this doesn't fall in either of these categories, so I think it won't really fit in this specific repo. Maybe RxSwiftExt could benefit from it, even though it is a relatively seldom used property that's quite easy for a developer to add to their own codebase without a library.

Issue #1874, opened 2019-02-12
DispatchQueue.isMain not returning as expected


ryu1006:
**Short description of the issue**:

  During XCUITests and scrolling a longer list of cells with a `rx.contentOffset` observable set in `awakeFromNib`, the DispatchQueue.isMain is returning false, causing a rxFatalError to be thrown when it actually is running on the main thread. The logic to check if the DispatchQueue is on the main thread seems incorrect.

![image](https://user-images.githubusercontent.com/5336143/52661326-caf9fe80-2ecf-11e9-86a5-800a827bfdcb.png)

![stacktrace](https://user-images.githubusercontent.com/5336143/52661318-c59cb400-2ecf-11e9-821f-9eaef6e6a9ef.png)

**Expected outcome**:

  DispatchQueue.isMain returns true.

**What actually happens**:

  DispatchQueue.isMain returns false.

**Self contained code example that reproduces the issue**:

```
class OffersTests: XCTestCase {
    func testOfferHide() {
        let tableView = Offers.offerTableView
        while !tableView.cells.element(boundBy: 12).isHittable {
            tableView.swipeUp()
        } 
        tableView.cells.element(boundBy: 12).swipeLeft()
    }
}

class TileTableViewCell: UITableViewCell {
    @IBOutlet fileprivate var scrollView: UIScrollView!

override public func awakeFromNib() {
        super.awakeFromNib()
        self.setupObservables()
    }

 fileprivate func setupObservables() {
        self.scrollView.rx.contentOffset
            .flatMap({ [weak self] offset -> Observable<Bool> in
                guard let weakSelf = self else {
                    return Observable.empty()
                }
                let progress = offset.x
                return Observable.just(progress > 200)
            })
            .subscribe(onNext: { [weak self] enabled in
                self?.isUserInteractionEnabled = enabled
            })
            .addDisposableTo(rx.disposeBag)
    }
}
```

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

  RxSwift 4.3.1

**Platform/Environment**

  - [X] iOS

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [X] easy, 100% repro

**Xcode version**:

```
Version 10.1 (10B61)
```

:warning: Fields below are optional for general issues or in case those questions aren't related to your issue, but filling them out will increase the chances of getting your issue resolved. :warning:

**Installation method**:
  - [X] CocoaPods

[...]

PR #1882, opened 2019-02-19
Fixed DelegateProxy to check if it's running on main thread

ryu1006:
Addressing issue: https://github.com/ReactiveX/RxSwift/issues/1874

### DIFF
--- a/Platform/DispatchQueue+Extensions.swift
+++ b/Platform/DispatchQueue+Extensions.swift
@@ -7,10 +7,15 @@
 //
 
 import Dispatch
-import Foundation
 
 extension DispatchQueue {
+    private static var token: DispatchSpecificKey<()> = {
+        let key = DispatchSpecificKey<()>()
+        DispatchQueue.main.setSpecific(key: key, value: ())
+        return key
+    }()
+
     static var isMain: Bool {
-        return Thread.isMainThread
+        return DispatchQueue.getSpecific(key: token) != nil
     }
 }
--- a/Platform/DispatchQueue+Extensions.swift
+++ b/Platform/DispatchQueue+Extensions.swift
@@ -7,6 +7,7 @@
 //
 
 import Dispatch
+import Foundation
 
 extension DispatchQueue {
     static var isMain: Bool {
--- a/Platform/DispatchQueue+Extensions.swift
+++ b/Platform/DispatchQueue+Extensions.swift
@@ -9,13 +9,7 @@
 import Dispatch
 
 extension DispatchQueue {
-    private static var token: DispatchSpecificKey<()> = {
-        let key = DispatchSpecificKey<()>()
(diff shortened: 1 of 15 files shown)


RxSwift/Schedulers/MainScheduler.swift
```diff
@@ -44,6 +45,13 @@ public final class MainScheduler : SerialDispatchQueueScheduler {
         }
     }
 
+    /// In case this method is running on a background thread it will throw an exception.
+    public class func ensureRunningOnMainThread(errorMessage: String? = nil) {
+        if !Thread.isMainThread {
```
freak4pc: I think a guard is nicer in this case
```suggestion
        guard Thread.isMainThread else { rxFatalError ... }
```

Platform/DispatchQueue+Extensions.swift
```diff
@@ -7,15 +7,10 @@
...
 extension DispatchQueue {
-    private static var token: DispatchSpecificKey<()> = {
-        let key = DispatchSpecificKey<()>()
-        DispatchQueue.main.setSpecific(key: key, value: ())
-        return key
-    }()
-
     static var isMain: Bool {
-        return DispatchQueue.getSpecific(key: token) != nil
+        return Thread.isMainThread
```
freak4pc: I believe this entire file should be reverted to develop?

RxSwift/Schedulers/MainScheduler.swift
ryu1006: Fixed

Platform/DispatchQueue+Extensions.swift
ryu1006: Fixed

kzaher:
Hi @ryu1006 ,

could you please take a look at Linux unit tests. I would like to release this with a patch release this weekend.

Otherwise looks ok to me.

@freak4pc Could you please smoke test this in a real app. I don't currently have access to any bigger codebase.

freak4pc:
@kzaher I'll give it a whirl and let you know.

ryu1006:
@kzaher checked and saw that isMainThread is not implemented in Linux's Foundation. Fortunately enough, majority of the instances for `ensureRunningOnMainThread` is only on files that ensure it's not running on Linux or is in Debug. Only `VirtualTimeScheduler` does not have these checks, so reverted it to the previous `ensureExecutingOnScheduler` function.

freak4pc:
I've run ~500 unit tests of a larger codebase and everything seems to run smoothly. 
I guess the main pieces we have counting on DelegateProxy is RxDataSources - and that seems to work and pass tests with no issues.

Platform/DispatchQueue+Extensions.swift
douknow: @freak4pc Just confused, can you explain why check by token is better than `Thread.isMainThread`?

Issue #1875, opened 2019-02-13
first and firstOrError operators lost


potemkinaleksey:
Hello!

At this commit: https://github.com/ReactiveX/RxSwift/commit/d4a30201a9bfa2e1d53d7f4a15f2c4a6c7b54e53 first and firstOrError operators was added. 
But then at this commit: https://github.com/ReactiveX/RxSwift/commit/95af87af0a30da05d37df7576e84175116d93004 they were lost.

It seems like it was by accident when merging, because there is still FirstSink class.

Can we get them back? Maybe i can do cherry pick and make a pull request?

kzaher:
Hi @potemkinaleksey ,

yeah, we can do that.

freak4pc:
Closing this to keep discussion running in #1886.

PR #1886, opened 2019-02-23
Return 'first' operator

potemkinaleksey:
Fixes issue #1875. Thanks @viniciussoares for original implementation. I just copied parts that were lost.


Tests/RxSwiftTests/Observable+PrimitiveSequenceTest.swift
```diff
@@ -172,6 +172,141 @@ extension ObservablePrimitiveSequenceTest {
...
+            next(300, 5)
+            ])
+        
+        let res = scheduler.start { () -> Observable<Int> in
+            let single: Single<Int> = xs.first().map { $0 ?? -1 }
+            return single.asObservable()
+        }
+        
+        XCTAssertEqual(res.events, [
+            next(250, 4),
```
freak4pc: These should all use the static method variation, e.g. 

```suggestion
            .next(250, 4),
```

potemkinaleksey: Fixed

kzaher:
@freak4pc

Looks good.

RxSwift/Traits/ObservableType+PrimitiveSequence.swift
```diff
@@ -18,6 +18,18 @@ extension ObservableType {
     public func asSingle() -> Single<E> {
         return PrimitiveSequence(raw: AsSingle(source: self.asObservable()))
     }
+    
+    /**
+     The `first` operator emits only the very first item emitted by this Observable,
+     or nil if this Observable completes without emitting anything.
```
gobetti: hi! sorry for too late comment, just checking, do you prefer returning an optional within a Single trait instead of a non-optional element as Maybe? why? thank you for taking the time! :pray:

freak4pc: Single is guaranteed to not emit a `completed` event (just `success` or `failure`) - that single value is an optional.

Maybe represents something else - the ability to have 0 or 1 elements. Meaning either `success` (1 element) or `complete` (0 elements) I don't think it really fits in this specific case. It would be much harder to consume and it doesn't relay the same meaning here IMO, because an Optional element is still an element, and not "0 elements".

gobetti: right, but we are forcing an element (nil) just to make it conform to the Single type, and the tradeoff is that we're forcing a type which was originally non-optional to be optional (or an originally optional type to become a "double optional")

also, the documentation written above explicitly says "or nil if this Observable completes without emitting anything" - isn't "complete without emitting anything" the reason why Maybe exists?

other than that, I agree that in some cases the Single with optional will be easier to use than the Maybe, just like in other cases I believe Maybe would be easier as Rx has signatures that are specific to the Maybe type and allows us to consider the "complete without emitting anything" scenario without having to unwrap a value on a so-called "success"

freak4pc: It makes sense "forcing" it to be optional IMO, given Swift's own Collection.first API

freak4pc: Consuming a Maybe.completed is much harder then a Single<E?>, if you'd ask more consumers (also personal opinion)

gobetti: I like your point comparing to Array, on the other hand streams are a bigger concept so much that the distinction between Maybe and Single has been introduced in this library. If Maybe should be replaced by Single with optional element because it's easier, then it wouldn't exist at all, imho. I still think that the behavior here expressed is exactly a Maybe, and if we want to make available a possibly easier (and the definition of easy will depend on the context, because it's also easier not having to unwrap), an alternative could be to have an extension method for Maybe called asSingle() that does just that, instead of forcing this conversion in a lower level.

freak4pc: I understand your thoughts but I tend to disagree :) I don't think we'll come to an agreement on it, and it's not my decision anyways. 

Being a purist is all good and nice but there is also context to consider. Usage of Maybe is very rare - for a reason, I guess.

gobetti: sure, it was a good discussion :) but apparently in your point of view, Maybe could be deleted. I've used it a lot in the projects I participated, I know RxJava guys do use it too, and we can't really know what's the reason why it's rare according to you. I think perhaps Traits as a whole are rare, inside RxSwift itself we find cases where Observable is used where it should have been Single, and Single is already consolidated as the way to represent network responses making it probably one of the most used Traits. Maybe doesn't have such an immediate real world association, but it does make sense and prevents having to unwrap a "success" case. What would you do if success is nil? Maybe guard and return? Not handling completed in Maybe would do the same with zero code. I won't try to convince you any further, but I'll be happy to have input on this from people who do find Maybe useful, if in this case they'd still rather the optional Single. Thanks for your patience!

gobetti: @kzaher any thoughts?

freak4pc: Please don't put words in my mouth :) I didn't say it should be deleted. 
I'm just saying that there are pieces inside RxSwift that are less used, and others that are more popular, much like operators that are lesser-used (such as window, or amb for example)

PR #1887, opened 2019-02-25
boundary-buffer

simpers:
Implemented a generic Boundary Buffer and a default Boundary Buffer with debounce on a given TimeInterval.

The latter of the above is quite common in other rx implementations so I was surprised to not find it here, but it encouraged me to implement it myself and just make a PR. We've used this since November so I thought to make a PR as it seems to be stable :)

I'll get to fixing any issues ASAP if there's any feedback.


freak4pc:
hey @simpers - thank you very much for the contribution. 
This PR has been open for a significant amount of time (10 months) with no specific community interest.

I still do appreciate the amount of work you've done here. Would you mind sharing a couple of substantial use cases for this boundary buffer? Thank you!

simpers:
Sorry, I had missed the bot comments, unfortunately.

This boundary buffer is something that exists in many other implementations of Rx, for instance rxjs. That's why I implemented this when it was missing, as we were porting some logic from some logic that used this feature, which was available in that original implementation's Rx-library, which I think was JavaScript or TypeScript.

After submitting this though I was told on Slack that there was a similar PR for the same feature that hadn't been merged for a long time either. This person told me 'good luck getting it merged', implying that it's impossible to get it merged into RxSwift main, and after a brief discussion with him I moved on.

So that's that. If you do want it merged though, then I could do as the bot wants and add some info to the changelong. What say you?

freak4pc:
Hey @simpers :) Thanks for the reply.

Those sort of comments from Slack member are unfortunate and needlessly negative IMO. Even before I was helping maintain this project I got a lot of content merged in. It needs to follow some basic rules:

1. You need to have a significant (preferably wide-spread) use-case for an operator, not just add it because its in a different implementation (for example a PR I had on min/max was rejected basically since they're quite useless for practical uses). 

2. If you don't have a specific use case but gather wide community support for a PR, that helps as well as a replacement to consider merging the code in.

3. Patience helps, but people also need to be responsible for their PRs. Sometimes pinging the library maintainers helps. I've been trying to notice more and clean up old PRs so that sort of thing isn't needed, but it's the nature of open-source libraries with a lot of.

tl;dr nothing is "impossible to get merged", just looking at `develop` you'll see dozens of commits and features merged from many users over time. They just need to be proper features for this repo, since it's a heavily used project and not every addition is necessarily worth the additional code impact. 

With that — I'd be really happy to know if you can share a use case (perhaps the one you migrated) for this operator. I'd love to take a deeper look into it, but usually if there's no specific community interest we at least want to know the code isn't gonna be dead-weight that nobody actually uses and only solves a very niche use-case (in those cases its better fitted to RxSwiftExt or otherwise)

Hope this clears things up, and again thank you so much for your contribution!

simpers:
The use-case is any time you want to listen to a sequence of events but not necessarily react to them one by one. Currently we use this to when listening to printer results in an integration step that will trigger UI changes. A task can produce many printer tasks on many printers, and they will all report their results to a service that will push an event for each printer task result. The idea was thus to listen to all kinds of results and debounce on the publisher to batch-update rather than update for each one, as it can be quite busy under certain circumstances.

This same procedure is used for instance when user input is used for a query, instead of listening for each character change, we debounce and wait for the "finished string". This last case is quite normal in Angular, where I used the original implementation of this from RxJS.

I think we also use this when batching HTTP requests and the like.

That should give you an idea when and where it is used, and it is mostly to minimise the amount of requests or view updates when not necessary.

freak4pc:

simpers:
@freak4pc hello, we're looking at the possibilities to have this branch merged so as to not have to rely on a special version of RxSwift in our app. Could you take a look at the commits and let me know if there's anything further I need to do?

I have already seen that I need to update the documentation on the `buffer(debounce:..)` version but there might be other things too.

freak4pc:
Hey @simpers - 
I don't think this had enough traction or that it would be popular enough to pull it into the framework. I'm happy to accept it into RxSwiftExt.

simpers:
Well, if I recall correctly this current PR uses private APIs so it'd have to be reimplemented without them. Could have wrong or it might have changed but that's what I recall. 

We did not at the time want to add another dependency either just for this little thing (which is mainline in the other RX implementations I've used), as we would probably have just implemented it ourselves as our own extension in that case.

But if that's the case then we're closing this.

freak4pc:
Not sure how is it mainline - I've never seen a boundary buffer in ReactiveX, RxJava or Rx.NET. Where have you seen it available?

freak4pc:
I've looked further and found precedence of others having this, so I'd like to add it. 
I'm wondering about the term "boundary". Trying to think if there's a more Swift-friendly name to this. Maybe `trigger`? Maybe it's worth keeping it as "boundary" ?

What I'll ask you to do if you have a moment. I've changed the base of this PR to `rxswift6` - do you mind rebasing on top of it? 

Thanks!

simpers:
> I've looked further and found precedence of others having this, so I'd like to add it.
> I'm wondering about the term "boundary". Trying to think if there's a more Swift-friendly name to this. Maybe `trigger`? Maybe it's worth keeping it as "boundary" ?
> 
> What I'll ask you to do if you have a moment. I've changed the base of this PR to `rxswift6` - do you mind rebasing on top of it?
> 
> Thanks!

Sorry, I missed the previous response here.

I got the terminology from here, if I recall correctly:

https://xngiser.gitbooks.io/rxjs/content/api/core/operators/buffer.html

The section called "With boundaries".

I'll rebase and fix this tomorrow.

simpers:
I have rebased the branch and updated the code and tests so it compiles now. I did get some mails regarding cancellations and failures but I am not sure they're related to my code?

Regardless, should I keep it as `boundary` or refactor it to trigger, @freak4pc?

freak4pc:
Sorry for the mess, a lot has changed. I think renaming it to `buffer(on trigger:)`, `buffer(by trigger:)` or `buffer(trigger:)` would be more informative for a consumer. I'll leave the choice to you :) 

This would need another rebase though. If you get it today or tomorrow I'll be happy to release it as part of RC1. 

Thanks!

freak4pc:
Yup, rebase onto `main` would be great. Sorry!

simpers:
Can you change the branch up there? It seems I cannot.

From Caspeco:simpers/boundary-buffer -> main

Will rename to `buffer(trigger:)` and update tests and then I'm done.

simpers:
I've updated the labels and tests and they passed before pushing. Should be ready to go from my side

freak4pc:
Hey, it doesn't actually let me change the base branch :( can you just open a new PR from the same branch to `main` ?

simpers:
Sure, I just did and it is found at #2241.

Issue #1892, opened 2019-03-06
force unwrap make crash


chaoskyme:
**Short description of the issue**:

 When I use disposeBag, but there is a crash with under code. It's show that `self._disposable!.dispose()` has a fatal error: `Fatal error: Unexpectedly found nil while unwrapping an Optional value`

**Expected outcome**:

  no crash

**What actually happens**:

  crash

**Self contained code example that reproduces the issue**:
The code in the RxSwift/ScheduledDisposable.swift
```swift
func disposeInner() {
    if !isFlagSet(&self._isDisposed, 1) {
        self._disposable!.dispose()
        self._disposable = nil
    }
}
```

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

  4.4.1

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [ ] easy, 100% repro
  - [x] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
 10.1
```

:warning: Fields below are optional for general issues or in case those questions aren't related to your issue, but filling them out will increase the chances of getting your issue resolved. :warning:

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [ ] yes (which ones)
  - [x] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [ ] I have a small code base
  - [x] I have a significant code base

freak4pc:
Can you provide your code where you have your dispose bag and subscription or it’s someyhing proprietary ?

chaoskyme:
Hi, sorry. My code's calling function is complex. When I use RxSwiftCommunity's RxASDataSources and update to the latest version 4.4.1, crash here.

kirillsh:
Have same issue, reproducing is easy:
`
let disposable = ScheduledDisposable(scheduler: MainScheduler.instance, disposable: Disposables.create())
disposable.dispose()
disposable.dispose()
`
[...]

PR #1894, opened 2019-03-06
Fix multiple dispose() on SheduledDisposable causing crash (#1892)

kirillsh:
Changed isFlagSet to fetchOr function call because of previously made typo in https://github.com/ReactiveX/RxSwift/commit/65efb404c80fe1b23108c8fbb57d05b615e21c93#diff-71e325b5b737d68192bbdea267e6d7e7

### DIFF
--- a/RxSwift/Disposables/ScheduledDisposable.swift
+++ b/RxSwift/Disposables/ScheduledDisposable.swift
@@ -42,7 +42,7 @@ public final class ScheduledDisposable : Cancelable {
     }
 
     func disposeInner() {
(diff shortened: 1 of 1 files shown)


freak4pc:
This seems to be fixing a breaking crash. The fix itself looks good to me but I think we should be quick to merge and push a patch out. 

Thoughts? @kzaher

kzaher:
@freak4pc yep, looks correct to me.

kzaher:
I'll do a patch this weekend if there are no other issues.

PR #1895, opened 2019-03-06
Helper methods for UIApplication and UIViewController

lsn21:
Helper methods for UIApplication and UIViewController

### DIFF
--- a/RxCocoa/iOS/UIApplication+Rx.swift
+++ b/RxCocoa/iOS/UIApplication+Rx.swift
@@ -20,5 +20,38 @@
             }
         }
     }
+
+    extension Reactive where Base: UIApplication {
+        public static var willResignActive: ControlEvent<()> {
+            let source = NotificationCenter.default.rx.notification(UIApplication.willResignActiveNotification).map { _ in return () }
+            
+            return ControlEvent(events: source)
+        }
+        
+        public static var didEnterBackground: ControlEvent<()> {
+            let source = NotificationCenter.default.rx.notification(UIApplication.didEnterBackgroundNotification).map { _ in return () }
+            
+            return ControlEvent(events: source)
+        }
+        
+        public static var willEnterForeground: ControlEvent<()> {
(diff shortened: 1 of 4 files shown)


RxCocoa/iOS/UIApplication+Rx.swift
```diff
@@ -20,5 +20,38 @@
             }
         }
     }
+
+    extension Reactive where Base: UIApplication {
```
kzaher: Could you please add comments for these extensions and add tests.

RxCocoa/iOS/UIViewController+Rx.swift
```diff
@@ -21,4 +21,54 @@
         }
     
     }
+
+    extension Reactive where Base: UIViewController {
```
kzaher: We can't add this set of extensions unfortunately.

jegnux: @kzaher out of curiosity, why so? I have a subset of these extensions in my project, are they unsafe to use?

gobetti: I don't know if that's @kzaher's reasoning, but in a previous project we had a crash report library that applied swizzling to UIViewController methods, then using Rx to check for method invocation would produce crashes. I'd say then that this usage should be project-specific instead of being part of the library, as it won't be safe for everybody to use

teameh: We're using similar extensions in our project. Do you know in which cases it might be unsafe?

freak4pc:
@lsn21 seems to me you are not still working on this PR but some of these extensions are good. 

Is anyone interested in pursuing this? 

Also, @kzaher - what was the issue with the extensions at the bottom ?

lordzsolt:
> Is anyone interested in pursuing this?

I don't mind picking up this PR. 
I can just add `UIApplication+Rx.swift` until @kzaher clarifies `UIViewController+Rx.swift`.

freak4pc:
@lordzsolt sounds good. I'd appreciate it :)

freak4pc:

PR #1897, opened 2019-03-12
Update for Swift 5 compiler and Xcode 10.2

alanzeino:
The only part I'm not sure about is the `hasElements(_:)` functions in `Zip+arity.swift`. The Swift compiler is now smarter at knowing that `rxFatalError()` returns `Never` and now those `return false` statements are redundant as they are never called, _unless_ the user comments out the `fatalError()` inside `rxFatalError()`. 

I think that overall this pattern is odd, and these `hasElements(_:)` functions should just throw an error if they didn't work, leaving it up to the call site to decide how to handle the error, rather than using a `fatalError()` buried into the library.

Happy to make changes on how you think the above should be handled.

All these changes are likely required for the GM of 10.2.

### DIFF
--- a/RxSwift/Observables/Zip+arity.swift
+++ b/RxSwift/Observables/Zip+arity.swift
@@ -71,7 +71,9 @@ final class ZipSink2_<E1, E2, O: ObserverType> : ZipSink<O> {
             rxFatalError("Unhandled case (Function)")
         }
 
+        #if !swift(>=4.2.2)
         return false
+        #endif
     }
 
     func run() -> Disposable {
@@ -183,7 +185,9 @@ final class ZipSink3_<E1, E2, E3, O: ObserverType> : ZipSink<O> {
             rxFatalError("Unhandled case (Function)")
         }
 
+        #if !swift(>=4.2.2)
         return false
+        #endif
     }
 
     func run() -> Disposable {
@@ -303,7 +307,9 @@ final class ZipSink4_<E1, E2, E3, E4, O: ObserverType> : ZipSink<O> {
             rxFatalError("Unhandled case (Function)")
         }
 
+        #if !swift(>=4.2.2)
         return false
+        #endif
     }
 
     func run() -> Disposable {
@@ -431,7 +437,9 @@ final class ZipSink5_<E1, E2, E3, E4, E5, O: ObserverType> : ZipSink<O> {
             rxFatalError("Unhandled case (Function)")
         }
 
+        #if !swift(>=4.2.2)
         return false
+        #endif
     }
 
     func run() -> Disposable {
@@ -567,7 +575,9 @@ final class ZipSink6_<E1, E2, E3, E4, E5, E6, O: ObserverType> : ZipSink<O> {
             rxFatalError("Unhandled case (Function)")
         }
 
+        #if !swift(>=4.2.2)
         return false
+        #endif
     }
 
     func run() -> Disposable {
@@ -711,7 +721,9 @@ final class ZipSink7_<E1, E2, E3, E4, E5, E6, E7, O: ObserverType> : ZipSink<O> 
             rxFatalError("Unhandled case (Function)")
         }
 
+        #if !swift(>=4.2.2)
         return false
+        #endif
     }
 
     func run() -> Disposable {
@@ -863,7 +875,9 @@ final class ZipSink8_<E1, E2, E3, E4, E5, E6, E7, E8, O: ObserverType> : ZipSink
             rxFatalError("Unhandled case (Function)")
         }
 
+        #if !swift(>=4.2.2)
         return false
+        #endif
     }
 
     func run() -> Disposable {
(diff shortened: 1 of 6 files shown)


RxSwift/Observables/Zip+arity.swift
```diff
@@ -71,7 +71,9 @@ final class ZipSink2_<E1, E2, O: ObserverType> : ZipSink<O> {
             rxFatalError("Unhandled case (Function)")
         }
 
+        #if compiler(<5.0)
```
kzaher: `Zip+arity.swift` is generated from `Zip+arity.tt`. There is a Preprocessor in `Preprocessor/Preprocessor.xcodeproj` inside of this repo :)

RxSwift/Rx.swift
```diff
@@ -39,6 +39,7 @@ func rxAbstractMethod(file: StaticString = #file, line: UInt = #line) -> Swift.N
 
 func rxFatalError(_ lastMessage: @autoclosure () -> String, file: StaticString = #file, line: UInt = #line) -> Swift.Never  {
     // The temptation to comment this line is great, but please don't, it's for your own good. The choice is yours.
+    // If you comment it out, you must also modify functions that call this and make sure they return a value.
```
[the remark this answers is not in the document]
alanzeino: Yep I agree with that. Will remove the comments.

alanzeino:
Any chance we can update the oldest supported Swift version to 4.1 or 4.2, and conversely the Xcode version used for CI to 10.x?

Right now my change can't support the `#if compiler()` directive added in [SE-0212](https://github.com/apple/swift-evolution/blob/master/proposals/0212-compiler-version-directive.md). It's not a big deal, but it means we'll have to change the code to do this instead:

```
    #if !swift(>=4.2.2)
    var hashValue: Int {
        return rawValue.hashValue
    }
    #else
    func hash(into hasher: inout Hasher) {
        hasher.combine(rawValue)
    }
    #endif
```

Having the compiler directive makes it a little cleaner than having to use the upcoming 4.2 version shipping with Xcode 10.2 (`4.2.2`).

Also, do you mind if I either update the documentation on how to run the preprocessor app or maybe commit this scheme to make it easier:

Screen Shot 2019-03-14 at 2 16 37 PM

kzaher:
> Any chance we can update the oldest supported Swift version to 4.1 or 4.2, and conversely the Xcode version used for CI to 10.x?

The reason why we are currently supporting the ancient version of Xcode on CI is because that's the lowest version we support at the moment, so I wouldn't want to break this for now. When we'll release the 5.0 we'll change it to lowest Xcode 10.

> Right now my change can't support the #if compiler() directive added in SE-0212. It's not a big deal, but it means we'll have to change the code to do this instead:

Let's use this for now then, like you've said, shouldn't be a big deal.
```swift
    #if !swift(>=4.2.2)
    var hashValue: Int {
        return rawValue.hashValue
    }
    #else
    func hash(into hasher: inout Hasher) {
        hasher.combine(rawValue)
    }
    #endif
```

> Also, do you mind if I either update the documentation on how to run the preprocessor app or maybe commit this scheme to make it easier:

Any improvement is welcome.

RxTest/Subscription.swift
```diff
@@ -38,9 +38,16 @@ extension Subscription
     : Hashable
     , Equatable {
     /// The hash value.
-    public var hashValue: Int {
+    #if !swift(>=4.2.2)
+    public var hashValue : Int {
```
freak4pc:
```suggestion
    public var hashValue: Int {
```

nitpick

freak4pc:
@alanzeino LGTM. Just rebase over develop and we should be good to go.

alanzeino:
I think @keith did most of these in bc3021591346940988c6fefe9c9dcc6ae372020e and 8fc9f7492f4ed187ec279f49426c28cb47b4e7b3. I'll update this PR if there's anything left to do.

alanzeino:
I updated it to be based off `origin/develop`.

freak4pc:
@alanzeino By the way, about the documentation update suggestion for Pre-processor, are you still interested in doing this? It can also be a separate PR if you prefer.

alanzeino:
> @alanzeino By the way, about the documentation update suggestion for Pre-processor, are you still interested in doing this? It can also be a separate PR if you prefer.

Yep I'll do this in a separate commit.

alanzeino:
One more thing, I'm not sure that the `#if !swift(>=4.2.2)` is working. In Xcode 10.2, I think '4.2 mode' reports as `4.2.0` and in 10.1, '4.2 mode' reports as `4.2.1`. When I put up the diff, I used 10.1 to figure out what the 4.2 version was. Still investigating.

tarunon:
Maybe... `#if swift(>=4.2) & compiler(>=5.0)` instead?

freak4pc:
Can’t we just go >4.2?

keith:
Here's a follow up to this for RxBlocking https://github.com/ReactiveX/RxSwift/pull/1903

PR #1902, opened 2019-03-24
Added setDelegate(_:) in UISearchBar

TaeJoongYoon:
I'm big fan of RxSwift.

I used UISearchBar and UISearchBarDelegate but there's nothing RxSwift setDelegate method.

It is not important but I hope it helps others :)


kzaher:
Hi,

can you please add some tests for this like for 
```
func testSetDelegateUsesWeakReference() {
        let scrollView = UIScrollView()

        var delegateDeallocated = false

        autoreleasepool {
            let delegate = MockScrollViewDelegate()
            _ = scrollView.rx.setDelegate(delegate)

            _ = delegate.rx.deallocated.subscribe(onNext: { _ in
                delegateDeallocated = true
            })

            XCTAssert(delegateDeallocated == false)
        }
        XCTAssert(delegateDeallocated == true)
    }
```

TaeJoongYoon:
I should have write test code about that. 

I added it!

PR #1903, opened 2019-03-24
Fix RxBlocking Swift 5 warning

keith:
This is required for Swift 5, in this case we can't really assume what
the new case will be, but for this use case it seems logic to return
instead of crashing.

### DIFF
--- a/RxBlocking/RunLoopLock.swift
+++ b/RxBlocking/RunLoopLock.swift
@@ -84,6 +84,8 @@ final class RunLoopLock {
                     return
                 case .timedOut:
                     throw RxError.timeout
+                default:
+                    return
                 }
             #endif
         }


RxBlocking/RunLoopLock.swift
```diff
@@ -84,6 +84,8 @@ final class RunLoopLock {
                     return
                 case .timedOut:
                     throw RxError.timeout
+                @unknown default:
```
freak4pc: Will this actually work in Swift 4? 🤔

keith: I'm not sure how far back it works, but it at least works with 4.2. We'll see how CI feels about it

keith: Looks like not in Xcode 9.3, I'll gate it some how

keith: Seems like just `default` might work here

keith:
Thanks!

PR #1905, opened 2019-03-25
Fix Swift 5 compiler #ifdefs

alanzeino:
So in #1897 I used Xcode 10.1 to figure out what the version of Swift 4.2 that shipped with 10.1 (and conversely in 10.2) was. But later I found out that Apple has changed the Swift 4.2 'version' when in 4.2 mode in the 5.0 compiler to 4.2.0, whereas it was 4.2.1 in Xcode 10.1.

Because we are supporting older versions of Xcode, we gotta make sure that the compiler #ifdef is only checked when compiling for 4.2. Therefore we mix this using `#if swift(>=4.2) && compiler(>=5.0)`.

However in some places, we gotta do something funky like this:

```
    #if swift(>=4.2)
        #if !compiler(>=5.0)
        return false
        #endif
    #else
    return false
    #endif
```

This is because we can't do `#if !swift(>=4.2) && !compiler(>=5.0)`, because these changes are compatible with 4.2 on Xcode 10.0 and 10.1, but not needed with 4.2 on 10.2 — and the first #ifdef will pass, but the second one will fail on Xcode 9.3.

### DIFF
--- a/RxTest/Subscription.swift
+++ b/RxTest/Subscription.swift
@@ -38,14 +38,14 @@ extension Subscription
     : Hashable
     , Equatable {
     /// The hash value.
-    #if !swift(>=4.2.2)
-    public var hashValue: Int {
-        return self.subscribe.hashValue ^ self.unsubscribe.hashValue
-    }
-    #else
+    #if swift(>=4.2)
     public func hash(into hasher: inout Hasher) {
         hasher.combine(self.subscribe)
         hasher.combine(self.unsubscribe)
+    }
+    #else
+    public var hashValue: Int {
+        return self.subscribe.hashValue ^ self.unsubscribe.hashValue
     }
     #endif
 }
--- a/Platform/DataStructures/Bag.swift
+++ b/Platform/DataStructures/Bag.swift
@@ -171,13 +171,13 @@ extension Bag {
 }
 
 extension BagKey: Hashable {
-    #if !swift(>=4.2.2)
-    var hashValue: Int {
-        return rawValue.hashValue
-    }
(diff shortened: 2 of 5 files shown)


freak4pc:
Sheeesh Apple. Was is that hard to make normal ranged compilation expressions ? 😂

RxTest/Subscription.swift
```diff
@@ -38,15 +38,15 @@ extension Subscription
     : Hashable
     , Equatable {
     /// The hash value.
-    #if !swift(>=4.2.2)
-    public var hashValue: Int {
-        return self.subscribe.hashValue ^ self.unsubscribe.hashValue
-    }
-    #else
+    #if swift(>=4.2) & compiler(>=5.0)
```
kzaher: &&?

kzaher:
Looks like `compiler(>=5.0)` won't work with 9.3.

tarunon:
Yeah, so I guess it should be next expression.
```swift
#if swift(>=4.2)
  #if compiler(>=5.0)
    // write Xcode10.2 process
  #else
    // write Xcode 10 process
  #endif
#else
  // write Xcode 9 process
#endif
```

It's anoying🤷‍♂️

freak4pc:
Not entirely sure the build issue is real. Retriggering CI.

keith:
@kzaher do you think you can make a Swift 5 compatible release soon?

kzaher:
@keith If you are asking can I cut a new build with these 5.0 changes, yes, I can release it this weekend  if everything goes well.

PR #1908, opened 2019-04-01
Add `firstOrError` operator to ObservableType

PetarMarijanovic:
The `firstOrError` operator emits only the very first item emitted by a specific Observable or throws a `RxError.noElements` if the Observable is empty.

http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Observable.html#firstOrError--

### DIFF
--- a/RxSwift/Observables/First.swift
+++ b/RxSwift/Observables/First.swift
@@ -10,6 +10,13 @@ fileprivate final class FirstSink<Element, O: ObserverType> : Sink<O>, ObserverT
     typealias E = Element
     typealias Parent = First<E>
 
+    private let _parent: Parent
+
+    init(parent: Parent, observer: O, cancel: Cancelable) {
+        _parent = parent
+        super.init(observer: observer, cancel: cancel)
+    }
+
     func on(_ event: Event<E>) {
         switch event {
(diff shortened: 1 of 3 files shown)


kzaher:
Hi @PetarMarijanovic ,

I think that `first` should be enough. Take a look at the return type. It's different then Java. It returns optional. Our goals was to have a single type safe operator which is consistent with Swift standard library and which replaces a lot of different operators from RxJava.

I can understand why you might need something like this, but I don't think it's vital.

After `first` you will need to have either some mappable operator or subscribe. In both cases you can easily handle the error in type safe way. Compiler errors will guide you.

petarmarijanovicfive:
Hi :)

I can see your point and can agree that it is not a vital part of RxSwift. I'll create an extension function which will be enough for my project.

Thank you for your time and I hope you are well :)

AlessandroMulloni:
I tend to disagree on this. You are turning the Single to a Maybe, which are semantically different. The lack of firstOrError() in RxSwift results in a lot of error-handling code that could be spared, because I need to handle both the "nil" case on success and the "error" case on error. In most cases that results in duplicate code for handling the error in two places.

PR #1913, opened 2019-04-03
Update Package.swift to support Swift 5

devxoul:
As upgrading to Swift 5, without having `platforms` the default deployment target is set to the latest version of each OS. It causes a compile-time minimum deployment target error.

* Add `platforms`([SE-0236](https://github.com/apple/swift-evolution/blob/master/proposals/0236-package-manager-platform-deployment-settings.md))
* Add `.v5` to `swiftLanguageVersions`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.2
+// swift-tools-version:5.0
 
 import PackageDescription
 
@@ -42,6 +42,9 @@ extension Target {
 
 let package = Package(
   name: "RxSwift",
+  platforms: [
+    .macOS(.v10_10), .iOS(.v8), .tvOS(.v9), .watchOS(.v3)
+  ],
   products: ([
     [
       .library(name: "RxAtomic", targets: ["RxAtomic"]),
@@ -64,5 +67,6 @@ let package = Package(
       .target(name: "RxTest", dependencies: ["RxSwift"]),
     ],
     Target.allTests()
-  ] as [[Target]]).flatMap { $0 }
+  ] as [[Target]]).flatMap { $0 },
+  swiftLanguageVersions: [.v4_2, .v5]
 )


devxoul:
@freak4pc, thanks for review. I've added `.v4_2` to `swiftLanguageVersions`. For CocoaPods, `swift_versions` is still in beta so we shouldn't use it for now. Any suggestion?

kzaher:
>  I've added .v4_2 to swiftLanguageVersions

Hm, don't see it.

devxoul:
@kzaher, I forgot to force push the branch 😅

Package.swift
```diff
@@ -42,6 +42,9 @@ extension Target {
 
 let package = Package(
   name: "RxSwift",
+  platforms: [
+    .macOS(.v10_10), .iOS(.v8), .tvOS(.v9), .watchOS(.v3)
```
freak4pc: Not sure how this works but shouldn't we have Linux as well? Or is that by default?

devxoul: I thought so but I couldn't find the linux platform definition.

https://github.com/apple/swift-package-manager/blob/b060e8e46ea1d8ae15c909405bb6ff7e7251a8cb/Sources/PackageDescription4/SupportedPlatforms.swift#L28-L90

devxoul:
Hmm I'm not sure how to configure `.travis.yml` file 🤔

kzaher:
> Hmm I'm not sure how to configure .travis.yml file 🤔

Yeah, that's the tricky one. We would need to make our tests more generic.

@freak4pc I can do that if you want, just don't modify travis for your PR.

freak4pc:
Sure, no problems. Feel free to make the changes and I’ll amend my PR.

kzaher:
Hi @devxoul ,

I'm not sure how this would work. We need to maintain backwards compatibility. Do we need to use 5.0 in the definition? Isn't `// swift-tools-version:4.0` compatible with client in 5.0?

kzaher:
@devxoul We have changed package a bit. Could you please provide some command that is failing for you?
Can you please try the `develop` branch?

freak4pc:
> Isn't // swift-tools-version:4.0 compatible with client in 5.0?

4.2 should be compatible with the Swift 5 tooling and compiler. I'm not sure 4.0 would be 🤔

kzaher:
I think we need to release 5.0 just for Xcode >= 10.2. This is hopeless.

freak4pc:
@kzaher I've made some test on `citest` branch, it looks better when just dropping Xcode 9 while still keeping Swift 4 (hopefully). Crossing fingers. 

https://travis-ci.org/ReactiveX/RxSwift/builds/516565086

devxoul:
It's like SDK version and minimum deployment target. As you can support iOS 9 with the latest iOS SDK 12.2, you can still use Swift 4.2 with Xcode 10.2 (Swift toolchain 5). I don't think it works with Xcode version < 10.2.

freak4pc:
How did apps worked without this definition up until now? I'm just wondering if it's worth making such a drastic change for SPM alone.

kzaher:
@devxoul we'll release 5.0 and clean up everything. Otherwise we'll go insane :) At least I will.

freak4pc:
This was done in #1920

PR #1915, opened 2019-04-04
RunLoopMode to RunLoop.mode

freak4pc:
Trying to build RxSwift on Ubuntu 18.04 fails due to this:

image

### DIFF
--- a/RxBlocking/RunLoopLock.swift
+++ b/RxBlocking/RunLoopLock.swift
@@ -12,7 +12,12 @@ import RxSwift
 
 #if os(Linux)
     import Foundation
-    let runLoopMode: RunLoopMode = RunLoopMode.defaultRunLoopMode
+    #if swift(>=4.2)
+    let runLoopMode: RunLoop.Mode = .default
+    #else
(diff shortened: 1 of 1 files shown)


freak4pc:
I've also fixed the CI to run Unix build on Xcode 10.2 @kzaher 

It'd be great to run everything on 10 and 10.2 on Swift 4.2 and Swift 5 modes but with the fact we're supporting Xcode 9.4 that would be a super-lengthy CI process.

.travis.yml
```diff
@@ -42,6 +42,10 @@ matrix:
         os: osx
         env: TEST=iOS
 
+      - osx_image: xcode10.2
+        os: osx
+        env: TEST=Unix
+
```
kzaher: @freak4pc We are already running tests on Xcode 9.3, 9.4,10.0 and 10.2.

This shouldn't be needed.

https://github.com/ReactiveX/RxSwift/blob/master/.travis.yml#L41-L43

kzaher: You can fix travis and add tests for Linux and 5.0. This is the test we don't have. Adding this won't add Linux tests.

PR #1918, opened 2019-04-06
CI: Drop support for Xcode 9. Test Swift 4, 4.2 and 5.

freak4pc:
This lets CocoaPods and Carthage builds still work for Swift 4/4.2 and 5.
The only downside is Swift 5 - there's no way to support `platforms` without only supporting Swift 5 in SPM. I think it's fine in general as SPM is in such an early stage.

If you'd want, we can bump a major version, as you originally suggested, but this way keeps 4.x and still supports all Swift versions, aside for in SPM. 

WDYT @kzaher ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 language: generic
 sudo: required
 dist: xenial
-osx_image: xcode9.3
+osx_image: xcode10
 
 env:
   - DANGER=1
@@ -25,17 +25,13 @@ matrix:
         os: linux
         env: TEST=Unix swift=5.0
 
-      - osx_image:
-        os: linux
-        env: TEST=Unix swift=4.2
-        
-      - osx_image: xcode9.4
+      - osx_image: xcode10.2
         os: osx
-        env: TEST=iOS
+        env: TEST=iOS swift=4.2
 
       - osx_image: xcode10.2
         os: osx
-        env: TEST=iOS
+        env: TEST=iOS swift=5.0
 
       - osx_image: xcode10
         os: osx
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ os:
 language: generic
 sudo: required
 dist: xenial
-osx_image: xcode9.3
+osx_image: xcode10
```
kzaher: Why did you change this?

kzaher:
When I say clean slate I mean only Swift >= 5.0 and Xcode 10.2.

freak4pc:
> Are you saying we should ditch Xcode < 10? 

Yes, that's exactly what I'm saying. 

>  If somebody doesn't want to update their software, they can just use 4.x indefinitely.

I'm OK with that.

> When I say clean slate I mean only Swift >= 5.0 and Xcode >= 10.2.

There's no need to support only 5.0. Xcode 10.2 supports Swift 4.2 as well.

What we need to support is only Xcode 10.2 and the Swift 5 **compiler**.

Makes sense?

kzaher:
> There's no need to support only 5.0. Xcode 10.2 supports Swift 4.2 as well.
> What we need to support is only Xcode 10.2 and the Swift 5 compiler.
> Makes sense?

But why would we support Swift 4.x? What do we gain? If you compile RxSwift with Swift 5 you can link it with other libraries compiled with 4.x.

Am I missing something?

freak4pc:
Uhm that's a fair point - I actually haven't considered it. 
But basically it means that users of Xcode 10.1 would still be able to consume the framework on most Dependency Managers. It's not a necessity, it seems.

kzaher:
> But basically it means that users of Xcode 10.1 would still be able to consume the framework on most Dependency Managers. It's not a necessity, it seems.

I would just ignore anything less then 10.2, since that's the latest version, and clean everything. It's a crappy version of Xcode, but we don't have a newer one :)

freak4pc:
I'm OK with that. Do you want me to do the cleanup ?

kzaher:
> I'm OK with that. Do you want me to do the cleanup ?

If you can, that would be great.

freak4pc:
Sure. As a sidenote, you have a few RxAtomic leftovers. Removed those in #1919.

PR #1919, opened 2019-04-06
Remove RxAtomic leftovers

freak4pc:
Hey @kzaher. It seems there were a couple of leftovers when removing RxAtomic from the project, wanted to clean this up.


freak4pc:
Uhm, are we still using RxAtomic on Linux?

kzaher:
> Uhm, are we still using RxAtomic on Linux?

We shouldn't.

Tests/RxSwiftTests/Atomic+Overrides.swift
```diff
@@ -1,5 +1,5 @@
 //
```
kzaher: You can delete this entire file and all depending on it.

freak4pc: It's used in tests - should I just directly use the implementation ?

PR #1920, opened 2019-04-06
RxSwift 5.0.0

freak4pc:
@kzaher As discussed, this PR bumps to project to RxSwift 5.0.0:

- Remove all conditional compilation.
- Update projects to Swift 5 where needed.
- Cleanup Travis. Use Xcode 10.2.
- Lock Podspecs etc to Swift 5. 
- General cleanup.

I'm wondering if we should release this as `5.0.0-beta.1`, up to you. 

Let me know if you have any thoughts on this.


freak4pc:
I think we might want to alter the README to point to RxSwift 4.x as well, just noting this here to not forget.

.travis.yml
```diff
@@ -15,61 +15,16 @@ env:
...
+
 matrix:
     exclude:
       - os: osx
         env: DANGER=1
 
-    include:
-      - osx_image:
-        os: linux
-        env: TEST=Unix swift=5.0
```
kzaher: I don't see any Linux target. We should have at least:
```
    include:	
      - osx_image:	
        os: linux	
        env: TEST=Unix swift=5.0
```

as far as I can tell.

RxExample/RxDataSources/Differentiator/Diff.swift
```diff
@@ -126,14 +126,15 @@ public enum Diff {
...
+        func hash(into hasher: inout Hasher) {
+            hasher.combine(identity.pointee)
+        }
 
-        let hashValue: Int
         let identity: UnsafePointer<E>
 
         init(_ identity: UnsafePointer<E>) {
             self.identity = identity
-            self.hashValue = identity.pointee.hashValue
```
kzaher: This was an optimization to memoize `hashValue`. `let hashValue: Int` should be saved regardless and used in ` func hash(into hasher: inout Hasher) {`. This is just a copy of RxDataSources, but still ...

freak4pc: I'm gonna give it a different name to avoid the Hashable warning.

devxoul:
Awesome work! How about creating a new branch named `rxswift-5.0` so people can target the specific branch to test RxSwift 5?

freak4pc:
@devxoul You can direct to `develop` for now. In a week or so `master` will be RxSwift 5.

devxoul:
@freak4pc I thought it would be more comfortable to target `rxswift-5.0` branch than `develop`. But it just works with the `develop` branch so nevermind!

PR #1923, opened 2019-04-09
Return Single trait for toArray() method

lucas34:


RxSwift/Observables/ToArray.swift
```diff
@@ -10,17 +10,17 @@
...
     - seealso: [toArray operator on reactivex.io](http://reactivex.io/documentation/operators/to.html)
     
-    - returns: An observable sequence containing all the emitted elements as array.
+    - returns: A Single sequence containing all the emitted elements as array.
     */
     public func toArray()
-        -> Observable<[E]> {
-        return ToArray(source: self.asObservable())
+        -> Single<[E]> {
+        return ToArray(source: self.asObservable()).asSingle()
```
freak4pc: Basically I think you can change this one to:

```suggestion
        return PrimitiveSequence(raw: self.asObservable())
```

And it'd be good to go

RxSwift/Observables/ToArray.swift
```diff
@@ -10,17 +10,17 @@
...
     - seealso: [toArray operator on reactivex.io](http://reactivex.io/documentation/operators/to.html)
     
-    - returns: An observable sequence containing all the emitted elements as array.
+    - returns: A Single sequence containing all the emitted elements as array.
     */
     public func toArray()
-        -> Observable<[E]> {
-        return ToArray(source: self.asObservable())
+        -> Single<[E]> {
+        return PrimitiveSequence(raw: self.asObservable())
```
freak4pc: That's not what I wrote :) This should be 

```suggestion
        return PrimitiveSequence(raw: ToArray(source: self.asObservable()))
```

freak4pc: Oh I did write that, my mistake :O

lucas34:
@freak4pc Thanks for the suggestion ! All checks passed.

Issue #1921, opened 2019-04-08
👷‍♂️👷‍♀️RxSwift 5 Pre-Release discussion


freak4pc:
Hey @kzaher & others - 

Since RxSwift 5 is around the corner, I took the liberty of making a list of PRs/Issues that were delayed for RxSwift 5.

The goal here is to see if:

* Any of them are easy enough to get into RxSwift 5; or,
* Any are hard to add but are needed for RxSwift 5 (and worth delaying the release for).

Since this is a major version bump we should take some time to think about any breaking changes we might want to introduce. 

This is what I gathered so far: 

* [x] doAfter* overloads to match RxJava. #1898 @freak4pc 
* [x] Move Relays to their own framework. #1924 @freak4pc 
* [x] Allow binding to multiple observers. #1702 @freak4pc 
* [x] Using `DispatchTimeInterval` instead of `TimeInterval` for Schedulers. #1472 @kzaher 
* [x] `combineLatest` of an empty array completes immediately. #1879 
* [x] Change `toArray` to return `Single`. #1923 @freak4pc 
* [x] Complete once any Completable completes. #1674~
* [x] Decide how to deal with soft-deprecated things (`Variable`, global `RxTest` things) @freak4pc #19223
* [x] Add `compactMap` #1925. @hmlongco
* [x] Fix `take` amount vs `take` interval ambiguity. @freak4pc (blocked by #1472)
* [x] Add `resultsSelector` missing closure labels for some overloads of combineLatest & zip.
* [x] Deprecate `Completable.merge` in favor of `Completable.zip`. #1929 #1931 
* [x] Carthage Static Library Support. #1940

Generic Renames Checklist:
* [x] E and ElementType to Element, TraitType to Trait. #1945
* [x] R to Result #1948 
* [x] C to Collection #1949
* [x] S to Sequence #1949
* [x] CompatibleType to ReactiveBase #1947 
* [x] S to SharingStrategy #1951
* [x] S to Subject #1950
* [x] `SubjectType.SubjectObserverType` to `SubjectType.Observer` #1950
* [x] O to Observer #1952 
* [x] O to Observable

If there are any other PRs/Issues you feel should participate in this discussion, please add a comment below.

freak4pc:
@kzaher the purpose is mainly to see which of these we want to do for RxSwift 5, as long as we're making breaking changes here.

valerianb:
[...]

PR #1924, opened 2019-04-09
Move Relays to separate framework - RxRelay

### LINKED ISSUE #1501: Moving *Relay into their own Framework.
**No deprecation warning in Xcode on using Variable**:

As per [release notes ](https://github.com/ReactiveX/RxSwift/releases/tag/4.0.0-rc.0) provided here, `Variable` is deprecated in the favor of `BehaviorRelay`. But I dont see any deprecation warning on using Variable in Xcode

**Expected outcome**:
Xcode should alert the deprecation of Variable.

**What actually happens**:
I am using `RxSwift4`. My pod specs looks like 

```
pod 'RxSwift',    '~> 4.0'
pod 'RxCocoa',    '~> 4.0'
```
My project is making use of `Swift4`. When I declare a variable 

`var emyFilter = Variable<[MyFilterModel]>([MyFilterModel(data: "{:}")])`

No warning is shown, realized the deprecation of `Variable` only when somebody specified it in their answer on [SO](https://stackoverflow.com/questions/47449056/how-to-get-rid-of-initial-value-of-variable-in-rxswift/47450618#47450618
)

Just realized that I cant even access `BehaviorRelay` with RxSwift (4.0.0) in pod repo. Am I missing anything here? Is `BehaviorRelay` available in RxSwift4.0 and what should be our course of action? Should we still stick with `Variable` or `BehaviorRelay`? There are loads of content on web regarding `Variable` but dont see much on `BehaviorRelay` yet, so little confused of its usage as well. As I cant access it in my Xcode, cant even access the source code

```
Using RxCocoa (4.0.0)
Using RxSwift (4.0.0)
```

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  Version 9.1 (9B55)
```

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

  - [ ] yes (which ones)
  - [x] no

  - [x] just starting
  - [ ] I have a significant code base

freak4pc:
As discussed in #1921, this PR moves Relays to a separate framework, making RxSwift structure as follows:

```none
┌──────────────┐    ┌──────────────┐
│   RxCocoa    ├────▶   RxRelay    │
└───────┬──────┘    └──────┬───────┘
        │                  │        
┌───────▼──────────────────▼───────┐
│             RxSwift              │
└───────▲──────────────────▲───────┘
        │                  │        
┌───────┴──────┐    ┌──────┴───────┐
│    RxTest    │    │  RxBlocking  │
└──────────────┘    └──────────────┘
```

A few things to note:

- RxRelay depends only on RxSwift which was the initial goal - so people can only use Relays with RxSwift only, without depending on RxCocoa.

- RxCocoa depends on RxSwift **and** RxRelay, which makes sense. It also does an `@_exported import RxRelay` to retain existing behavior and not force people into importing `RxRelay` manually.

- I've added information to the README about this new structure, and reverted the "revert" you did earlier - my point with this is to keep it in `develop` and not merge it to `master` until we're ready to release RxSwift 5. Let me know if this makes sense to you.

Let me know if you have any questions.

Fixes #1501.


kzaher:
@freak4pc I wonder should we also extract `SharedSequence` from `RxCocoa`. I think it would make sense.

freak4pc:
@kzaher I was thinking about this, but how would it work and what good would it do? 

`SharedSequence` is basically useless without its concrete types as of today (e.g. `Driver`, `Signal`), right? 

I think nothing uses `SharedSequence` aside for this. What problem would it solve?

DevAndArtist:
```
┌────────────────┐   ┌────────────────┐   ┌────────────────┐
│    RxTrait     ├───▶    RxCocoa     ◀───┤    RxRelay     │
└────────▲───────┘   └────────▲───────┘   └────────▲───────┘
         │                    │                    │        
         │                    │                    │        
┌────────┴────────────────────┴────────────────────┴───────┐
│                         RxSwift                          │
└─────────────────────────────┬────────────────────┬───────┘
                              │                    │        
                              │                    │        
                     ┌────────▼───────┐   ┌────────▼───────┐
                     │     RxTest     │   │   RxBlocking   │
                     └────────────────┘   └────────────────┘
```

I think it might be good idea to move all traits into something like `RxTrait`. I updated the diagram a little (the arrows represent the 'import into' direction). In `RxTrait` we could further extend more traits such as traits that do not require the main queue.

freak4pc:
I don't really see a reasoning to separate this. 

The point of separating Relay to a separate framework was to allow people to use RxRelay with RxSwift alone, without RxCocoa. 

Traits can still live in their respective frameworks. Kruno mentioned moving SharedSequence to a framework. That would allow perhaps having more SharedSequences that are not bound to the MainScheduler. 

I believe it would require some refactoring to allow it to work on any scheduler at runtime (e.g. with `subscribeOn`), because right now it requires a Scheduler as an associated type which means a single SharedSequence-type always works on single Scheduler (be it main, test, or other).

Issue #1764, opened 2018-10-11
Suggestion: `compactMap` operator


takasek:
We use `flatMap` to transform and filter events.

```swift
stream.flatMap { x -> Observable<Foo> in
    do {
        try x.validate()
        guard x.isOnCertainCondition else { return .empty() }
        return x.fooOrNil().map(Observable.just) ?? .empty()
    } catch let e {
        return .error(e)
    }
}

optionalStream
    .flatMap(Observable.from(optional:))
```

But flatMapping seems too much as a solution in some cases.
What do you think to introduce `compactMap` to RxSwift?
With `func compactMap<U>(transform: @escaping (E) throws -> U?) -> Observable<U>` we can replace the former code like below.

```swift
stream.compactMap { x -> Foo? in
    try x.validate()
    guard x.isOnCertainCondition else { return nil }
    return x.fooOrNil()
}

optionalStream
    .compactMap { $0 }
```

This solution seems more natural for me.
It is similar to [FilterMap](https://github.com/RxSwiftCommunity/RxSwiftExt/blob/2204a685f7761244f311161f6c35ea0406488e17/Source/RxSwift/filterMap.swift) but `compactMap` provides more swifty way.

However, I saw a discussion on `Rxswift's flatMap operator works differently as it is in Swift` #1712
@freak4pc said:

> There is no correlation between Swift's compactMap and flatMap situation to what Rx's flatMap aims to solve, since they work on entirely different data types and solve very different things.

Is it negative to conform RxSwift to Swift style?

freak4pc:
Hey @takasek - Thanks for taking the time to draft this :-)

I'm personally -1 on this suggestion for several reason: 

1. We should strive to only add useful operators that are *part of the ReactiveX.io* standard. Adding too much code (especially relatively simplistic features) just bloats the codebase, thus also the maintenance and supports the open source team needs to provides.

2. In the issue you linked two you can see the second issue which is even worst for me. The `flatMap` term is shared between what we know from Swift and `flatMap` the Rx operator. This is already confusing for some people. Adding `compactMap` into the fire would further increase that confusion in my honest opinion. 

tl;dr - I would think more on this if this operator would provide extremely significant value, but it is quite trivial to create in your own codebase if you need it. 

> Is it negative to conform RxSwift to Swift style?

[...]

PR #1925, opened 2019-04-10
Add CompactMap to RxSwift

hmlongco:
Adds compactMap operator to RxSwift ObservableType.

CompactMap allows you to map an event stream and automatically filters any optional results that may be nil.
```
class AccountInfoViewModel {
    
    var accountInformation: Observable<AccountInformation>!
    var processing: Observable<Bool>!
    var error: Observable<String>!

    init(load: Observable<Void>) {
        let loading = load
            .flatMapLatest { 
                AccountInformation.load()
                    .materialize()
            }
            .share()

        accountInformation = loading
            .compactMap { $0.element }
            .share()

        error = loading
            .compactMap { $0.error?.localizedDescription }

        processing = Observable
            .merge(load.map{_ in true}, loading.map{_ in false})
    }

}
```
In the above example, the accountInformation assignment is functionally equivalent to:
```
        accountInformation = loading
            .map { $0.element }  // element may be nil
            .filter { $0 != nil }
            .map { $0! }
            .share()
```
Or creating your own extension:
```
extension ObservableType {
    func compactMap<R>(_ transform: @escaping (Self.E) throws -> R?) -> RxSwift.Observable<R> {
        return self.map { try? transform($0) }.filter { $0 != nil }.map { $0! }
    }
}
```
CompactMap also provides builtin functionality similar to the unwrap extension.
```
optionalStringObservable
    .compactMap { $0 }
    .bind(to: requiredStringObserver)
```
Optionals are a major part of the Swift language, and compactMap makes manipulating optional streams of data much, much easier.

### DIFF
--- a/RxSwift/Observables/CompactMap.swift
+++ b/RxSwift/Observables/CompactMap.swift
@@ -3,7 +3,7 @@
 //  RxSwift
 //
 //  Created by Michael Long on 04/09/2019.
-//
+//  Copyright © 2015 Krunoslav Zaher. All rights reserved.
 //
 
 extension ObservableType {
--- a/RxSwift/Observables/CompactMap.swift
+++ b/RxSwift/Observables/CompactMap.swift
@@ -4,6 +4,7 @@
 //
 //  Created by Michael Long on 04/09/2019.
 //
+//
 
 extension ObservableType {
 
--- a/RxSwift/Observables/CompactMap.swift
+++ b/RxSwift/Observables/CompactMap.swift
@@ -1,0 +1,81 @@
+//
+//  CompactMap.swift
+//  RxSwift
+//
+//  Created by Michael Long on 04/09/2019.
+//
+
+extension ObservableType {
+
+    /**
+     Projects each element of an observable sequence into an optional form and filters all optional results.
+
+     Equivalent to:
+
+     func compactMap<R>(_ transform: @escaping (Self.E) throws -> R?) -> RxSwift.Observable<R> {
+        return self.map { try? transform($0) }.filter { $0 != nil }.map { $0! }
+     }
+
+     - parameter transform: A transform function to apply to each source element and which returns an element or nil.
+     - returns: An observable sequence whose elements are the result of filtering the transform function for each element of the source.
+
+     */
+    public func compactMap<R>(_ transform: @escaping (E) throws -> R?)
+        -> Observable<R> {
+            return CompactMap(source: self.asObservable(), transform: transform)
+    }
+}
+
+final private class CompactMapSink<SourceType, O: ObserverType>: Sink<O>, ObserverType {
+    typealias Transform = (SourceType) throws -> ResultType?
+
+    typealias ResultType = O.E
+    typealias Element = SourceType
+
+    private let _transform: Transform
+
+    init(transform: @escaping Transform, observer: O, cancel: Cancelable) {
+        self._transform = transform
+        super.init(observer: observer, cancel: cancel)
+    }
+
+    func on(_ event: Event<SourceType>) {
+        switch event {
+        case .next(let element):
+            do {
+                if let mappedElement = try self._transform(element) {
+                    self.forwardOn(.next(mappedElement))
+                }
+            }
+            catch let e {
+                self.forwardOn(.error(e))
+                self.dispose()
+            }
+        case .error(let error):
+            self.forwardOn(.error(error))
+            self.dispose()
+        case .completed:
+            self.forwardOn(.completed)
+            self.dispose()
+        }
+    }
+}
+
+final private class CompactMap<SourceType, ResultType>: Producer<ResultType> {
+    typealias Transform = (SourceType) throws -> ResultType?
+
+    private let _source: Observable<SourceType>
+
+    private let _transform: Transform
+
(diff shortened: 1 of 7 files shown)


SlashDevSlashGnoll:
The `unwrap` operator in the community extensions already does this:
https://github.com/RxSwiftCommunity/RxSwiftExt#unwrap

I'm not sure what the general design goals/policies are on what goes into the base package vs the community extensions but since this is entirely a Swift-specific operator vs something that would be useful in other language bindings, I'm not sure it belongs here rather than in the extensions repo.

hmlongco:
@SlashDevSlashGnoll The general rationales for adding this were discussed in [issue 1764](https://github.com/ReactiveX/RxSwift/issues/1764). Note takasek's comments in particular.

> Referring to a proposal examined to introduce compactMap to Swift standard library, you can see same functionality in other languages:
> ```
> Haskell, Idris
> mapMaybe :: (a -> Maybe b) -> [a] -> [b]
> Ocaml (Core and Batteries)
> filter_map : 'a t -> f:('a -> 'b option) -> 'b t
> F#
> List.choose : ('T -> 'U option) -> 'T list -> 'U list
> Rust
> fn filter_map<B, F>(self, f: F) -> FilterMap<Self, F> where F: FnMut(Self::Item) -> Option<B>
> Scala
> def collect[B](pf: PartialFunction[A, B]): List[B]
> ```
> It may indicate that compactMap is a general and useful idea for sequence across platforms.

And as mentioned, optionals are a major and distinct part of the Swift language. CompactMap makes manipulating optional streams of data much, much easier. In addition to the above, the concept doesn't pertain purely to Swift, as the same operator could be used to deal with nullable data in Java.

Unwrap is available in RxSwiftExt, but it doesn't allow you to map the result as does compactMap and it itself is a composite operator combining filter and map.

As shown above, executing compactMap's equivalent (and common) functionality in RxSwift requires three distinct operations (map, filter, map) with all of the corresponding overhead of creating three observable objects, three sinks, and the corresponding disposables linking the sinks and subscriptions together.

Here it's a single operator with a single corresponding sink with a rather significant improvement in performance and memory footprint.

freak4pc:
If we have compactMap, we should have a default overload where the closure just does `{ $0 }`.

This is basically what the community knows as `unwrap`, but `compactMap` is really the "Swift" name to express this. 

@hmlongco Generally though, `compactMap { $0 }` is sort of a filter as well in that sense.

Tests/RxSwiftTests/Observable+CompactMapTests.swift
```diff
@@ -0,0 +1,191 @@
...
+            .next(230, 3),
+            .next(270, 4),
+            .next(340, 5),
+            .next(380, 6),
+            .next(390, 7),
+            .next(450, 8),
+            .next(470, 9),
+            .next(560, 10),
+            .next(580, 11),
+            .completed(600)
```
freak4pc: Can we add a for an Optional stream, e.g. the "common" usecase of `compactMap`

RxSwift/Observables/CompactMap.swift
```diff
@@ -0,0 +1,82 @@
...
+
+     func compactMap<R>(_ transform: @escaping (Self.E) throws -> R?) -> RxSwift.Observable<R> {
+        return self.map { try? transform($0) }.filter { $0 != nil }.map { $0! }
+     }
+
+     - parameter transform: A transform function to apply to each source element and which returns an element or nil.
+     - returns: An observable sequence whose elements are the result of filtering the transform function for each element of the source.
+
+     */
+    public func compactMap<R>(_ transform: @escaping (E) throws -> R?)
```
freak4pc: I think this would be better off as 

```suggestion
    public func compactMap<R>(_ transform: @escaping (E) throws -> R? = { e in return e })
```

WDYT?

freak4pc: (or `{ $0 }` if that works for default)

hmlongco: Compiler barfs at that one. Have to add...
```
    public func compactMap<R>()
        -> Observable<R> where Self.E == R {
            return CompactMap(source: self.asObservable(), transform: { $0 })
    }
```

freak4pc: I'm OK with having an extra overload, and that would also solve the test concern I mentioned.

LucianoPAlmeida: @freak4pc  :))
Just an opinion here, since this will allow to call it like `compactMap()`, I think for the naming, it will be confusing since is not a map anymore(because it doesn't have explicity the transform map function T -> U). Maybe in this case is more clear to define another function with a clearer naming like `compactElements()` or simply `compact()`. Not sure, maybe it makes sense :))

freak4pc: I'm OK with either. TBH, even having `compactMap` only with a closure is also fine and aligns with what Swift does.

tonyarnold: Did a variant of this method get added that handles `.compactMap { $0 }`?

freak4pc:
Accidentally resolved ... no, there’s no overload that defaults the closure
to that

PR #1929, opened 2019-04-14
Add Completable.zip (alias for Completable.merge)

freak4pc:
@kzaher We've discussed this before in a different thread and I bumped into this as well, when someone asked how to do this in a Slack group.

I was actually wondering if we should just entirely remove `merge` (or deprecate it) and leave only `zip`, as long as we're doing breaking changes. Since `zip` better describes the behavior, and `merge` is about merging of elements, which are not in the context of Completable.

WDYT ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 * Xcode 10.2 is the minimum supported version (or Swift 5 on Linux).
 * Changes the return type of `ObservableType.toArray` to `Single`. 
 * Adds `compactMap`.
+* Adds `Completable.zip` (alias of `Completable.merge`).
 
 If you're using Xcode 10.1 and below, please use [RxSwift 4.5](https://github.com/ReactiveX/RxSwift/releases/tag/4.5.0).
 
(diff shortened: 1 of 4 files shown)


kzaher:
I don't think we can do this yet. We first need to change `zip` operator.

freak4pc:
How so? It seems the current behavior represents `zip` accurately - e.g. "pairs" of completion (given there are no elements, just completion). 

`merge` really doesn't accurately describe the behavior IMO.

kzaher:
The answer is here f962ddd5269c98ca35ca8a16b37f1d65dffe7c11. I've fixed it now.

freak4pc:
Oh, that's very interesting. Thanks !

freak4pc:
@kzaher One last though on this topic - I still think it might be worth to entirely deprecate `Completable.merge` and leave only `Completable.zip` - merge doesn't properly describe what this does IMO. What do you think? Worth doing, or better to leave as is?

kzaher:
>  I still think it might be worth to entirely deprecate Completable.merge

@freak4pc Let's deprecate it.

PR #1932, opened 2019-04-15
RxExample now uses DispatchTimeInterval

freak4pc:
RxExample was still using regular TimeInterval(s).

If we change the label as discussed in #1921, I'll change this from: 

```swift
 .throttle(.milliseconds(500), scheduler: MainScheduler.instance)
```

to either
```swift
 .throttle(for: .milliseconds(500), scheduler: MainScheduler.instance)
```

or 
```swift
 .throttle(duration: .milliseconds(500), scheduler: MainScheduler.instance)
```


kzaher:
`throttle(for` sounds like the throttling will be performed for the first X period.
`throttle(duration`, could also be read as throttle for the first X period.

I like the version without parameter because if somebody knows what throttle does, he doesn't need additional info, and if he doesn't nothing can't create a wrong impression.

Let's not do a hasty rename for temporal operators, doesn't sound like an improvement to me.

freak4pc:
Makes sense. Thanks !

PR #1938, opened 2019-04-18
Replace instance of RxCocoa with RxRelay

sberrevoets:
Found this in some of the documentation.

As a side question: is a new release with RxRelay planned?

### DIFF
--- a/Documentation/Subjects.md
+++ b/Documentation/Subjects.md
@@ -6,7 +6,7 @@ All of behave exactly the same like described [here](http://reactivex.io/documen
 Relays
 ======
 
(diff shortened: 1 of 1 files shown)


freak4pc:
Hey Scott, I think I already did this in develop :) 
I think we’re trying to see if we can support static libraries for Carthage, but aside from that RxSwift 5 should be released very soon.

sberrevoets:
Oops, this was a bit impulsive, didn't check the branching strategy for the project first 😅 Thanks for the update!

freak4pc:
No worries! Thank you anyways for taking the time :)

Issue #1799, opened 2018-11-12
Add support for static frameworks in Carthage


mfcollins3:
**Short description of the issue**:

RxSwift currently includes only dynamic frameworks when building with Carthage. As of Carthage 0.30.0, Carthage is capable of building static frameworks without the use of external tools. This can be achieved by duplicating the current set of RxSwift targets and changing the Mach-O Type field to Static Library instead of Dynamic Framework. The targets still build as static frameworks, however. The two sets of targets can co-exist and are output to different directories at build time (for example `Carthage/Build/iOS` for the dynamic framework and `Carthage/Build/iOS/Static` for the static framework).

It's my preference to use the static frameworks in my iOS applications instead of dynamic frameworks in order to improve launch times by reducing the number of frameworks that I am consuming that require dynamic linking.

I currently fork RxSwift and implement this on my fork. I'd be happy to contribute it back via pull request if the interest is there to add it to RxSwift.

**Expected outcome**:

Carthage builds produce static frameworks alongside the dynamic frameworks.

**What actually happens**:

Carthage is currently only producing dynamic frameworks for RxSwift when running `carthage update`.

**Platform/Environment**

  - [x] iOS
  - [x] macOS
  - [x] tvOS
  - [x] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  Xcode 10.0
```

:warning: Fields below are optional for general issues or in case those questions aren't related to your issue, but filling them out will increase the chances of getting your issue resolved. :warning:

**Installation method**:
  - [ ] CocoaPods
  - [x] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [ ] yes (which ones)
  - [x] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [ ] I have a small code base
  - [x] I have a significant code base

kzaher:
Hi @mfcollins3 ,

I guess we can try to add it. It guess it depends on the PR will we be able to accept it. I'm fine with adding that to the main repo.

Feel free to make a PR and we can see how to proceed.

[...]

PR #1933, opened 2019-04-15
Update CHANGELOG for RxSwift 5

freak4pc:
This should align the CHANGELOG to everything worked on for RxSwift 5 and discussed in #1921.

PR #1940, opened 2019-04-18
Support Static Libraries for Carthage

freak4pc:
Following the discussion in #1799, this seems to work now. The only downside here is that the "validate-playground" script fails, and I'm not sure if there's an easy way to fix it here. Any suggestions welcomed.

cc/ @bobgodwinx @mfcollins3

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,8 @@ $ pod install
 
 ### [Carthage](https://github.com/Carthage/Carthage)
 
+Officially supported: Carthage 0.33 and up.
+
 Add this to `Cartfile`
 
 ```
--- a/Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Combining_Operators.xcplaygroundpage/Contents.swift
@@ -7,7 +7,7 @@
  ----
  [Previous](@previous) - [Table of Contents](Table_of_Contents)
  */
-import RxSwift
+import RxPlaygrounds
 /*:
 # Combination Operators
 Operators that combine multiple source `Observable`s into a single `Observable`.
--- a/Rx.playground/Pages/Connectable_Operators.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Connectable_Operators.xcplaygroundpage/Contents.swift
@@ -7,7 +7,7 @@
  ----
  [Previous](@previous) - [Table of Contents](Table_of_Contents)
  */
-import RxSwift
+import RxPlaygrounds
 
 playgroundShouldContinueIndefinitely()
 /*:
--- a/Rx.playground/Pages/Creating_and_Subscribing_to_Observables.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Creating_and_Subscribing_to_Observables.xcplaygroundpage/Contents.swift
@@ -7,7 +7,7 @@
  ----
  [Previous](@previous) - [Table of Contents](Table_of_Contents)
  */
-import RxSwift
+import RxPlaygrounds
 /*:
  # Creating and Subscribing to `Observable`s
  There are several ways to create and subscribe to `Observable` sequences.
--- a/Rx.playground/Pages/Debugging_Operators.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Debugging_Operators.xcplaygroundpage/Contents.swift
@@ -7,7 +7,7 @@
  ----
  [Previous](@previous) - [Table of Contents](Table_of_Contents)
  */
-import RxSwift
+import RxPlaygrounds
 /*:
  # Debugging Operators
  Operators to help debug Rx code.
--- a/Rx.playground/Pages/Error_Handling_Operators.xcplaygroundpage/Contents.swift
+++ b/Rx.playground/Pages/Error_Handling_Operators.xcplaygroundpage/Contents.swift
@@ -7,7 +7,7 @@
  ----
  [Previous](@previous) - [Table of Contents](Table_of_Contents)
  */
-import RxSwift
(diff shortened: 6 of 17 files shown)


mfcollins3:
@freak4pc It looks like the `swift` command is trying to load `RxSwift` as if it's a dynamic framework and is failing. It's interpreting the code and trying to use dyld to dynamically make the calls to the framework's methods. I switched to using `swiftc` to build the program and link with the RxSwift framework, then run the program and that works:

```bash
. scripts/common.sh

PLAYGROUND_CONFIGURATIONS=(Release)

# make sure macOS builds
for scheme in "RxSwift"
do
  for configuration in ${PLAYGROUND_CONFIGURATIONS[@]}
  do
    PAGES_PATH=${BUILD_DIRECTORY}/Build/Products/${configuration}/all-playground-pages.swift
    rx ${scheme} ${configuration} "" build
    cat Rx.playground/Sources/*.swift Rx.playground/Pages/**/*.swift > ${PAGES_PATH}
    swiftc -v -D NOT_IN_PLAYGROUND -target x86_64-apple-macosx10.10 -F ${BUILD_DIRECTORY}/Build/Products/${configuration} -framework RxSwift ${PAGES_PATH}   
    ./all-playground-pages
  done
done
```

I added the `-framework RxSwift` to the command line for `swiftc`. I then added the call to execute the `all-playground-pages` program. I'm assuming that you're just looking to make sure that the playgrounds output without failing, correct?

kzaher:
@freak4pc I think that besides making out tests pass we should figure out how to make playgrounds work with these static libraries.

I think that we can add a dynamic framework called `Playgrounds` in RxExample which imports and reexports all static frameworks and use that for playgrounds.

I'm thinking of putting it to RxExample because otherwise Carthage will build it.

freak4pc:
@kzaher Absolutely. Actually @mfcollins3's solution seems to work locally for me, which is awesome! I didn't think it would be that simple to fix. Let's wait to see if CI agrees with this :)

freak4pc:
@kzaher this seems good to go. 

I think maybe we should update the README to mention we support Static Libraries with Carthage 0.30.1 and up.

freak4pc:
I've updated the README. 
I added a CHANGELOG entry in the mani CHANGELOG PR - #1933

kzaher:
Hi @freak4pc ,

It looks to me like playgrounds are still broken.

freak4pc:
@kzaher Seems to be running just fine ... what am I missing ? 

https://imgur.com/hKwRg7R

kzaher:
@freak4pc The purpose of playground tests is not for the CI to pass, but to make sure that we didn't break playgrounds accidentally :)

The fact that CI passes is great, but users aren't able to use them because playgrounds don't use the command that you are using.

Screen Shot 2019-04-20 at 12 13 40 PM

freak4pc:
Ah, I missed that. I tried playing with this a bit in https://github.com/ReactiveX/RxSwift/pull/1940/commits/cea92be78be41939744a3eec9007d70c65687127 but couldn't get it to work with a wrapper dynamic framework. 

I won't have much time to dig into this in the following days, so if anyone can chime in and has any thoughts, that'd be great.

kzaher:
For some reason I also couldn't get it to work with a dynamic framework, but that's the only plausible idea I have.

☹️

kzaher:
I can't even get the example app to work. I think that if we get RxExample to work, playgrounds will probably work :)

`-all_load` is useless as far as I can tell. Xcode 10.2.1 (10E1001)

@mfcollins3 Any ideas?

This is extremely annoying.

freak4pc:
I was actually able to get the example project itself to work. Already heading out today so will try to share tomorrow.

mfcollins3:
@kzaher I did a fresh clone of the repo and checked out the `static lib` branch. In RxExample-iOS, I removed the embedded frameworks (which also removed the linked frameworks). I added back RxSwift and RxCocoa to the linked frameworks (they don't need to be embedded since they're now linked into the executable). I also changed the **Other Linker Flags** to `-ObjC`. I couldn't figure out what the `-objc_loadall` flag did, and once I removed it, it didn't seem to be necessary. I think that `-ObjC` is necessary when using RxCocoa because of the Objective-C code used for the delegate proxies. I didn't need the `-all_load` flag. This worked for RxExample-OSX as well.

For RxPlaygrounds, I just added RxSwift, RxCocoa, and RxRelay to the linked libraries list on the **General** tab. Once I did that, they playgrounds worked. I saw that `-all_load` and `-ObjC` were already specified in the **Other Linker Flags** build setting.

After that, it looks like you just need to update `scripts/validate-playgrounds.sh` to build `RxPlaygrounds` before attempting to build/run the playgrounds program. I'm not the best bash person, but here's the modified script that I got to work:

```bash
. scripts/common.sh

PLAYGROUND_CONFIGURATIONS=(Release)

# make sure macOS builds
for configuration in ${PLAYGROUND_CONFIGURATIONS[@]}
do
  for scheme in "RxSwift"
  do
    rx ${scheme} ${configuration} "" build
  done

  rx RxPlaygrounds ${configuration} "" build
  PAGES_PATH=${BUILD_DIRECTORY}/Build/Products/${configuration}/all-playground-pages.swift
  cat Rx.playground/Sources/*.swift Rx.playground/Pages/**/*.swift > ${PAGES_PATH}
  swiftc -v -D NOT_IN_PLAYGROUND -target x86_64-apple-macosx10.10 -F ${BUILD_DIRECTORY}/Build/Products/${configuration} -framework RxSwift ${PAGES_PATH}   
  ./all-playground-pages
done
```

I was trying to get `RxPlaygrounds` to work in the `for scheme in...` loop, but it wouldn't work for me for some reason.

I put these in a PR for the `staticlib` branch. See #1942.

PR #1943, opened 2019-04-21
Update Geolocation.swift

SURYAKANTSHARMA:
dd case for whenInUse, as we required the only whenInUse  location in our app to its better that our demo for geolocation subscription should work when user give the whenInUse permission. Currently its is showing error in case user given the whenInUse permission.
Thanks

### DIFF
--- a/RxExample/RxExample/Services/GeolocationService.swift
+++ b/RxExample/RxExample/Services/GeolocationService.swift
@@ -37,6 +37,8 @@ class GeolocationService {
                 switch $0 {
                 case .authorizedAlways:
                     return true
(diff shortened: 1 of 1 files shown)


freak4pc:
Hey :) please target `develop` instead of master.

@kzaher I think something is wrong with CI - Danger left no warning on incorrect branch.

kzaher:
@freak4pc Have no idea.

SURYAKANTSHARMA:
@freak4pc Thanks 👍  . Will make sure for future to target develop branch if found any correction.

PR #1945, opened 2019-04-22
Align associated types

freak4pc:
This was a discussion started initially by @devandartist. 
Basically we’re using random associated generic naming everywhere and this PR aims to change everything to Element. It will assist in future generic features that will hopefully be shipped with later Swift versions. 

This PR also provides a deprecation, keeping code backwards-compatible.

Thanks !

### DIFF
--- a/RxCocoa/Traits/SharedSequence/SharedSequence.swift
+++ b/RxCocoa/Traits/SharedSequence/SharedSequence.swift
@@ -20,7 +20,6 @@ import RxSwift
     To find out more about units and how to use them, please visit `Documentation/Traits.md`.
 */
 public struct SharedSequence<S: SharingStrategyProtocol, Element> : SharedSequenceConvertibleType {
-    public typealias Element = Element
     public typealias SharingStrategy = S
 
     let _source: Observable<Element>
--- a/RxCocoa/Traits/SharedSequence/SharedSequence.swift
+++ b/RxCocoa/Traits/SharedSequence/SharedSequence.swift
@@ -20,42 +20,42 @@ import RxSwift
     To find out more about units and how to use them, please visit `Documentation/Traits.md`.
 */
 public struct SharedSequence<S: SharingStrategyProtocol, Element> : SharedSequenceConvertibleType {
-    public typealias E = Element
+    public typealias Element = Element
     public typealias SharingStrategy = S
 
-    let _source: Observable<E>
+    let _source: Observable<Element>
 
-    init(_ source: Observable<E>) {
+    init(_ source: Observable<Element>) {
         self._source = S.share(source)
     }
 
-    init(raw: Observable<E>) {
+    init(raw: Observable<Element>) {
         self._source = raw
     }
 
-    #if EXPANDABLE_SHARED_SEQUENCE
+    #if EXPANDABLE_SHARED_SEQUENC
     /**
      This method is extension hook in case this unit needs to extended from outside the library.
      
      By defining `EXPANDABLE_SHARED_SEQUENCE` one agrees that it's up to him to ensure shared sequence
      properties are preserved after extension.
     */
-    public static func createUnsafe<O: ObservableType>(source: O) -> SharedSequence<S, O.E> {
-        return SharedSequence<S, O.E>(raw: source.asObservable())
+    public static func createUnsafe<O: ObservableType>(source: O) -> SharedSequence<S, O.Element> {
+        return SharedSequence<S, O.Element>(raw: source.asObservable())
     }
     #endif
 
     /**
     - returns: Built observable sequence.
     */
-    public func asObservable() -> Observable<E> {
+    public func asObservable() -> Observable<Element> {
         return self._source
     }
 
     /**
     - returns: `self`
     */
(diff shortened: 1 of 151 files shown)


DevAndArtist:
@kzaher I also would like to know if we can clean up other associated or generic types. Renaming local generic types like `C` and `O` should be a no brainer as they don't break any code and better names would provide more semantic description. Furthermore it would be great to change `*Type` suffixed names to align more with stdlib name convention. For example it does not make sense a protocol to require an associated type `ElementType` when the conforming type is known to have `Element` where the implementation then just adds `typealias ElementType = Element`. We can soft deprecate old names the same way this PR already does it and remove the deprecated type aliases in RxSwift 6.

kzaher:
Hi @DevAndArtist ,

>  I also would like to know if we can clean up other associated or generic types.

Sure

> For example it does not make sense a protocol to require an associated type ElementType when the conforming type is known to have Element where the implementation then just adds typealias ElementType = Element

I'm not sure how to solve this in a better way. We have protocol associated values that we want to name the same as generic argument. Maybe I'm overlooking something obvious.

RxCocoa/Traits/SharedSequence/SharedSequence.swift
```diff
@@ -20,42 +20,41 @@ import RxSwift
...
         self._source = S.share(source)
     }
 
-    init(raw: Observable<E>) {
+    init(raw: Observable<Element>) {
         self._source = raw
     }
 
-    #if EXPANDABLE_SHARED_SEQUENCE
+    #if EXPANDABLE_SHARED_SEQUENCElement$
```
kzaher: This seems wrong.

DevAndArtist:
> We have protocol associated values that we want to name the same as generic argument.

Reading this two scenarios come to my mind what you could mean.

1. Associated types having the same name as the generic type parameter.

```swift
protocol P {
  associatedtype Element
}
struct Observable<Element>: P {} // done nothing more we need to do
```

2. Naming generic types the same way as they constraining protocols.

This could be a little verbose but we still can do this and improve in the future.

```swift
func foo<Collection: Swift.Collection>(_ collection: Collection) where Collection.Element == Int
```

In the future we could potentially re-write some generic members differently and even with constraints. One pitched syntax form on SE is:

```swift
func foo(_ collection: some Collection<.Element == Int>)
```

kzaher:
> Associated types having the same name as the generic type parameter.

What the hell, I was trying to do this and I got compiler errors. Since when is this supported?

kzaher:
After this there are also some additional renames like @DevAndArtist suggested.

`R -> Result`
`C -> Collection`
`O -> Other`

DevAndArtist:
> What the hell, I was trying to do this and I got compiler errors. Since when is this supported?

Not sure what went wrong for you, but this was definitely supported for a while now. This is basically how retroactive conformance works where you can 'extract' the generic type, as currently swift does not support *qualified type lookup for generic types* (I pitched this [here](https://forums.swift.org/t/allow-member-lookup-to-find-generic-parameters/7768)).

DevAndArtist:
One example why I don't like the `*Type` suffixed name convention.

```swift
/// A type that can be converted to `Event<Element>`.
public protocol EventConvertible {
    /// Type of element in event
    associatedtype ElementType

    /// Event representation of this instance
    var event: Event<ElementType> { get }
}
```

This protocol is implemented on `Event` which has generic type called `Element`. To satisfy the protocol requirement it requires to add `typealias ElementType = Element` which is basically redundant.

I would like to take the chance and rethink **all** `*Type` suffixed names. Some of them can potentially just drop the suffix, other may be renamed. 

Renaming example:
```diff
public protocol ReactiveCompatible {
    /// Extended type
-    associatedtype CompatibleType
+   associatedtype ReactiveBase

    /// Reactive extensions.
-    static var rx: Reactive<CompatibleType>.Type { get set }
+    static var rx: Reactive<ReactiveBase>.Type { get set }

    /// Reactive extensions.
-    var rx: Reactive<CompatibleType> { get set }
+    var rx: Reactive<ReactiveBase> { get set }
}
```

RxCocoa/Traits/SharedSequence/SharedSequence.swift
freak4pc: Whoops :)

freak4pc:
@kzaher Done. 
I also renamed ElementType to Element, and TraitType to Trait. 

I can do others as well but should we do them as this PR? For example actually renaming single-letter-constraints (O, R, C) would probably be lengthier, but could also be in the scope of this PR. Up to you.

kzaher:
Hi @freak4pc ,

I appreciate your help with this, but if we are doing these kind of big refactorings please don't add a lot of other commits on top of the reviewed commits, but just address the comments first and open new PRs for additional changes.

I don't mind this with small PRs, but for big ones it's frustrating to need to go through all of the changes multiple times or going throught each of the commit one by one.

freak4pc:
Hey, no problem. I was about to suggest moving those to a separate PR but you’ve already merged :) thanks.

PR #1947, opened 2019-04-24
Renaming `CompatibleType` to `ReactiveBase`

DevAndArtist:
Semantically this associated type refers the generic type parameter `Base` from `Reactive` lens. Thus the associated type is renamed from being called `CompatibleType` to `ReactiveBase` to resemble the targeting type. The old type is deprecated to prevent any source breakage.

### DIFF
--- a/RxSwift/Reactive.swift
+++ b/RxSwift/Reactive.swift
@@ -39,7 +39,7 @@ public protocol ReactiveCompatible {
     /// Extended type
     associatedtype ReactiveBase
 
-    @available(*, deprecated, message: "Use `CompatibleType` instead.")
+    @available(*, deprecated, message: "Use `ReactiveBase` instead.")
     typealias CompatibleType = ReactiveBase
 
     /// Reactive extensions.
--- a/RxSwift/Reactive.swift
+++ b/RxSwift/Reactive.swift
@@ -37,13 +37,16 @@ public struct Reactive<Base> {
 /// A type that has reactive extensions.
 public protocol ReactiveCompatible {
     /// Extended type
-    associatedtype CompatibleType
+    associatedtype ReactiveBase
+
+    @available(*, deprecated, message: "Use `CompatibleType` instead.")
+    typealias CompatibleType = ReactiveBase
 
     /// Reactive extensions.
-    static var rx: Reactive<CompatibleType>.Type { get set }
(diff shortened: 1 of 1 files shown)


DevAndArtist:
I would argue that the precedence for this name would be `associatedtype ArrayLiteralElement` from `ExpressibleByArrayLiteral`.

freak4pc:
In this specific case I'm not sure this necessarily increases understandability - The protocol basically says "If you conform to this, you are reactive compatible, and `CompatibltType` is the actual compatible type".  

I mean, I don't necessarily care too much to rename this, but I'm not sure it really makes a difference in this specific case.

DevAndArtist:
Compared to other protocols in stdlib, there are also plenty of `*Type` suffixed associated types, I think this would make more sense as it describes the type the protocol requires the conforming type to provide. For example `StringLiteralType` is more clear that `T.StringLiteralType` describes the string literal type for `T`, while `T.CompatibleType` seems to have no semantical anchor. Therefore it either should be `T.ReactiveCompatibleType` or as proposed `T.ReactiveBase`.

freak4pc:
The protocol itself is ReactiveCompatible, so ReactiveCompatible.CompatibleType does have semantic meaning in my eyes. But if we do a rename I prefer the former option (`ReactiveCompatibleType`). Anyways again, this is basically just an internal construct so it doesn't have that much meaning. Everywhere where you extend you use `Base` (from `Reactive`). 

e.g. 
`extension Reactive where Base: Something`.

DevAndArtist:
It's not internal as you can re-route the associated type on the conforming type and ignore the default implementation. 

`ReactiveCompatible.CompatibleType` has not much meaning in Swift as referencing a nested type as a member of a protocol is not valid.

kzaher:
Hm,

I was lost at lenses. Don't think our reactive extension is a lens.

I'm not sure is `Base` more clear, but we do have 
```
public struct Reactive<Base> {
    /// Base object to extend.
    public let base: Base

    /// Creates extensions with base object.
    ///
    /// - parameter base: Base object.
    public init(_ base: Base) {
        self.base = base
    }
}
```

This PR would make protocol more consistent with the main implementation, and `CompatibleType` is not mentioned anywhere, so I'm fine merging this.

@jegnux named it. It's not a bad name I guess.

PR #1948, opened 2019-04-24
Rename R generic constraint to Result

freak4pc:


freak4pc:
I'll add a CHANGELOG once tests are done, I don't want to break the build over that :)

PR #1956, opened 2019-04-26
Trivial: Updating playground examples

LucianoPAlmeida:
Updating some playground .interval() deprecated calls.


freak4pc:
Unix failed, rerunning

PR #1959, opened 2019-04-28
Fix wrong deprecated message

mvolpato:


freak4pc:
Nice catch! thank you! :)

kzaher:
@mvolpato @freak4pc 
Interesting, I was just creating 5.0.0 tag. Should I wait if somebody is doing some additional tests?

freak4pc:
This wouldn't cause any real issues. We can wait a bit anyways. 
I suggest taking a look in the dynamic/static discussion - I think there is a question to be answered there @kzaher #1601

freak4pc:
@kzaher Was referring to this one: 
https://github.com/ReactiveX/RxSwift/issues/1799

Maybe we should make dummy branches on RxDataSources and RxSwiftExt to make sure there are no glaring issues with that release? Anyways the dynamic/static question is a difficult one to answer as well.

mvolpato:
I did not find anything else, but I am not really testing, just starting a new project using Swift 5.

kzaher:
@mvolpato That was what I was thinking of :)

@freak4pc 
> Maybe we should make dummy branches on RxDataSources and RxSwiftExt to make sure there are no glaring issues with that release?

I have a dummy branch of RxDataSources for integration tests, so I am doing this.

kzaher:
I guess if something will be wrong, we can make a patch release, I don't have any additional way of testing that I already didn't use.

PR #1962, opened 2019-05-04
drive() and emit() for multiple observers

freak4pc:
To align with `bind(to: Observer...)` (#1702) this PR adds the same abilities to `drive()` and `emit()`.  

I also cleaned up some of the implementation (removed an unneeded private method) and removed some duplicate tests for `Driver`.


kzaher:
Hi @freak4pc ,

wouldn't this change our ABI?

freak4pc:
The code is backward compatible (e.g. `drive(observer)` works alongside `driver(observer1, observer2)`) 

The method signature itself is different which might mean it's not binary stable - but I'm not sure that's an entirely huge issue as long as we retain the same execution context. 

If it's a big deal we can easily leave the old signature and relay it into the variadic one.

freak4pc:
@kzaher I think this can be a good time to add this.

pluddy:
This would clean up quite a bit of code repetition we have going on as well.  Any way we can keep the conversation going on this @kzaher @freak4pc ?

freak4pc:
This would be a good addition to RxSwift 6 which should be quite soon. ABI breakage shouldn’t be an issue in that case.

juanjo-ramos:
I just saw this https://github.com/ReactiveX/RxSwift/commit/c3c0cac3d4c176b04404e3574d62b51776277384.

Does it mean this PR won't be included in RxSwift 6?

freak4pc:
No:
1. It’s just a beta
2. It wasn’t release yet

PR #1963, opened 2019-05-05
Documentation: Fixed compile errors on 'Automatic input validation' code from Examples.md

marcosgriselli:
This PR fixes a few compile errors on the [Automatic input validation](https://github.com/ReactiveX/RxSwift/blob/master/Documentation/Examples.md#automatic-input-validation) code sample from the **Examples.md** document. 

Screen Shot 2019-05-05 at 20 26 39 _Error from on Xcode 10.2.1_

There are no errors on the RxSwift code but we can agree that it's best having error-free code blocks when possible.

Would it be useful to add the implementations of the complementary functions under a **detail** tag? It's the only code block from that document that doesn't compile but it's already long enough. 

Example: 
<details>
  <summary>Extra code</summary>
  <p>

```swift
class API {
    static func usernameAvailable(_ username: String) -> Observable<Bool> {
        return Observable.just(Bool.random())
    }
}

func validationColor(_ validity: Availability) -> UIColor {
    switch validity {
    case .taken, .invalid:
        return .red
    case .pending:
        return .orange
    case .available:
        return .green
    }
}
```
  </p>
</details>

### DIFF
--- a/Documentation/Examples.md
+++ b/Documentation/Examples.md
@@ -136,7 +136,7 @@ self.usernameOutlet.rx.text
             // Convenience for constructing synchronous result.
             // In case there is mixed synchronous and asynchronous code inside the same
             // method, this will construct an async result that is resolved immediately.
-            return Observable.just(Availability.invalid(message: "Username can't be empty."))
+            return Observable.just(.invalid(message: "Username can't be empty."))
         }
 
         // ...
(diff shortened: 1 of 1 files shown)


Documentation/Examples.md
```diff
@@ -169,8 +169,8 @@ self.usernameOutlet.rx.text
 // Good old `subscribe(onNext:)` can do that.
 // That's the end of `Observable` chain.
     .subscribe(onNext: { validity in
```
freak4pc:
```suggestion
    .subscribe(onNext: { [weak self] validity in
```

Documentation/Examples.md
```diff
@@ -153,7 +153,7 @@ self.usernameOutlet.rx.text
                   return Availability.available(message: "Username available")
```
freak4pc: If we're explicit about the return type we can just return

```suggestion
                  return .available(message: "Username available")
```

freak4pc:
> Would it be useful to add the implementations of the complementary functions under a detail tag? 

I don't think so - it's there for "theoretical" purposes, not so much for an actual "copy and paste" example.

PR #1968, opened 2019-05-09
UITextField.isSecureTextEntry binder

MortyMerr:
"Show password" button is a quite common task to implement. It can be implemented easily with `textField.isSecureTextEntry.toggle()` however, it is not a reactive way. With this binder things will become better:
```swift
showPasswordButton.rx.tap
  .scan(true) { isSecureTextEntry, _ in !isSecureTextEntry }
  .bind(to: passwordTextField.rx.isSecureTextEntry)
  .disposed(by: disposeBag)
```
Probably, i have to implement similar for `UITextView`, however i don't think it will be useful. What do you think?

Also added .idea to .gitignore for AppCode users. Can do it in a separate PR if u wish :)

### DIFF
--- a/Tests/RxCocoaTests/UITextField+RxTests.swift
+++ b/Tests/RxCocoaTests/UITextField+RxTests.swift
@@ -46,6 +46,16 @@ final class UITextFieldTests : RxTest {
             XCTAssertEqual(textField.attributedText!, attributedText)
         }
     }
+
+    func test_isSecureTextEntryObserver() {
+        // because of leak in iOS 11.2
+        if #available(iOS 11.3, tvOS 11.3, *) {
+            let textField = UITextField()
+            XCTAssertFalse(textField.isSecureTextEntry)
+            textField.rx.isSecureTextEntry.onNext(true)
+            XCTAssertTrue(textField.isSecureTextEntry)
+        }
+    }
 }
 
 private extension String {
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -51,3 +51,7 @@ Carthage/Build
 
 .build/
 Packages/
+
+# AppCode
+
+.idea/
```
taher-mosbah: I prefer having this in another PR

Tests/RxCocoaTests/UITextField+RxTests.swift
```diff
@@ -46,6 +46,16 @@ final class UITextFieldTests : RxTest {
             XCTAssertEqual(textField.attributedText!, attributedText)
         }
     }
+
+    func test_isSecureTextEntryObserver() {
+        // because of leak in iOS 11.2
```
taher-mosbah: You are referring this leak right ? maybe add a link ?
https://forums.developer.apple.com/thread/94323

.gitignore
MortyMerr: Removed

Tests/RxCocoaTests/UITextField+RxTests.swift
MortyMerr: All tests in this file have similar comment about this leak. To be honest, I don't know which exactly leak is mention, i just implement my test in a similar way

PR #1975, opened 2019-05-14
add `object` function to RxCocoa URLSession extension

gringoireDM:
This pulls request adds a new function to take advantage of Decodable in `URLSession` Rx extension


freak4pc:
Hey, thank you for the PR :) Not sure it's something that really fits RxSwift Core, and even if so, it would be better to add it on `Observable<Data>`. 

Since it's just a simple transformation (e.g. `map`) - I'm not sure its worth an entire operator in the core library. But it could possibly live inside RxSwiftExt. What do you think?

gringoireDM:
I wanted to take advantage of the already existing `RxCocoaURLError` since the decode can throw. If you think it's not a good fit it's ok for me.

Since there is a specific case for json, i thought that to take advantage of decodable would be a good idea. Let me know if you think i should close the PR

PR #1976, opened 2019-05-14
Fix warnings/deprecations

MortyMerr:
As a potential contributor, I would like to see only my warnings in CI logs. However, right now there are few deprecation messages, swiftlint warnings and type-check warnings. This PR fix them all.

### DIFF
--- a/RxCocoa/macOS/NSSlider+Rx.swift
+++ b/RxCocoa/macOS/NSSlider+Rx.swift
@@ -16,7 +16,7 @@ extension Reactive where Base: NSSlider {
     /// Reactive wrapper for `value` property.
     public var value: ControlProperty<Double> {
         return self.base.rx.controlProperty(
-            getter: { control in
+            getter: { control -> Double in
                 return control.doubleValue
             },
             setter: { control, value in
--- a/Tests/RxSwiftTests/Observable+MergeTests.swift
+++ b/Tests/RxSwiftTests/Observable+MergeTests.swift
@@ -1222,7 +1222,7 @@ extension ObservableMergeTest {
     func testFlatMapFirst_Complete() {
         let scheduler = TestScheduler(initialClock: 0)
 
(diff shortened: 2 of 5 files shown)


Tests/RxSwiftTests/Observable+MergeTests.swift
```diff
@@ -1222,7 +1222,7 @@ extension ObservableMergeTest {
     func testFlatMapFirst_Complete() {
         let scheduler = TestScheduler(initialClock: 0)
 
-        let xs = scheduler.createHotObservable([
+        let xs: TestableObservable<TestableObservable<Int>> = scheduler.createHotObservable([
```
freak4pc: Why is this needed? Seems to work without.

RxCocoa/macOS/NSSlider+Rx.swift
```diff
@@ -16,7 +16,7 @@ extension Reactive where Base: NSSlider {
     /// Reactive wrapper for `value` property.
     public var value: ControlProperty<Double> {
         return self.base.rx.controlProperty(
-            getter: { control in
+            getter: { control -> Double in
```
freak4pc: Why is this needed?

Tests/RxSwiftTests/Observable+MergeTests.swift
MortyMerr: This line cause type-check time limit warning:
```
/Users/travis/build/ReactiveX/RxSwift/Tests/RxSwiftTests/Observable+MergeTests.swift:1225:28: expression took 155ms to type-check (limit: 100ms)
```
This fix prevents it. If u think this warning is ok, i am ready to discard changes and leave only deprecation fix + swiftlint disabling

RxCocoa/macOS/NSSlider+Rx.swift
MortyMerr: Same as above

PR #1977, opened 2019-05-15
Add onDisposed to Maybe, Single and Completable

MortyMerr:
Based on https://github.com/ReactiveX/RxSwift/pull/1681 which is looks like forgotten


kzaher:
I think we should add this, but we should either create new convenience with non optional `onDisposed` or wait for 6.0 because this breaks the ABI.

MortyMerr:
@kzaher thank you for your attention :)
Why `onDisposed` should be non-optional?

kzaher:
Because we already have `subscribe(onCompleted: (() -> Void)? = nil, onError: ((Swift.Error) -> Void)? = nil)`.

MortyMerr:
Just to clarify, you mean, we have to create an additional `subscribe` method which will take `onDisposed` as non-optional parameter to prevent breaking the ABI?

RxSwift/Traits/Completable.swift
```diff
@@ -76,27 +76,46 @@ extension PrimitiveSequenceType where Trait == CompletableTrait, Element == Swif
...
         #else
                 let callStack = [String]()
         #endif
 
-        return self.primitiveSequence.subscribe { event in
+        let disposable: Disposable
+        if let disposed = onDisposed {
+            disposable = Disposables.create(with: disposed)
+        }
+        else {
```
freak4pc: Change to `} else {` please (e.g. move up a line)
```suggestion
        else {
```

RxSwift/Traits/Completable.swift
```diff
@@ -76,27 +76,46 @@ extension PrimitiveSequenceType where Trait == CompletableTrait, Element == Swif
...
+                          onDisposed: (() -> Void)? = nil) -> Disposable {
         #if DEBUG
                 let callStack = Hooks.recordCallStackOnError ? Thread.callStackSymbols : []
         #else
                 let callStack = [String]()
         #endif
 
-        return self.primitiveSequence.subscribe { event in
+        let disposable: Disposable
+        if let disposed = onDisposed {
```
freak4pc:
```suggestion
        if let onDisposed = onDisposed {
```

t4rj4n:
Hello,
could it be, that this feature never managed to get into the master branch?
I cannot find it here:
https://github.com/ReactiveX/RxSwift/blob/master/RxSwift/Traits/Completable.swift
And if im reading it here correctly, the pull request errored in to CI:
https://travis-ci.org/github/ReactiveX/RxSwift/builds/632257143

Greetings from t4rj4n

freak4pc:
I think this commit is part of the RxSwift 6 branch. I might cherry-pick some of the commits into the master branch since they can work absolutely fine on RxSwift 6.

PR #1978, opened 2019-05-16
Add compact map for shared sequences, single and maybe

gringoireDM:
This Pull request adds `compactMap` operators for `Single`, `SharedSequences`, and `Maybe`.

Since `compactMap` is now part of `ObservableType` I think that like other operators it should be included in the implementation of other Traits.


NKorchagin:
👍

RxSwift/Traits/Single.swift
```diff
@@ -213,6 +213,18 @@ extension PrimitiveSequenceType where Trait == SingleTrait {
             return Single(raw: self.primitiveSequence.source.map(transform))
     }
     
+    /**
```
kzaher: We can not add `compactMap` to `Single` because it breaks the contact.

gringoireDM: That's why it returns a `Maybe`, just like `filter` does.
Don't you think it's right?

acecilia:
Can we also add it to Driver?

NKorchagin:
> @gringoireDM Can you also add it to Driver?

It's already added via `SharedSequenceConvertibleType`

gringoireDM:
@kzaher unit tests added

RxSwift/Traits/Single.swift
kzaher: Oh, that's fine. I've misread that you are returning `Single`. My bad.

PR #1979, opened 2019-05-18
Allow to define custom traits

ajjnix:
In our project is more obviously in some cases create trait `Errorless`. I want to explicitly state that the error will not be returned. I found preposition https://github.com/ReactiveX/RxSwift/issues/1272 with reply 

> You can define your own traits if you like, including that have guarantees you want

So I tried to write `Errorless` trait but it's not possible cause traits it's a typealias on `PrimitiveSequence`. And `PrimitiveSequence` has `internal` init. So in firstly you just can't make a instance of our custom trait. And in secondly you can't define basic operations like `filter`/`map` and other because `source` in `PrimitiveSequence` also defined as internal.

I think, we should allow to create custom traits.

### DIFF
--- a/RxSwift/Traits/PrimitiveSequence.swift
+++ b/RxSwift/Traits/PrimitiveSequence.swift
@@ -8,9 +8,9 @@
 
 /// Observable sequences containing 0 or 1 element.
 public struct PrimitiveSequence<Trait, Element> {
-    let source: Observable<Element>
+    public let source: Observable<Element>
 
-    init(raw: Observable<Element>) {
+    public init(raw: Observable<Element>) {
         self.source = raw
     }
 }


kzaher:
We can't make this `public`. That would allow anyone to break the contract.

If somebody wants to extend `Single` or `Maybe` or `Completable` they can use the normal extension.

```
func myExtension() {
    return Single.create { observer in
         return self.subscribe { event in
         } 
    }
}
```

or create their own traits from scratch.

ajjnix:
@kzaher
 > If somebody wants to extend

But if I don't want to extend? I want to define **custom** trait. How can I do it?

ajjnix:
> or create their own traits from scratch.

I should create own `PrimitiveSequence` with all logics for it?

ajjnix:
For example my problem:

```swift
public enum ErrorlessTrait { }

public typealias Errorless<Element> = PrimitiveSequence< ErrorlessTrait, Element>

public enum ErrorlessEvent<Element> {
    case success(Element)
    case completed
}

extension PrimitiveSequenceType where TraitType == ErrorlessTrait {
```

So how can I make instance of `Errorless` without `public init`?

kzaher:
@ajjnix 

Yes, you would do the same we've done.

Personally I don't think it makes much sense for two reasons:
* `Single` usually represents some operation and it is extremely unlikely that you'll handle every possible case except for trivial cases. That means that you'll probably have just one big fat `.unknown` enum anyway. That is not a big improvement here.
* You won't need to do this if we introduce additional generic arguments in the base library. Then it will just be a typealias. We are investigating this, but this won't be soo soon.

PR #1988, opened 2019-05-23
Add Hooks documentation, remove unexpected logs

MortyMerr:
`Hooks` enum encapsulates quite useful logic for error handling/debugging, however, I didn't find any documentation about that. I am sure this enum could help users with debugging their apps, so my proposal is to add a description of it to `GettingStarted.md`

Also, I add check on empty array. If the callstack array is empty user will see just "subscription called from:" and empty line, which might be quite confusing for him.

### DIFF
--- a/Documentation/GettingStarted.md
+++ b/Documentation/GettingStarted.md
@@ -727,6 +727,19 @@ You can recover from failure of observable by using `catch` operator. There are 
 
 There is also `retry` operator that enables retries in case of errored sequence.
 
+### Hooks and Default error handling
+
+RxSwift offers a global Hook that provides a default error handling mechanism for cases when you don't provide your own `onError` handler.
+
+Set `Hooks.defaultErrorHandler` with your own closure to decide how to deal with unhandled errors in your system, if you need that option. For example, sending the stacktrace or untracked-error to your analytics system. 
+
+By default, `Hooks.defaultErrorHandler` simply prints the received error in `DEBUG` mode, and does nothing in `RELEASE`. However, you can add additional configurations to this behavior.
+
+In order to enable detailed callstack logging, set `Hooks.recordCallStackOnError` flag to `true`.
+
+By default, this will return the current `Thread.callStackSymbols` in `DEBUG` mode, and will track an empty stack trace in `RELEASE`. You may customize this behavior by overriding `Hooks.customCaptureSubscriptionCallstack` with your own implementation.
+
+
 ## Debugging Compile Errors
 
 When writing elegant RxSwift/RxCocoa code, you are probably relying heavily on compiler to deduce types of `Observable`s. This is one of the reasons why Swift is awesome, but it can also be frustrating sometimes.
--- a/RxSwift/ObservableType+Extensions.swift
+++ b/RxSwift/ObservableType+Extensions.swift
@@ -93,7 +93,10 @@ extension Hooks {
     fileprivate static var _defaultErrorHandler: DefaultErrorHandler = { subscriptionCallStack, error in
         #if DEBUG
             let serializedCallStack = subscriptionCallStack.joined(separator: "\n")
-            print("Unhandled error happened: \(error)\n subscription called from:\n\(serializedCallStack)")
+            print("Unhandled error happened: \(error)")
+            if !serializedCallStack.isEmpty {
+                print("subscription called from:\n\(serializedCallStack)")
+            }
         #endif
     }
     fileprivate static var _customCaptureSubscriptionCallstack: CustomCaptureSubscriptionCallstack = {


Documentation/GettingStarted.md
```diff
@@ -727,6 +727,36 @@ You can recover from failure of observable by using `catch` operator. There are
...
+        print($0)
+    })
+
+```
+
+Implement `Hooks.customCaptureSubscriptionCallstack` closure, if you need custom way of grabbing logs.
+
+Implement `Hooks.defaultErrorHandler` if you need custom reaction on unhandled errors (e.g. send stacktrace to analytics). 
+
+As error logs are quite inconspicuous, you might want raise fatal error if unhandled error occured. There is already `bind(onNext)` method for that kind of needs.
```
freak4pc: This specific section is quite opinionated and I’d prefer to remove it. usage of bind on next is reserved to the context of dealing with UI and it’s not just a way to capture errors every where.

MortyMerr: If you say that, I will remove it :)
But could u please explain to me, how potential user should understand that? There are no docs about "reserved to the context of dealing with UI". It's just a wrapper which calls rxFatalErrorInDebug in `onError` block. Why can't I use it not only for UI, if I just want to get this "fatalError" behavior?

Maybe I will add an additional note like "bind(onNext:) is reserved only for UI"? What do you think?

freak4pc: Because `subscribe` is for subscribing to a stream and `bind` is always used to create some connection. I guess it's not really "reserved for UI" but it's where you would most often use it. I don't think it's something that needs documentation, it's just an API.

freak4pc: My point was that its entirely not related to documenting Hooks so I would prefer not having it there personally.

MortyMerr: Sure, get it. Delete this section.

MortyMerr:
If there are English grammar issues, I will appreciate any help with fixing them :)
Thanks

Documentation/GettingStarted.md
```diff
@@ -727,6 +727,29 @@ You can recover from failure of observable by using `catch` operator. There are
 
 There is also `retry` operator that enables retries in case of errored sequence.
 
+### Default error handling
```
freak4pc: I rewrote this a bit, what do you think? 

### Hooks and Default error handling

RxSwift offers a global Hook that provides a default error handling mechanism for cases when you don't provide your own `onError` handler.

Set `Hooks.defaultErrorHandler` with your own closure to decide how to deal with unhandled errors in your system, if you need that option. For example, sending the stacktrace or untracked-error to your analytics system. 

By default, `Hooks.defaultErrorHandler` simply prints the received error in `DEBUG` mode, and does nothing in `RELEASE`. However, you can add additional configurations to this behavior.

In order to enable detailed callstack logging, set `Hooks.recordCallStackOnError` flag to `true`.

By default, this will return the current `Thread.callStackSymbols` in `DEBUG` mode, and will track an empty stack trace in `RELEASE`. You may customize this behavior by overriding `Hooks.customCaptureSubscriptionCallstack` with your own implementation.

```md
### Hooks and Default error handling

RxSwift offers a global Hook that provides a default error handling mechanism for cases when you don't provide your own `onError` handler.

Set `Hooks.defaultErrorHandler` with your own closure to decide how to deal with unhandled errors in your system, if you need that option. For example, sending the stacktrace or untracked-error to your analytics system. 

By default, `Hooks.defaultErrorHandler` simply prints the received error in `DEBUG` mode, and does nothing in `RELEASE`. However, you can add additional configurations to this behavior.

In order to enable detailed callstack logging, set `Hooks.recordCallStackOnError` flag to `true`.

By default, this will return the current `Thread.callStackSymbols` in `DEBUG` mode, and will track an empty stack trace in `RELEASE`. You may customize this behavior by overriding `Hooks.customCaptureSubscriptionCallstack` with your own implementation.
```

MortyMerr: I believe you know better :)
Thank you for your attention, applied all changes to PR

Documentation/GettingStarted.md
```diff
@@ -727,6 +727,31 @@ You can recover from failure of observable by using `catch` operator. There are
...
+
+RxSwift offers a global Hook that provides a default error handling mechanism for cases when you don't provide your own `onError` handler.
+
+Set `Hooks.defaultErrorHandler` with your own closure to decide how to deal with unhandled errors in your system, if you need that option. For example, sending the stacktrace or untracked-error to your analytics system. 
+
+By default, `Hooks.defaultErrorHandler` simply prints the received error in `DEBUG` mode, and does nothing in `RELEASE`. However, you can add additional configurations to this behavior.
+
+In order to enable detailed callstack logging, set `Hooks.recordCallStackOnError` flag to `true`.
+
+```swift
```
freak4pc: Hey, I don't think this needs code sample - it's just a simple flag, not sure what value is gained from showing how to set it.

MortyMerr: Fixed, thanks

PR #1989, opened 2019-05-24
Extended BehaviorRelay with KeyPath and Collection.

cristoppotelli:
Hello there, as I moved away my code base from Variable, which was extensively used by my team, I figured out a way to let my code behave as it was with variables, plus behavior relay immutability.

I have implemented some methods, which I believe can be useful to others, that "mimic" the collections functions signatures when Element is a collection (a very small set of these), for manipulating a subrange of the value, and accepting the updated one.

Also, two methods which make use of key paths to update a value's store property, prior to knowing who Element will be, thanks to generics.

Thanks in advance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [5.0.1](https://github.com/ReactiveX/RxSwift/releases/tag/5.0.1)
 
 * Reverts Carthage integration from using static to dynamic libraries. #1960
+* Extended `BehaviorRelay` with KeyPath and Collection.
 
 ## [5.0.0](https://github.com/ReactiveX/RxSwift/releases/tag/5.0.0)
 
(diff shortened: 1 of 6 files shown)


freak4pc:
Hey @cristoppotelli, 
I see you closed but just wanted to comment regardless,

I appreciate you taking the time to write the code but it would be a bad idea to add it to core since it would promote a bad practice of abusing relays for data/stateful and also abstract the fact the user is actually adding another element and not modifying the collection. 

If someone uses it that way in RxSwift he’s welcome to do so, but it’s not a practice we want to encourage with code helpers IMO.

cristoppotelli:
Hey, thanks for replying.

I did close the pull request as I noticed I had to change a subset of unit test I added, looking forward to reopen the pull request, if applicable.

These extension I wrote tackle the problem which incurs, for example, when you have an array of section models, and you might want to update just the section, as a whole, at the nth position in the array, but still accepting the whole value array to update a table view.

What, in your opinion, would be a better Rx-esque approach for those kind of situations?

Thanks.

freak4pc:
In general, Rx is about composition of streams of data, and not manually modifying a "stateful" object (like BehaviorRelay or Variable). Any code additions to this repo shouldn't support that kind of flow as it's basically against what Rx offers. 

Sections in a table change based on specific triggers, For example a combineLatest of a network requests, a user tapping on a specific cell, etc. could define what the Sections should be at any given moment, without managing any intermediate variable/property like a relay/state.

cristoppotelli:
Thanks, I appreciated your feedback.

PR #1258, opened 2017-05-17
[Feature] - Add RepeatWhen operator to Observable

sergdort:
This PR adds repeatWhen operator to observable sequence #396 

### Description of change

* Implement adopt almost all logic from the `RetryWhen` operator
* Add `RepeatWhenSequence`, `RepeatWhenSequenceSink`, `RepeatWhenSequenceSinkIter`, `RepeatTriggerSink`
* Add `repeatWhen` extension methods on `ObservableType`

### Questions

* Is it worth to include to the main repo? :) 
* I feel a little bit sorry for copy pasting almost everything from `RetryWhen`, so I'm thinking maybe there is a way to refactor so we can share logic between both of the operators ?

PR #1990, opened 2019-05-25
Add repeatWhen operator

MortyMerr:
Based on https://github.com/ReactiveX/RxSwift/pull/1258 (original author implementation, name and dates saved)
This PR has no updates since 2017, however, there is still activity in it. People asking about this operator. So, I think it will be quite useful to have it in core
I updated implementation according to new guidelines and added playground example

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -231,6 +231,7 @@ custom_categories:
   - Range
   - Reduce
   - Repeat
+  - RepeatWhen
   - RetryWhen
   - Sample
   - Scan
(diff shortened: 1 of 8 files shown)


juanxme:
Since 2017 ? 🙏

freak4pc:
I still find this to be a very niche operator. I don't really think it fits any common use case that I know of, but am happy to hear more. Definitely nice to know so much work has already been done on this, but we should be smart on adding operators that are only used by ~3-5% of the community, for example (IMO) 

I'm happy to move this into the repo if you post this PR on Slack and gather some community interest here.

MortyMerr:
If u don't mind, let's just close it

PR #1998, opened 2019-06-02
Replace custom Result types on native

MortyMerr:
Swift 5.0 introduced Result type, so there is no need in custom implementations now


RxSwift/Traits/Single.swift
```diff
@@ -15,13 +15,9 @@ public enum SingleTrait { }
...
-public enum SingleEvent<Element> {
-    /// One and only sequence element is produced. (underlying observable sequence emits: `.next(Element)`, `.completed`)
-    case success(Element)
-    
-    /// Sequence terminated with an error. (underlying observable sequence emits: `.error(Error)`)
-    case error(Swift.Error)
-}
+/// .success: One and only sequence element is produced. (underlying observable sequence emits: `.next(Element)`, `.completed`)
+/// .failure: Sequence terminated with an error. (underlying observable sequence emits: `.error(Error)`
+public typealias SingleEvent<Element> = Result<Element, Swift.Error>
```
freak4pc: The other changes make sense but I'm not sure this one necessarily does. 
Yes, the type signatures do - but they mean different things. Not too sure, I'll let @kzaher decide on this.

RxExample/RxExample/Result.swift
```diff
@@ -1,12 +0,0 @@
-//
-//  Result.swift
-//  RxExample
-//
-//  Created by Krunoslav Zaher on 3/18/17.
-//  Copyright © 2017 Krunoslav Zaher. All rights reserved.
-//
-
-enum Result<T, E: Error> {
```
freak4pc: This removal makes sense to me

kzaher:
Hm ...,

I think that we could think about doing that in the future (6.0). We can't do anything at the moment because it would break ABI compatibility and semantic versioning.

We can only remove `RxExample/RxExample/Result.swift`.

MortyMerr:
@kzaher ok thanks, i will separate that PR

RxExample/RxExample/Result.swift
MortyMerr: Moved to separate PR

RxSwift/Traits/Single.swift
MortyMerr: IMO, Result and SingleEvent have the same meaning here. I can't prove that but please just think about that a little bit more
Feel free to close that PR, if it's not suitable :)

PR #2002, opened 2019-06-07
Use WKWebView over deprecated UIWebView

iamDecode:
`UIWebView` is deprecated and will inevitably be removed at some point in the future. Apple urges developers to adopt `WKWebView` instead. This PR replaces the existing bindings for `UIWebView` with similar bindings for `WKWebView`.

This PR is required in order to support Project Catalyst builds (iOS apps on MacOS), as `UIWebView` is the cause of nasty errors:
```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_UIWebView", referenced from:
      objc-class-ref in RxWebViewDelegateProxy.o
ld: symbol(s) not found for architecture x86_64
```

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -100,7 +100,7 @@ custom_categories:
   - UITextView+Rx
   - UIView+Rx
   - UIViewController+Rx
-  - UIWebView+Rx
+  - WKWebView+Rx
 - name: RxCocoa/iOS/Protocols
   children:
   - RxCollectionViewDataSourceType
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 
 ---
+## Master
+
+* Replace bindings on deprecated UIWebView with WKWebView.
+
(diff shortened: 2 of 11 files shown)


iamDecode:
My bad, I thought all tests passed locally, but I missed the DelegateProxy tests. Seems that deallocating the `WKWebView` does not trigger `onComplete` or `rx.deallocated`. I cant seem to figure out why this is the case; I am not too familiar with the RxSwift codebase. Help appreciated :)

freak4pc:
Thanks for the contribution! This is a bit of a problem. We can't simply pull out support for an entire UI Component without bumping a major version (e.g. RxSwift 6) which we don't plan to do soon. 

I'm wondering if it's possible doing `#if canImport(UIKit)` on some of these, instead. So they are only triggered in that specific case. WDYT ?

noahsark769:
I'm not sure that `#if canImport(UIKit)` will work, since Catalyst apps can still import UIKit but don't have UIWebView. I think that surrounding all the UIWebView stuff with `#if !targetEnvironment(UIKitForMac)` might work! Not sure if this will break existing builds though, we should check to see if it compiles on Xcode 10

PR #2003, opened 2019-06-08
bind(to keypath:on root:) similar to Combine's `assign(to:on:)`

freak4pc:
I wanted to add this operator to mirror [Combine's `assign(to:on:)`](https://developer.apple.com/documentation/combine/currentvaluesubject/3235789-assign) operator.
It's a great addition because it makes `Binder`(s) not as needed for every property you want to bind/drive/emit to. 

It means that instead of having to create a custom Binder for your UI Object, you could just do 

```swift
stream
 .bind(to: \MyObject.property, on: myObjectInstance)
```

Which is much more flexible. 
Any thoughts are extremely welcomed :)

### DIFF
--- a/RxCocoa/Common/Observable+Bind.swift
+++ b/RxCocoa/Common/Observable+Bind.swift
@@ -70,7 +70,6 @@ extension ObservableType {
          return binder(self)(curriedArgument)
     }
 
-
     /**
     Subscribes an element handler to an observable sequence.
     In case error occurs in debug mode, `fatalError` will be raised.
@@ -82,6 +81,32 @@ extension ObservableType {
     public func bind(onNext: @escaping (Element) -> Void) -> Disposable {
         return self.subscribe(onNext: onNext, onError: { error in
             rxFatalErrorInDebug("Binding error: \(error)")
+        })
+    }
+
+    /**
+    Bind emitted elements to the referenced key path of a specific `Root` object.
+
+    This allows writing code as follows:
+
+    ```
+    stream.bind(to: \UITextField.text, on: myTextField)
+    ```
+
+    In case error occurs in debug mode, `fatalError` will be raised.
+    In case error occurs in release mode, `error` will be logged.
+
+    - parameter keyPath: The key path of the property to assign.
+    - parameter object: The object on which to assign the value.
+    - returns: Subscription object used to unsubscribe from the observable sequence.
+    - note: If a weak reference to the Root object cannot be retained, elements will
+            not be bound, and will simply be ignored.
+    */
+    public func bind<Root: AnyObject>(to keyPath: ReferenceWritableKeyPath<Root, Element>,
+                                      on object: Root) -> Disposable {
+        return self.bind(onNext: { [weak object] element in
+            guard let object = object else { return }
+            object[keyPath: keyPath] = element
         })
     }
 }
--- a/RxCocoa/Traits/Driver/Driver+Subscription.swift
+++ b/RxCocoa/Traits/Driver/Driver+Subscription.swift
@@ -117,6 +117,29 @@ extension SharedSequenceConvertibleType where SharingStrategy == DriverSharingSt
         MainScheduler.ensureRunningOnMainThread(errorMessage: errorMessage)
         return self.asObservable().subscribe(onNext: onNext, onCompleted: onCompleted, onDisposed: onDisposed)
     }
+
+    /**
+     Drive emitted elements to the referenced key path of a specific `Root` object.
+
+     This allows writing code as follows:
+
+     ```
+     driver.drive(\UITextField.text, on: myTextField)
+     ```
+
+     - parameter keyPath: The key path of the property to assign.
+     - parameter object: The object on which to assign the value.
+     - returns: Subscription object used to unsubscribe from the observable sequence.
+     - note: If a weak reference to the Root object cannot be retained, elements will
+     not be bound, and will simply be ignored.
+     */
+    public func drive<Root: AnyObject>(_ keyPath: ReferenceWritableKeyPath<Root, Element>,
+                                       on object: Root) -> Disposable {
+        return drive(onNext: { [weak object] element in
+            guard let object = object else { return }
+            object[keyPath: keyPath] = element
+        })
+    }
 }
 
 
--- a/RxCocoa/Traits/Signal/Signal+Subscription.swift
+++ b/RxCocoa/Traits/Signal/Signal+Subscription.swift
@@ -95,6 +95,29 @@ extension SharedSequenceConvertibleType where SharingStrategy == SignalSharingSt
     public func emit(onNext: ((Element) -> Void)? = nil, onCompleted: (() -> Void)? = nil, onDisposed: (() -> Void)? = nil) -> Disposable {
         return self.asObservable().subscribe(onNext: onNext, onCompleted: onCompleted, onDisposed: onDisposed)
     }
+
+    /**
+     Emit elements to the referenced key path of a specific `Root` object.
+
+     This allows writing code as follows:
+
+     ```
+     signal.emit(to: \UITextField.text, on: myTextField)
+     ```
+
+     - parameter keyPath: The key path of the property to assign.
+     - parameter object: The object on which to assign the value.
+     - returns: Subscription object used to unsubscribe from the observable sequence.
+     - note: If a weak reference to the Root object cannot be retained, elements will
+     not be bound, and will simply be ignored.
+     */
+    public func emit<Root: AnyObject>(to keyPath: ReferenceWritableKeyPath<Root, Element>,
+                                      on object: Root) -> Disposable {
+        return emit(onNext: { [weak object] element in
+            guard let object = object else { return }
+            object[keyPath: keyPath] = element
+        })
+    }
 }
 
 
--- a/Sources/AllTestz/Observable+BindTests.swift
+++ b/Sources/AllTestz/Observable+BindTests.swift
@@ -1,0 +1,1 @@
+../../Tests/RxCocoaTests/Observable+BindTests.swift
--- a/Sources/AllTestz/main.swift
+++ b/Sources/AllTestz/main.swift
@@ -248,6 +248,8 @@ final class DriverTest_ : DriverTest, RxTestCase {
     ("testDriveBehaviorRelay1", DriverTest.testDriveBehaviorRelay1),
     ("testDriveBehaviorRelay2", DriverTest.testDriveBehaviorRelay2),
     ("testDriveBehaviorRelay3", DriverTest.testDriveBehaviorRelay3),
+    ("testDriveKeyPath", DriverTest.testDriveKeyPath),
+    ("testDriveKeyPathObjectGone", DriverTest.testDriveKeyPathObjectGone),
     ] }
 }
 
@@ -393,6 +395,26 @@ final class ObservableAmbTest_ : ObservableAmbTest, RxTestCase {
     ("testAmb_LoserThrows", ObservableAmbTest.testAmb_LoserThrows),
     ("testAmb_ThrowsBeforeElectionLeft", ObservableAmbTest.testAmb_ThrowsBeforeElectionLeft),
     ("testAmb_ThrowsBeforeElectionRight", ObservableAmbTest.testAmb_ThrowsBeforeElectionRight),
+    ] }
+}
+
+final class ObservableBindTest_ : ObservableBindTest, RxTestCase {
+    #if os(macOS)
+    required override init() {
+        super.init()
+    }
+    #endif
+
+    static var allTests: [(String, (ObservableBindTest_) -> () -> Void)] { return [
+    ("testBindToObserver", ObservableBindTest.testBindToObserver),
+    ("testBindToObservers", ObservableBindTest.testBindToObservers),
+    ("testBindToOptionalObserver", ObservableBindTest.testBindToOptionalObserver),
+    ("testBindToOptionalObservers", ObservableBindTest.testBindToOptionalObservers),
+    ("testBindToOptionalObserverNoAmbiguity", ObservableBindTest.testBindToOptionalObserverNoAmbiguity),
+    ("testBindToCurried1", ObservableBindTest.testBindToCurried1),
+    ("testBindToCurried2", ObservableBindTest.testBindToCurried2),
+    ("testBindToKeyPath", ObservableBindTest.testBindToKeyPath),
+    ("testBindToKeyPathObjectGone", ObservableBindTest.testBindToKeyPathObjectGone),
     ] }
 }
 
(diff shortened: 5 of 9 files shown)


freak4pc:
@wddwycc also suggested another interesting addition for "dynamic `Binder`" which I won't mind adding in this PR if you like it, @kzaher: 

image

RxCocoa/Common/Observable+Bind.swift
```diff
@@ -84,4 +83,30 @@ extension ObservableType {
...
+    stream.bind(to: \UITextField.text, on: myTextField)
+    ```
+
+    In case error occurs in debug mode, `fatalError` will be raised.
+    In case error occurs in release mode, `error` will be logged.
+
+    - parameter keyPath: The key path of the property to assign.
+    - parameter object: The object on which to assign the value.
+    - returns: Subscription object used to unsubscribe from the observable sequence.
+    - note: If a weak reference to the Root object cannot be retained, elements will
```
icanzilb: Since Root is a type, please use backpacks

RxCocoa/Common/Observable+Bind.swift
```diff
@@ -84,4 +83,30 @@ extension ObservableType {
             rxFatalErrorInDebug("Binding error: \(error)")
         })
     }
+
+    /**
+    Bind emitted elements to the referenced Keypath of a specific Root object.
```
icanzilb: Since KeyPath is not a type name, please write it as key path (or to reference the parameter with that name use the spelling of the parameter name and backticks)

danielt1263:
If the goal is to have something similar to what Combine has, wouldn't it make more sense to have the method as an extension on Driver rather than Observable? Combine's method assumes no error is possible and the assumption is that we would mainly be using this on UI right?

freak4pc:
Daniel, I plan to add Driver as well
I’m not sure a replay is necessary for binding though, is it ?

danielt1263:
> I’m not sure a replay is necessary for binding though, is it ?

Maybe on Signal? The main point to me is that it shouldn't be possible to feed errors into the binder in the first place, that way you don't need to worry about the fatalError(). Maybe I should have said "as well as" instead of "rather than" because I see value in having it on ObservableType too.

freak4pc:
@danielt1263 Added on `Driver` and `Signal`. 

I think it's still worthwhile to have the `bind(to:on:)` variation with the added note/debug-crash/runtime-warning, but if community thinks otherwise I'm totally open for any other option.

RxCocoa/Common/Observable+Bind.swift
```diff
@@ -84,4 +83,30 @@ extension ObservableType {
             rxFatalErrorInDebug("Binding error: \(error)")
         })
     }
+
+    /**
+    Bind emitted elements to the referenced key path of a specific `Root` object.
```
LucianoPAlmeida: Since the object is passed as a parameter, maybe 
```suggestion
    Bind emitted elements to the referenced key path of the given `Root` object.
```
What do you think?

freak4pc: Don’t think it changes the meaning that much but sure, thanks

jasdev:
(Chiming in to say this is a wonderful change, thanks @freak4pc! 🙏🏽)

kzaher:
@freak4pc 

> @wddwycc also suggested another interesting addition for "dynamic Binder" which I won't mind adding in this PR if you like it, @kzaher:

There are tradeoffs to both approaches. The reason Apple took this approach is because in this way they don't need to pollute the namespace of the object being bound to.

Since we've already crossed that bridge and polluted the namespace already it seems like the option @wddwycc suggested is more elegant and less verbose. We can add `drive<Observer>(_ observer: Observer)` and `emit`.

freak4pc:
The only problem with that one is that I believe `@dynamixcMemberLookup` with keypath is only available in Swift 5.1 compiler. I'll make sure.

freak4pc:
@kzaher Yup, Swift 5.1 and up only: 
https://github.com/apple/swift-evolution/blob/master/proposals/0252-keypath-dynamic-member-lookup.md

We can have both versions, or only the keypath version for `#compiler(>=5.1)`. What do you think?

kzaher:
@freak4pc 

I don't think we should worry that that feature is available only in 5.1. We can add version guard.

I think that we should probably decide on a single solution. We shouldn't take into account `Combine` similarity as an argument. If we think that something is a better API, we should pick that.

We can add some API to RxCombine is we want to add some API to enable easy migration to `Combine` for users.

freak4pc:
@kzaher Makes sense to me, but it means two annoying facts:

1. If we use a Binder, it would make this a RxCocoa-only thing, a bit of a bummer.
2. If we use Reactive, only `ReactiveCompatible` objects will get binders. 
3. Since `@dynamicMemberLookup` has to be on the type (e.g. can't be on an extension), we'll need to have two implementations based on Swift version compile-check. 

Are these issues OK by you ?

kzaher:
> Since @dynamicMemberLookup has to be on the type (e.g. can't be on an extension), we'll need to have two implementations based on Swift version compile-check.

I'm hoping that this is a minor nuisance, but  I'm not sure how does Swift version compile-check interact with binary stability. There could be problems here. It does seem risky.

> 1. If we use Reactive, only ReactiveCompatible objects will get binders.

I guess the users would need to decide on that. 

I think that there isn't a lot of benefits of using something like:

`drive(object.rx.\Type.prop1.prop2)` or `xs.drive(\Type.prop1.prop2, on:object)`
to
`drive { object.prop1.prop2 = $0 }`

The only significant difference is the weakify part, but I think that >99% of cases are binding child views or view controllers so the weakify part doesn't matter much.
Sure, there is a difference regarding error handling if you are using the raw `Observable` API, but we've added unhandled errors handling hooks anyway.
Even if you have weakify part it is still recommended to dispose the subscription for fragility reasons.

`NSObject` is already `ReactiveCompatible`. The question is what are people using this API for and what do they want to improve.

If this was my private project I would just put @dynamicMemberLookup on `Reactive` and be done with it.

Since this is slightly bigger project, I'm fine with us having this because it's more "correct" in the sense (it doesn't pollute the namespace further) and doesn't make any assumptions regarding the object being bound to, you don't have to conform the object to `ReactiveCompatible` if it's not `NSObject` derived, but is more verbose and has yoda ordering which is unnatural to me when you read the API.

If you think that we should merge this extension, I'm fine with it, but the `drive` extension is an exception, which is again weird (without `to:` label). 

If we merge this, I don't think that it makes to also add the `Reactive` extension.

kamitchell:
This is great! I happen to have just added a `binder(to: ReferenceWritableKeyPath<ModelType, T>)` to one of my model classes, just to make this kind of thing easy. I found this pull request while looking to see if any RxSwift changes were forthcoming.

This brings up a question, that perhaps should be another issue. I have a coding rule for myself that only GUI classes import `RxCocoa`. The `bind(to:)` method is generically useful enough that I was wondering if it could be factored out as the relay classes were.

freak4pc:
@kzaher We have `#compiler` vs `#swift` - so we can target the Swift 5.1 compiler which might help with the issue. I'll look into it.

I'll also poke around with other thoughts around Reactive/ReactiveCompatible that you mentioned earlier. 

As @kamitchell mentioned - I would've loved having this on RxSwift because it's useful beyond the land of UI - but ReactiveCompatible is only in RxCocoa, so I don't think that's feasible unless we have _both_ approaches (seems like overkill and not too elegant, but it means Linux users can use the bind to Keypath thing)

kamitchell:
> ReactiveCompatible is only in RxCocoa

That's true.

The code in `Observable+Bind.swift` only extends `ObservableType`, and could be in `RxSwift`. The `bind()` operators, especially with `keyPath`, would be very useful for data flows that don't involve the UI, and on Linux.

kzaher:
@freak4pc If we are going to release 6.0, then we don't have to worry about not adding `@dynamicMemberLookup`. Thoughts?

freak4pc:
So release 6.0 will only support Swift 5.1 and up ? 

Let's just not do anything under time pressure here, there's nothing pushing us to do a 6.0 immediately since the UIWebView thing is just a warning. Perhaps let's mark which of the open PRs / features we want for 6.0 and do our best to roll these out in the next few week(s) ?

kzaher:
@freak4pc Sure, but let's start considering 6.0 and what should we put in. It seems like adding bumping version might help us with these issues.

kzaher:
> So release 6.0 will only support Swift 5.1 and up ?

Yes, let's make our life easier :)

freak4pc:
@kzaher I played with this today, this implementation mostly works fine for RxSwift and doesn't require RxCocoa, but - Since Reactive.Base can be anything, and not necessarily a class-bound object, there is no way to weakify the object (that I know of). 

That means it could possibly create some retain cycle the consumer might not expect. 

```swift
@dynamicMemberLookup
public struct Reactive<Base> {
    /// Base object to extend.
    public private(set) var base: Base

    /// Creates extensions with base object.
    ///
    /// - parameter base: Base object.
    public init(_ base: Base) {
        self.base = base
    }

    public subscript<Property>(dynamicMember keyPath: ReferenceWritableKeyPath<Base, Property>) -> AnyObserver<Property> {
        return AnyObserver { event in
            guard case .next(let value) = event else { return }

            self.base[keyPath: keyPath] = value
        }
    }
}
```

If I make `Base: AnyObject`, that would break a lot of code for consumers that count on it to make Reactive Extensions on protocols, etc.

I think what might be a good middle ground is heaving this extension, but also having a `bind(to:)` in RxSwift itself (not RxCocoa). In this case, if the KeyPath is explicit, the consumer has better knowledge of what might happen. 

Thoughts?

freak4pc:
Happy to get some more thoughts about my last comment here - @kzaher, @wddwycc, @danielt1263 + anyone else who commented on this thread.

danielt1263:
Implementing code that could create a retain cycle feels like a very bad thing. No amount of documentation will save users. I would rather field questions asking why it won't compile and then telling them to make their protocol a class type then to be in a position where newbies aren't even asking because they haven't noticed the retain cycle... whether the code is in RxSwift or RxCocoa.

I think of this as a cool little feature for people to discover that makes things easier in some situations rather than something that will be in every tutorial and used extensively. As such, having it only work on class based types (much like rx.disposeBag) makes sense, especially if it's in RxCocoa.

freak4pc:
tl;dr I think there are two main solutions, but in both case I still think it would be good to have `bind(to:on:)` in RxSwift.

1. Leave `.rx` namespace as is, this means it can be used for class-bound objects, but as well value-types such as structs, or non-class-bound protocols.

2. Make `.rx` namespace with `@dynamicMemberLookup` which probably fits >90% of use-cases. This would mean it would only work on class-bound objects or class-bound protocols (because of `ReferenceWritableKeypath`, basically). 

## More details

It works quite well for reference-types:

image

image


 But it means it won't work for anything that isn't class-typed. e.g. Structs, Protocols (unless the protocol is bound to `class` or `AnyObject`). 

I'm worried this might be too big of a change. 

We could also have the "best of both worlds" but it would mean we would need to have

`object.rx.*`
`object.something-else.*`

I think that would be quite ugly unfortunately :(

It does provide a lot of value having `@dynamicMemberLookup` this way, but I can't think of any good way to target non-classes as well.

For example `Binder` also works only with reference types.

SlashDevSlashGnoll:
`For example Binder also works only with reference types`  

To me this encapsulates the value of this.   Just like Binder this is just syntactic sugar that can be used if your types allow for it but it wouldn't be the way to teach or learn how to use Rx.   Given that I don't see it as a big deal if it was limited to reference types only.

jegnux:
Hey, jumping in this following @freak4pc call on the Slack. 
TBH I don’t really understand how the debate moved away from a simple, universal and elegant solution (the `bind to: KeyPath on:` binder) to something more complicated and less discoverable, working on reference types only and working on types having a `.rx.` namespace only. 

I completely ++ for the original proposed feature and - - for the rest. 🤷🏻‍♂️

freak4pc:
Thanks for the two comments. 🎉 

@SlashDevSlashGnoll that's my current thinking process on why this "makes sense" - even today when making a `extension Reactive where Base:` - you would usually expose a Binder that takes the `base` in. That `base` also can only be a class-bound object even today (`Binder.Target: AnyObject`)

@jegnux Since we're releasing a major version it was an opportunity to rethink on how some new language features could ease the use of binding etc. 

It seems to me that this is a big change, yes, but: 

1. From discoverability it actually seems nicer - because every property of the `rx`'d object will already be exposed as a `AnyObserver` without the need to manually expose properties and create `Binder`s individually.

2. From the reference-only/object-bound only see note to @SlashDevSlashGnoll above. It seems, in a way, this is already the case today.

3. Lastly, I don't think these two proposals collide. Having `bind(to:on:)` is still great IMO. 

I'll run this branch on some of my big projects and see how much of a "migration" it creates. Would be some sort of indication, I guess. 

Happy for more thoughts!

freak4pc:
On the downside of this - 
It means that if you want something entirely custom for a property, you would need to name it differently. I'm not sure how common of a case this is, but it's worth considering.

freak4pc:
I just ran some tests on my big projects. It all compiles with all changes.

Surprisingly it seems like:

1. If you have your own property (for example `isEnabled`) in the reactive extension, it would take precedence over the dynamic key path. Which means customizability is still right at your finger tips. You can still make binders as you wish, but you can also get rid of a bunch of redundant things such as:

```swift
var isEnabled: Binder<Bool> {
    return Binder(base) { view, isEnabled in 
        view.isEnabled = isEnabled
    }
}
```

2. At least on my projects, as I mentioned earlier, there were no issues with the class-bound-only behavior, as that seems to be the general consensus. Would be great getting more people to test this, though.

freak4pc:
I just migrated one small module and got a -380 Diff PR just from removing redundant `Binder`s this catches automatically. The more I play with it, I actually really like this idea. Great use of this feature IMO.

freak4pc:
I've moved the specific portion referring to `@dynamicMemberLookup` over to https://github.com/ReactiveX/RxSwift/pull/2110

Would be great continuing the discussion there with a working implementation @jegnux @SlashDevSlashGnoll @danielt1263

kamitchell:
> but also having a bind(to:) in RxSwift itself (not RxCocoa)...

Yes, I'm still for moving the bind functions to RxSwift.

farzadshbfn:
@freak4pc @kzaher
@dynamicMemberLookup is in Swift5.1 but it doesn’t affect ABI and is back deployable (since it’s only code generation) just like DSL and PropertyWrappers.

I shipped an application with this (wrote a wrapper over ReactiveCompatible to enable such functionality)

freak4pc:
#2110 brings `@dynamicMemberLookup`-based binding, which makes this PR quite redundant. closing for the time being.

PR #2005, opened 2019-06-09
Add breakpoint operator

MortyMerr:
Combine framework introduced breakpoint operator https://developer.apple.com/documentation/combine/publisher/3204689-breakpointonerror
I think it might be quite useful for debugging. It will also help to satisfy requirements of users, who want to try reactive programming not only with iOS 13+

I am not sure about tests and playground examples (it's not possible). What do u think?

### DIFF
--- a/RxSwift/Observables/Breakpoint.swift
+++ b/RxSwift/Observables/Breakpoint.swift
@@ -1,0 +1,175 @@
+//
+//  Breakpoint.swift
+//  RxSwift
+//
+//  Created by Anton Nazarov on 09/06/19.
+//  Copyright © 2017 Krunoslav Zaher. All rights reserved.
+//
+
+import Foundation
+
+extension ObservableType {
+    /**
+     Raises a debugger signal when a provided closure needs to stop the process in the debugger.
+     - seealso: [breakpoint operator on Combine](https://developer.apple.com/documentation/combine/publisher/3204688-breakpoint)
+
+     - parameter onNext: Predicate to invoke for each element in the observable sequence.  Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter afterNext: Predicate to invoke for each element after the observable has passed an onNext event along to its downstream. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter onError: Predicate to invoke upon errored termination of the observable sequence. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter afterError: Predicate to invoke after errored termination of the observable sequence. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter onCompleted: Predicate to invoke upon graceful termination of the observable sequence. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter afterCompleted: Predicate to invoke after graceful termination of the observable sequence. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter onSubscribe: Predicate to invoke before subscribing to source observable sequence. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter onSubscribed: Predicate to invoke after subscribing to source observable sequence. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter onDispose: Predicate to invoke after subscription to source observable has been disposed for any reason. It can be either because sequence terminates for some reason or observer subscription being disposed. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - returns: The source sequence that raises a debugger signal when one of the provided closures returns `true`.
+     */
+    public func breakpoint(
+        onNext: @escaping (Element) throws -> Bool = { _ in false },
+        afterNext: @escaping (Element) throws -> Bool = { _ in false },
+        onError: @escaping (Swift.Error) throws -> Bool = { _ in false },
+        afterError: @escaping (Swift.Error) throws -> Bool = { _ in false },
+        onCompleted: @escaping @autoclosure () throws -> Bool = false,
+        afterCompleted: @escaping @autoclosure () throws -> Bool = false,
+        onSubscribe: @escaping @autoclosure () -> Bool = false,
+        onSubscribed: @escaping @autoclosure () -> Bool = false,
+        onDispose: @escaping @autoclosure () -> Bool = false
+    ) -> Observable<Element> {
+        return Breakpoint(
+            source: self.asObservable(),
+            eventPredicate: {
+                switch $0 {
+                case let .next(element):
+                    return try onNext(element)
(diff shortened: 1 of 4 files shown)


RxSwift/Observables/Breakpoint.swift
```diff
@@ -0,0 +1,182 @@
+//
+//  AsMaybe.swift
```
freak4pc:
```suggestion
//  Breakpoint.swift
```

RxSwift/Observables/Breakpoint.swift
```diff
@@ -0,0 +1,182 @@
...
+      Raises a debugger signal upon receiving an error.
+      When the upstream Observable fails with an error, this operator raises the `SIGTRAP` signal, which stops the process in the debugger.
+      Otherwise, this Observable passes through values and completions as-is.
+      - seealso: [breakpointOnError operator on Combine](https://developer.apple.com/documentation/combine/publisher/3204689-breakpointonerror)
+      - returns: An Observable that raises a debugger signal upon receiving a failure.
+      */
+    public func breakpointOnError() -> Observable<Element> {
+        return Breakpoint(
+            source: self.asObservable(),
+            eventPredicate: {
```
freak4pc: You can also do `$0.error != nil` if you want to shorten this. Leaving as-is is also 100% fine :)

RxSwift/Observables/Breakpoint.swift
```diff
@@ -0,0 +1,182 @@
...
+                }
+            },
+            afterEventPredicate: { _ in false }
+        )
+    }
+}
+
+#if DEBUG
+private func raiseBreakpoint() {
+    raise(SIGINT)
```
freak4pc: You can test this by setting a signal event handler, something like 

```swift
var trappedSignal = false
signal(SIGINT) { _ in
  trappedSignal = true
}

XCTAssert...
```

At least IIRC... worth trying :)

RxSwift/Observables/Breakpoint.swift
```diff
@@ -0,0 +1,182 @@
...
+import Foundation
+
+extension ObservableType {
+    /**
+     Raises a debugger signal when a provided closure needs to stop the process in the debugger.
+     - seealso: [breakpoint operator on Combine](https://developer.apple.com/documentation/combine/publisher/3204688-breakpoint)
+
+     - parameter onNext: Predicate to invoke for each element in the observable sequence.  Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter afterNext: Predicate to invoke for each element after the observable has passed an onNext event along to its downstream. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
+     - parameter onError: Predicate to invoke upon errored termination of the observable sequence. Return `true` from this closure to raise `SIGTRAP`, or false to continue.
```
freak4pc: You're writing SIGTRAP everywhere but signaling a SIGINT - which is the right one? I think the former?

RxSwift/Observables/Breakpoint.swift
freak4pc: Here's another option: 
https://developer.apple.com/documentation/dispatch/dispatchsource/2300045-makesignalsource

RxSwift/Observables/Breakpoint.swift
jdisho: or maybe `@unknown default`? :)

RxSwift/Observables/Breakpoint.swift
MortyMerr: My bad. Fixed

RxSwift/Observables/Breakpoint.swift
MortyMerr: Sure, done

RxSwift/Observables/Breakpoint.swift
MortyMerr: The problem with both solutions - Xcode debugger handles signals and prevent application handlers to catch them. 
I can fix that with `process handle SIGTRAP -p true`. But I need to do it manually every time. I can also disable debugger for the whole test target, but I doubt that it's a correct solution.

RxSwift/Observables/Breakpoint.swift
MortyMerr: Changed to SIGTRAP, just for consistency with Combine

MortyMerr:
> If this operator only signals on DEBUG, maybe the entire operator should be available just in DEBUG?

I thought about that. But then I imagine my feelings after a release build error because of undefined `breakpoint` (might be quite annoying). So I decided just to make it empty `do` in Release. I can see three ways:
1) Combine behavior, same in debug/release (pretty scared, cause app will crash in case of forgotten `breakpoint`)
2) Implemented behavior, just do nothing in release
3) Compile error in release
I am ok with three of them, so just pick the number :)

> For Carthage users this will be unusable

I am not a big expert/fan of Carthage, but I hope that won't stop us from introducing this useful operator...

freak4pc:
I'm leaning towards option 1 vs 3. 

* Option 1: Yes, it would suck if someone would leave it in production, but it would also suck if someone leaves a fatalError in production. The fact someone has a knife doesn't mean they have to stab themselves. Also the fact it traps means you, as a developer, would know very early in the process. This would also prevent binary build issues like in Carthage.

* Option 3: If we want this to be a debug-only thing, we should make the entire operator DEBUG-only, so the code doesn't build at all with `breakpoint` in release. 

Leaning towards 1 at this point. WDYT?

RxSwift/Observables/Breakpoint.swift
freak4pc: Yeah I understand. that's a bit tricky, I would want at least some sort of test here... with xcodebuild I guess this would work though, rihgt?

MortyMerr:
Agree. Will change implementation to 1 option

kzaher:
Hi guys, 

I understand what you want to do, but I don't think that this way of debugging works.

1. If you are debugging something the callstack is usually not that useful.
2. If the callstack is useful you can use `do` operator to just create lambda and put a breakpoint there, log and inspect the value, error etc. Just having a breakpoint without the ability to inspect values or errors when the breakpoint hits is less useful than just using `do` operator with breakpoint. You still need to provide closure that returns a boolean (and you don't get the important value in it) so the readability isn't improved at all imho.
3. If you are debugging using breakpoints people are toggling breakpoints during the debugging process in the debugger, they don't recompile the process when they want to disable breakpoint, that kills the process state. (Better alternative is to use `do` operator, it already has all of the callbacks and ).
4. Combine framework is in beta and it doesn't make sense to just say "we should include something because they have" for multiple obvious reasons.

MortyMerr:
Ok, feel free to close it, if it's not needed :)

freak4pc:
Let's close this for now and revisit if the need ever pops back (or if/when Combine becomes popular and the community want this operator). Thanks for your work regardless!

Issue #2007, opened 2019-06-10
Adopt few Combine operators


MortyMerr:
I understand that this repo doesn't aim to embed all operators from Combine. But IMO absence of Combine operators might be confusing for new users, which want to try being reactive before iOS 13.0
Of course "add operator because there is a similar one in Combine" is a bad reason, but they might quite useful. Here is a list, which can be appended with new Combine release. Even if we that repo doesn't want to adopt them - it will be good to see "No" explicit. 

- [ ] [`func replaceNil(with: T)`](https://developer.apple.com/documentation/combine/publisher/3204749-replacenil) - which is already implemented in https://github.com/RxSwiftCommunity/RxOptional but in a less performant way. I think that might be quite useful even if we have `.compactMap` already.

- [ ] [`func mapError(transform:)`](https://developer.apple.com/documentation/combine/publishers/maperror) - yes, we don't have typed errors in RxSwift, but sometimes there is a need to catch an error and replace it with another. I saw a few times implementation with `catchError` and throwing new inside of it. `.mapError` would be more descriptive.

- [ ] [`func output(at:) / func output(in:)`](https://developer.apple.com/documentation/combine/publishers/output) - `enumerated().skipWhile()` might be useful, but don't provide exactly same behaviour. 

- [ ] [`func measureInterval()`](https://developer.apple.com/documentation/combine/publishers/measureinterval) - might be quite tricky, but looks useful for building stream of events based on the frequency of upstream

- [ ] `max/min` operators, which is always added to [reactiveX](http://reactivex.io/documentation/operators/max.html)

- [ ] [`func contains(_ :)`](https://developer.apple.com/documentation/combine/publishers/contains)

- [ ] [count](https://github.com/RxSwiftCommunity/RxSwiftExt/pull/190/files) - maybe rethink solution about placing it to RxSwiftExt?

- [ ] [last(where:)](https://developer.apple.com/documentation/combine/publisher/3204716-last)

- [ ] [allSatisfy](https://developer.apple.com/documentation/combine/passthroughsubject/3204581-allsatisfy?changes=latest_minor)

As always I am really interested in your opinion and ready to contribute :)

freak4pc:
I want to start with saying thanks for your work on this. 
My personal opinion is that we shouldn't add an operator **just because** it's part of Combine if it doesn't make sense to us. This is a large project and I think we should be careful with adding new operators without a good enough reason.

## Operators that I think might be a good fit: 

[...]

PR #2010, opened 2019-06-14
Add AllSatisfy operator

MortyMerr:
No words about Combine. 
`allSatisfy` is a Swift.Sequence member. It might be useful for dealing with sequences.
https://developer.apple.com/documentation/swift/sequence/2996794-allsatisfy

If you find it useful, I will also add it for `SharedSequence`
https://github.com/ReactiveX/RxSwift/issues/2007

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -191,6 +191,7 @@ custom_categories:
 - name: RxSwift/Observables
   children:
   - AddRef
+  - AllSatisfy
   - Amb
   - AsMaybe
   - AsSingle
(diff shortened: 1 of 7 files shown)


MortyMerr:
@freak4pc what do you think?)

freak4pc:
@MortyMerr Thanks for the work you've put into this. 

It's a nice idea in general - But I'm trying to think if there is a practical use-case to it. Yes, it exists in Combine and Swift Foundation, but I think the approach lately was to figure out "Does someone in the community actually find this useful for a real-life application?" 

What do you think? I can find many ideas for using this on a collection, but not as many real applications for a stream of values over time.

freak4pc:
Hey @MortyMerr - I'm still contemplating about this one. I'm wondering if you found a practical use case to this. Thought about this some more but still find this very hard to reason about. 

There is a huge difference between allSatisfy for finite sequences / collections and something like a possibly infinite stream — not from the technical perspective, but from the usability perspective. I can't really think of a feature where this ability would help, but am happy to learn otherwise. 

Thank you!

MortyMerr:
Feel free to close 

PR #2012, opened 2019-06-19
Support project Catalyst

iamDecode:
Using RxSwift in an iOS app compiled for Mac currently yields nasty errors because `UIWebView` is deprecated and not supported in UIKit for Mac:
```
Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_UIWebView", referenced from:
      objc-class-ref in RxWebViewDelegateProxy.o
ld: symbol(s) not found for architecture x86_64
```

The official documentation states `#if !targetEnvironment(UIKitForMac)` is the best way of dealing with these issues ([source](https://developer.apple.com/documentation/uikit/creating_a_mac_version_of_your_ipad_app)).

This PR adds some extra conditions to support compiling with project Catalyst.

### DIFF
--- a/RxCocoa/Deprecated.swift
+++ b/RxCocoa/Deprecated.swift
@@ -220,6 +220,9 @@ extension ObservableType {
             fatalError()
         }
     }
+#endif
+
+#if os(iOS) && !targetEnvironment(UIKitForMac)
     extension UIWebView {
         @available(*, unavailable, message: "createRxDelegateProxy is now unavailable, check DelegateProxyFactory")
         public func createRxDelegateProxy() -> RxWebViewDelegateProxy {
--- a/RxCocoa/iOS/Proxies/RxWebViewDelegateProxy.swift
+++ b/RxCocoa/iOS/Proxies/RxWebViewDelegateProxy.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2016 Krunoslav Zaher. All rights reserved.
 //
 
-#if os(iOS)
+#if os(iOS) && !targetEnvironment(UIKitForMac)
 
 import UIKit
 import RxSwift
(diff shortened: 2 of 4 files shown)


freak4pc:
This is excellent! If this works in Xcode 10 and builds,  I'm all for it. Let's wait for CI.

freak4pc:
So, RxCocoa fails build because of a warning that the targetEnvironment is unknown. 

I'd really want to find away around this, if you have any ideas. 

image

I thought wrapping it in a `#if compiler(>=5.1)` would work but it doesn't seem like it.

iamDecode:
I've stared at this for a while, but I cant seem to find my way around this. It is parsed, even if we do something like `#if false && targetEnvironment(UIKitForMac)` or nesting as @freak4pc suggested. I was not able to silence that specific warning; that does not seem possible for Swift other than silencing all warnings for the target ([relevant source](https://twitter.com/jckarter/status/972894889221414912)).

freak4pc:
@iamDecode I opened a bug in Swift Compiler:
https://bugs.swift.org/browse/SR-10973

There seems to be a fix here, we might need to pull this in and bite the bullet, but let me think of options until tomorrow: 
https://github.com/apple/swift/pull/25593

kzaher:
I think it should be ok to just add `--allow-warnings` to podfile verification. 

It's probably better to allow users to use it with warnings than have it fail.

freak4pc:
Yup. That would be the only option here - it’s a parser but which was fixed but only for Xcode 11. Can you add it to Travis ? @iamDecode

iamDecode:
@freak4pc sure thing, I updated it! I added a bit of logic to only add `--allow-warnings` argument for the RxCocoa target, would be a bit much to add it globally.

PR #2018, opened 2019-06-29
Add `share()` for Single and Maybe

MortyMerr:
Added `share()` similar to `Observable` for `Single` and `Maybe`. 

I removed `replay` from `Single.share`, cause passing any value besides 1 would be irrational. It's impossible to replay more than one event from `Single`. `replay: 0` would break the contract, cause `Single` sequence must have one event.

I replaced `replay` in `Maybe.share` on `shouldReplay` flag for the same reason (can't replay more than one element).

If that implementation is ok, I also will add `.share` for `Completable`.

### DIFF
--- a/RxSwift/Observables/ShareReplayScope.swift
+++ b/RxSwift/Observables/ShareReplayScope.swift
@@ -156,6 +156,81 @@ extension ObservableType {
     }
 }
 
+extension PrimitiveSequence where Trait == SingleTrait {
+    /**
+     Returns a Single that **shares a single subscription to the underlying sequence**, and immediately upon subscription replays element in buffer.
+
+     This operator is equivalent to:
+     * `.whileConnected`
+     ```
+     // Each connection will have it's own subject instance to store replay event.
+     // Connections will be isolated from each another.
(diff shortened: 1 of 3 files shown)


kzaher:
Hi @MortyMerr ,

I don't think we should add sharing operators to finite sequences. Using sharing operators on them alone is usually a code smell IMHO.

Sharing operators make sense for sequences that don't complete where resources want to be optimised (observers, web sockets ...).

MortyMerr:
Hi @kzaher thank u for your comments :)

What's the required way of handling multiple subscriptions with `Single`? Use `.asObservable()`?

freak4pc:
> What's the required way of handling multiple subscriptions with Single? Use .asObservable()?

As Krunoslav mentioned, having to do a `share` on something that has zero or one values doesn't make a lot of sense to me. If you're referring to an API that is wrapped inside Single, the call itself wouldn't be `share()`d, but the stream invoking that call (usually would be some flatMap to that Single). 

Closing this for now, comment if you have any more thoughts on this. Thanks !

PR #2022, opened 2019-07-01
Normalizing `map` implementation

MortyMerr:
This PR is probably more like a question than solution. `map` operator has optimization, which composes two `map` into one. According to comment, it's implemented in a kind of ugly way. I would suggest to remove it because:

- this place is not going to become better

Methods declared in extensions (and not backed by protocol) use static dispatch, because extensions have their own witness table. As I know, Swift core team has no plans to break that optimization and support overriding methods, declared in extensions.

- this looks like a hack in a most popular place of the repo

True story: my friend started to look at this project from my advice. After two hours of digging in code, he sent me screenshot of that place with sad emoji.
I understand, that it's a non-commercial project which has no task to attract people. However, hacks confuse users. And `Observable.swift` is in top-3 place, which every developer will check.

- the approach used only for one operator, which might be confusing for contributors

This approach can be also be applied to other popular operators (`filter`, `compactMap` and `flatMap`). It will be a pleasure for me to do this. It will unify operators implementation style.
 However, I doubt it worth it, because:

- IMO, this is very micro-optimization

I run `Benchmarks.testMapFilterPumping` after these changes and get 1-2% worse performance. Does it really worth it?


MortyMerr:
@freak4pc thanks for your comments. Yes, it's rational. As I already mentioned in the description, this PR is more like a question, not real change suggestion.
But what do you think about expanding such kind of optimization on other operators? Will it be useful to increase the performance of a few other operators with same style optimization? Or `map` is the only one which desires that?

`TRACE_RESOURCES` were removed just because map operators count was used only for `composeMap` test.

freak4pc:
If these kind of optimizations would provide substantial, measurable value (benchmarks), we could opt in to it. But TBH I believe Kruno examined some of the other optimizations on other operators, so it makes sense to me he chose not to do it for a reason.

kzaher:
Hi guys,

These were some optimisations we've made. If it's just a couple of % maybe they don't matter now.

If you guys want to remove them. I'm fine with it. It seems that Apple is exposing implementation details for so the Swift compiler can perform performance optimisations.

I guess it seems that if we wanted to perform performance optimisations we should just copy what Combine is doing since it looks like they are faster.

Just don't merge it directly to `master` and go through `develop`.

freak4pc:
I don't think they actually shared any source / implementation details. Where did you see any ?

kzaher:
@freak4pc I think that the return type from the operator extension is a concrete type for which you need to do type erasure manually. E.g. `map` operator, something like (`MapPublisher`).

freak4pc:
I wouldn't call that implementation details, it's just a structure. I'm not sure what kind of optimizations they do. If they pull the events through the chain of operators upon subscription, is it really that different? You don't have to type erase, also...

kzaher:
@freak4pc 

I haven't tested it, but I would assume that it makes the life easier for the compiler.

kzaher:
Since this optimizes for a really special case and potentially slows down the common one, l'm fine with removing this.

I think that there is probably some other low-hanging fruit that makes a more significant difference.

PR #2023, opened 2019-07-01
Prefer private over fileprivate

MortyMerr:
Swift 3 is gone and we can use private properties in extensions. Also, there are few other cases when `fileprivate` can be replaced with `private`. 
Pros:
- compiler will provide optimizations
- contributor will be less confused with `fileprivate` usage

### DIFF
--- a/Platform/DataStructures/Bag.swift
+++ b/Platform/DataStructures/Bag.swift
@@ -35,7 +35,7 @@ struct Bag<T> : CustomDebugStringConvertible {
     
     typealias Entry = (key: BagKey, value: T)
  
-    fileprivate var _nextKey: BagKey = BagKey(rawValue: 0)
+    private var _nextKey: BagKey = BagKey(rawValue: 0)
 
     // data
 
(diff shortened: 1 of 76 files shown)


MortyMerr:
Sorry, change target branch

MortyMerr:
CI is green now :)

kzaher:
I think it might be worthwhile to investigate how can we make sure we don't break ABI with changes. Unsure how to do that.

freak4pc:
I'm not sure it necessarily matters if we release new versions - binary stability is nice, but people are still locked in to specific versions of the library, that's what versions are for. WDYY?

kzaher:
@freak4pc 

> I'm not sure it necessarily matters if we release new versions - binary stability is nice, but people are still locked in to specific versions of the library, that's what versions are for. WDYT?

It would be nice to maintain backwards stable ABI because it seems some people expect that and are depending their closed source libraries on us.

I would say that would we want to maintain backwards stable ABI depends on the amount of effort we need to put into it. If we can automate this in some easy way, then I would say yes, otherwise no.

freak4pc:
@kzaher A bit confused on why, since closed source libraries are also versioned (usually) and possibly locked to a specific API version. ABI is useful in case of Swift where things come bundled on devices, etc, but is it really making a huge difference here? 

About automation I guess the .swiftmodule option suggested by @LucianoPAlmeida (I think) could do it, but not 100% sure it's worth it in any way.

kzaher:
Let's continue this here #2026.

PR #2027, opened 2019-07-04
Add takeUntil for SharedSequence

MortyMerr:
Added takeUntil similar to Observable. It doesn't complete, but stops emit events.

Use case:
Landing screen which has animations, which is triggered by `Driver.timer`. I want animation stops when user taps button
```swift
Driver<Int>.interval(.seconds(5))
  .takeUntil(UIButton().rx.tap.asObservable())
  .drive { *run animation* }
```

### DIFF
--- a/RxCocoa/Traits/SharedSequence/SharedSequence+Operators.swift
+++ b/RxCocoa/Traits/SharedSequence/SharedSequence+Operators.swift
@@ -425,6 +425,22 @@ extension SharedSequence {
     }
 }
 
+// MARK: - TakeUntil
+extension SharedSequence {
+    /**
+     Returns the elements from the source shared sequence until the other observable sequence produces an element.
+
+     - seealso: [takeUntil operator on reactivex.io](http://reactivex.io/documentation/operators/takeuntil.html)
+
+     - parameter other: Observable sequence that terminates propagation of elements of the source sequence.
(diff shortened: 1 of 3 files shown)


freak4pc:
Hey Anton. None of the take-style operators are currently part of SharingStrategy AFAIK. I don't think we'd want to add more implementations or variations for very specific / one-off use cases.

Usually a `Driver` is what you'd use to expose a regular observable stream. 

For example in a ViewModel - 

```swift
animateScreen = Observable<Int>
    .interval(.seconds(5))
    .takeUntil(buttonTapped)
    .asDriver()
```

Your specific code seems have all of this logic in the View Controller which isn't necessarily a best practice of using `Driver` to compose logic.

In this specific case it is even less of a good idea because Driver can't complete and they replay upon subscription, so `takeUntil` has a wrong meaning on that specific entity.

MortyMerr:
Thank you for architectural consultation, however, that was sample code to show idea.

freak4pc:
It wasn't architectural consultation :) You showed an example to showcase your point and I mentioned Driver is usually not something you transform upon, but is the "end piece", and also in this specific case takeUntil doesn't have meaning a consumer would expect. 

I can't "guess" your use case - whatever you write is what I reply to.

PR #2032, opened 2019-07-12
Fix redundant typealias specification

FuYouFang:
self._clock will eventually be set to virtualTime, so I removed useless settings.


freak4pc:
CI Seems to feel after your changes

LucianoPAlmeida:
The title of this PR should change, now that it is just a trivial type change to a typealias and not a remove settings :))

kzaher:
This has the potential to break the ABI.

Issue #2035, opened 2019-07-15
DelegateProxy bug


DanboDuan:
**Short description of the issue**:

There is a bug in RxCocoa when UIScrollView's delegate is swizzled

**Expected outcome**:

  UIScrollView.rx.delegate or   UIScrollView.rx.setDelegate should work

**What actually happens**:
 if I hook the delegate, the RxCocoa works wrong

**Self contained code example that reproduces the issue**:

The `BDWebViewDecorator` is subclass of NSProxy to decorate the delegate.

```Objective-C
// sample codes, not all, the Decorator will forward all methods calling except bd_isCollectionViewTrackerDecorator  to the target
@implementation BDCollectionViewDelegateDecorator

- (BOOL)respondsToSelector:(SEL)aSelector {
    if (aSelector == @selector(bd_isCollectionViewTrackerDecorator)) {
        return YES;
    }

    return [self.target respondsToSelector:aSelector];
}

@end

// setXXXDelegate: and setDelegate: are swizzled
- (void)setXXXDelegate:(id)delegate {
    if (delegate == nil) {
        [self setXXXDelegate:delegate];
    }
    // once is ok
    if ([delegate respondsToSelector:@selector(bd_isCollectionViewTrackerDecorator)]) {
         [self setXXXDelegate:delegate]; 
         return;
    }
    BDWebViewDecorator *proxy = [[BDWebViewDecorator alloc] initWithTarget:delegate];
    [self setXXXDelegate:proxy];
    //  setWebViewDelegate first then hold proxy
    self.bd_TrackerDecorator = proxy;
}
```

RxCocoa works wrong here

```Swift
public static func proxy(for object: ParentObject) -> Self {
        let maybeProxy = self.assignedProxy(for: object)
        let proxy: AnyObject
        if let existingProxy = maybeProxy {
            proxy = existingProxy
        } else {
            proxy = castOrFatalError(self.createProxy(for: object))
            self.assignProxy(proxy, toObject: object)
            assert(self.assignedProxy(for: object) === proxy)
        }
        let currentDelegate = self._currentDelegate(for: object)
        let delegateProxy: Self = castOrFatalError(proxy)

        if currentDelegate !== delegateProxy {
            delegateProxy._setForwardToDelegate(currentDelegate, retainDelegate: false)
            assert(delegateProxy._forwardToDelegate() === currentDelegate)
            self._setCurrentDelegate(proxy, to: object)
[...]

PR #2037, opened 2019-07-17
Fix delegate proxy error

DanboDuan:
Fix [issue 2035](https://github.com/ReactiveX/RxSwift/issues/2035)

This issue occurs when some one does swizzle the delegate


DanboDuan:
no body care about the crash  in RxCocoa when some one hook the delegate? This means RxCocoa can't work with some kinds of hooks

RxCocoa/Common/DelegateProxy.swift
```diff
@@ -225,19 +225,37 @@
...
+            }
+
+            let res = super.responds(to: aSelector)
                 || (self._forwardToDelegate?.responds(to: aSelector) ?? false)
                 || (self.voidDelegateMethodsContain(aSelector) && self.hasObservers(selector: aSelector))
+            return res
+        }
+
+        @objc open func isDelegateProxy(_ proxy: AnyObject) -> NSNumber {
+            if self === proxy {
```
powerje: `return self === proxy`

RxCocoa/Common/DelegateProxy.swift
```diff
@@ -225,19 +225,37 @@
         }
 
         override open func responds(to aSelector: Selector!) -> Bool {
-            return super.responds(to: aSelector)
+            if  aSelector == #selector(self.isDelegateProxy(_:)) {
+                return true;
+            }
+
+            let res = super.responds(to: aSelector)
```
powerje: No need to store this into a value, just return directly

freak4pc:
Hey @DanboDuan - would you mind:

1. Providing a reproducible project that causes an issue.
2. Fix this PR so CI passes

Thanks!

RxCocoa/Common/DelegateProxy.swift
DanboDuan: I had tried, but it doesn't work saying bool is not NSNumber while i need a NSNumber

RxCocoa/Common/DelegateProxy.swift
DanboDuan: You are right

DanboDuan:
@freak4pc I provided the case in the [issue 2035](https://github.com/ReactiveX/RxSwift/issues/2035).

this case happens when some one does the same thing as Rxcocoa. When some hooks the setDelegate of the UIScrollView and wapper the delegate with a proxy, and then set the proxy to be the delegate of the UIScrollView. it's just like this

```Objective-C
// sample codes, not all, 
//the Decorator will forward all methods calling 
//except bd_isCollectionViewTrackerDecorator  to the target
@implementation BDCollectionViewDelegateDecorator

- (BOOL)respondsToSelector:(SEL)aSelector {
    if (aSelector == @selector(bd_isCollectionViewTrackerDecorator)) {
        return YES;
    }

    return [self.target respondsToSelector:aSelector];
}

@end

// setXXXDelegate: and setDelegate: are swizzled
- (void)setXXXDelegate:(id)delegate {
    if (delegate == nil) {
        [self setXXXDelegate:delegate];
    }
    // once is ok
    if ([delegate respondsToSelector:@selector(bd_isCollectionViewTrackerDecorator)]) {
         [self setXXXDelegate:delegate]; 
         return;
    }
    BDWebViewDecorator *proxy = [[BDWebViewDecorator alloc] initWithTarget:delegate];
    [self setXXXDelegate:proxy];
    //  setWebViewDelegate first then hold proxy
    self.bd_TrackerDecorator = proxy;
}
```
I will provide a demo project in https://github.com/DanboDuan/SwiftDemo

very simple demo with a tableview.
Just Rxcocoa works, but if with TestHook fails.

RxCocoa/Common/DelegateProxy.swift
DanboDuan: image

DanboDuan: i had tried and i need a nsnumber. i will fix it is another way
image

DanboDuan:
@freak4pc  and I don't know how to fix the PR to make the CI passed. I had tried but failed. I'm sorry I haven't figured out why the CI fails

freak4pc:
You can't use NSSelectorFromString because it doesn't work in Swift Package Manager / Linux. You might want to limit that code with compile-time conditionals.

kzaher:
`UI{Table,Collection}View` can only have a single delegate. That's an unfortunate design decision from Apple. I can see how that would work for data source maybe, but not for delegate.

In any case, if you have multiple libraries which want to do this, then I would suggest picking one of them.

We can't add workarounds like this in this library.
* There is no standard for chaining delegates.
* It adds more complexity to already complex code.
* We can't test it correctly in isolation.
* It subtly breaks `reset` logic and will cause maintenance burden.
* Swizzling is a bad idea and as far as I can tell we don't swizzle `UI{Table,Collection}Views`.
https://github.com/DanboDuan/SwiftDemo/blob/master/TestHook/UITableView%2BDecorator.m#L50
* There is no need to have two libraries doing the same thing. Just pick one.

DanboDuan:
@kzaher it not the same thing. it is the same style of swizzle.

As you can Rxcocoa is kind of reactive programming. It's not aimed to swizzle the delegate or the datasource. but it should take this case into consideration.

`assert(self._currentDelegate(for: object) === proxy)`

this code assumes no one has swizzled the setDelegate method, it is not reasonable.

Rxcocoa doesn't swizzle in fact, but it doesn't mean that it should not work with swizzled delegate.

The case I happened to learn because I have to build a SDK to track the tableview or collection view item clicking event. This is aimed to swizzle the delegate. And there would be other people who are using Rxcocoa might also want to do. And some day we might want track datasource which will lead to swizzle the datasource in the same way. it will also fail if it's not in consideration.

DanboDuan:
@freak4pc Thanks. I just fix the problem

freak4pc:
@kzaher Knows this code much much better than I do so I'd have to defer to his choice on whether or not to merge it as is, thanks for making the changes to fix the CI.

DanboDuan:
OK, I will fix it in my repo after Rxcocoa 5.0.2 release.

kzaher:
As far as I can tell this code:
https://github.com/DanboDuan/SwiftDemo/blob/master/TestHook/UITableView%2BDecorator.m#L66
not only swizzles but it also fails to swizzle `delegate` method to strip the proxy.

> this code assumes no one has swizzled the setDelegate method, it is not reasonable.

I would say that this code assumes that `setDelegate` and `delegate` return the same object. It has nothing to do with swizzling. If you want to swizzle `setDelegate` and decorate the delegate, then swizzle the `delegate` method to strip that decoration.

> The case I happened to learn because I have to build a SDK to track the tableview or collection view item clicking event. This is aimed to swizzle the delegate. And there would be other people who are using Rxcocoa might also want to do. And some day we might want track datasource which will lead to swizzle the datasource in the same way. it will also fail if it's not in consideration.

That SDK is doing it wrong. Like I've said, our code only assumes that when you perform `setDelegate`, `delegate` getter returns the same object. If their swizzling breaks this assumption their code should be fixed.

Swizzling is always a bad idea and should be discouraged. If somebody is doing it, then it's up for them to make sure that the existing behaviour is unbroken. It's not up to this library to cleanup poorly executing swizzling.

DanboDuan:
> our code only assumes that when you perform setDelegate, delegate getter returns the same object. If their swizzling breaks this assumption their code should be fixed.

But Reactive's setDelegate and delegate also return different objects. The UIScrollViewDelegate is set and the DelegateProxy is returned. you can check the code in `UIScrollView+Rx.swift`.

DanboDuan:
> That SDK is doing it wrong. Like I've said, our code only assumes that when you perform setDelegate, delegate getter returns the same object. If their swizzling breaks this assumption their code should be fixed.

@kzaher even if we don't break the assumption, RxCocoa still work wrong in this case. When we add a 
 instance method `tableView:didSelectRowAtIndexPath:` to the delegate which is in fact the `RxCocoa.RxTableViewDelegateProxy` or `RxCocoa.RxCollectionViewDelegateProxy`, the code won't callback

```Swift
tableView.rx.itemSelected.subscribe(onNext: { indexPath in
            print("select index ：\(indexPath)")
        }).disposed(by: disposeBag)
```
So we should take some work to fix it

kzaher:
> But Reactive's setDelegate and delegate also return different objects. The UIScrollViewDelegate is set and the DelegateProxy is returned. you can check the code in UIScrollView+Rx.swift.

@DanboDuan If you are referring to `rx.setDelegate` and `rx.delegate` it's because that's how we've defined our API. We can define our API in any way we want. The code you are referring to is **changing the observable behavior of Apple defined APIs which is wrong**. 

This is wrong because Apple defined how these APIs should behave. If you are swizzling the Apple APIs and changing their behavior then you are to blame if you broke somebody expecting the API to behave as they are defined by Apple.

It's simple. Do with you APIs as you please. If you are swizzling Apple's APIs you better make sure their observable behavior is left unchanged otherwise you are doing something wrong. If somebody doesn't understand this, they shouldn't be going around swizzling Apple's APIs.

DanboDuan:
@kzaher I am not referring to `rx.delegate`. I mean the `UITableView.delegate`.

I just swizzle the `setDelegate:` and then check the `delegate`:

1. if the `delegate` has method `tableView:didSelectRowAtIndexPath:`, I just swizzle it and call the origin. This doesn't break anything.
2. if not add one. This might be some problem. But why?

it is just a very common action.

freak4pc:
@kzaher when you have a few minutes, could you make a call on this PR? I'm trying to do some house cleaning here.

freak4pc:

PR #2039, opened 2019-07-17
Add RelayType & AnyRelay

MortyMerr:
## Motivation:

Implementing my ViewModel protocol I want to have input ("action-like") properties, which is useful for binding user intents (touches/swipes). Such sequences of events obviously can't be failed / completed, so `Relay` perfect fit for them.
```swift
protocol ViewModel {
  // Input
  var toggle: PublishRelay<Bool> { get }
}
```

However, I don't want to expose my underlaying type of Relay. Maybe I need to store the latest state with `BehaviourRelay`, maybe not. Moreover, I want to prevent a subscription on my input property `toggle`. I need something like `AnyObserver`, but only accepting `Element` instead of `Event<Element>`.

This PR introduces `RelayType` and `AnyRelay` which help to achieve such behavior:

```swift
protocol ViewModel {
  // Input
  var toggle: AnyRelay<Bool> { get }
}
```
Also I reduce code dublication for Publish/Behaviour relays.

### DIFF
--- a/.jazzy.yml
+++ b/.jazzy.yml
@@ -136,9 +136,11 @@ custom_categories:
   - NSView+Rx
 - name: RxRelay
   children:
+  - AnyRelay
   - BehaviorRelay
   - Observable+Bind
   - PublishRelay
+  - RelayType
   - Utils
 - name: RxSwift
   children:
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -50,6 +50,8 @@
 		6B9CA56E202A206A002C2D11 /* KeyPathBinder+RxTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6B9CA56D202A1F96002C2D11 /* KeyPathBinder+RxTests.swift */; };
 		6B9CA56F202A206B002C2D11 /* KeyPathBinder+RxTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6B9CA56D202A1F96002C2D11 /* KeyPathBinder+RxTests.swift */; };
 		6B9CA570202A206C002C2D11 /* KeyPathBinder+RxTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6B9CA56D202A1F96002C2D11 /* KeyPathBinder+RxTests.swift */; };
+		7889220322DF24C700BA25F4 /* RelayType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7889220222DF24C700BA25F4 /* RelayType.swift */; };
+		7889220522DF250900BA25F4 /* AnyRelay.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7889220422DF250900BA25F4 /* AnyRelay.swift */; };
 		7EDBAEB41C89B1A6006CBE67 /* UITabBarItem+RxTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7EDBAEAB1C89B1A5006CBE67 /* UITabBarItem+RxTests.swift */; };
 		7EDBAEBC1C89B9B7006CBE67 /* UITabBarItem+Rx.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7EDBAEB71C89B9B7006CBE67 /* UITabBarItem+Rx.swift */; };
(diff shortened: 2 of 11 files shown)


LucianoPAlmeida:
Hey @MortyMerr 
This seems a little bit the like #1813 except for some different implementation details and also, related to #1551. There were some discussions on 1813 that maybe can be applied here too.

MortyMerr:
@LucianoPAlmeida thank you for links :)
I saw that discussion about over-abstraction. However, I provided strong motivation for introducing that feature. It's very useful and I can't imagine a way to work around this case without `AnyRelay`. It doesn't look for me too complicated, because this repo already uses similar implementation for `Observable` and `ObservableType`. I can see no reasons why `Relay` can't also adopt that feature.

`AnyRelay` similar to `AnyObserver` would be extremely useful in usual reactive MVVM or MVI (feedback loop) architecture implementation. That's why I decided to revisit that question :)

LucianoPAlmeida:
Hey @MortyMerr :))
You mean a work around to not expose the underlying relay right?  
I see the point, maybe some day when the language supports it we can do it without need a `AnyRelay` type just with the protocol itself like:
```swift 
protocol ViewModel {
  // Input
  var toggle: some RelayType where _.Element==Bool { get }
}

```
But TBH I really don't see a use case where one would use a protocol defining a generic relay and have two different implementations with a `Publish` and `Behavior`. Maybe it could be good just for the sake of not exposing the underlying type but not to have in fact use this to have 1+ implementations.

MortyMerr:
Yes, i hope someday) but for now we can use type erasure.

Of course, I didn’t mean really having two implementation of viewModel. But viewController should not know about underpaying relay type. I just want to disable opportunity to subscribe on viewModel input in view controller. 

With `BehaviorRelay` it’s possible. Moreover, I can get `viewModel.toggle.value` in viewController, which is code smell. But with `RelayType` I can only push new Element, which is required behavior. Does that make sense? :)

kzaher:
Hi @MortyMerr ,

> However, I don't want to expose my underlaying type of Relay. Maybe I need to store the latest state with BehaviourRelay, maybe not.

`AnyObserver` exists because `ObserverType` is a protocol. That problem doesn't exist with Relays. There is not disadvantage to exposing a `Relay`. Relay is already abstracting it's consumers. I don't see a point of hiding type type of relay.

If you want to create a layer to hide the Rx implementation, then you would create a local protocol.

MortyMerr:
Hi @kzaher,

I don’t want to hide Rx. My purpose is declaring type which can only consume new elements. I can’t use Relay directly, because it also has `subscribe` method, which is not required.

I can achieve that with Observer, using `AnyObserver` protocol. But I can’t do that with `Relay`, because it doesn’t implement generic consuming protocol (like `RelayType`).

I thought that such layer of abstraction, which allows to hide underlaying type, should be in that repo. It also helps to reduce code, because there should be no difference between `PublishRelay` and `BehaviorRelay` `bind` implementation (now this code is copy pasted).

If u think, that copypasting code for both relays is ok and user should implement such level of abstraction by himself - please feel free to close this PR. Thanks 😊 

freak4pc:
Hey Anton - I can understand why basically this sounds like a solid idea. Actually myself and @mosamer wanted to do something similar, adding some Binndable protocol to unify everything that could be bound a stream of values. 

Even though that theoretically sounded like a good idea at the time, it falls apart in many places - specifically those around rules. 

* A PublishRelay/PublishSubject represents *events* (no replay or initial value) so you shouldn’t be able to bind a Behavior* to it, or driver a Driver to it, etc. since that breaks the contract of these subjects. 

* Likewise, BehaviorSubject/BehaviorRelay represent **state** where things sort of have a cache/stack (replay of 1 with initial value). It would make no sense to bind a Publish* to it or emit to it via a Signal as that breaks that contract as well.

* There are many more cases where these specialization occur and in my experience, trying to generalize something that isn’t really generic is usually a bad plan. It’s less DRY to “copy paste” some impl’s as you mentioned but it makes much more sense to have the ability to easily specify the rules, traits and reletionships of these different entities. 

* Like kruno mentioned, having another abstraction on top of Relay that already abstracts Subjects seems like a huge overkill for really little value:)

Obviously this is just my $0.02 but I feel it would be wrong to try adding something around this even though I myself though it to be a good idea before really diving into it :)

freak4pc:
We should make a decision on this one way or the other so we can close this with a decision. Please share your final thoughts / arguments here @MortyMerr @kzaher

MortyMerr:
Hey @freak4pc i trust your opinion and don't want to argue anymore. I still need protocol which provides only `accept(_ next: Element)` behavior. However, if you don't think it's a good idea, let's just close it

freak4pc:
Thanks. To be honest it’s quite trivial to add to your own codebase if it’s a use case you have but I haven’t seen any other demand for it.

PR #2051, opened 2019-08-16
Remove dubbed errors policy from docs #trivial

MortyMerr:
I thought that:
> it never fails

and 

> it never errors out

mean the same thing - no `.error` even can be emitted.

Please fix me if I am wrong :)


freak4pc:
I would actually opt to swap those. e.g. leave "it never errors out" and remove "it never fails" since we're always discussing "Error events"

MortyMerr:
Sure, done

PR #2062, opened 2019-09-13
Removed support to UIWebView

gringoireDM:
Removed references to UIWebView in RxCocoa and Tests to address https://github.com/ReactiveX/RxSwift/issues/2055


freak4pc:
Hey there - Thank you :) We'll probably address this a bit later in a separate PR, but good to have this open for tracking.

gringoireDM:
The build is failing, but I'm not sure for which reason to be honest. 
All tests are passing.

freak4pc:
Because the docs are not updated. Run `Please run ./scripts/update-jazzy-config.rb`, please.

freak4pc:
Also run `./scripts/all-tests.sh` to make sure the tests around this have been properly removed for SPM as well.

gringoireDM:
Done, Thank you so much.

zishanj:
I wonder why Apples's new `WKWebView` library not used instead of removing the complete support for WebView?

freak4pc:
WKWebView should be added separately, unrelated to this. Adding stuff is easier than removing stuff :)

KashifJavaid:
How can I use this branch to test if it stop the warning on my project ?

gringoireDM:
@KashifJavaid I don't see why you shouldn't. There is 0 reference in this branch to UIWebView, if you still get the warning, you have other references to that class, somewhere.

tommytian:
Thanks you for fixing this. ;)

kzaher:
Merging https://github.com/ReactiveX/RxSwift/pull/2064 will at help people to reference the changes if they really need to deal with the warning while we think about what will 6.0 contain.

jormungand2006:
Hi, Apple announced that

> The App Store will no longer accept new apps using UIWebView as of April 2020 and app updates using UIWebView as of December 2020.
> https://developer.apple.com/news/?id=12232019b

, so I expect RxSwift 6.0 release before April 2020.

keisyrzk:
I do have the Rx v 5.0.1 and the problem stil exists. When submitting to Apple I receive the error every time. When we can expect the fix? It seems like not be fixed though.

keisyrzk:
Unfortunately I am still facing the same problem when submitting to Apple. I did the search with 'grep' command and have many results to RxCocoa. Could you please tell us if the problem is really solved by your side? 
When can we expect the new Rx lib version release? :)

freak4pc:
Hey, as other threads mention, this is just a warning. Apple will only start enforcing this in April. We'll have an official release before that, but this isn't an issue for the time being.

kamokaizen:
Is there any release for depreciated frameworks?

freak4pc:
UIWebView was already removed in 5.1.0

PR #2064, opened 2019-09-13
Rename UIKitForMac to macCatalyst

freak4pc:
Xcode 11 GM changed UIKitForMac to macCatalyst.

### DIFF
--- a/RxCocoa/Deprecated.swift
+++ b/RxCocoa/Deprecated.swift
@@ -222,7 +222,7 @@ extension ObservableType {
     }
 #endif
 
-#if os(iOS) && !targetEnvironment(UIKitForMac)
+#if os(iOS) && !targetEnvironment(macCatalyst)
     extension UIWebView {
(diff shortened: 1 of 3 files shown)


kzaher:
Maybe the simplest really would be to just bump the major version for Xcode 11 and remove this completely.

freak4pc:
@kzaher Not sure if you're still on Slack but messaged you over there with some thoughtsl. 

I guess we can possibly do that - I feel like bumping a major for so few changes is "dangerous", because it means any new change to the API will cause even more bumps. I don't want to rush features into 6.x just to get it done before a major bump, it's a bit tricky... not sure what choice we have.

But also we have some more time since it's just a warning, I'd still like to at least decide on a path of solution here so we can quickly pull it off if anything happens.

kzaher:
Hey, saw your messages there now. 

I've in general turned off all notifications on all of my devices for the past year so it might be harder to reach me :)

freak4pc:
At least I still have Slack to bug you on 🤣🤣🤣🤣

PR #2071, opened 2019-09-22
Add conversion from Maybe to Completable and Single

kokhong20:
## Description
This pull request exposed four new functions in `Maybe` trait, allow `Maybe` to be converted to `Single` and `Completable`.

## Motivation
I have encountered a limitation in RxSwift that I am not able to chain `Maybe` traits with `Single` and `Completable`. The example below demonstrate my use cases.
 
### Example
I have 3 function for getting and setting the objects.
```
func getInt() -> Single<Int>
func getInt2(_ id: Int) -> Single<Int>
func setInt(_ id: Int) -> Completable
```
This pull request enable us to do the following chaining.
```
getInt().filter { $0 == "" } // Single.filter coverts Single to Maybe trait
    .flatMapSingle(getInt2)
    .flatMapCompletable(setInt)
    .subscribe()
```


sidepelican:
related to this: https://github.com/ReactiveX/RxSwift/issues/1580

tamulis:
Is this PR still going to be merged sometime?

freak4pc:
Thanks for this PR.

1. Converting Maybe to Single shouldn't possible - there is no compatibility between them as far as I can tell. The other way can be possible. 
2. Seems like this table could help with figuring out all the possible conversions: https://github.com/ReactiveX/RxSwift/issues/1580#issuecomment-371803063
3. This is really diverged from `main` at this point. Not sure if you can rebase etc, but if yes, I'm happy to take a more serious look into it.

Thanks!

freak4pc:
Closed due to inactivity

Issue #2073, opened 2019-09-25
PublishSubject as PropertyWrapper


gringoireDM:
With 5.1 version swift got empowered with `propertyWrapper` language feature.
Thinking about common patterns where propertyWrapper could be used I started to think about `PublishSubject`.

```swift
class Foo {
    var bar: Observable<Int> { _bar.asObservable() }
    private let _bar = PublishSubject<Int>()
}
```

This example is a classic common pattern that with property wrapper would reduce the amount of boilerplate the users has to write to reach this goal.

Having

```swift
@propertyWrapper
public final class PublishSubject<Element>
    : Observable<Element>
    , SubjectType
    , Cancelable
    , ObserverType
    , SynchronizedUnsubscribeType {

    ...
    public var wrappedValue: Observable<Element> { self.asObservable() }
    ...

}
```

we can rewrite `Foo` in this way.

```swift
class Foo {
    @PublishSubject var bar: Observable<Int>
}
```

The compiler will synthesise the private var `_bar` which will be a private publish subject. The compiler will stop the users from defining bar to be anything different than an `Observable`.

unfortunately, `@propertyWrapper` can't be defined as extensions, so a support to PropertyWrapper has to be done in RxSwift and it can't be done on a third party.

What do you think?

If you like the idea I can open a PR.

gringoireDM:
Similarly BehaviourSubject and BehaviorRelay can have property wrappers too:

```swift
@propertyWrapper
public final class BehaviorSubject<Element>
    : Observable<Element>
    , SubjectType
    , ObserverType
    , SynchronizedUnsubscribeType
    , Cancelable {

    ....
    public var wrappedValue: Observable<Element> { self.asObservable() }
    public var projectedValue: Element {
        get { try! self.value() }
        set { self.onNext(newValue) }
    }
    ....
}
```

so that 
```swift
class Foo {
    var bar: Observable<Int> { _bar.asObservable() }
    let _bar = BehaviorSubject(value: 0)
    var $bar: Int { 
        get { try! _bar.value() }
        set { _bar.onNext(newValue) }
    }
}
```

would simply become

```swift
class Foo {
    @BehaviorSubject(value: 0) var bar: Observable<Int>
}
```

freak4pc:
Saw your PR on this - first of all its best to wait for feedback on something this drastic _before_ making a PR. Even though I appreciate your work, obviously :) 

[...]

PR #2074, opened 2019-09-25
Property wrappers

gringoireDM:
This PR makes `PublishSubject`, `BehaviorSubject` and `BehaviorRelay` to be propertyWrappers.

As described in #2073 this will enable users to reduce the amount of boilerplate code that would be needed while implementing common patterns involving these RxSwift types.


gringoireDM:
I see that the history is broken. closing this

PR #2075, opened 2019-09-25
PublishSubject, BehaviorRelay and BehaviorSubject are now property wrappers

gringoireDM:
This PR makes PublishSubject, BehaviorSubject and BehaviorRelay to be propertyWrappers.

As described in #2073 this will enable users to reduce the amount of boilerplate code that would be needed while implementing common patterns involving these RxSwift types.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os:
 language: generic
 sudo: required
 dist: xenial
-osx_image: xcode10.2
+osx_image: xcode11
 
 env:
   - DANGER=1
@@ -29,7 +29,7 @@ matrix:
     include:	
       - osx_image:	
         os: linux	
-        env: TEST=Unix swift=5.0
+        env: TEST=Unix swift=5.1
 
 notifications:
   slack: rxswift:3ykt2Z61f8GkdvhCZTYPduOL
--- a/RxRelay/BehaviorRelay.swift
+++ b/RxRelay/BehaviorRelay.swift
@@ -11,9 +11,16 @@ import RxSwift
 /// BehaviorRelay is a wrapper for `BehaviorSubject`.
 ///
 /// Unlike `BehaviorSubject` it can't terminate with error or completed.
+@propertyWrapper
 public final class BehaviorRelay<Element>: ObservableType {
     private let _subject: BehaviorSubject<Element>
-
+    
+    public var wrappedValue: Observable<Element> { self.asObservable() }
+    public var projectedValue: Element {
+        get { value }
+        set { accept(newValue) }
+    }
+    
     /// Accepts `event` and emits it to subscribers
     public func accept(_ event: Element) {
         self._subject.onNext(event)
--- a/RxSwift/Subjects/BehaviorSubject.swift
+++ b/RxSwift/Subjects/BehaviorSubject.swift
@@ -9,6 +9,7 @@
 /// Represents a value that changes over time.
 ///
 /// Observers can subscribe to the subject to receive the last (or initial) value and all subsequent notifications.
+@propertyWrapper
 public final class BehaviorSubject<Element>
     : Observable<Element>
     , SubjectType
@@ -44,7 +45,13 @@ public final class BehaviorSubject<Element>
     public var isDisposed: Bool {
         return self._isDisposed
     }
- 
+    
+    public var wrappedValue: Observable<Element> { self.asObservable() }
+    public var projectedValue: Element {
+        get { try! self.value() }
+        set { self.onNext(newValue) }
+    }
+    
     /// Initializes a new instance of the subject that caches its last value and starts with the specified value.
     ///
     /// - parameter value: Initial value sent to observers when no other value has been received by the subject yet.
--- a/RxSwift/Subjects/PublishSubject.swift
+++ b/RxSwift/Subjects/PublishSubject.swift
@@ -9,6 +9,7 @@
 /// Represents an object that is both an observable sequence as well as an observer.
 ///
 /// Each notification is broadcasted to all subscribed observers.
+@propertyWrapper
 public final class PublishSubject<Element>
     : Observable<Element>
     , SubjectType
@@ -44,6 +45,8 @@ public final class PublishSubject<Element>
     public var isDisposed: Bool {
         return self._isDisposed
     }
+    
+    public var wrappedValue: Observable<Element> { self.asObservable() }
     
     /// Creates a subject.
     public override init() {


freak4pc:
Let's move the discussion here and close the issue, copying my comment from there: 

First of all its best to wait for feedback on something this drastic _before_ making a PR. Even though I appreciate your work, obviously :) 

From a technical perspective this is nice, but it reminds me of #1639 where I personally had a very strong thought against hiding such implementation details. 

The fact people don't know `@Published` probably uses a CurrentValueSubject under the hood is strange to me, but here its even worse because this code base specifically had changes to remove `.value` setters, to get rid of the notion that "a stream can have a value". Streams don't have a value, only a snapshot in some specific time point. Having `Variable` before that worked with regular setters created a lot of confusion and abuse of how RxSwift can be best used. 

So, I'm torn. 

From one side - technologically this advancement of using property wrappers is really nice. BUT, it creates a massive hiding of implementation details which means reduced "technical responsibility" of how to compose things and make them work together (e.g. understand what these subjects actually do in some sense). 

All of this is speaking of Beahvior* - I don't really see how this fits Publish* stuff since they don't have a "current" notion at all (Event-based, not state-based).

Appreciate your work, and your thoughts on this :)

gringoireDM:
> The fact people don't know @Published probably uses a CurrentValueSubject under the hood is strange to me, but here its even worse because this code base specifically had changes to remove .value setters

Totally agree with that, the difference here is that this does not add a `value` setter to the behaviour subject. It's the projected value to have it, which means that actually is the class having the BehaviorSubject that has a "shortcut". We can remove the projected value from Behaviors. no big deal.

> From one side - technologically this advancement of using property wrappers is really nice. BUT, it creates a massive hiding of implementation details which means reduced "technical responsibility" of how to compose things and make them work together (e.g. understand what these subjects actually do in some sense).

The choice of using a property wrapper instead of a function is already an acknowledgement of intentions: expose an observable and hide the ObserverType from the rest of the world.

As defensive developer I feel risky to expose publish subjects as public, the way to avoid it is the well known pattern that this PR is trying to simplify the adoption. I believe that this PR would lead towards writing safer code with less effort (and unfortunately it can't be done as plugin... property wrappers can be adopted just where the object declaration is).

> All of this is speaking of Beahvior* - I don't really see how this fits Publish* stuff since they don't have a "current" notion at all (Event-based, not state-based).

The purpose is to have a private PublishSubject and a public observable for free. :) 

`@PublishSubject public var bar: Observable<Int>` will make public the observable, but private the publishSubject. to do that without property wrappers you need to double your efforts by having a publishSubject declaration as private and an observable computed properties.

gringoireDM:
Here is an example of how you would use it:

```swift
import RxSwift

class Foo {
    @PublishSubject var bar: Observable<Int>
    @BehaviorSubject(value: 0) var baz: Observable<Int>
    
    func next(_ value: Int) {
        _bar.onNext(value)
        $baz = value * 2 
        // or
        _bas.accept(value * 2)
    }
}

example("propertyWrapper example") {
    let disposeBag = DisposeBag()
    let foo = Foo()
    foo.bar
        .subscribe(onNext: { print($0) })
        .disposed(by: disposeBag)
    
     //foo._baz can't access because it's private
     //foo.$baz is an Int
     //foo.baz is an Observable<Int>
    foo.baz
        .subscribe(onNext: { print($0) })
        .disposed(by: disposeBag)

    foo.next(1)
    foo.next(2)
    foo.next(10)
}
```

freak4pc:
Regardless of the decision we make around this, I think we need different, separate types, for wrappers. I don't think we should turn the existing types into property wrappers, much like @Published wraps a different subject type. 

If we're going down that path, that is, maybe we should consider that. 

I'm a bit confused with how this: 
`@PublishSubject public var bar: Observable<Int>`

How would consumers bind into it, for example? Manually sending values to the subject is one use case but it's not the majority use case necessarily.

gringoireDM:
The property wrapper should serve cases where you want something like this:
```
class Foo {
    public var bar: Observable<Int> { return _bar.asObservable() }
    private let _bar = PublishSubject<Int>()
}
```

If you need to manually inject values to `_bar` from outside the class, then you don't need this pattern and you can use `PublishSubject` the way you always did. Usually I would say it's not desirable to manually send value to a publish subject owned by another class. Obviously it depends on the use case, but typically I would use public subject internally as implementation detail, to expose a plain observable to the rest of the world.

freak4pc:
Sure, but again what makes this implementation private ? Can’t I access $baz externally ?

gringoireDM:
the way that property wrapper works is that the compiler synthesise for you a private property which is the Wrapper class (in `baz` case the `BehaviorSubject`, in `bar` case would be `PublishSubject`). you can access the private synthesised properties by prepending `_` to the var name (so `_bar`, `_baz`). These are private, you can't access it from outside the class implementation. IF the property wrapper expose a variable called `projectedValue`, then a new property will be synthesised as public that can be accessed by prepending `$` (in my example `$baz`, PublishSubject does not have a projectedValue, so it won't have `$bar`).

So, writing
```swift
class Foo {
    @PublishSubject public var bar: Observable<Int>
}
```

is exactly like writing 
```swift
class Foo {
    public var bar: Observable<Int> { return _bar.asObservable() }
    private let _bar = PublishSubject<Int>()
}
```

And for the case of Behavior*
```swift
class Foo {
    @BehaviorSubject(value: 0)
    public var bar: Observable<Int>
}
```

the way I implemented in this PR BehaviorSubject Property wrapper, is exactly like writing 
```swift
class Foo {
    public var bar: Observable<Int> { return _bar.asObservable() }
    private let _bar = BehaviorSubject<Int>(value: 0)
    public var $bar: Int {
        get { return try! _bar.value() }
        set { _bar.onNext(newValue }
    }
}
```

Thanks to swift compiler magic I get `_bar` and `$bar` for free.

If we don't like `$bar` we can remove the projected value from the Behavior* objects so that we get for free just the private Behavior*

freak4pc:
Yup, understood. Despite that, I still feel like the property wrapper shouldn't be the type itself (like Published isn't the Subject but only a wrapper). 

Maybe a more proper idea here would be to make separate types that wrap these subjects in a similar fashion to the way you suggested (this also means that if you'd like, you can do it yourself even if it's not part of this codebase)

Also, not sure if a property-wrapper for something that inherently doesn't have a value necessarily makes sense (e.g. Publish*). The idea you've brought force makes PublishSubject a propertyWrapper for a very specific use case, a case that isn't named properly right now (e.g. the subject's name) and isn't intuitive in my personal opinion.

gringoireDM:
I understand your point, and it makes sense, even if to make a wrapper around Publish subject would mean basically to rewrite publish subject. The wrapper type would still need to be an observable and observer type.

Do you want to wait for other feedbacks or do you prefer I close this PR?

carlosypunto:
@gringoireDM, maybe applying the `@propertyWrapper` to subclasses of `BehaviorRelay`, `BehaviorSubject` and `PublishSubject` with other names.

carlosypunto:
being final classes, the only thing that can work is to make the current private ones, not final and renowned (a _ at begining). Make a subclass for each final and public. One with actual name and the other with the `@propertyWrapper` chosen name. I don't know what to say the team

gringoireDM:
@carlosypunto that would work. @freak4pc what do you think?

gringoireDM:
after all, being non final but still public will prevent subclassing outside of RxSwift library (you need open for that). I actually think it's a really good idea.

freak4pc:
Sorry, but I'm a tad confused how would that help. 
@carlosypunto are you referring to something like this? 

```swift
private class _PublishSubject ...

public final class PublishSubject: _PublishSubject { ... }

@proeprtyWrapper
public class PublishWhatever: _PublishSubject { ... }
```

gringoireDM:
I think it's more like

```swift
public class PublishSubject { ... }

@proeprtyWrapper
public final class PublishWhatever: PublishSubject { ... }
```

It would help, because this way `PublishWhatever` don't have to implement PublishSubject again.

freak4pc:
Right, so wasn't sure what was the private offer around here. Regardless, good names would be hard for these wrappers. 

Publish models events where Behavior models state — so maybe something around that, but that might be confusing. 

If it's additive and not replacing the existing types, I think this could work possibly.

gringoireDM:
The existing `PublishSubject` should pass from being final to non final. Which is not API breaking.

Now, let's find a good name for the wrapper. :D That's going to be the toughest part.
`@PublishedObservable`? `@PrivatePublishSubject`?

yuriferretti:
@gringoireDM 

First of all congratulations for bringing this up! I think a well crafted implementation can make our life way easier when dealing with this kind of pattern you are trying to solve and is/should be very common in well designed Rx APIs.

I have some doubts about the implementation that I would like you to clarify in order to help you improving the final result:

1. Why is required to make the current subject classes/subclasses the wrappers themselves?
2. How using the compiler generated `_storage`, that exposes the subject will behave in subclasses?

Please, let me know you thoughts sou I can try helping you with some nice suggestions =]

carlosypunto:
Yes @freak4pc  , my approach was like your, but being PublishWhatever final too.

gringoireDM:
> Why is required to make the current subject classes/subclasses the wrappers themselves?

Because any new wrapper type should behave like `PublishSubject`. Also having a publish subject under the hood, it will have a bunch of methods and variables that proxy the underlying `PublishSubject`. Not a very elegant solution.

> How using the compiler generated _storage, that exposes the subject will behave in subclasses?

It's private, so you can't use it in subclasses, which is the correct behaviour. The purpose is to hide implementation details. You shouldn't be able to alter the stream from outside the class.

yuriferretti:
> Because any new wrapper type should behave like `PublishSubject`. Also having a publish subject under the hood, it will have a bunch of methods and variables that proxy the underlying `PublishSubject`. Not a very elegant solution.

Got it. However subclassing would allow declarations like the following:
```swift
@PublishWhatever var p1: Observable<X>
let p2: PublishWhatever<X> = PublishWhatever()
let p3: PublishSubject<X> = PublishWhatever()
```
IMO Rx already have a very steep learning curve for people migrating from imperative to declarative world and those subjects can many times be used to cross the line between both worlds, and in this case more options would make choosing the right one even more difficult and confuse.

Another point is that so far I haven't seen a single implementation that uses `@propertyWrapper` as the type itself like this one. Right now they lean towards wrapping values and given them some sort of super power or default behavior, please take a look at [this cool repo](https://github.com/guillermomuntaner/Burritos).  Also I'm not telling this is wrong, just unusual.

> It's private, so you can't use it in subclasses, which is the correct behaviour. The purpose is to hide implementation details. You shouldn't be able to alter the stream from outside the class.

I think this PR is using the wrong tool for the right job, property wrappers aren't an abstraction mechanism. IMHO they're more about back storage behavior and the issue you are trying to solve arises because swift compile doesnt allow covariance when fulfulling protocol requirements, e.g.

```swift
protocol Proto {
    var obs: Observable<Int> { get }
}

class SomeProto: Proto {
     // PublishSubject<Int> is an Observble<Int> but compiler complains because it isn't 
     // an exact match
     let obs: PublishSubject<Int>
}
```

 Also this relies in the current private access of `_storage`as a way of "hiding implementation details", but one of the [future directions](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md#finer-grained-access-control) already provides a glimpse to more finner grained access control that can broke this assumption.

Despite of those problems I'm still feeling that this PR has much value and it should be implemented without subclassing in mind.

My suggestion is the following:
```swift
// I didn't thinking about naming
@propertyWrapper
struct PublishBacked<Element> {
    private let subject = PublishSubject<Element>()
    var wrappedValue: Observable<Element> { subject }
    var projectedValue: PublishSubject<Element> { subject }
}

@propertyWrapper
struct BehaviorBacked<Element> {
    private let subject: BehaviorSubject<Element>
    var wrappedValue: Observable<Element> { subject }
    var projectedValue: BehaviorSubject<Element> { subject }
    
    init(value: Element) {
        subject = .init(value: value)
    }
}
```
Which allows:
```swift
protocol Proto {
    var observed: Observable<Int> { get }
}

class SomeProto: Proto {
    
    // compiler doesn't complains and Proto requirements are fulfilled
    @BehaviorBackedObservable(value: 10) var observed: Observable<Int>
    
    func someMethodThatEmits() -> Observable<String> {
        // use as it is, an observable
        return observed.map { $0.description }
    }
    
    func someMethodThatSendEvents(value: Int) {
        // use the backing subject
        $observed.onNext(value)
    }
    
    func someMethodThatSubscribes(observable: Observable<Int>) {
        // allows subscription, its projection is a common subject
        _ = observable.subscribe($observed)
    }
}
```
Also other wrapper types can be added like `Replay`, `Async` and so on.

IMHO the advantages of this approach are:
* doesn't require subclassing which means that it can be copied to the code base right now, or even to a dedicated wrapper package that RxSwift can provides in the future.
* uses the `$` syntax to access the backing subject and inform something different is happening like [@State](https://developer.apple.com/documentation/swiftui/state), [@Binding](https://developer.apple.com/documentation/swiftui/binding)
* doesn't relies on `_storage` that will receive more granular access control in the future which can make it private, the same way it will allow hiding the `$projectedValue` and make `SomeProto` usesless.

What do you think @gringoireDM @freak4pc ?

gringoireDM:
> Got it. However subclassing would allow declarations like the following:
> ```
> @PublishWhatever var p1: Observable<X>
> let p2: PublishWhatever<X> = PublishWhatever()
> let p3: PublishSubject<X> = PublishWhatever()
> ```

I'm not following. 

> IMO Rx already have a very steep learning curve for people migrating from imperative to declarative world and those subjects can many times be used to cross the line between both worlds, and in this case more options would make choosing the right one even more difficult and confuse.

This should not stop re-innovation. With an evolving language we should evolve too with latest advance and features. There is a steep learning curve in advanced swift programming too, but this is not stopping the community from introducing Function builders for DSL, and other advanced topics.

> Another point is that so far I haven't seen a single implementation that uses @propertyWrapper as the type itself like this one. Right now they lean towards wrapping values and given them some sort of super power or default behavior, please take a look at this cool repo. Also I'm not telling this is wrong, just unusual.

Again, this doesn't mean that they are forbidden. If you follow the swift forum discussion around `@propertyWrapper` you will see that `projectedValue` was introduced specifically for these use cases were you might not want the public `$foo` in order to expose the public `wrappedVale` and **not** the property wrapper.

> I think this PR is using the wrong tool for the right job, property wrappers aren't an abstraction mechanism. IMHO they're more about back storage behavior and the issue you are trying to solve arises because swift compile doesnt allow covariance when fulfulling protocol requirements, e.g.

Wrong. Property wrappers are born to allow synthesis of commonly repeated patterns. You can read that from the first lines of the official [swift community proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0258-property-wrappers.md):

> There are property implementation patterns that come up repeatedly. Rather than hardcode a fixed set of patterns into the compiler (as we have done for lazy and @NSCopying), we should provide a general "property wrapper" mechanism to allow these patterns to be defined as libraries.

Which is exactly what this PR is about: taking a common pattern and create a property wrapper mechanism to generalise it.

Notice that in the examples of the proposal there is also our beloved `BehaviorRelay` here proposed in this PR with a divergent wrappedValue.

> Also this relies in the current private access of _storageas a way of "hiding implementation details", but one of the future directions already provides a glimpse to more finner grained access control that can break this assumption.

Wrong again. Future direction will always have to take in consideration backwards compatibility, meaning that the future will give us (like specifically stated in document) "finer control" which means that you will be able to explicitly define the access control of synthesised properties. For this PR purposes, future directions won't affect our purposes: private PublishSubject, public Observable.

> My suggestion is the following:

projectedValue is public for now. We can do it when (if) we will ever gain finer control over access level of the projected value which we don't have.

The purpose of this PR is to use at best a tool and allows new comers to easily implement safer patterns. To expose a publish subject is generally bad. it means that external classes might be able to inject values in a sequence that is meant to be controlled by the class that is exposing it. Sometimes this happens for laziness in exposing the computed observable property for each publish subject you have in your code. In cases where you have a mix due to transition from imperative to reactive programming, these use cases are very frequent. New comers won't think about the dangers of exposing a PublishSubject. This PR helps new comers to implement a safer pattern that will also bring them to the light of learning that to alter a stream is pretty often bad, and they should rethink the way they are using RxSwift before blaming the framework for their frustration whenever they lose control over their streams and run in racing conditions.

> doesn't relies on _storage that will receive more granular access control in the future which can make it public, the same way it will allow hiding the $projectedValue and make SomeProto usesless.

Again, "receive more granular access control" doesn't mean that the swift community might turn it to be public. it will never happen. it would break existing contract and we don't want that. "receive more granular access control" means that in the future you as writer of the property wrapper might be able to have any combination of public, internal, private for get and set of `wrappedValue` and `projectedValue`. I know that because I followed the whole discussion on the swift forum and I was one of the first to raise my hand and talk about public vs private.

yuriferretti:
> I'm not following.

If you use the wrapper as a wrapper it doesn't allow the following:

```swift
@propertyWrapper
struct Wrapper<T> {
    let wrappedValue: T
}
@Wrapper v1: Int = 10
let v2: Int = Wrapper(wrappedValue: 10) // this is not possible
```
But adding the wrapper as subclasse will allow the construct above and use the wrapper as the subject and make the following possible:
```swift
 @PublishWhatever var p1: Observable<X>
 let p2: PublishWhatever<X> = PublishWhatever()
 let p3: PublishSubject<X> = PublishWhatever()
 let p4: Observable<X> = PublishWhatever<X>()
```
Who is the wrapper?
Who is being wrapped?
Don't you think Rx already have a lot of subjects and adding more with different behaviors and syntax depending on where and how you declare them them will benefit this library?


> This should not stop re-innovation. With an evolving language we should evolve too with latest advance and features. There is a steep learning curve in advanced swift programming too, but this is not stopping the community from introducing Function builders for DSL, and other advanced topics.

I'm not saying to stop innovation here. BUT given RxSwift is not our private playground we should adopt the new technologies in a meaningful way, observing all trade-offs and picking the right tools for the right jobs, and not just using them because they are the latest and most hyped.

> Exactly because they are newcomers, we should promote safer patterns since the beginning making them easy to implement so that they can learn and find smaller walls to climb instead of blaming the framework or the paradigm because they are using PublishSubject the wrong way and loose control over their streams.

And you really think adding a new `@PublishWrapper` that in fact is a masked subject and force its manipulation by using `_storage` will grant all this control over the streams? I'm seeing that this is adding more choices that can be exposed the same way and just enlarging the wall for newcomers to climb.


> Again, this doesn't mean that they are forbidden. 

Yeah that's why I clearly wrote `ununsual`.


> If you follow the swift forum discussion around `@propertyWrapper` you will see that `projectedValue` was introduced specifically for these use cases were you might not want the public `$foo` in order to expose the public `wrappedVale` and **not** the property wrapper.

I think you are mixing things here......
Quoting directly the `projection` definition from the propetyWrappers proposal:

> A property wrapper type can choose to provide a projection property (e.g., $foo) to expose more API for each wrapped property by defining a projectedValue property.
As with the wrappedValue property and init(wrappedValue:), the projectedValue property must have the same access level as its property wrapper type.

I just can't find anything that states that the `projectedValue` should be used only in the case you mentioned. However what I can find is that  right now`projectedValue` MUST have the SAME access level as the `wrappedValue`. So it clearly is not meant to circunvent any access level issues it only allows more API to be exposed to `wrappedValue` by using `$` 

> Notice that in the examples of the proposal there is also our beloved BehaviorRelay here proposed in this PR with a divergent wrappedValue.

I have no problem with divergent `wrappedValues`, I also think the vast majority of then, if not all, will have divergent `wrappedValue`
 
I suggest you exposing all your quotes and citations to allow other people that don't follow Swift evolution to read and analyze them in full context.

Here's the example you mentioned:
```swift
@BehaviorRelay
var myValue: Int = 17

let observer = $myValue.subscribe(...)   // subscribe an observer
$myValue.accept(42)  // set a new value via the synthesized storage property

print(myValue)   // print the most recent value
```
Which by the way is a beautiful and meaningful usage of wrappers in Rx that doesn't tryes to use wrappers as abstraction mechanism it's only wrapping an `Int`. 

> Wrong again. Future direction will always have to take in consideration backwards compatibility, meaning that the future will give us (like specifically stated in document) "finer control" which means that you will be able to explicitly define the access control of synthesised properties. For this PR purposes, future directions won't affect our needs: private PublishSubject, public Observable.

I think you are being selectively negligent here given you are advocating in favor of providing a "safer" pattern.  When the finner access control comes to light everyone will be able to declare:

```swift
@PublishSubject 
 public(storage) private(projection) foo: Observable<X>
```
And then I will start seeing again the same patterns that this PR is trying to avoid because user will be able to access the subject using `_storage`.

> The purpose of this PR is to use at best a tool and allows new comers to easily implement safer patterns. To expose a publish subject is generally bad. it means that external classes might be able to inject values in a sequence that is meant to be controlled by the class that is exposing it. 

Again! Why do you need a new wrapper/class that is not different in any way from a vanilla subject in order to enable a safer pattern?
Wouldn't be easier to guide newcomers in how to use a `protocol` and why in general is good to hide implementation details from consumers being this `protocols` requirements related or not to Rx, subjects and etc.?

> Sometimes this happens for laziness in exposing the computed observable property for each publish subject you have in your code. In cases where you have a mix due to transition from imperative to reactive programming, these use cases are very frequent. 

Yeah and this PR is adding at least more 3 subject choices to newcommers at this imperative declarative border as I mentioned before. IMO, in this specific context, more choices will turn into  more doubts, more insecurities and more problems. 
Also deprecating `Variable` and moving relays to a separate package put RxSwift in the path of reducing the choices and errors caused by misuse and abuse of those tools, which is the total oposite adding more subject subclasses.

> This PR helps new comers to implement a saffer pattern that will also bring them to the light of learning that to alter a stream is pretty often bad, and they should rethink the way they are using RxSwift before blaming the framework for their frustration whenever they lose control over their streams and run in racing conditions.

I just can't see how just adding a wrapper will teach a newcommer why expose a Subject is generally bad. 

OTOH What I can see is Rx providing source code that teaches how to use inheritance + compiler synthesis to expose a type as its supertype, which in fact can lead to property wrapper abuse in a more broad manner that surpasses Rx scopes.

> Again, "receive more granular access control" doesn't mean that the swift community might turn it to be public. 

Perfect! I think I couldn't express myself clearly here.
But In you own words:

> Future direction will always have to take in consideration backwards compatibility, meaning that the future will give us (like specifically stated in document) "finer control" which means that you will be able to explicitly define the access control of synthesised properties.

And being repetitive: this finner access control will enable us hidding the `$` projection **if desired**. In other words the `$` cannot become more exposed than it is right now.  However the same improvements in access control will allow anyone who declares any wrapped property to expose the `_storage` **if desired**, the same way they already can expose the subjects right now.
In summary this PR trusts in a restriction that anyone will be able to lift in the near future.

> I know that because I followed the whole discussion on the swift forum and I was one of the first to raise my hand and talk about public vs private.

So you will also be the first one to notice that relying on current state of wrappers access control in an attempt to avoid subjects abuse can backfire in the near future and let us with useless wrappers and classes poluting the namespace. 

In addition the problems you are mentioning are caused by the lack of knowledge about `Subjects`, `Streams` and Rx stuff. Those errors will keep arising with or without this wrapper if users don't know when and how to use a subject, and IMHO adding more of them doesn't seems to help.

gringoireDM:
> If you use the wrapper as a wrapper it doesn't allow the following:
> ```swift
> @propertyWrapper
> struct Wrapper<T> {
>     let wrappedValue: T
> }
> @Wrapper v1: Int = 10
> let v2: Int = Wrapper(wrappedValue: 10) // this is not possible
> ```

Ehmm... a property wrapper is still a type and can be initialised. of course it is still possible. Useless, but possible. open any playground and write `var state = State(initialValue: 10)` the compiler won't stop you.

> Who is the wrapper?
Who is being wrapped?
Don't you think Rx already have a lot of subjects and adding more with different behaviors and syntax depending on where and how you declare them them is not going to harm this library?

This PR is not introducing any new Subject.

> I'm not saying to stop innovation here. BUT given RxSwift is not our private playground we should adopt the new technologies in a meaningful way, observing all trade-offs and picking the right tools for the right jobs, and not just using them because they are the latest and most hyped.

This **is** the right tool for what I can do. The swift Core team says that in the very first 3 lines of why wrappers are introduced in the language. I identified a common pattern and I generalised it to make it easy to use.

> And you really think adding a new @PublishWrapper that in fact is a masked subject and force its manipulation by using _storage will grant all this control over the streams?

The compiler will make sure that this is the case. If you try to access the private synthesised property **your code just won't compile**. If the publish subject is private, you can alter the stream just within the class/struct scope that implements it. End of the story. can't escape it. It is as simple as that: private can't be accessed from outside.

> Yeah that's why I clearly wrote ununsual.

Why? because you didn't see any open sourced? It's something that the core team thought possible otherwise the projectedValue was a mandatory requirement. don't see the problem to be honest. It is possible, it does the job, it does it well... can't see any reason why that would be unusual...

> I just can't find anything that states that the projectedValue should be used only in the case you mentioned. However what I can find is that right nowprojectedValue MUST have the SAME access level as the wrappedValue. So it clearly is not meant to circunvent any access level issues it only allows more API to be exposed to wrappedValue by using $

Never mentioned how projectedValue should be used. I said why it was introduced. In early stages of propertyWrapper design the projectedValue was inexistent and the famous `$` property was in fact the propertyWrapper itself. People complained saying"I might want to hide access to the wrapper" then the projectedValue was added, and the documentation says "**you can** (not you must) add a projectedValue" ultimately the piece of proposal you quoted says ** to expose more API for each wrapped property by defining a projectedValue property.** which is not what I want to do here.

For the rest I deserve the right to say that we can agree to disagree. I'm tired. We are going around saying the same over and over and to be honest I don't fancy to repeat something hoping to get a different result. We both expressed our point of view. Thank you for your time.

yuriferretti:
Let's make the things more objective here to avoid killing this idea.

Things that I totally agree with you:
1. I would be nice to have a way to use a subject and expose it as an observable without the extra typing we need today;
2. Wrappers does seem to make it more approachable;

Things that I don't agree with you:
1. Trusting only in the current locked private access of `_storage` will make everything safer forever.
2. Subclassing is the way to go because it forces the `_storage` to be private

Why?:
Currently the state of wrappers can make subclassing appealing because `_storage` access is private. However is possible that swift will provide ways to make those storages public and all the effort will be easily circumvented. As you said, people are lazy and would do the minimum effort to achieve their goals which could be enabling the public storage in the wrappers declaration. This way we'd be back in the exposed subject jungle and with a lot of `_storage` access from outer classes.

What I suggested you is to wrap the subject in a pure wrapper that is not a subclass. This wrapper would be totally useless as a type itself as you mentioned, so no confusion will happen when choosing which subject to use because it is meant to be solely a wrapper.

Finally what all of us would have to do and I'm 100% sure you already do is to make those classes conform to a protocol. Then the observable that's wrapped satisfies the conformance and you have a more future proof solution that you can broadly use right now without having to wait for the next Rx release, get the results, improve it based on that and then add it to Rx repo confident of the great job you did.

Also, I think you should reopen this PR! It's valuable! Take the `propertyWrapper` proposal itself as example...It had a lot of back and forth and discussions that led to more refined design. I'm just a single person that disagrees with you in some points, but I think there are a lot of other people that can have different ideas then we did, or can agree with the current design and I have no problem with that. ✌️

PR #2077, opened 2019-09-26
Move to Swift 5.1 / Xcode 11 

freak4pc:
This bumps minimum Swift version to 5.1, updates CI, removes all deprecated methods (as we don't need to support if we're bumping major), and removes some unnecessary return statements as part of SE-0255. 

SPM-related tests are still failing, seems the v5_1 version isn't recognized for some reason. Could be wrong version of SPM, will need to check.


Package.swift
```diff
@@ -68,5 +68,5 @@ let package = Package(
     ],
     Target.allTests()
   ] as [[Target]]).flatMap { $0 },
-  swiftLanguageVersions: [.v5]
+  swiftLanguageVersions: [.v5_1]
```
gringoireDM: for swift 5.1 there is not a new language version (odd... but that's it), so languageVersions can remain v5. I guess that the constraint will come from swift-tools-version.

RxBlocking.podspec
```diff
@@ -26,5 +26,5 @@ Waiting for observable sequence to complete before exiting command line applicat
   s.exclude_files         = 'RxBlocking/Platform/**/*.swift'
 
   s.dependency 'RxSwift', '~> 5'
-  s.swift_version = '5.0'
+  s.swift_version = '5.1'
```
gringoireDM: I'm not sure it's needed

Tests/RxCocoaTests/NSTextView+RxTests.swift
```diff
@@ -12,7 +12,7 @@ import AppKit
 import XCTest
 
 final class NSTextViewTests: RxTest {
-    static let timeout: TimeInterval = 0.5
+    static let timeout: TimeInterval = 2
```
gringoireDM: why was this change needed?

RxBlocking.podspec
freak4pc: How come? We're targeting Swift 5.1 and up.

Tests/RxCocoaTests/NSTextView+RxTests.swift
freak4pc: Seems the time the control takes to focus is longer when I ran the tests on Xcode 11, so i increased the timeout.

Package.swift
freak4pc: Yup, noticed this yesterday. Unintuitive but I guess can make some sense :)

RxBlocking.podspec
gringoireDM: My understanding is that this value is going to set the equivalent SWIFT_VERSION as build setting. There is no 5.1 in Xcode build settings. Apparently, if your project is compatible with swift 5, and you are running xcode11, then you are on swift 5.1;

From cocoa pods documentation:
> Note The Swift compiler mostly accepts major versions and sometimes will honor minor versions. While CocoaPods allows specifying a minor or patch version it might not be honored fully by the Swift compiler.

So I would say that I might be wrong and it's ok to put 5.1, even if Xcode is going to ignore it and mark it simply as swift 5... I wonder if there is a way to constrain to Xcode 11.... doing some research.

gringoireDM: didn't find a way to constrain Xcode version... I'd say that 5.1 is probably fine.

gringoireDM: I searched, :D I swear I did, but I didn't find a way to impose xcode11 specifically. (which I suppose it also makes sense). I would say that we can keep 5.1 in the spec, Xcode will probably honour 5.0 requirement, but at the very least the spec is clear about requirements.

freak4pc:
Ok, Unix tests are good now. 

Interestingly, URLSession and other APIs have moved from Foundation to a framework called FoundationNetworking: https://github.com/apple/swift-corelibs-foundation/blob/a264bfd40af8412161cdc05be11c1cf35f6a56f1/Docs/ReleaseNotes_Swift5.md#dependency-management

@kzaher I've commented a test here that seems to deadlock on Unix. I really don't have enough knowledge to debug this but it seems only the first concurrency task actually runs. Any ideas? This PR is blocking any other PR related to this release. 

Commented out: 
https://github.com/ReactiveX/RxSwift/pull/2077/commits/afeef37ae815479bb95018e2532fd6844a8af86f#diff-504a8e9a107d0a1ece221fa82cce0a4bR1050

Deadlocked test:
https://github.com/ReactiveX/RxSwift/blob/cce95dd704bc08cd3d69c087a05a6fc3118e2722/Tests/RxSwiftTests/Observable%2BObserveOnTests.swift#L444-L493

kzaher:
Hi @freak4pc ,

I don't think we can simply remove that test. As far as I can tell this is not a deadlock but a real regression in either OperationQueue or in OperationQueueScheduler on Linux.

Here is a proof.
https://github.com/ReactiveX/RxSwift/tree/debugging-concurrent-scheduler-hang

I've added prints. `let concurrent = { () -> Disposable in` is not being executed twice when doing `_ = scheduler.schedule((), action: concurrent)` twice in succession.

If you just checkout ed90e60a you will see it prints.
```
Before
Dispatching 1
Dispatching 2
Dispatching 3
Started
Waiting
```

If there were 2 threads concurrently executing `concurrent` there would be two `Before` prints. But there aren't.

If somebody has more information that would be great.

I definitely don't see a reason to remove out tests, because as far as I can tell it is working properly.

If we find that this is Apple's bug with `OperationQueue` we should put a conditional compilation to use normal `ConcurrentDispatchQueueScheduler` vs `OperationQueueScheduler` for this version of Swift compiler and Linux os.

kzaher:
Oh, one can easily test this on https://github.com/ReactiveX/RxSwift/tree/debugging-concurrent-scheduler-hang if they have docker locally.

`docker pull swift`
`./scripts/test-linux.sh`

freak4pc:
OK @kzaher - seems you're right, this is definitely a regression in OperationQueue on Linux.

I've opened a bug on the Swift compiler with a clean reproduction (without Rx) and added a workaround for the time being as we discussed: https://bugs.swift.org/browse/SR-11647

kzaher:
@freak4pc  I can't explain why cocoapods fail with your branch. 

Develop branch succeeded 13 days ago.
https://travis-ci.org/ReactiveX/RxSwift/builds/600083555

You probably changed something regarding CocoaPods configuration in this branch.

freak4pc:
I made no changes to CocoaPods configuration. I think Travis used to hold some sort of cache perhaps and it was cleared. It seems `pod repo update` basically does nothing and I'm not sure why.

freak4pc:
@kzaher This finally works. Would be great to merge this so we can start accumulating more content on top of RxSwift 6.0.0-beta.1

freak4pc:
Hey @kzaher - I know you're very busy, but wondered if you have time to look into this? I'd like to accumulate as much content as possible into a releasable beta

RxCocoa/Foundation/URLSession+Rx.swift
```diff
@@ -7,18 +7,26 @@
...
 import struct Foundation.TimeInterval
-import class Foundation.HTTPURLResponse
-import class Foundation.URLSession
-import class Foundation.URLResponse
 import class Foundation.JSONSerialization
 import class Foundation.NSError
 import var Foundation.NSURLErrorCancelled
 import var Foundation.NSURLErrorDomain
 
+#if canImport(FoundationNetworking)
```
kzaher: I've been reading about this, geez. I don't know how does this affect ABI. Does this mean that if this becomes magically available on {i,mac,tv,watch}OS that it will break our ABI?

If we don't have a better workaround, it's probably fine.

scripts/validate-podspec.sh
```diff
@@ -16,9 +16,17 @@ function cleanup {
...
+    COCOAPODS_REPO_COUNT=`bundle exec pod repo list --count-only | cut -c 1`
+
+    if [[ $COCOAPODS_REPO_COUNT == "0" ]]; then
+        # No repo spec yet, we have to set it up
+        bundle exec pod setup
+    else
+        # We have some remote repo (cached or otherwise), so we can simply update it 
+        bundle exec pod repo update
+    fi
+fi
```
kzaher: Interesting, glad you've been able to solve it :)

RxCocoa/Foundation/URLSession+Rx.swift
freak4pc: I'm really not sure, but there's no other solution right now. 
About ABI stability, @keith has a PR around this, starting with removing all of these individual imports in favour of importing all of Foundation. WDYT of that? https://github.com/ReactiveX/RxSwift/pull/2103

PR #2082, opened 2019-10-03
DisposeBag function builder initializer

freak4pc:
This is blocked until we merge #2077. 

With function builders, we can enhance the standard initializer to get a nicer SwiftUI-like syntax, as follows:

```swift
self.disposeBag = DisposeBag { 
    thing.bind(to: otherThing)
    driver.drive(to: textField.rx.text)
    observable.debug("x").subscribe()
}
```


LucianoPAlmeida:
This is awesome @freak4pc \o/ 
Just a question is if this wouldn't need some kind of availability check like if `swift >= 5.1`?

freak4pc:
RxSwift 6 will support Swift 5.1 and up only :)

yuriferretti:
@freak4pc that's very nice addition! What do you think also adding a new method on `DisposeBag` that accepts the same `functionBuilder` syntax?

Like:
```swift
// didn't thhink about naming
disposeBag.disposing {
    thing.bind(to: otherThing)
    driver.drive(to: textField.rx.text)
    observable.debug("x").subscribe()
}
```
This way we can leverage the easier syntax with already created dispsoe bags....

Tests/RxSwiftTests/DisposeBagTest.swift
```diff
@@ -73,7 +72,32 @@ extension DisposeBagTest {
...
+}
+
+// DisposeBag function builder initializer tests
+extension DisposeBagTest {
+  func testDisposableBuilder() {
+    let disposable1 = TestDisposable()
+    let disposable2 = TestDisposable()
+    let disposable3 = TestDisposable()
+
+    var disposeBag = DisposeBag {
```
LucianoPAlmeida: @freak4pc Just an idea that this might open the possibility of having a syntax like this for `zip`, `combineLastest` ... and so on? 
Maybe don't add much value, but it could feel nice to have something like
```swift
let observable = Zip {
   observable1
   observable2
}
```

freak4pc: Interesting in general. I think @MortyMerr's idea of leveraging an ArrayBuilder would be good because it would mean having a possibility to have this for many things. I'll make some attempts :)

MortyMerr: What about it? Can I open separate PR for it with adding function builder overload for all array accepting methods?

freak4pc: No need as I already have a PR for dispose bag.

MortyMerr: I mean, what about other operators which accepts array as a parameter (like zip, merge)? Will it be ok to add function builder overload for them?

freak4pc: Let's think about this separately. After re-thinking this yesterday I'm wondering if it's really useful, since most use cases don't really do so much multi-line, where for a `DisposeBag` probably all cases do. I guess we _could_ provide both options, but I'm not sure it would provide any specific value.

MarcSteven:
@freak4pc Hi when do you release RxSwift 6? Thank for your sharing and contributions.

MortyMerr:
Hey @freak4pc please check out this discussion before merging that https://twitter.com/M0rtyMerr/status/1180419001689657345
It looks like using propertyWrapper just to reduce commas number (passing variadic vs passing array) is not best practice.

freak4pc:
Hey @MortyMerr - saw your tweet, but not sure how that denotes this isn't a best practice? Regardless I don't think there's a "practice" issues here as it's just syntactic sugar. Maybe I misunderstood the tweet.

Tests/RxSwiftTests/DisposeBagTest.swift
freak4pc: I thought about this some more — I don't want to over-abuse this for things like operators (because it's too wide). 

For DisposeBag I think it's mostly OK since it's just syntactic, there are no multiple types in the array (only `Disposable`) so the tweet @MortyMerr mentioned is not an issue really cc/ @LucianoPAlmeida

freak4pc:
I'll be adding this in the next version, once I patch up these conflicts :)

freak4pc:
This will be part of RxSwift 6 (877ad35b)

PR #2083, opened 2019-10-05
RxTest function builder

MortyMerr:
Inspired by https://github.com/ReactiveX/RxSwift/pull/2082/files
Just a proposal to use more Swift sugar (less commas).
Sadly we can't omit `Recorded` type, but we can use factory methods.

### DIFF
--- a/Tests/RxSwiftTests/Observable+CombineLatestTests.swift
+++ b/Tests/RxSwiftTests/Observable+CombineLatestTests.swift
@@ -463,11 +463,11 @@ extension ObservableCombineLatestTest {
         for factory in factories {
             let scheduler = TestScheduler(initialClock: 0)
 
-            let e0 = scheduler.createHotObservable([
-                .next(150, 1),
-                .next(210, 2),
-                .error(220, testError1),
-                ])
+            let e0: TestableObservable<Int> = scheduler.createHotObservable {
+                Recorded.next(150, 1)
+                Recorded.next(210, 2)
+                Recorded.error(220, testError1)
+            }
             
             let e1 = scheduler.createHotObservable([
                 .next(150, 1),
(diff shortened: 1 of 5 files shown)


Tests/RxSwiftTests/Observable+CombineLatestTests.swift
```diff
@@ -675,14 +675,16 @@ extension ObservableCombineLatestTest {
...
-                .next(230, 4 + 5),
-                .next(235, 4 + 6),
-                .next(240, 4 + 7),
-                .completed(250)
-            )
+            let messages: [Recorded<Event<Int>>] = Recorded.events {
+                // Sadly we can't ommit `Recorded` here
+                Recorded.next(220, 2 + 3)
+                // but we can use deprecated factory methods
+                next(225, 3 + 4)
```
freak4pc: These methods will be removed in RxSwift 6, plus we don't want to encourage using deprecated methods.  
Wondering if there's a way to do this so you won't have to repeat `Recorded` every time 🤔

MortyMerr: BTW, u can use same generic builder in ur disposable pr, no need to specialize for every concrete type
Yes, it’s clear about deprecation, I just wanted to show all options (if user don’t want to repeat Recorded, he can wrap it into factory method) 
Looks like, omitting enum type is incorrect syntax for now. Maybe, you know someone who is in the context of swift evolution progress. I wonder if it’s a bug or feature

freak4pc: Yeah I saw that, it's an interesting though. I think I might do that. Thought it would be nice to have a specific builder for each usage, but you're right in that its generic enough.

MortyMerr: In modern swift everything is function builder 😀
Did I get it right - we won’t proceed with this idea until we can omit enum type?

freak4pc: Yes, I think it makes things more verbose, so the sugar isn’t really too sweet 😂

PR #2087, opened 2019-10-09
Fix use of articles in the ControlProperty / ControlEvent section

afonsograca:
Just a minor documentation update.


freak4pc:
Thanks !

PR #2088, opened 2019-10-11
Sweeten binds between observers and observables

sebastianvarela:
When someone want to bind an `observer` to an `observable` have to do:
```swift
observable
    .bind(to: observer)
    .disposed(by: disposeBag)
```

with these sweet operators, just have to do:
```swift
observer ~> observable ~~> disposeBag
``` 
or
```swift
disposeBag <~~ observer ~> observable
```
or
```swift
disposeBag <~~ observable <~ observer
```

You can also use a closure or a func to just observe an observable by using:
```swift
observable ~> { value in
    // Do something with value
} ~~> disposeBag
```
or 
```swift
func updateValue(value: Any) {
    // Do something with value
}
observable ~> updateValue ~~> disposeBag
```

### DIFF
--- a/Rx.xcodeproj/project.pbxproj
+++ b/Rx.xcodeproj/project.pbxproj
@@ -792,6 +792,11 @@
 		ECBBA59E1DF8C0D400DDDC2E /* RxTabBarControllerDelegateProxy.swift in Sources */ = {isa = PBXBuildFile; fileRef = ECBBA59D1DF8C0D400DDDC2E /* RxTabBarControllerDelegateProxy.swift */; };
 		ECBBA5A11DF8C0FF00DDDC2E /* UITabBarController+RxTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ECBBA5A01DF8C0FF00DDDC2E /* UITabBarController+RxTests.swift */; };
 		ECBBA5A21DF8C0FF00DDDC2E /* UITabBarController+RxTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ECBBA5A01DF8C0FF00DDDC2E /* UITabBarController+RxTests.swift */; };
+		F2688C722351231C0043FA21 /* ObservableBinding.swift in Sources */ = {isa = PBXBuildFile; fileRef = F2688C712351231C0043FA21 /* ObservableBinding.swift */; };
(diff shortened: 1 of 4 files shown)


freak4pc:
Thanks for your first contribution here @SebastianOsinski ! I hope to see more of your code in our repo :) 

This is an interesting thought similar to RAS's <~ binding operator. I think there have been several discussions around this here that mention it doesn't really fit into the RxSwift core repo. I'm trying to think if it could fit into RxSwiftExt, or some other portion. 

Of course if there is some serious public demand around adding these, we should discuss further, but remembering previous similar PRs, this is a relatively lesser wanted feature, as custom operators usually cause some confusion :)

SebastianOsinski:
@freak4pc wrong Sebastian 😃 i got really confused for a moment ;)

freak4pc:
Ah woops, sorry ! @sebastianvarela ! :)

sebastianvarela:
Thanks @freak4pc, I made this PR because I move from ReactiveSwift to RxSwift and I miss these operators because they are quite simple and easy to read. I think it will help other users to adapt too.

If you think that this operator don't belong to this repo I can make a suggestion for RxSwiftExt, this is the report that you say? https://github.com/RxSwiftCommunity/RxSwiftExt
thanks you, again!

> Thanks for your first contribution here @SebastianOsinski ! I hope to see more of your code in our repo :)
> 
> This is an interesting thought similar to RAS's <~ binding operator. I think there have been several discussions around this here that mention it doesn't really fit into the RxSwift core repo. I'm trying to think if it could fit into RxSwiftExt, or some other portion.
> 
> Of course if there is some serious public demand around adding these, we should discuss further, but remembering previous similar PRs, this is a relatively lesser wanted feature, as custom operators usually cause some confusion :)

freak4pc:
Yup, that's Ext! We never actually had operators added there, so it's an interesting discussion. I suggest opening an issue with the example to gather some community interest first, so you don't have to rework your code for that repo :) 

There are some operators here that don't live in RAS and are a bit confusing and worth discussing, for example an expression where you have errors to both sides indicates two-way binding and not necessarily what you'd want to convey here :)

sebastianvarela:
Thanks @freak4pc.
I made an issue with this case to see the community interest:
https://github.com/RxSwiftCommunity/RxSwiftExt/issues/244

PR #2089, opened 2019-10-11
Introduce Weakify and weakTargetMap

adlerj:
Since instance functions are curried with the instances that they are called on the below code creates a retain cycle on `self` when `transform(_:)` is passed into the `map()` function.

```
class BadVC: UIViewController {
    let disposeBag = DisposeBag()

    init(with stream: Observable<Int>) {
        super.init(nibName: nil, bundle: nil)

        Observable<Int>
            .of(1, 2, 3)
            .concat(stream)
            .map(transform(_:)) // Captures self
            .subscribe(onNext: { (value) in
                print(value)
            })
            .disposed(by: disposeBag)
    }
    
    func transform(_ number: Int) -> String {
        return "\(number)"
    }
}
```

This issue exists due to `MapSink` retaining the transformation function until the stream is disposed. In cases where we have a `disposeBag` bag living on self we end up with a retain cycle.

This solution allows for targets to be weak before there is an attempt to curry the target with a given function:

```
Observable<Int>
    .of(1, 2, 3)
    .concat(stream)
    .map(weakifyTarget(self, function: BadVC.transform(_:)))
    .subscribe(onNext: { (value) in
        print(value)
    })
    .disposed(by: disposeBag)
```

Even shorter hand wrappers can be introduced:
```
    Observable<Int>
            .of(1, 2, 3)
            .concat(stream)
            .weakTargetMap(self, BadVC.transform(_:))
            .subscribe(onNext: { (value) in
                print(value)
            })
            .disposed(by: disposeBag)
```

Note: I recognize that if `func transform(_ number: Int)` is implemented as a class function than this issue wouldn't exist. However there are cases where access to self is needed.

### DIFF
--- a/RxSwift/Observables/Map.swift
+++ b/RxSwift/Observables/Map.swift
@@ -21,6 +21,24 @@ extension ObservableType {
         -> Observable<Result> {
         return Map(source: self.asObservable(), transform: transform)
     }
+
+    /**
+     Projects each element of an observable sequence into a new form.
+
+     - seealso: [map operator on reactivex.io](http://reactivex.io/documentation/operators/map.html)
+
+     - parameters:
+       - target: The target the transform function lives on. Will be held weakly.
+       - transform: A transform function to apply to each source element.
+     - returns: An observable sequence whose elements are the result of invoking the transform function on each element of source. 
+                An error is returned if the target is nil.
+     */
+    public func weakTargetMap<ClassType: AnyObject, Result>(
+        _ target: ClassType,
+        _ transform: @escaping (ClassType) -> (Self.Element) throws -> Result
+    ) -> RxSwift.Observable<Result> {
+        map(weakifyTarget(target, function: transform))
+    }
 }
 
 final private class MapSink<SourceType, Observer: ObserverType>: Sink<Observer>, ObserverType {
--- a/RxSwift/Weakify/Weakify.swift
+++ b/RxSwift/Weakify/Weakify.swift
@@ -1,0 +1,25 @@
+public func weakifyTarget<ClassType: AnyObject, InputType, OutputType>(
+    _ target: ClassType,
+    function: @escaping (ClassType) -> (InputType) throws -> OutputType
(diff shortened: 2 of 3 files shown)


freak4pc:
Thanks @adlerj ! Happy to see a bunch of new contributors, wondering if it's because of Hacktoberfest, but happy about it regardless :)

There's already a `withunretained` operator in RxSwiftExt which solves this very problem. Something like your `weakifyTarget` doesn't pertain to RxSwift specifically but to Swift and ARC in general, so maybe a more general purpose repo (like SwifterSwift, or something similar) would fit better. In general these sort of things usually don't go into RxSwift core, but in other community repos such as Ext.

Do you think `withUnretained` solves this problem for you? https://github.com/RxSwiftCommunity/RxSwiftExt#withunretained

adlerj:
Thanks for the speedy reply! I'm pretty new to RxSwift and was unaware that `RxSwiftExt` existed, cool stuff! Makes perfect sense that this type of stuff should live there :)

The implementation with `withUnretained` looks like:
```
Observable<Int>
            .of(1, 2, 3)
            .concat(stream)
            .withUnretained(self)
            .map { (badVC, value) -> String in
                return badVC.transform(value)
            }
            .subscribe(onNext: { (value) in
                print(value)
            })
            .disposed(by: disposeBag)
```

vs my proposal of :
```
Observable<Int>
    .of(1, 2, 3)
    .concat(stream)
    .weakTargetMap(self, BadVC.transform(_:))
    .subscribe(onNext: { (value) in
        print(value)
    })
    .disposed(by: disposeBag)
```

At the end of the day it's just syntactic sugar but it does feel like a bit less boilerplate.

freak4pc:
The `weakTargetMap` isn't ideal because it means making duplicates of all operators just for this case. This is really why operators exist, there was a similar discussion around this here: https://github.com/RxSwiftCommunity/RxSwiftExt/pull/236

adlerj:
Makes sense, thanks for the replies :)

Issue #2081, opened 2019-10-02
DelegateProxy crashes with UITableView in some scenarios


svenfy:
# Description

run demo, click the first row, navigate to `Example1_CustomizationUsingTableViewDelegate`

> 2019-10-02 23:37:03.091294+0800 Example[6933:116418] [TableView] Warning once only: UITableView was told to layout its visible cells and other contents without being in the view hierarchy (the table view or one of its superviews has not been added to a window). This may cause bugs by forcing views inside the table view to load and perform layout without accurate information (e.g. table view bounds, trait collection, layout margins, safe area insets, etc), and will also cause unnecessary performance overhead due to extra layout passes. Make a symbolic breakpoint at UITableViewAlertForLayoutOutsideViewHierarchy to catch this in the debugger and see what caused this to occur, so you can avoid this action altogether if possible, or defer it until the table view has been added to a window. Table view: <UITableView: 0x7f9dd2063600; frame = (-19.5 -146.5; 414 896); clipsToBounds = YES; autoresize = RM+BM; gestureRecognizers = <NSArray: 0x600001416af0>; layer = <CALayer: 0x600001a203c0>; contentOffset: {0, 0}; contentSize: {414, 0}; adjustedContentInset: {0, 0, 0, 0}; dataSource: <RxCocoa.RxTableViewDataSourceProxy: 0x600003e606c0>>

# Advice
修改`/RxCocoa/Common/DelegateProxyType.swift`中的
```Swift
object.layoutIfNeeded()
```
为
```Swift
if let tv = object as? UITableView {
    if tv.window != nil {
        object.layoutIfNeeded()
    }
} else {
    object.layoutIfNeeded()
}
```

freak4pc:
What demo are you referring to? @svenfy

chiliec:
I think @svenfy means this https://github.com/RxSwiftCommunity/RxDataSources/blob/master/Examples/Example/Example1_CustomizationUsingTableViewDelegate.swift

freak4pc:
Can you clarify the fix you posted? Is it related to UIScene in any way or its been broken this whole time?

iNoles:
Similar problem on https://github.com/RxSwiftCommunity/RxDataSources/issues/331

Sharath8805:
[...]

PR #2076, opened 2019-09-26
Fix iOS 13 UITableViewAlertForLayoutOutsideViewHierarchy warning

ronzohan:
Fixes `UITableViewAlertForLayoutOutsideViewHierarchy` warning on **iOS 13** by checking if `layoutIfNeeded()`  needs to be called by performing a check if the `object` is currently attached to a window.

PR #2090, opened 2019-10-13
Call layoutIfNeeded() on table only if table has a window

chiliec:
demo fix advice #2081


yunuserenguzel:
Looks like this PR is a duplicate of https://github.com/ReactiveX/RxSwift/pull/2076 ?

chiliec:
I think yes, it's the same.

PR #2091, opened 2019-10-15
Fix a bug that resource count doesn't decrease because of static locks

devxoul:
## Background

Using with `TRACE_RESOURCES` option, `RecursiveLock` increases the resource count on initialization and decreases it on deallocation. But if you create a static lock instance, it will never get deallocated so the resource count will not decrease.

This PR introduces `StaticRecursiveLock` class that doesn't increase or decrease the resource count on initialization and deallocation.

## Discussion

Theoretically, the static instance is not deallocated so the resource count should not be decreased. But I think `Resources.total` is intended to be used for debugging unexpectedly retained objects.


freak4pc:
Hey Jeon, long time! Hope everything's going great :) 

I'm a bit confused on why this is part of RxSwift, though? Seems to be an implementation detail for some specific app... Is this used in any way by Rx itself?

Platform/RecursiveLock.swift
```diff
@@ -29,6 +29,24 @@ import class Foundation.NSRecursiveLock
             _ = Resources.decrementTotal()
         }
     }
+
+    /// A recursive lock for static instances. It doesn't increase or decrease the resource count
+    /// on initialization and deallocation.
+    ///
+    /// Use this class for static lock instances or the resource count will never decrease since
```
LucianoPAlmeida: Hey :) 
I'm not sure `the resource count will never decrease` is the correct statement here. Because it would decrease on lock/unlock it would just never decrease the one incremented on `init` right? That's at least what I could understand, maybe I'm getting this wrong :)
Other thing is if the instance is always allocated, isn't correct to have a resource count always taking it into account? I mean have it incremented would be correct?

devxoul:
@freak4pc Hey, how is it going?

Absolutely. It is used in `Hooks._lock`. It holds the static reference so the resource count will never get decreased. Theoretically it could be the *correct* behavior but I thought that is it not *intended* behavior.

Platform/RecursiveLock.swift
devxoul: Hey, thanks for your comment. 

This behavior might be theoretically correct. But in perspective of that [`Resources.total` API is designed for debugging memory leaks](https://github.com/ReactiveX/RxSwift/blob/6b2a406b928cc7970874dcaed0ab18e7265e41ef/Documentation/GettingStarted.md#debugging-memory-leaks), this might be confused because the resource count remains on `1` even if every resources **_seems to be_** deallocated.

LucianoPAlmeida: That makes sense,  the comment `never decremented` got me a little bit confused.
Thank's for the answer :))

freak4pc:
I'm wondering if it makes more sense to have an initializer on RecursiveLock to mention whether or not it should trace resources (defaults to `true` and made `false` in that specific instance). 

Adding another specialization/subclass doesn't feel needed to me here. WDYT @devxoul ?

devxoul:
@freak4pc That was exactly my first attempt. But if you're not using `TRACE_RESOURCES` option `RecursiveLock` is equal to `NSRecursiveLock`, which means that you cannot pass the parameter to the initializer 😅

LucianoPAlmeida:
Maybe always use the custom class (even for no trace resources builds), and let the trace resources logic encapsulated inside it. 
```swift
class RecursiveLock: NSRecursiveLock {
        override init(flag: Bool = false) {
            # if TRACE_RESOURCES
            if flag {
               _ = Resources.incrementTotal()
            }
            #endif 
            super.init()
        }

       # if TRACE_RESOURCES
        override func lock() {
            super.lock()
            _ = Resources.incrementTotal()
        }
        override func unlock() {
            super.unlock()
            _ = Resources.decrementTotal()
        }
        deinit {
            _ = Resources.decrementTotal()
        }
        #endif
    }

```
A bit ugly but ...  Just an idea :))

LucianoPAlmeida:
@devxoul I believe you referenced the wrong person hahah 😂 

The overhead you mention makes total sense. So the StaticRecursiveLock seems the better approach in this case :))

devxoul:
GitHub auto completion always suggests me a wrong person 🤦‍♂🤦‍♂🤦‍♂

Platform/RecursiveLock.swift
devxoul: I added some words to make more sense!

```diff
- will never decrease
+ will never decrease under 1
```

kzaher:
Hi @devxoul,

It's been a while since we've met :) Hope you are doing well.

I'm not sure what problem are we really solving. Even if we remove that static lock reference, I'm assuming that everyone has at least some lazy loaded resource referencing publish/replay subject, and that won't be solved with this patch.

Can you maybe explain what problem are you trying to solve?

> Theoretically, the static instance is not deallocated so the resource count should not be decreased. But I think Resources.total is intended to be used for debugging unexpectedly retained objects.

Sure, but why that bugs you?

devxoul:
@kzaher Hey, long time no see. Hope you're doing good as well :)

I was trying to ensure that there is no memory leak after each test case is executed. For example:

```swift
func setUp() {
  super.setUp()
  self.expectedResourceCount = Resources.total
}

func tearDown() {
  let currentResourceCount = Resources.total
  XCTestAssert(currentResourceCount, self.expectedResourceCount)
  super.tearDown()
}

func testFoo() {
  _ = Observable.interval<Int>(.seconds(1), scheduler: MainScheduler.instance)
    .subscribe()
    .dispose()
}
```

But with the static lock it will fail even if the user did everything correct. This behavior could look like a bug in the user's perspective.

kzaher:
If you want to test generic code in this way you would need to measure the count the second time because during the first run you could lazy load static members. 

I have no idea how to do this with XCTest though except not referencing static members or lazy loading them before the first count. Fixing RX code is not enough if you want to make a foolproof system.

devxoul:
Yeah I have an option to preload the lazy resources like the code below:

* https://github.com/ReactiveX/RxSwift/blob/5.0.1/Tests/RxTest.swift#L90-L91
    ```swift
    func setUpActions(){
        _ = Hooks.defaultErrorHandler // lazy load resource so resource count matches
        _ = Hooks.customCaptureSubscriptionCallstack // lazy load resource so resource count matches
        #if TRACE_RESOURCES
            self.startResourceCount = Resources.total
            ...
        #endif
    }
    ```
* https://github.com/ReactiveX/RxSwift/blob/5.0.1/Tests/RxTest.swift#L115
    ```swift
    func tearDownActions() {
        #if TRACE_RESOURCES
            ...
            XCTAssertEqual(self.startResourceCount, Resources.total)
        #endif
    }
    ```

But I'm worrying about that it will deeply depends on RxSwift's internal implementation.

There have been only one static resource(`Hooks._lock`) that increase the resource count, so I don't think it's a big deal to make it foolproof. And we already have a great test that prevents from unexpected resource allocations.

freak4pc:

PR #2095, opened 2019-10-27
Remove unneeded return statements

sticksen:
This PR removes all unneeded return statements as described in https://github.com/apple/swift-evolution/blob/master/proposals/0255-omit-return.md
This code change can be done when using Swift 5.1, so I´m targetting this branch with my PR.


freak4pc:
Thanks for the PR @sticksen ! 
While this is a nice ability and we _can_ omit returns, we still haven't decided if this is the style we want to follow in this repo. 

@kzaher wdyt? I know you like explicitness in these kind of things, so let me know your thoughts.

kzaher:
@freak4pc I'm fine with these `return` removals. Seems like having `return` doesn't improve readability much.

On the other hand, I don't think that changes to `scripts/validate-podspec.sh` are good. These changes are there to allow CI to process major version bumps.

freak4pc:
It was one of my attempts, I'm not sure why he has that commit in his PR. Can you remove it @sticksen ?

sticksen:
No problem, will remove it!

sticksen:
It´s ready for reviewing again, @kzaher, @freak4pc

freak4pc:
I believe I addressed all of these in https://github.com/ReactiveX/RxSwift/commit/8dd25b4637cb64d44b0d644f37427803c24f1343

PR #2102, opened 2019-11-07
Switch to Swift Module Interface for generated libraries

asaltvld30:
Set the new Build Settings flag `BUILD_LIBRARIES_FOR_DISTRIBUTION` in order to generate stable module interfaces between different Swift compiler versions.

https://developer.apple.com/videos/play/wwdc2019/416/


freak4pc:
I don't have enough information on what this does. Can you tl;dr this for me ? Would be hard watching an hour-long session soon :)

lordzsolt:
I went through the video to see what it's about. They mentioned 2 changes:
- This will build an `.xcframework` file, instead of the old `.framework`.
- The `.xcframework` will contain all the frameworks for iOS/macOS/tvOS/Simulator in a single file
- Instead of the previous binary `.swiftmodule` format, the .xcframework will contain a new format, so if you built a framework with Swift 5.0.3, it will be compatible if have a newer version of the compiler, instead of seeing `Compiled module was created by a newer version of the compiler`.

However, there are some cons that come with enabling this flag:
- All the frameworks you depend on must enable this flag (ie. to enable it on RxCocoa, you need to enable it on RxSwift and RxRelay)
- Supposedly, it has some negative performance implications?
- `enums` need to be marked with `@frozen` keyword, otherwise consumers will have to have a `@unknown default:` in their `switch` cases

All in all, it probably requires a bit more consideration than just enabling a flag in the `.xcodeproj` file

freak4pc:
It seems interesting, but it also seems too early to do this while we don't know the full scope. I agree that just turning on the flag makes no real sense ATM. 

Closing this for now. If you'd like to pursue this further, please make a new PR with more content and details so we can decide accordingly :) Thanks !

Issue #2026, opened 2019-07-04
Figure out how to add tests for breaking the ABI backwards compatibility.


kzaher:
Somebody knows how can we do this?

LucianoPAlmeida:
@kzaher I don't know exactly how to do it, but looking to the apparently way the swift tests the stability of the ABI/API of the swift standard library. Seems like a way to do it after swift 5.1 (that have module stability) to compare the swiftinterface files for the module. Before the module stability they seems to dump the module interface to a Json format and compare them. I'm not 100% sure about all this, I just got these ideas looking [here](https://github.com/apple/swift/tree/master/test/api-digester), but check the way they do it for the stdlib can be a way to check if an similar approach can be applicable here.

kzaher:
@freak4pc 
> A bit confused on why, since closed source libraries are also versioned (usually) and possibly locked to a specific API version.

I guess I can see these cases. There could potentially be more:
* If we want to distribute binaries in RxSwiftCommunity for example and we don't keep RxSwift ABI stable then the released binaries will break for using different versions.
* If somebody makes a closed source binary which is depending on RxSwift and wants to redistribute it. The consumer of the closed source binary will make their own build of RxSwift so without ABI stability that could break.

It's hard for me to say how much effort would this take and how many users are there, but I would rather we make a conscious decision do we want this library to be semantically ABI stable or not. I would probably say yes, but if that takes a lot of work and we can't automate it I wouldn't be in favour of making it ABI stable. 

We can probably just check what other libraries are doing and do the same.

kzaher:
@LucianoPAlmeida Thanks, I would need to take a look.

LucianoPAlmeida:
> * If somebody makes a closed source binary which is depending on RxSwift and wants to redistribute it. The consumer of the closed source binary will make their own build of RxSwift so without ABI stability that could break.
> 

One case related to that may be a client that import two closed binary libraries A and B that depends on different versions of RxSwift. So if without ABI to guarantee backward compatibility, that client might not be able to build it because that one that depends on the old version might not be able to link against the Rx binary if it has changed the ABI.

LucianoPAlmeida:
@kzaher @freak4pc 
Found two files that maybe relevant to this discussion.
One is the [Library Evolution](https://jrose-apple.github.io/swift-library-evolution/) 
[...]

PR #2103, opened 2019-11-14
Remove scoped imports for library evolution

keith:
In order to support `-enable-library-evolution` and `swiftinteface`
files, we have to use non-scoped imports. I've filed an upstream bug to
track this https://bugs.swift.org/browse/SR-11782

This is a step towards https://github.com/ReactiveX/RxSwift/issues/2026

### DIFF
--- a/Platform/AtomicInt.swift
+++ b/Platform/AtomicInt.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2018 Krunoslav Zaher. All rights reserved.
 //
 
-import class Foundation.NSLock
+import Foundation
 
 final class AtomicInt: NSLock {
     fileprivate var value: Int32
--- a/Platform/Platform.Darwin.swift
+++ b/Platform/Platform.Darwin.swift
@@ -9,8 +9,7 @@
 #if os(macOS) || os(iOS) || os(tvOS) || os(watchOS)
 
     import Darwin
-    import class Foundation.Thread
-    import protocol Foundation.NSCopying
+    import Foundation
 
     extension Thread {
         static func setThreadLocalStorageValue<T: AnyObject>(_ value: T?, forKey key: NSCopying) {
--- a/Platform/Platform.Linux.swift
+++ b/Platform/Platform.Linux.swift
@@ -8,7 +8,7 @@
 
 #if os(Linux)
 
-    import class Foundation.Thread
+    import Foundation
 
     extension Thread {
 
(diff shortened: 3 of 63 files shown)


keith:
Note I've only changed this for RxSwift and RxBlocking. I'd be happy to update this for the other modules as well if folks would like!

keith:
Note I didn't add `@frozen` to enums here so technically this isn't supported entirely yet. There's a question of which enums should be frozen and which should not. But also Swift produces a warning if you have `@frozen` but are not compiling with `-enable-library-evolution` which I expect to be unacceptable for this repo since that would be produced for the SwiftPM build (I believe) and for CocoaPods and possible debug builds. I've submitted a PR for discussion about removing that warning here https://github.com/apple/swift/pull/28227

freak4pc:
Hey @keith - interesting proposal! I've went ahead and done this for all of the other modules as well in https://github.com/ReactiveX/RxSwift/pull/2103/commits/5b76d487e07bd19e01d2bff249f635d4894b910e. 

I'm abroad right now but happy to keep discussing this. I agree that without frozen enums, this is just a step but not a resolution of how to test binary stability. 

Thanks for your work and thought of this :)

keith:
Thanks! Yea I think from our perspective we would use this + a much smaller patch to add `@frozen` in some places until we could figure out how to acceptably add those here.

freak4pc:
Putting this here as a reference for the discussion: 
https://github.com/apple/swift-evolution/pull/1025/files

keith:
Do you think this is something we want to wait on, or should we merge this and stop using scoped imports going forward, and then add the `@frozen` stuff in the future?

freak4pc:
Hey @keith, sorry for not replying first. I'm waiting on this to be merged first #2077 and I'm happy to merge this after barring @kzaher's opinion. 

Would be great if you could rebase your branch on top of that one and make it the base of this PR, perhaps

keith:
Sounds good. I'll rebase when I'm back in the office this week!

keith:
I'll actually wait for the linked PR to be merged, and then revisit

freak4pc:
Yup, sounds like a good idea. That PR also had to make some changes to scoped import (due to `FoundationNetworking` on Linux), so your changes would potentially clean up a lot of that mess

kzaher:
@freak4pc @keith This is fine by me. If you can merge this that would be great.

freak4pc:
I rebased but tests are failing, I think we need to explicitly import FoundationNetworking on Linux for some cases, will take care of this later @keith

keith:
Thanks!

PR #2110, opened 2019-12-21
Automatic property binders with @dynamicMemberLookup

freak4pc:
Moved this specific portion of the discussion from #2003 over here.

The more I play with this implementation, the more I like it:

1. You can already see how many redundant Binder(s) I could completely get rid of thanks to this (I'm still not sure if I'll remove the tests as well, in favor of more generic tests, even though it's nice having such broad coverage).

2. It still lets you define same-named properties that override the "automatic" ones you get from `@dynamicMemberLookup` which is awesome and still allows for cusomizability. 

One point I'm still not sure of:

Should we perhaps just move `Binder` to RxSwift itself? I'm not sure if it's that specific to RxCocoa, and we can use it here (instead of `AnyObserver`) since it already covers this very specific usecase. Not a necessity, though.


freak4pc:
Uhm, Travis tests do indicate there are ambiguity issues. But locally it runs fine... I'll have to take another look at this.

kamitchell:
> Should we perhaps just move Binder to RxSwift itself?

I'm in favor of refactoring general-purpose code into RxSwift.

kamitchell:
Let me see if I understand how this works in the broader sense...if I were to use this with my own class, I'd have to either

- inherit from `NSObject`
- extend my class with `ReactiveCompatible`

Then, I can:

- imperatively refer to a property in my object with `instance.propertyName`
- bind to the property with `.subscribe(instance.rx.propertyName)`

Right?

freak4pc:
Yup @kamitchell

lordzsolt:
Oh wow, I was about to write out a page long rant about how `@dynamicMemberLookup` is unsafe, but using `ReferenceWritableKeyPath` is pretty amazing! 😅 
- You have autocompletion
- You have compiler errors in case the property doesn't exist

However, some drawbacks:
- Xcode will autocomplete non-writeable keypaths (ex: `constraint.rx.multiplier`) and produce a gibberish error `Ambiguous reference to 'rx'`
- Suddenly you have access to all writable properties (ex: `constraint.rx.shouldBeArchived`), including all the shit you inherit from `NSObject` and `NSAccessibility`

freak4pc:
All green here. I tested this on 3 large projects and it works beautifully — I think it will really enhance the development experience of wrapping non-Rx constructs. 

I've constructed a "Bleeding Edge" group on Slack to gather testers for these sort of features. If you have a big project and want to help drive the future of RxSwift by simply testing branches locally on your project, let me know. It would be invaluable for the entire community :)

freak4pc:
I've sent this to a few more testers and this seems stable enough. I'm going to merge this in and gather more feedback when we release the beta.

Issue #2109, opened 2019-12-16
Why there is no ReplayRelay?


evgeny-sureev:
Of course we can define `EXPANDABLE_SHARED_SEQUENCE` and create such Relay, but I think it is more convenient to have it in framework.

I can make a PR with ReplayRelay, as I am using it in my projects.

winstondu:
Seconded. Please do

lordzsolt:
Wondering about this also.

freak4pc:
This is currently in progress in #2111. Closing.

PR #2111, opened 2019-12-22
Add ReplayRelay

lordzsolt:
Fix issue #2109 

Add ReplayRelay wrapper around ReplaySubject.
Based on the previous Relay implementations.


evgeny-sureev:
Please add ability to drive ReplayRelay (in `Driver+Subscription.swift`) and ability to convert ReplayRelay to Driver. You can borrow code from my branch at https://github.com/evgeny-sureev/RxSwift/tree/add-replay-relay

freak4pc:
Hey @evgeny-sureev - converting ReplayRelay to Driver doesn't make sense necessarily - you can't guarantee the bufferSize would be 1, and you can't guarantee it would have any replay value at the time of subscription, so it doesn't fulfil the guarantees of a Driver.

lordzsolt:
@freak4pc I guess the first part, allowing a `Driver` to drive a `ReplayRelay` makes sense?
And I'm also wondering, should I add it for the PublishRelay also?

evgeny-sureev:
@freak4pc but Driver trait documentation doesn't say anything about requiring initial value.

lordzsolt:
@evgeny-sureev The issue isn't with the initial value. The issue is the buffer size:

`ReplayRelay<Bool>.create(bufferSize: 5).asDriver().drive(activityIndicator.rx.isActive)`

ReplayRelay will buffer 5 elements, which will be replayed to the activityIndicator. Doesn't really make sense.

evgeny-sureev:
@lordzsolt no, it will replay only last value, in Driver.swift:
```
    public static func share<Element>(_ source: Observable<Element>) -> Observable<Element> {
        return source.share(replay: 1, scope: .whileConnected)
    }
```
All previous values (if any) will be discarded.

freak4pc:
Hey @evgeny-sureev - i'm not takling about a starting value guarantee. 

But `Driver` represents state, so it must have a value to replay, which is why we don't use it for `PublishRelay`, as well - since there's no way to guarantee there will be a value available for replay. 

You're better off using something like Signal for this, perhaps. But I'd argue non e of the traits fit "Replay*" too well because it's such a strange beast.

evgeny-sureev:
So why we can convert regular Observables to Drivers, without any starting values whatsoever?

freak4pc:
Observable specifically can be converted to anything because it's the "core unit". When we're referring to traits we try to be strict as to not allow "illegal" connections. 

For example you can't drive a publish relay, you can't "emit to" a behavior relay,  tou can't turn a BehaviorRelay into a Signal, you can't turn a PublishRelay into a Driver, etc.

lordzsolt:
> Observable specifically can be converted to anything because it's the "core unit". When we're referring to traits we try to be strict as to not allow "illegal" connections.
> 
> For example you can't drive a publish relay, you can't "emit to" a behavior relay, tou can't turn a BehaviorRelay into a Signal, you can't turn a PublishRelay into a Driver, etc.

Makes sense. My intention was also to use the _strictest_ trait, while expressing what I need (hence `BehaviorRelay` was too much, but I couldn't enqueue with `PublishRelay`, so I needed `ReplayRelay` 😄  )

Should I still add Drivers driving `ReplayRelay`, in `Driver+Subscription.swift`?

freak4pc:
No. As I mentioned, since there is no guarantee ReplayRelay will have any value, it doesn't seem relevant in my eyes.

RxRelay/ReplayRelay.swift
```diff
@@ -0,0 +1,50 @@
...
+//
+//  Created by Zsolt Kovacs on 12/22/19.
+//  Copyright © 2019 Krunoslav Zaher. All rights reserved.
+//
+
+import RxSwift
+
+/// ReplayRelay is a wrapper for `ReplaySubject`.
+///
+/// Unlike `ReplaySubject` it can't terminate with error or completed.
```
freak4pc:
```suggestion
/// Unlike `ReplaySubject`, it can't terminate with an error or complete.
```

RxRelay/ReplayRelay.swift
```diff
@@ -0,0 +1,50 @@
...
+    // Accepts `event` and emits it to subscribers
+    public func accept(_ event: Element) {
+        self._subject.onNext(event)
+    }
+
+    private init(subject: ReplaySubject<Element>) {
+        self._subject = subject
+    }
+
+    /// Creates new instance of `ReplayRelay` that replays at most `bufferSize` last elements of sequence.
```
freak4pc:
```suggestion
    /// Creates new instance of `ReplayRelay` that replays at most `bufferSize` last elements sent to it.
```

RxRelay/ReplayRelay.swift
```diff
@@ -0,0 +1,50 @@
...
+    ///
+    /// - parameter bufferSize: Maximal number of elements to replay to observer after subscription.
```
freak4pc:
```suggestion
    /// - parameter bufferSize: Maximal number of elements to replay to observers after subscription.
```

RxRelay/ReplayRelay.swift
```diff
@@ -0,0 +1,50 @@
...
+    /// - returns: New instance of replay relay.
+    public static func create(bufferSize: Int) -> ReplayRelay<Element> {
+        return ReplayRelay(subject: ReplaySubject.create(bufferSize: bufferSize))
```
freak4pc:
```suggestion
        ReplayRelay(subject: ReplaySubject.create(bufferSize: bufferSize))
```

RxRelay/ReplayRelay.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /// Creates a new instance of `ReplayRelay` that buffers all the elements of a sequence.
+    /// To avoid filling up memory, developer needs to make sure that the use case will only ever store a 'reasonable'
+    /// number of elements.
+    public static func createUnbound() -> ReplayRelay<Element> {
+        return ReplayRelay(subject: ReplaySubject.createUnbounded())
```
freak4pc:
```suggestion
        ReplayRelay(subject: ReplaySubject.createUnbounded())
```

RxRelay/ReplayRelay.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /// Subscribes observer
+    public func subscribe<Observer: ObserverType>(_ observer: Observer) -> Disposable where Observer.Element == Element {
+        return self._subject.subscribe(observer)
```
freak4pc:
```suggestion
        self._subject.subscribe(observer)
```

RxRelay/ReplayRelay.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    /// - returns: Canonical interface for push style sequence
+    public func asObservable() -> Observable<Element> {
+        return self._subject.asObserver()
```
freak4pc:
```suggestion
        self._subject.asObserver()
```

RxRelay/ReplayRelay.swift
```diff
@@ -0,0 +1,50 @@
...
+
+    /// Creates new instance of `ReplayRelay` that replays at most `bufferSize` last elements of sequence.
+    ///
+    /// - parameter bufferSize: Maximal number of elements to replay to observer after subscription.
+    /// - returns: New instance of replay relay.
+    public static func create(bufferSize: Int) -> ReplayRelay<Element> {
+        return ReplayRelay(subject: ReplaySubject.create(bufferSize: bufferSize))
+    }
+
+    /// Creates a new instance of `ReplayRelay` that buffers all the elements of a sequence.
```
freak4pc:
```suggestion
    /// Creates a new instance of `ReplayRelay` that buffers all sent to it.
```

Tests/RxCocoaTests/Signal+Test.swift
```diff
@@ -527,3 +527,123 @@ extension SignalTests {
...
+        })
+
+        _ = (Signal.just(1) as Signal<Int>).emit(to: relay1, relay2)
+
+        XCTAssertEqual(latest1, 1)
+        XCTAssertEqual(latest2, 1)
+    }
+
+    func testEmitOptionalReplayRelay1() {
+        let relay = ReplayRelay<Int?>.create(bufferSize: 1)
```
freak4pc: not seeing any tests for big buffer sizes - without these, this is basically identical to Behavior Relay.

Please make tests for some random amount of elements showcasing having less than max, more than max, zero elements, etc. This seems to me like the essence worth testing of this relay.

lordzsolt: I added the tests I could think of.

I noticed there are not tests for ReplaySubject with bigger buffer size, so I added some there as well.
Considering ReplayRelay is a wrapper around ReplaySubject, the tests I added in `ReplayRelayTests.swift` seem a bit redundant. 

Let me know what you think and if you have any cases I didn't think of.

freak4pc:
Hey @lordzsolt, there were some naming refactors so you'll have to rebase. 

Also — I'm still positive we shouldn't have a `asDriver()` for ReplayRelay, but I'm not sure if we shouldn't allow `drive(replayRelay)`. WDYT ?

lordzsolt:
> Hey @lordzsolt, there were some naming refactors so you'll have to rebase.
> 
> Also — I'm still positive we shouldn't have a `asDriver()` for ReplayRelay, but I'm not sure if we shouldn't allow `drive(replayRelay)`. WDYT ?

After reading through Driver and Relay again, I think neither `asDriver()` nor `drive(replayRelay)` should be allowed, with the current implementation.

1. `drive(replayRelay)` should not be allowed, because a `Driver` can `complete`, while a `Relay` shouldn't, so it's an illegal connection.

2. My reasoning for not allowing `asDriver()` is a bit different than yours.

I think the fact you mentioned, that `Driver` is used to represent state and `ReplayRelay` does not guarantee to have a value, is not necessarily correct.
- The example from the [Traits documentation for the Driver](https://github.com/ReactiveX/RxSwift/blob/master/Documentation/Traits.md#driver) is using `Driver` without having a value, inside the `flatMapLatest`, `fetchAutoCompleteItems(query).asDriver(onErrorJustReturn: [])`. 
- The documentation also only mentions that `Driver` is doing `.share(replay: 1, scope: .whileConnected)`, not that it must currently have a value.

On the other hand, the same documentation says `Driver` will only replay, at most, **1** element, but `ReplayRelay` can have a buffer size > 1. So:
- `ReplayRelay<T>.create(bufferSize: 1)` **should** have `asDriver()`
- `ReplayRelay<T>.create(bufferSize: 2...)` **shouldn't** have `asDriver()`

But this is not possible with the current implementation, so I would rather not allow illegal connections.

**However**, `PublishRelay` could have an `asDriver()`.

freak4pc:
> drive(replayRelay) should not be allowed, because a Driver can complete, while a Relay shouldn't, so it's an illegal connection.

That's not a valid concern or you wouldn't be able to bind an observable to a relay as well. Relays don't complete, it doesn't mean you can't bind things that _themselves_ complete. It's the Relay's responsibility to ignore these based on its rules. Otherwise you wouldn't be able to use `asDriver` on _anything_.

> The example from the Traits documentation for the Driver 

I worked on that documentation, I think your understanding of it is wrong. `asDriver()` and `asSignal()` are both valid on `Observable` because it is the core unit (as mentioned above). When we have type-safe units (traits) that have specific guarantees, we try as hard as possible to not allow illegal conversions from there to things that don't follow their guaranteed properties. Like when you do `asSingle()` from and `Observable` that emits more than one value, you would get a warning. 

If `Driver` has a replay 1, it makes sense allowing to make one from a core unit (Observable) where it is unknown how many values can happen, but it makes no sense converting from Publish where we have absolute guarantee that it can only have no replay. (That's what `Signal` is for)

> PublishRelay could have an asDriver().

No, it can't based on the notes I said above. Publish represents Events, Behavior represents State. As mentioned above, the proper trait for PublishRelay is Signal.

lordzsolt:
> No, it can't based on the notes I said above. Publish represents Events, Behavior represents State. As mentioned above, the proper trait for PublishRelay is Signal.

Ah yes. I forgot to take `Signal` into the picture. You're right, Publish -> Driver and Replay -> Driver are illegal connections.

lordzsolt:
Updated.

winstondu:
Will this be merged in soon? I have my own implementation, but obviously will prefer switching to a standard implementation from this repo

freak4pc:
Probably, but being merged doesn’t mean it’s immediately released.

abdulhaq-e:
I'm late to the discussion, regarding converting a `Relay` (other than `BehaviourRelay`) to a `Driver`. I have the following situation:

```swift
// for certain reasons, the business logic is not completely reactive,
// so I have to use Relays/Subjects

// UI layer
// some ui being driven by the driver:
myDriver.drive(....)

// logic layer 1
var myRelay = PublishRelay<String>()
myDriver: Driver<String> = myRelay.asDriver()

// logic layer 2 
var myString = "Hello, World!"
myRelay.accept(myString)
```

The line `myRelay.asDriver` doesn't work in the current implementation (if using a `PublishRelay`, I have to provide an error recovery strategy, which I usually do with `...asDriver(onErrorRecover: { _ in .empty() }`

I cannot provide an initial value because it's not available at the time  of creating the relay.

I'm kind of lost here, @freak4pc you mentioned here and in other threads that `PublishRelay` represents events and `BehaviourRelay` represents state. Doesn't the example I have above represent state? What should I use.

Thanks

freak4pc:
@abdulhaq-e I don't know what you use your relay represents so I can't say if it's events or state :) Usually all state has initial state. 

Regardless, if you need a Driver-like for `Publish*`, you're looking for `Signal`. 

I suggest checking out our Slack for more assistance / discussions around this.

PR #2114, opened 2020-01-15
Add observeOn, subscribeOn main thread help method

huiping192:


huiping192:
Always need main thread to do some UI things, and every time writing`observeOn(MainScheduler.instance)` is too heavy to me  :)

freak4pc:
Hey @huiping192 - first of all, thank you very much for your contribution :)
I'm sorry but we don't usually add minor helpers this way because:

1. They're easy to make - so anyone can just add them if they need them
2. They're opinionated - it might be something that makes sense for you, but for many developers and teams it might not (so it means it's not generic enough)

tl;dr the only difference in length is 9 characters which is really not a reason to add it into RxSwift core itself, but if its something that benefits your team, you might want to just add it internally. 

Thanks again!

huiping192:
@freak4pc  thank you for the review~ 
1,2 make sense to me.  :)

PR #2117, opened 2020-01-19
Model based wrapper around DataSourcePrefetching

lordzsolt:
RxCocoa has model based wrappers around a lot of events (`modelSelected`, `modelDeselected`, etc...), but `UITableViewDataSourcePrefetching` and `UICollectionViewDataSourcePrefetching` have only index based wrappers.

----

These can be useful, for example when the `RxTableViewDataSourceType` is constructed from an array of objects, which store image URLs:

Instead of having to write:
`tableView.rx.prefetchRows.map { $0.compactMap { try? tableView.rx.model(at: $0) as? MyModel }.subscribe { models in  ... }`

This allows to simply write:
```tableView.rx.prefetchModels(MyModel.self).subscribe { models in ... } ```


freak4pc:
Hey There, thanks for the code. My only consideration here is that I don't think this is a common enough use case that people look for in RxCocoa (this is the first time I've ever heard of someone interested in this construct). 

We're looking to reduce the amount of new code that is barely used inside RxCocoa.

I need to consider this some more and also gauge some company interest around this addition.

gringoireDM:
I agree with @lordzsolt and I am also interested in this PR.
Wondering if this PR should be moved to RxDataSources, but I have no doubt that it's a good thing to have in Rx

freak4pc:
I feel like two comments might not be enough interest for the change/addition. I think I'll be closing this for now, given how long its been. Thanks!

PR #2124, opened 2020-02-02
Implement pairwise operator

ihorvovk:
It emits a pair of last two elements for every element from the observable sequence


freak4pc:
Hey there, thank you but `pairwise` is already part of RxSwiftExt. We usually don't add operators in RxSwift if they don't exist in ReactiveX *and* have wide use cases and community needs :)

freak4pc:
Hey there, thank you but `pairwise` is already part of RxSwiftExt. We usually don't add operators in RxSwift if they don't exist in ReactiveX *and* have wide use cases and community needs :)

PR #2130, opened 2020-02-10
Add `TRACE_RESOURCES` to Package.swift

ahaubenstock:
Defines the `TRACE_RESOURCES` macro for debug builds as recommended by @danielt1263.


danielt1263:
I've started using SPM recently and this will be helpful.

freak4pc:

In the case of this specific PR, we might have to remove SPM support soon as it is riddled with massive bugs and we cannot force Apple to fix these. We've sent multiple reports and are mainly hoping right now :)

PR #2134, opened 2020-02-20
Raise the minimum deployment target for macOS to 10.12

### LINKED ISSUE #2125: Xcode 11.4 beta: 'NSTextView' is incompatible with 'weak' references

When building using the just released Xcode 11.4 beta, RxCocoa fails to build on the following line: https://github.com/ReactiveX/RxSwift/blob/c3c0cac3d4c176b04404e3574d62b51776277384/RxCocoa/macOS/NSTextView%2BRx.swift#L20

**Expected outcome**:

RxCocoa compiles without errors or warnings.

**What actually happens**:

The compile fails with the following error:

```
RxSwift/RxCocoa/macOS/NSTextView+Rx.swift:20:12: 'NSTextView' is incompatible with 'weak' references
```

Version 5.0.1

**Platform/Environment**

  - [ ] iOS
  - [x] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  Version 11.4 beta (11N111s)
```

**Installation method**:
  - [ ] CocoaPods
  - [x] Carthage
  - [ ] Git submodules

  - [x] yes (11.3.1, 11.4 beta 1)
  - [ ] no

  - [ ] just starting

tonyarnold:
This PR proposes to raise the minimum deployment version for macOS targets to macOS 10.12 "Sierra". Fixes #2125.

I also investigated using some combination of availability checks, but these approaches all required duplication of the `RxTextViewDelegateProxy` type. 

Given that there's a major release of RxSwift in development at the time of writing this PR, this seems to me to be a good time to make a change to the base requirements of the library.

I understand that folks (:wave: @freak4pc) would prefer to wait to see whether Apple chooses to leave the error diagnostic that is causing #2125 in future builds of Xcode 11.4, but in either case the PR is here if you can make use of it. 

As mentioned in #2125, I've filed FB7566796 about this with Apple, however the diagnostic is relevant and correct in its advice.


freak4pc:
Thanks for making this! Overall I’m not against, but 6.0 will take some time either way so it will not immediately solve the issues Apple randomly broke, which makes this a bit tricky. Regardless, labeling this for 6.0.

PR #2143, opened 2020-03-11
Added Bindable Sink for prompt property in UINavigationItem

iDevid:


freak4pc:
Hey @iDevid, thanks for the PR! 
We've made an active decision not to merge really simple additions to the core libraries unless they are extremely popular / have a huge core experience for a specific element.

We're doing this because:
1. Having Binders for every tiny property is hard to maintain, along all of the tests related to these.
2. These lesser-used binders are really easy one-liners, that anyone can just add to their own code-base as needed.
3. Starting with RxSwift 6, none of these will be needed thanks to the new [dynamic lookup binders](https://github.com/ReactiveX/RxSwift/pull/2110), which will basically give you automatic reactive binders for every property. 

Hope this makes sense, and again - thank you for your contribution. 

Shai.

PR #2144, opened 2020-03-11
Reactive extension WKWebView navigation delegate

gringoireDM:
This PR add a new delegate proxy for WKNavigationDelegate for WKWebView.
Losing UIWebView delegate proxy, this PR now offers a replacement.

Please note that WKWebView has multiple delegates. This PR offers a proxy just for the WKNavigationDelegate. Other delegates will be subject of future PRs

### DIFF
--- a/RxCocoa/iOS/WKWebView+Rx.swift
+++ b/RxCocoa/iOS/WKWebView+Rx.swift
@@ -25,7 +25,6 @@ extension Reactive where Base: WKWebView {
         navigationDelegate
             .methodInvoked(#selector(WKNavigationDelegate.webView(_:didCommit:)))
             .map { a in try castOrThrow(WKNavigation.self, a[1]) }
-
     }
     
     /// Reactive wrapper for `navigationDelegate` message.
@@ -51,7 +50,7 @@ extension Reactive where Base: WKWebView {
                     try castOrThrow(WKNavigation.self, a[1]),
                     try castOrThrow(Error.self, a[2])
                 )
-        }
+            }
     }
 }
 
--- a/RxCocoa/iOS/WKWebView+Rx.swift
+++ b/RxCocoa/iOS/WKWebView+Rx.swift
@@ -1,0 +1,58 @@
+//
+//  WKWebView+Rx.swift
+//  RxCocoa
+//
+//  Created by Giuseppe Lanza on 14/02/2020.
+//  Copyright © 2020 Krunoslav Zaher. All rights reserved.
+//
+
+#if os(iOS) || os(macOS) || os(tvOS)
+
+import RxSwift
+import WebKit
+
+@available(iOS 10.0, tvOS 10.0, OSXApplicationExtension 10.10, *)
+extension Reactive where Base: WKWebView {
+    
+    /// Reactive wrapper for `navigationDelegate`.
+    /// For more information take a look at `DelegateProxyType` protocol documentation.
+    public var navigationDelegate: DelegateProxy<WKWebView, WKNavigationDelegate> {
+        RxWKNavigationDelegateProxy.proxy(for: base)
+    }
+    
+    /// Reactive wrapper for `navigationDelegate` message.
+    public var didCommit: Observable<WKNavigation> {
+        navigationDelegate
+            .methodInvoked(#selector(WKNavigationDelegate.webView(_:didCommit:)))
+            .map { a in try castOrThrow(WKNavigation.self, a[1]) }
+
+    }
+    
+    /// Reactive wrapper for `navigationDelegate` message.
+    public var didStartLoad: Observable<WKNavigation> {
+        navigationDelegate
+            .methodInvoked(#selector(WKNavigationDelegate.webView(_:didStartProvisionalNavigation:)))
+            .map { a in try castOrThrow(WKNavigation.self, a[1]) }
+    }
+
+    /// Reactive wrapper for `navigationDelegate` message.
+    public var didFinishLoad: Observable<WKNavigation> {
+        navigationDelegate
+            .methodInvoked(#selector(WKNavigationDelegate.webView(_:didFinish:)))
+            .map { a in try castOrThrow(WKNavigation.self, a[1]) }
+    }
+
+    /// Reactive wrapper for `navigationDelegate` message.
+    public var didFailLoad: Observable<(WKNavigation, Error)> {
+        navigationDelegate
+            .methodInvoked(#selector(WKNavigationDelegate.webView(_:didFail:withError:)))
+            .map { a in
+                (
+                    try castOrThrow(WKNavigation.self, a[1]),
+                    try castOrThrow(Error.self, a[2])
+                )
+        }
+    }
+}
+
+#endif
--- a/RxCocoa/iOS/WKWebView+Rx.swift
+++ b/RxCocoa/iOS/WKWebView+Rx.swift
@@ -6,12 +6,12 @@
 //  Copyright © 2020 Krunoslav Zaher. All rights reserved.
 //
 
-#if os(iOS) || os(macOS) || os(tvOS)
+#if os(iOS) || os(macOS)
 
 import RxSwift
 import WebKit
 
-@available(iOS 10.0, tvOS 10.0, OSXApplicationExtension 10.10, *)
+@available(iOS 10.0, OSXApplicationExtension 10.10, *)
 extension Reactive where Base: WKWebView {
     
     /// Reactive wrapper for `navigationDelegate`.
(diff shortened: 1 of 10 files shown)


RxCocoa/iOS/WKWebView+Rx.swift
```diff
@@ -0,0 +1,58 @@
...
+    public var navigationDelegate: DelegateProxy<WKWebView, WKNavigationDelegate> {
+        RxWKNavigationDelegateProxy.proxy(for: base)
+    }
+    
+    /// Reactive wrapper for `navigationDelegate` message.
+    public var didCommit: Observable<WKNavigation> {
+        navigationDelegate
+            .methodInvoked(#selector(WKNavigationDelegate.webView(_:didCommit:)))
+            .map { a in try castOrThrow(WKNavigation.self, a[1]) }
+
```
freak4pc:
```suggestion
```

RxCocoa/iOS/WKWebView+Rx.swift
```diff
@@ -0,0 +1,58 @@
...
+    /// Reactive wrapper for `navigationDelegate` message.
+    public var didFailLoad: Observable<(WKNavigation, Error)> {
+        navigationDelegate
+            .methodInvoked(#selector(WKNavigationDelegate.webView(_:didFail:withError:)))
+            .map { a in
+                (
+                    try castOrThrow(WKNavigation.self, a[1]),
+                    try castOrThrow(Error.self, a[2])
+                )
+        }
```
freak4pc: Keep curly braces aligned to opening please

```suggestion
            }
```

Tests/RxCocoaTests/DelegateProxyTest+WebKit.swift
```diff
@@ -0,0 +1,49 @@
+//
+//  DelegateProxyTest+WebKit.swift
+//  Tests
+//
+//  Created by Giuseppe Lanza on 14/02/2020.
+//  Copyright © 2020 Krunoslav Zaher. All rights reserved.
+//
+
+#if os(iOS) || os(macOS)
```
freak4pc: I think you don't need this entire file, the delegate proxy is already tested on UIKit level. (For example there isn't a specific `DelegateProxyTest+UITableView`). 

As long as you have a `WKWebViewTests`, that should suffice:]

Tests/RxCocoaTests/DelegateProxyTest.swift
```diff
@@ -446,6 +447,9 @@ extension DelegateProxyTest {
         autoreleasepool {
             control = nil
         }
+        
+        wait(for: [expectation], timeout: 1)
```
freak4pc: why is this needed? seems like the tests pass without it as well

Tests/RxCocoaTests/DelegateProxyTest.swift
```diff
@@ -770,3 +774,13 @@ extension MockTestDelegateProtocol
...
+
+#if os(iOS) || os(macOS)
+import WebKit
+
+@available(iOS 10.0, OSXApplicationExtension 10.10, *)
+extension MockTestDelegateProtocol
+    : WKNavigationDelegate
+{
+}
+#endif
```
freak4pc: Same, probably not needed

Tests/RxCocoaTests/WKWebView+RxTests.swift
```diff
@@ -0,0 +1,110 @@
...
+import XCTest
+
+// Any WKNavigation object manually created on dealloc crashes the program.
+// This class overrides the deinit method of the WKNavition to avoid crashes.
+@available(iOS 10.0, OSXApplicationExtension 10.10, *)
+private class SafeWKNavigation: WKNavigation {
+    static func toggleSafeDealloc() {
+        guard let current_original = class_getInstanceMethod(SafeWKNavigation.self, NSSelectorFromString("dealloc")),
+            let current_swizzled = class_getInstanceMethod(SafeWKNavigation.self, #selector(nonCrashingDealloc)) else { return }
+        method_exchangeImplementations(current_original, current_swizzled)
```
freak4pc:
```suggestion
        guard let current_original = class_getInstanceMethod(SafeWKNavigation.self, NSSelectorFromString("dealloc")),
              let current_swizzled = class_getInstanceMethod(SafeWKNavigation.self, #selector(nonCrashingDealloc)) else { return }
        method_exchangeImplementations(current_original, current_swizzled)
```

freak4pc: Could you also move the helpers below the test class itself under a mark? e.g.

```swift
final class WKWebViewTests { ... }

// MARK: - Test Helpers

private class SafeWKNavigation: WKNavigation  { }
```

Tests/RxCocoaTests/WKWebView+RxTests.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+    
+    @objc func nonCrashingDealloc() { }
+}
+
+extension String: Error {}
```
freak4pc: I prefer not using this trick. Can you wrap the error in something private for this test using an actual concrete type? even an enum case with a String associated value would be fine, or something similar. This would make all Strings in all tests conform to Error.

freak4pc:
P.S. @gringoireDM this also needs a CHANGELOG entry, please :)

Tests/RxCocoaTests/DelegateProxyTest+WebKit.swift
gringoireDM: WKWebView behaves differently in dealloc. Dealloc doesn't happen immediately as expected so I preferred to test as it was tested for UIWebView before.

Tests/RxCocoaTests/DelegateProxyTest.swift
gringoireDM: WKWebView is not deallocated immediately. Internally there is something else that retains the WKWebView for a little longer than expected:

https://stackoverflow.com/questions/39349911/memory-leak-in-wkwebview-instantiation

Tests/RxCocoaTests/DelegateProxyTest.swift
gringoireDM: same as https://github.com/ReactiveX/RxSwift/pull/2144#discussion_r391470310

gringoireDM:
@freak4pc is there a swiftFormat step I skipped? If not, should we add one?

Tests/RxCocoaTests/DelegateProxyTest+WebKit.swift
freak4pc: Right, but you are testing a lot of non-dealloc behavior (e.g. copying the "doThisTest" stuff etc). Maybe worth just testing the specific deallocation? Or is there any reason to test the surrounding?

freak4pc:
@gringoireDM We don't have SwiftFormat ATM. Anything specific you're referring to adding this for? Happy to accept a PR if it's anything helpful. 

Thanks !

Tests/RxCocoaTests/DelegateProxyTest.swift
freak4pc: Still somewhat confused - if you have tests in a separate file, maybe this conformance should be there, and not in the general Delegate Proxy Test file.

Tests/RxCocoaTests/DelegateProxyTest+WebKit.swift
gringoireDM: Code coverage mainly. It's a new file, with a new Delegate proxy. Without this test `RxWKNavigationDelegateProxy` won't be covered and while we can agree that it follows the same pattern as others we can't be sure about future refactoring and human errors.

gringoireDM:
@freak4pc a lot of changes for requests comes for spacings and listing stuff we can uniform using swiftFormat rules so we won't need to comment and fix those anymore in PRs. 

Anyway, it's something for a separate discussion not related to this PR specifically at the moment.

freak4pc:
Happy to. Don’t have a lot of experience with customizing it and Xcode’s default spacing is quite lacking.

gringoireDM:
I'll prepare one asap

Tests/RxCocoaTests/DelegateProxyTest+WebKit.swift
freak4pc: Alright. I don't mind much about coverage in this case (not everything has to be covered), but I don't mind the extra test.

Tests/RxCocoaTests/DelegateProxyTest.swift
freak4pc: So can we move this to **DelegateProxyTest+WebKit.swift** ?

gringoireDM: sure makes sense.

I felt safer developing the delegate proxy for the first time having a test confirming I did it right, you know what I mean?

freak4pc: Yup, that's fair. :)

PR #2149, opened 2020-03-17
Fix constraints warnings

Abedalkareem:
I'm facing a problem when I'm binding the view with the data in the view did load or when I'm removing a custom view from the window. The view is trying to update the constraint but there is as I understand some internal UIKit constraints which will be zero when the view is not attached to a window.

`_UITemporaryLayoutWidth`
`_UITemporaryLayoutHeight`
```
 "<NSLayoutConstraint:0x283e26b70 '_UITemporaryLayoutWidth' App.StringsActionSheet:0x1171c0a60.width == 0   (active)>"
```


freak4pc:
I'm wondering if this is related to #2081

Abedalkareem:
> I'm wondering if this is related to #2081

Yes, I'm also getting the same warning in case of a table view.

freak4pc:
It seems we fixed it in 9f377a65 already, but it's missing the Disposal part, i'll fix it manually.

freak4pc:
Added missing portion in d1622907041f560ac58a7ac8b4751c0b34f3fd1f

ameue:
it works

PR #2150, opened 2020-03-18
SharingScheduler.mock(scheduler:action:) can be throwing

sidepelican:
On Xcode 11.0, it's itroduced a throwing XCTFunction like `XCTUnwrap` ( https://developer.apple.com/documentation/xctest/3380195-xctunwrap ).  On Xcode 11.4 Beta, `XCTSkipIf` ( https://developer.apple.com/documentation/xctest/3521325-xctskipif ) is implemented.
These functions throw some `XCTCustomErrorHandling` or `XCTSkip`, and it needs to be handled by XCTestCase.

`SharingScheduler.mock(scheduler:action:)` is not marked `throws`, so it can't throw such error in `action` clusure. 

ex:

```swift
func callFoo() -> Bool? { ... }

func testFoo() throws {
	try SharingScheduler.mock(makeScheduler: { MyScheduler() }) {
		let fooResult = try XCTUnwrap(callFoo()) // compile error
		XCTAssertTrue(fooResult)
	}
}
```

This PR make it possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 * Add `distinctUntilChanged(at:)` for Key Paths.
 * Fix `DelegateProxy` call to `layoutIfNeeded` for an object without a window. #2076
 * Add `ControlEvent` wrappers to `UIApplication` Notifications. #2116
+* `SharingScheduler.mock(scheduler:action:)` can use throwing function for `action`. #2150
 
 ## [5.1.1](https://github.com/ReactiveX/RxSwift/releases/tag/5.1.1)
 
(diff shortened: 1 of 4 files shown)


freak4pc:
I kinda like this idea, but I'm wondering if it would add any breakage. Is your project large enough to test this @sidepelican ?

sidepelican:
We have a large project using RxSwift but the test cases using `SharingScheduler.mock` are not so many. Maybe about 10 cases.

freak4pc:
Yeah I understand. It would be super useful if we can confirm this doesn't cause any issues for existing codebases. 
Overall this seems solid, I'll put this on Slack to see if anyone can test this on their codebase.

freak4pc:
I ran a few tests and this seems safe enough to me. Can you rebase please? @sidepelican

sidepelican:
@freak4pc Thank you for the testing. I rebased.

freak4pc:
Thanks!

PR #2159, opened 2020-03-30
FEAT: Add Rx Error handing method to use Optional

O-O-wl:
when occured error, return `Observable` converting  .error  to .next(nil)  and `.completed`


freak4pc:
Hey @O-O-wl, thanks for your contribution but this is a very specific use-case to your code base probably, and not something generic enough or a "community standard" that we'd want to include in the repo. I really appreciate your time, though. Thanks!

PR #2162, opened 2020-04-01
Update DispatchQueue+Extensions.swift to use the Equality check on DispatchQueue's

jlaws:


freak4pc:
@jlaws Hey there, thanks for the contribution :) 
It doesn't seem like you can actually do that. CI is failing: 

image

PR #2163, opened 2020-04-02
Catalyst support requires iOS SDK

rjchatfield:
image


rjchatfield:
Unfortunately, this didn't solve the problem either. After compiling I get this: 

image

```
ld: in core/librealm-macosx-dbg.a(bptree.cpp.o), building for Mac Catalyst, but linking in object file built for macOS, for architecture x86_64
```

PR #2169, opened 2020-04-11
fix manual installation instructions

fizzy871:
RxCocoa has RxRelay dependency. So when manually adding Rx using submodules, RxRelay should be added too, otherwise App Store Connect will not accept build with 'ITMS-90562: Invalid Bundle - One or more dynamic libraries that are referenced by your app are not present in the dylib search path.' error.


fizzy871:
new pull request with correct user in commit created

PR #2184, opened 2020-05-16
Remove link to dead video

gagansingh:
This YouTube video is now set to private. This change removes the link to the private video.


freak4pc:
A bummer they remove it, it was great! Thanks :)

PR #2187, opened 2020-05-27
add propertyWrapper to BehaviorRelay

vovabndr:
**Description**
Add  propertyWrapper to BehaviorRelay

For now, it’s possible to create `BehaviorRelay` with:

`
  @BehaviorRelay var name = ""
`

Access:
  Without propertyWrapper:
     `name` ->  `BehaviorRelay<String>`
     `name.value` ->  `String`
    With propertyWrapper:
     `name` ->  `String`
     `$name` ->  `BehaviorRelay<String>`

Setting new value:
    Without propertyWrapper:
    ` name.accept("NEW")`
  With propertyWrapper:
     ` name = "NEW"` or ` $name.accept("NEW")`


freak4pc:
Hey @vovabndr, I [published this idea on Twitter a long time ago](https://twitter.com/freak4pc/status/1215375493819764743) but this isn't something we'll want to have as part of RxSwift itself probably since it in general encourages sub-optimal usage of Rx as a concept. It is also part of the reason we removed Variable which worked in the same way. 

Thanks for the idea, though!

PR #2188, opened 2020-05-30
Develop

Catch-XXII:
RxCocoaObjCRuntimeError+Extensions.swift file rxCocoaErrorForTarget(_:)

line 131 switch errorCode  
handle the default value for the switch

@unknown default:
                    print("Fatal Error")
                }


freak4pc:
Please open a proper issue/PR. We can't accepts PRs/Issues not following the template and not providing full explanations. Thanks.

Issue #2190, opened 2020-06-03
TextField placeholder property


demirciy:
**Short description of the issue**:

  There is no property for placeholder in UITextField

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

  5.1.1

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [x] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  11.5 (11E608c)
```
**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [x] I have a small code base
  - [ ] I have a significant code base

Fixed on #2191 and waiting for pr approve.

demirciy:
Closed by the reason --> https://github.com/ReactiveX/RxSwift/pull/2191#issuecomment-638154023

PR #2191, opened 2020-06-03
Add placeholder property to UITextField (#2190)

demirciy:


freak4pc:
Hey @demirciy - thanks or your contribution, but as a general rule we don't add really simple binders into RxCocoa core, since it's usually one line of code. Also, RxSwift 6 will automatically synthesize these so it won't be an issue anymore.

Thanks!

PR #2192, opened 2020-06-10
Unify coding style of Subjects

masamichiueta:
I found subjects used different coding style while I was reading RxSwift source code.
This PR unifies coding style of Subjects.

This is my first trial to contribute `RxSwift`. Please let me know if this PR is not worth reading.


masamichiueta:
@freak4pc Thank you for your review!

>1.It doesn't matter too much if different subjects use somewhat different styling

I got it! While reading the source code, I was just a little wondering why they were different 🙇‍♂️

>2. I'm worried making actual non-style changes since these could introduce unknown side-effects

That's true...

Thanks again for your attention about this PR, and let me close this.

Issue #2196, opened 2020-06-28
RxSwift binary releases are not forward compatible as they don't declare BUILD_LIBRARY_FOR_DISTRIBUTION


borland:
If you take an older binary framework release from github, and then use it on a version of Xcode that has a newer swift compiler (e.g. swift 5.1 vs swift 5.2) you get the compile error:

Module compiled with Swift 5.x cannot be imported by the Swift 5.y compiler

**Expected outcome**:

The code will compile

**What actually happens**:

Failure to compile with error: `Module compiled with Swift 5.x cannot be imported by the Swift 5.x compiler`

You can repro this if you download the current RxSwift.framework binary release from github, and import it into any iOS project using Xcode 12 beta - or if you download an older (5.0.1) binary release and import it into a project in Xcode 11.5

This is easy to fix; RxSwift should define BUILD_LIBRARY_FOR_DISTRIBUTION = YES; in the pbxproj, and there are a couple of non-public `init` methods that need to be made public because they are overriden in derived classes by public ones.

Pull request here:

https://github.com/ReactiveX/RxSwift/pull/2197

With the changes in the above PR, I can build a binary RxSwift.framework using Xcode11.5 (swift 5.2), and then import it into a project using Xcode12 beta (swift 5.3) and it works correctly.
Without those changes, we get the "Module cannot be imported by the new compiler" error

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

Probably all platforms but my app only runs on iOS so I do not have a good test harness available for the other platforms.

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [X] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

11.5

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [ ] I have a small code base
  - [X] I have a significant code base

Note: People who install RxSwift via CocoaPods, Carthage (and SPM?) should be unaffected, as I believe those cause the RxSwift source code to be downloaded and built locally, thus the compiler versions of RxSwift and the host app should always match. When using a pre-built binary framework though this issue becomes important

freak4pc:
Lets keep the discussion running in #2197

PR #2197, opened 2020-06-28
reactivex/RxSwift#2196 - enable BUILD_LIBRARY_FOR_DISTRIBUTION and fi…

borland:
…x a few minor quirks in the code that cause it not to work


RxSwift/Disposables/DisposeBase.swift
```diff
@@ -8,7 +8,7 @@
 
 /// Base class for all disposables.
 public class DisposeBase {
-    init() {
+    public init() {
```
borland, line 11: Required as there are various classes that derive from DisposeBase that have a public init method, and the generated .swiftinterface files don't like this.
Seems like it probably wants to be public anyway as the class is public

RxSwift/Observable.swift
```diff
@@ -10,7 +10,7 @@
 ///
 /// It represents a push style sequence.
 public class Observable<Element> : ObservableType {
-    init() {
+    public init() {
```
borland, line 13: Required as there are various classes that derive from Observable that have a public init method, and the generated .swiftinterface files don't like this.
Seems like it probably wants to be public anyway as the class is public

RxSwift/Subjects/ReplaySubject.swift
```diff
@@ -59,7 +59,7 @@ public class ReplaySubject<Element>
     }
     
     /// Returns observer interface for subject.
-    public func asObserver() -> SubjectObserverType {
```
borland, line 62: Required as otherwise when Xcode12 is looking at the generated .swiftinterface it sees that SubjectObserverType = ReplaySubject<Element> and gets confused with it's a circular reference.

Looking at all the other subjects, they look to have moved away from SubjectObserverType (which is marked deprecated); This was the only one left so presumably it is safe

RxSwift/Subjects/ReplaySubject.swift
```diff
@@ -59,7 +59,7 @@ public class ReplaySubject<Element>
...
+    public func asObserver() -> ReplaySubject<Element> {
```
freak4pc, line 62: This is source-breaking, we can't merge this

RxSwift/Observable.swift
freak4pc: This is actually a bit of an unwanted side-effect. It would be preferable to leave the initializer internal in this case. Is there any way to do this anyways? There is no value in initializing Observable directly this way.

RxSwift/Subjects/ReplaySubject.swift
freak4pc: This is a bit tricky since it's a source breaking change. I don't think we'll be able to merge it in because you switched away from an abstracted observer to the concrete type and some people might count on the abstraction. 

Perhaps we could trying AnyObserver here? What do we do in other subjects?

RxSwift/Disposables/DisposeBase.swift
freak4pc: Again, really preferable to avoid this. There is no value in init'ing this directly.

borland: ok. I'll see if I can come up with some other change which leaves this internal but stops it getting emitted into the .swiftinterface. It feels like this might be a bug in swift, but not sure yet

RxSwift/Subjects/ReplaySubject.swift
borland: ReplaySubject declares that SubjectObserverType is aliased to `ReplaySubject<Element>` already, so the return type of `asObserver()` is always `ReplaySubject<Element>` before or after the change.

I looked at `BehaviourSubject`
https://github.com/ReactiveX/RxSwift/blob/a580d07ed002217fd91d8446c3a852486e9beefa/RxSwift/Subjects/BehaviorSubject.swift#L148

`AsyncSubject` https://github.com/ReactiveX/RxSwift/blob/a580d07ed002217fd91d8446c3a852486e9beefa/RxSwift/Subjects/AsyncSubject.swift#L145

and `PublishSubject`
https://github.com/ReactiveX/RxSwift/blob/a580d07ed002217fd91d8446c3a852486e9beefa/RxSwift/Subjects/PublishSubject.swift#L128

and those all return their own concrete types from `asObserver`. They are final though and ReplaySubject is not, so yeah this would be source-breaking if anyone subclassed ReplaySubject and then called asObserver()
Are we trying to get covariant return types here?

A github search doesn't show any classes that subclass ReplaySubject other than a unit test? Is subclassing ReplaySubject a design goal?

Re: changing - If we changed it to `AnyObserver` then that seems  worse than returning `ReplaySubject`?
I.e. if we return `ReplaySubject` from a subclass, we lose some type information, but if we return `AnyObserver` then we lose *more* type information, and don't seem to benefit here?

OTOH this could also just be a bug in swift, it's hard to know what should and shouldn't work

RxSwift/Disposables/DisposeBase.swift
borland: FYI I couldn't infer any way work around this, I filed a bug on swift:
https://bugs.swift.org/browse/SR-13108

See what they say about it

RxSwift/Subjects/ReplaySubject.swift
freak4pc: Ah I see. I wasn’t near the code when commenting, I think this fix is fine in this case.
Let’s see what Apple has to say about the internal initializers

borland: Thanks. If there is a resolution for the internal initalizers, shall I make ReplaySubject sealed here to match the other subjects? Or leave it as overridable and just change the asObserver method?

freak4pc: I'd say we can make it final as well given its a relatively niche subject, and in favor of proceeding into binary stability :)

freak4pc:
@borland did you see the comment on the SR? seems like we need to remove `-enable-testing` unless it causes any issues.

RxSwift/Disposables/DisposeBase.swift
borland: @freak4pc Not sure if you saw the commentary from swift/apple over on their JIRA - it looks like this is a bug in the swift compiler and may be related to having "enable testability" set on the RxSwift project. We should not need to change the RxSwift source code here.

It seemed like a good solution would be to disable testability for release builds (as Apple suggest, this flag alters the meaning of public/private and is undesirable for release builds anyway). I went to validate this and found however that even without the testability flag, the bug still occurs and the swiftinterface file is still broken. I'm waiting now for apple to confirm that I've got my flags right.

If disabling testability proves not to work, what comes to my mind next is a post-build step which patches the generated swiftinterface files to work around the bug.
It's a hack, but I'll probably do this internally for my company's own copies of the RxSwift binary framework. How do you feel about the idea of that for RxSwift proper though?

RxSwift/Observable.swift
borland: discussion in DisposeBase.swfit

RxSwift/Subjects/ReplaySubject.swift
borland: Thanks, I'll make this change and update the PR, just waiting on that swiftinterface thing

RxSwift/Disposables/DisposeBase.swift
freak4pc: @borland Yeah I commented below I don't mind doing disable testability at all, this should be fine. It's unfortunate this doesn't solve it ...

A post-build step is not a great step in the right direction IMO and I'm worried it could cause issues for other dependency managers, or even worse for folks using submodules.

Lets keep our fingers crossed around these, seems like Robert is actually quite responsive given he already had a PR around the testability thing.

BTW did you consider removing the testability flag altogether just to test if it makes any change?

borland: I set the “enable testability” flag to No for all configurations in Xcode, and it was still bugged. I’m not 100% sure if that’s the right flag so waiting to see what Robert says. 
Agree a post build step is not good

borland: re-reviewed against the latest commits

RxSwift/Subjects/ReplaySubject.swift
borland: re-reviewed against the latest commits

Rx.xcodeproj/project.pbxproj
```diff
@@ -3893,6 +3893,7 @@
 		A2897D50225CA1E7004EA481 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				BUILD_LIBRARY_FOR_DISTRIBUTION = NO;
```
borland, line 3896: This is now disabled for Debug and Release-Test configurations, and enabled for the release configuration, doesn't seem like a good idea to generate swiftinterfaces from those anyway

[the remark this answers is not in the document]
freak4pc: Let's go with way of elimination. 

3 and 4 don't seem likely as both are manual and require a lot of maintenance. 

2 is easy, and also means "if you want this for your own use-case, you can achieve it manually using steps we can publish, etc."

1 is tricky. Changing the public API service just because of an Apple bug isn't a great idea today, at least. I'm wondering if we can mark these specific ones with deprecation availability of something, and if that would affect subclasses. Mind checking that?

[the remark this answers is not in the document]
borland: > I'm wondering if we can mark these specific ones with deprecation availability of something, and if that would affect subclasses. Mind checking that?

I'm not quite sure I understand what that means?
Do you mean can we annotate the code with some attribute or something, so it will start throwing compiler warnings in Xcode 12 so we don't forget to come back and fix it?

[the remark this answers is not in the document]
freak4pc: The reason is to provide people from directly using the initializer on these types.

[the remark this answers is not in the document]
freak4pc: What are your thoughts on this? @borland 

Based on the SR comment, the real fix might be right around the corner.

freak4pc:
@borland In the last [Swift development snapshot](https://swift.org/builds/development/xcode/swift-DEVELOPMENT-SNAPSHOT-2020-07-04-a/swift-DEVELOPMENT-SNAPSHOT-2020-07-04-a-osx.pkg), I'm not seeing the override, so a good chance that SR has fixed it.

freak4pc:
@borland Want to check Xcode 12 beta 3 ?

freak4pc:
@borland ping? Would be great having this in RxSwift 6 if everything around this is ready.

borland:
Hi @freak4pc - I can't reply to your "ping" above... I'm not really familiar with the quirks of the github UI
I've been slammed at work so have had no time to look into this. Getting back into it a little bit now.

I'm trying to build RxSwift again using Xcode 12 GM (12A7209).
I notice the following:

- When I build a framework that I have which references RxSwift, it now fails for simulator builds with this error:

failed to build module 'RxSwift' from its module interface; the compiler that produced it, 'Apple Swift version 5.2.4 (swiftlang-1103.0.32.9 clang-1103.0.32.53)', may have used features that aren't supported by this compiler, 'Apple Swift version 5.3 (swiftlang-1200.0.29.2 clang-1200.0.30.1)'
import RxSwift

I suspect that's a red herring though. Immediately above that error is this other one:

<unknown>:0: error: unable to load standard library for target 'arm64-apple-ios8.0'

When I build RxSwift itself using Xcode 12 for the simulator it complains about not supporting iOS 8 - seems like Xcode 12 only goes back as far as 9.0 now, so RxSwift might want to update that. Is that best done in a seperate issue?

borland:
@freak4pc I checked and when building RxSwift 5.1.1 using Xcode 12 GM (12A7209) with only the changes in this PR right now (i.e. no modifications to DisposeBase, only just the ReplaySubject change) that the swiftinterface files have the following for DisposeBase

@_hasMissingDesignatedInitializers public class DisposeBase {
  @objc deinit
}

This looks correct

freak4pc:
Xcode 12 would work on the `rxswift6` branch (where we dropped the deployment target), and we'll actually want to rebase this on that branch and merge it there.

Would be great if you have the time to apply this branch to that working branch, otherwise I can try doing it myself a tad later.

Issue #2172, opened 2020-04-19
Adding share to the Single


NohEunTae:
**Short description of the issue**:

One of the great examples of using Single is to apply it to the network requests.
Also, another great example of using share method is to prevent duplicate calls that reaches from those network requests.
Usually whileConnected attribute/option is recommended to be applied when using share method.
However, in my opinion, there might be a problem when applying it properly to the Single. The problem is, in Single, the complete attribute is set automatically after the connection succeeds, that means the connection is off.
So the way to solve this problem is to convert Single to Observable which ignores Single’s terminate options and returns that converted Observable.

**Expected outcome**:

Performing work ...
Performing work ...

**What actually happens**:

Performing work ...
Performing work ...
Performing work ...
Performing work ...

**Self contained code example that reproduces the issue**:

```swift

import RxSwift

var disposeBag = DisposeBag()
let shareResult = Single.deferred { () -> PrimitiveSequence<SingleTrait, Void> in
    print("Performing work ...")
    return Single.just(())
}.asObservable().share(replay: 1, scope: .whileConnected)

shareResult.subscribe().disposed(by: disposeBag)
shareResult.subscribe().disposed(by: disposeBag)
disposeBag = DisposeBag()

shareResult.subscribe().disposed(by: disposeBag)
shareResult.subscribe().disposed(by: disposeBag)
disposeBag = DisposeBag()
       
```

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

RxSwift 5.5.1

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
11.3.1
```

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [ ] yes (which ones)
  - [x] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [x] I have a small code base
  - [ ] I have a significant code base

freak4pc:
[...]

PR #2198, opened 2020-06-28
Adding shareUntilDisposed to the Single

NohEunTae:
Usually whileConnected attribute/option is recommended to be applied when using share method.

But there is a problem when applying share-whileConnected to the Single. The problem is, in Single, the complete attribute is set automatically after the connection succeeds, that means the connection is off.

I think the simplest way to solve this problem is converting Single to Observable which ignores Single's terminated options and returns that converted Observable. The returned Observable can share until that Observable has been disposed.

This PR might useful for applying share-whileConnected in Single.

### DIFF
--- a/RxSwift/Observables/ShareReplayScope.swift
+++ b/RxSwift/Observables/ShareReplayScope.swift
@@ -156,6 +156,29 @@ extension ObservableType {
     }
 }
 
+/// Single replay count
+public enum SingleReplayCount: Int {
+    case zero, one
+}
+
+extension PrimitiveSequenceType where Trait == SingleTrait {
+    /**
+     Returns an observable sequence that **shares a single subscription to the underlying sequence**, and immediately upon subscription replays  elements in buffer.
+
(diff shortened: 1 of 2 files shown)


freak4pc:
Hey @NohEunTae - 
This seems to me like a very niche operator and use-case. I'm not too sure it fits in RxSwift's code core base, but let me look into this a bit more and give you an answer in a couple of days.

Thanks for the contribution and idea!

freak4pc:
I thought about this some more and feel like it doesn't fit the spirit of RxSwift due to it solving a too-narrow use-case. Thank you so much for taking the time to make this, anyways. 

Let me know if you have anymore questions :)

NohEunTae:
Hey @freak4pc -
Thanks for the answer. It is narrow-use case, but I think it is useful when using single and share together.
For this reason how about adding this feature to RxSwiftExt?
Or can you give me a little direction to solve this issue? ([#2172](https://github.com/ReactiveX/RxSwift/issues/2172))

PR #2199, opened 2020-07-02
Use SwiftPM default minimum deployment targets

sharplet:
Xcode 12 raises the minimum known iOS deployment target from iOS 8.0 to iOS 9.0, resulting in a warning when using RxSwift via SwiftPM. Instead of explicitly specifying minimum deployment targets, we can use SwiftPM's default values which will stay up-to-date whenever the toolchain in use changes the valid deployment target range.

**Update:** Here's a citation for my claim about the default values for minimum deployment targets, in [SE-0236: Package Manager Platform Deployment Settings](https://github.com/apple/swift-evolution/blob/master/proposals/0236-package-manager-platform-deployment-settings.md#proposed-solution):

> A package will be assumed to support all platforms using a predefined minimum deployment version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ DerivedData
 *.ipa
 *.xcuserstate
 *.o
+.swiftpm/
 
 timeline.xctimeline
 .vscode
--- a/Package.swift
+++ b/Package.swift
@@ -42,9 +42,6 @@ extension Target {
 
 let package = Package(
   name: "RxSwift",
-  platforms: [
-    .macOS(.v10_10), .iOS(.v8), .tvOS(.v9), .watchOS(.v3)
-  ],
   products: ([
     [
       .library(name: "RxSwift", targets: ["RxSwift"]),
--- a/RxTest/XCTest+Rx.swift
+++ b/RxTest/XCTest+Rx.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2015 Krunoslav Zaher. All rights reserved.
 //
 
+#if !os(watchOS)
 import RxSwift
 import XCTest
 /**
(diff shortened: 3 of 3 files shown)


sharplet:
What's your thinking behind hard-coding the deployment target in SwiftPM? I'd be happy to make broader, less-specific-to-SwiftPM changes at some point when I have the time, but this seemed like a simpler and more straightforward change given that the compiler already enforces availability checks for newer APIs. My understanding is that it's not necessary to ever specify a platform in the package manifest unless you want a value _higher_ than the toolchain's minimum. Leaving the platforms directive out altogether has the advantage of being portable across both Xcodes 11 & 12.

freak4pc:
It's not a specific thinking, just an instinct. I'm always a fan of specificity (especially for configuration) over magic.

Regardless, as I said, this will come out in the next patch version closer to when Xcode 12 is actually in advanced betas.

freak4pc:
I just saw this: 
> Leaving the platforms directive out altogether has the advantage of being portable across both Xcodes 11 & 12.

This is a valid point, I'll think about this a bit :)

sharplet:
FYI, I've updated the original comment with a link to the SE that documents the defaults for minimum deployment targets.

sharplet:
Specifically, regarding:

> It's not a specific thinking, just an instinct. I'm always a fan of specificity (especially for configuration) over magic.

IMO this isn't relying on magic, but rather documented and intended behaviour.

sharplet:
Any thoughts on whether you'd like to proceed with this change?

freak4pc:
Hey Adam, sorry for the delay here, have been a bit preoccupied.
I think I’ll merge the in, let me take an extra look tomorrow to finalise.

Thanks!

sharplet:
No worries!

sharplet:
@freak4pc Thanks! What do you think about back-deploying this change in a patch release?

freak4pc:
That’s the plan, I’d like to add a few minor additions before cutting a release. I’ll update :)

Shai.

PR #2202, opened 2020-07-13
add tintColor to UIView

changm4n:
add tintColor to UIView


freak4pc:
Hey @changm4n - 
We prefer not adding relatively trivial Binders to RxCocoa core, since anyone can make them very easily if they need it, and adding them to core requires tests, maintaining, etc. 

Thanks for your contribution and time, regardless :)

PR #2207, opened 2020-07-21
update deployment target to iOS 9 to support Xcode 12

dvshelley:


freak4pc:
Hey @dvshelley - 
We're not going to drop iOS 8 support, especially not while Xcode 12 is still in beta. As far as I know, it's only a warning in Xcode 12 and doesn't have any practical meaning. 

Due to versioning concerns, _if_ we actually do this, it'd have to be on RxSwift 6, probably.

dvshelley:
> As far as I know, it's only a warning in Xcode 12 and doesn't have any practical meaning.

It is only a warning when you are doing a debug build. If you try to do an AdHoc/Archive build, the warning turns to an error. My goal is to create a build of our app for testing so our QA folks can find bugs before iOS 14 releases. I can't create an Archive build using the master branch as it is now. The goal of the PR is to fix that.

freak4pc:
Yup, I understand, unfortunately it's not something we'll do in the coming months. We don't make dramatic changes for beta versions until it finalizes further down the road, and we can't drop iOS 8 without a major bump. 

My only suggestion would be using your own fork for the beta builds you need to make.

freak4pc:
Please see the `rxswift6` branch for future development where iOS 9 is the minimum deployment. 
Thanks!

PR #2212, opened 2020-07-30
Add the possibility to convert an observable to driver ignoring error

sebastianvarela:
## Description
This PR exposes a new method for all observables that allow to convert directly to driver ignoring the error event

## Motivation
We have had enough problems in our team to convert some subject to driver that, apart from never failing, many times we do not have or do not know what error value to send in the conversion

## Example

```swift
let someSubject = PublishSubject<String>()
let asDriver = someSubject.asDriverIgnoringError()
```

### DIFF
--- a/RxCocoa/Traits/Driver/ObservableConvertibleType+Driver.swift
+++ b/RxCocoa/Traits/Driver/ObservableConvertibleType+Driver.swift
@@ -54,4 +54,19 @@ extension ObservableConvertibleType {
             }
         return Driver(source)
     }
+
+    /**
+    Converts observable sequence to `Driver` trait. Ignoring Errors
+    - returns: Driver trait.
+    */
+    public func asDriverIgnoringError() -> Driver<Element> {
+        let source = self
+            .asObservable()
+            .map(Optional.init)
+            .observeOn(DriverSharingStrategy.scheduler)
+            .catchErrorJustReturn(nil)
+            .filterNil()
+        
+        return Driver(source)
+    }
 }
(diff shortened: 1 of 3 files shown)


freak4pc:
Hey @SebastianOsinski - 
I appreciate you taking the time to do this, but unfortunately it is too opinionated and specialized to go into the RxSwift core library. 

Choosing to ignore errors for something you convert to a `Driver` or `Signal` might be a great choice for some specific team, but it's definitely not a generic enough concept to have in the core repo, unfortunately. 

By the way, as a side-note, `PublishSubject` should usually be converted to `Signal` - because they both don't replay a value to new subscribers. `Driver` usually goes with `BehaviorSubject`.

freak4pc:
BTW you can simply use `asDriver(onErrorDriveWith: .never())` which would have the same effect as what you wrote above. 

If you want an alias for your own codebase, you can simply do: 

```
public extension ObservableConvertibleType {
    /// Converts observable sequence to `Driver` trait.
    /// In case of an error, the Driver will never emit.
    ///
    /// - returns: A Driver
    func asDriver() -> Driver<Element> {
        return asDriver(onErrorDriveWith: .never())
    }

    /// Converts observable sequence to `Signal` trait.
    /// In case of an error, the Signal will never emit.
    ///
    /// - returns: A Signal
    func asSignal() -> Signal<Element> {
        return asSignal(onErrorSignalWith: .never())
    }
}
```

sebastianvarela:
@freak4pc Thanks for the tip.
I would put this suggestion on our toolset of utilities for Rx:
https://github.com/masmovil/swift-mas-magic-pills-rx

Thanks.

freak4pc:
You can also suggest it for https://github.com/RxSwiftCommunity/RxSwiftExt

PR #2215, opened 2020-08-13
Display bind call location when binding error to relay

kustra:
This PR is about making it easier to debug when an error gets bound to a `PublishRelay` or a `BehaviorRelay`.
Should this happen, `rxFatalErrorInDebug` gets called, which incorporates the caller's file name and line number into its error message. But since `ObservableType.bind(to relays: ...)` doesn't take `file` and `line` arguments, this error message always just displays `Observable+Bind.swift`, with no clue as to where to find the real issue.
The stack doesn't help either, since we're in the `bind` operation's code, so all the frames in the stack trace refer to RxSwift project code locations.
With this change, it immediately points to the relevant `bind` operation call in the project code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 * Fix `DelegateProxy` call to `layoutIfNeeded` for an object without a window. #2076
 * Add `ControlEvent` wrappers to `UIApplication` Notifications. #2116
 * `SharingScheduler.mock(scheduler:action:)` can use throwing function for `action`. #2150
+* Display bind call location when binding error to relay.
 
 ## [5.1.1](https://github.com/ReactiveX/RxSwift/releases/tag/5.1.1)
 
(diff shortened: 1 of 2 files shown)


freak4pc:
Hey @kustra, 
I do understand the use-case but I don't feel comfortable adding this into RxSwift:

1. It doesn't seem to me like a proper use case or a best practice (even looking at other open source projects, but generally speaking there's stock API that does that) to inject the file and line throughout the chain like that. It seems more common in tests, etc.
2. This would be a ABI-breaking change since it would alter the signature of these methods
3. If a crash like this happens during debug, it is more appropriate to just catch it in a debugger

I think the only proper way to go here here is that, if it's really a common issue that happens to you during development, make your own wrapper around these methods perhaps to catch the crashing lines, but I don't see it fitting to add this into RxSwift.

Thank you for understanding

(I'm closing this but feel free to keep commenting here if you have more questions)

PR #2220, opened 2020-08-27
Add `Infallible`

freak4pc:
`Infallible` is an `Observable`-like type that guarantees no failure.

Think about it like `Driver` or `Signal`, but without any specific scheduler, without resource sharing and without replaying. 

This is pegged for RxSwift 6.


kean:
This looks pretty exciting, thank you.

Issue #2227, opened 2020-09-14
Xcode 12 issue - IPHONEOS_DEPLOYMENT_TARGET >= 9.0


StefaniOSApps:
We need to add small changes:
`Showing All Messages The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.0.99.`

freak4pc:
See `rxswift6` branch.

alouanemed:
@freak4pc your response is vague. I've checked that branch and there is no mention of the issue above.

freak4pc:
It's not very vague :) RxSwift 5 will keep supporting iOS 8. 

RxSwift 6 will drop iOS 8 and support iOS 9 and up, so I suggested using that branch.

alouanemed:
Ah, I see. Make sense, Thanks.

I tried that branch but the issue till persist. **Moya/Moya** uses RxSwift and I updated their cart file to use that branch here https://github.com/alouanemed/Moya/blob/master/Cartfile.

freak4pc:
That's strange, I'm using that branch as well. Can you provide a repro?

alouanemed:
Should I update the version to **6** like this : https://github.com/alouanemed/Moya/blob/master/Moya.podspec#L40

alouanemed:
It didn't work, I specified the version to **6.0.0-beta.1**
```
Pre-downloading: `Moya` from `https://github.com/alouanemed/Moya.git`
[!] CocoaPods could not find compatible versions for pod "RxSwift":
  In Podfile:
    Moya/RxSwift (from `https://github.com/alouanemed/Moya.git`) was resolved to 14.0.0, which depends on
      RxSwift (~> 6.0.0-beta.1)

None of your spec sources contain a spec satisfying the dependency: `RxSwift (~> 6.0.0-beta.1)`.

You have either:
 * out-of-date source repos which you can update with `pod repo update` or with `pod install --repo-update`.
 * mistyped the name or version.
 * not added the source repo that hosts the Podspec to your Podfile.


```

freak4pc:
You talked about Cartfile and now mentioned CocoaPods. I'm confused :)

alouanemed:
I am using Cocoa in my project. Moya/Moya uses Cartfile

freak4pc:
Your Podfile should link directly to the `rxswift6` branch: https://stackoverflow.com/a/39930762

[...]

PR #2228, opened 2020-09-14
Set IPHONEOS_DEPLOYMENT_TARGET minimum to 9.0 (Xcode 12 issue)

StefaniOSApps:
See #2227


freak4pc:
Hey @StefaniOSApps - Please check the `rxswift6` branch. We won't be dropping iOS 8 for RxSwift 5.x since it's a breaking change.

PR #2233, opened 2020-09-21
RxSwift#2196 - BUILD_LIBRARY_FOR_DISTRIBUTION=YES and ENABLE_TESTABIL…

borland:
As per the discussion in https://github.com/ReactiveX/RxSwift/pull/2197, brought this across to the RxSwift6 branch


RxSwift/Subjects/ReplaySubject.swift
```diff
@@ -59,7 +59,7 @@ public class ReplaySubject<Element>
     }
     
     /// Returns observer interface for subject.
-    public func asObserver() -> SubjectObserverType {
+    public func asObserver() -> ReplaySubject<Element> {
```
borland, line 62: As per the other PR, this was safe because ReplaySubject is `public` (but not `open`) and the only subclasses are internal to RxSwift and aren't public, such as ReplayBufferBase, ReplayOne, etc

borland:
BTW @freak4pc - not sure if you've hit this yet, but Xcode 12 is essentially forcing people into xcframeworks.

What I'd been doing using Xcode 11 is building a single RxSwift.framework which inside it had a fat binary + swiftmodule files for armv7, arm64, i368 and x86_64

I rebuilt the same and loading in Xcode 12, and I get this error compiling my app for the simulator

    ld: building for iOS Simulator, but linking in dylib built for iOS, file '<my-stuff>/RxSwift.framework/RxSwift' for architecture arm64

This seems to be because now with ARM Macs coming, arm64 is now a valid architecture for both the simulator and the real device, but build settings (#ifdefs etc) might be different across the two.

You can't have two arm64 slices in a single fat binary, so you need to build an xcframework instead, which ends up having two copies of RxSwift.framework

Cheers

freak4pc:
That's pretty interesting. I've excluded arm64 in other open-source projects.
The ZIP file is created using `carthage archive`. 

I might start addnig `.xcframework` files to the release process. Will have to see how iit works out.

Thanks!

freak4pc (approved):

drinkius:
@freak4pc really hope to see `.xcframework`  in the upcoming release!

PR #2241, opened 2020-10-08
rebased boundary-buffer

simpers:
Rebased version of #1887 - could for some reason not change branches in that one

### DIFF
--- a/RxSwift/Observables/Buffer.swift
+++ b/RxSwift/Observables/Buffer.swift
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
         nextTimer.setDisposable(disposable)
     }
 }
+
+extension ObservableType {
+    
+    /**
+     Projects each element of an observable sequence info a buffer that's sent out when the boundary Observable emits a .next event or a .complete event, if the accumulated buffer's count
+     is over zero.
+     
+     - seealso: [buffer operator on reactivex.io](http://reactivex.io/documentation/operators/buffer.html)
+     
+     - parameter boundary: Observable that will act as a boundary between each window.
+     - returns: An observable sequence of buffers.
+     */
+    public func buffer<BoundaryElement>(boundary: Observable<BoundaryElement>) -> Observable<[E]> {
+        return BufferBoundary(source: self.asObservable(), boundary: boundary)
+    }
+    
+    /**
+     Projects each element of an observable sequence info a buffer that's sent out when the boundary Observable emits a .next event or a .complete event, if the accumulated buffer's count
+     is over zero.
+     
+     A useful real-world example of this is when you have a resource that needs to be updated depending on some asynchronous tasks and you want to accumulate all of those tasks's results
+     and just PUT/PATCH the remote resource once per batch of results instead of once per result.
+     
+     - seealso: [buffer operator on reactivex.io](http://reactivex.io/documentation/operators/buffer.html)
+     
+     - parameter debounce: Amount of time to debounce the source as the boundary between each window.
+     - parameter scheduler: Scheduler to run debouncing on.
+     - returns: An observable sequence of buffers.
+     */
+    public func buffer(debounce: RxTimeInterval, scheduler: SchedulerType) -> Observable<[E]> {
+        let shared = self.share()
+        return shared.buffer(boundary: shared.debounce(debounce, scheduler: scheduler))
+    }
+}
+
+final fileprivate class BufferBoundary<Element, BoundaryElement> : Producer<[Element]> {
+    
+    fileprivate let _source: Observable<Element>
+    fileprivate let _boundary: Observable<BoundaryElement>
+    
+    init(source: Observable<Element>, boundary: Observable<BoundaryElement>) {
+        _source = source
+        _boundary = boundary
+    }
+    
+    override func run<O : ObserverType>(_ observer: O, cancel: Cancelable) -> (sink: Disposable, subscription: Disposable) where O.E == [Element] {
+        let sink = BufferBoundarySink(parent: self, observer: observer, cancel: cancel)
+        let subscription = sink.run()
+        return (sink: sink, subscription: subscription)
+    }
+}
+
+final fileprivate class BufferBoundarySink<Element, BoundaryElement, O: ObserverType>
+    : Sink<O>
+    , LockOwnerType
+    , ObserverType
+    , SynchronizedOnType where O.E == [Element] {
+    typealias Parent = BufferBoundary<Element, BoundaryElement>
+    typealias E = Element
+    
+    private let _parent: Parent
+    
+    let _lock = RecursiveLock()
+    
+    // state
+    private let _serialDisposable = SerialDisposable()
+    private var _buffer = [E]()
+    
+    init(parent: Parent, observer: O, cancel: Cancelable) {
+        _parent = parent
+        super.init(observer: observer, cancel: cancel)
+    }
+    
+    func run() -> Disposable {
+        let disposable = SingleAssignmentDisposable()
+        _serialDisposable.disposable = disposable
+        disposable.setDisposable(_parent._boundary.subscribe { (event: Event<BoundaryElement>) in
+            switch event {
+            case .next(_):
+                self.startNewWindowAndSendCurrentOne()
+                break
+            case .error(let error):
+                self._buffer = []
+                self.forwardOn(.error(error))
+                self.dispose()
+                break
+            case .completed:
+                if self._buffer.count > 0 {
+                    self.forwardOn(.next(self._buffer))
+                }
+                self.forwardOn(.completed)
+                self.dispose()
+                break
+            }
+        })
+        
+        return Disposables.create(_serialDisposable, _parent._source.subscribe(self))
+    }
+    
+    func startNewWindowAndSendCurrentOne() {
+        let buffer = _buffer
+        _buffer = []
+        if buffer.count > 0 {
+            forwardOn(.next(buffer))
+        }
+    }
+    
+    func on(_ event: Event<E>) {
+        synchronizedOn(event)
+    }
+    
+    func _synchronized_on(_ event: Event<Element>) {
+        switch event {
+        case .next(let element):
+            _buffer.append(element)
+        case .error(let error):
+            _buffer = []
+            forwardOn(.error(error))
+            dispose()
+        case .completed:
+            if _buffer.count > 0 {
+                forwardOn(.next(_buffer))
+            }
+            forwardOn(.completed)
+            dispose()
+        }
+    }
+}
--- a/RxSwift/Observables/Buffer.swift
+++ b/RxSwift/Observables/Buffer.swift
@@ -148,7 +148,7 @@ extension ObservableType {
      - parameter boundary: Observable that will act as a boundary between each window.
      - returns: An observable sequence of buffers.
      */
-    public func buffer<BoundaryElement>(boundary: Observable<BoundaryElement>) -> Observable<[E]> {
+    public func buffer<BoundaryElement>(boundary: Observable<BoundaryElement>) -> Observable<[Element]> {
         return BufferBoundary(source: self.asObservable(), boundary: boundary)
     }
     
@@ -165,7 +165,7 @@ extension ObservableType {
      - parameter scheduler: Scheduler to run debouncing on.
      - returns: An observable sequence of buffers.
      */
-    public func buffer(debounce: RxTimeInterval, scheduler: SchedulerType) -> Observable<[E]> {
+    public func buffer(debounce: RxTimeInterval, scheduler: SchedulerType) -> Observable<[Element]> {
         let shared = self.share()
         return shared.buffer(boundary: shared.debounce(debounce, scheduler: scheduler))
     }
@@ -181,7 +181,7 @@ final fileprivate class BufferBoundary<Element, BoundaryElement> : Producer<[Ele
         _boundary = boundary
     }
     
-    override func run<O : ObserverType>(_ observer: O, cancel: Cancelable) -> (sink: Disposable, subscription: Disposable) where O.E == [Element] {
+    override func run<O : ObserverType>(_ observer: O, cancel: Cancelable) -> (sink: Disposable, subscription: Disposable) where O.Element == [Element] {
         let sink = BufferBoundarySink(parent: self, observer: observer, cancel: cancel)
         let subscription = sink.run()
         return (sink: sink, subscription: subscription)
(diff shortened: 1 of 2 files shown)


simpers:
@freak4pc here you go :)

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+     
+     - seealso: [buffer operator on reactivex.io](http://reactivex.io/documentation/operators/buffer.html)
+     
+     - parameter debounce: Amount of time to debounce the source as the boundary trigger between each window.
+     - parameter scheduler: Scheduler to run debouncing on.
+     - returns: An observable sequence of buffers.
+     */
+    public func buffer(debounce: RxTimeInterval, scheduler: SchedulerType) -> Observable<[Element]> {
+        let shared = self.share()
+        return shared.buffer(trigger: shared.debounce(debounce, scheduler: scheduler))
```
freak4pc, line 170: Is there a way to do this without share? I don't think we can introduce it this way because it inevitably makes the `Observable` shared downstream which could cause bugs and be an unexpected side-effect for consumers

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+    }
+}
+
+final fileprivate class BufferTrigger<Element, TriggerElement> : Producer<[Element]> {
+
```
freak4pc, line 175:
```suggestion
```

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+     */
+    public func buffer(debounce: RxTimeInterval, scheduler: SchedulerType) -> Observable<[Element]> {
+        let shared = self.share()
+        return shared.buffer(trigger: shared.debounce(debounce, scheduler: scheduler))
+    }
+}
+
+final fileprivate class BufferTrigger<Element, TriggerElement> : Producer<[Element]> {
+    
+    fileprivate let _source: Observable<Element>
```
freak4pc, line 176:
```suggestion
    private let source: Observable<Element>
```

Please remove all underscores (we have `private` for that)
Make sure `private` isn't enough here - it's preferable over `fileprivate`

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+    fileprivate let _trigger: Observable<TriggerElement>
+    
+    init(source: Observable<Element>, trigger: Observable<TriggerElement>) {
+        _source = source
+        _trigger = trigger
+    }
+    
+    override func run<O : ObserverType>(_ observer: O, cancel: Cancelable) -> (sink: Disposable, subscription: Disposable) where O.Element == [Element] {
```
freak4pc, line 184:
```suggestion
    override func run<O : ObserverType>(_ observer: O, cancel: Cancelable)
     -> (sink: Disposable, subscription: Disposable) where O.Element == [Element] {
```

A bit of breathing room :)

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+        _parent = parent
+        super.init(observer: observer, cancel: cancel)
+    }
+    
+    func run() -> Disposable {
+        let disposable = SingleAssignmentDisposable()
+        _serialDisposable.disposable = disposable
+        disposable.setDisposable(_parent._trigger.subscribe { (event: Event<TriggerElement>) in
+            switch event {
+            case .next(_):
```
freak4pc, line 217:
```suggestion
            case .next:
```

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+    , ObserverType
+    , SynchronizedOnType where O.Element == [Element] {
+    typealias Parent = BufferTrigger<Element, TriggerElement>
+    typealias E = Element
+    
+    private let _parent: Parent
+    
+    let lock = RecursiveLock()
+    
+    // state
```
freak4pc, line 203: I'd remove this comment

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+    }
+    
+    override func run<O : ObserverType>(_ observer: O, cancel: Cancelable) -> (sink: Disposable, subscription: Disposable) where O.Element == [Element] {
+        let sink = BufferTriggerSink(parent: self, observer: observer, cancel: cancel)
+        let subscription = sink.run()
+        return (sink: sink, subscription: subscription)
+    }
+}
+
+final fileprivate class BufferTriggerSink<Element, TriggerElement, O: ObserverType>
```
freak4pc, line 191:
```suggestion
final fileprivate class BufferTriggerSink<Element, TriggerElement, Observer: ObserverType>
```

Please avoid the use of single-letter generics when possible (same for `E`)

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+    
+    init(parent: Parent, observer: O, cancel: Cancelable) {
+        _parent = parent
+        super.init(observer: observer, cancel: cancel)
+    }
+    
+    func run() -> Disposable {
+        let disposable = SingleAssignmentDisposable()
+        _serialDisposable.disposable = disposable
+        disposable.setDisposable(_parent._trigger.subscribe { (event: Event<TriggerElement>) in
```
freak4pc, line 215: Don't think the type hinting is needed here
```suggestion
        disposable.setDisposable(_parent._trigger.subscribe { event in
```

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+            switch event {
+            case .next(_):
+                self.startNewWindowAndSendCurrentOne()
+                break
+            case .error(let error):
+                self._buffer = []
```
freak4pc, line 221: Doesn't this need some sort of locking? e.g. making sure there's exclusive access to the buffer?

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+                self.dispose()
+                break
+            }
+        })
+        
+        return Disposables.create(_serialDisposable, _parent._source.subscribe(self))
+    }
+    
+    func startNewWindowAndSendCurrentOne() {
+        let buffer = _buffer
```
freak4pc, line 239: This method is used just once,  I'd move this code over to that call-site

RxSwift/Observables/Buffer.swift
```diff
@@ -136,3 +136,131 @@ final private class BufferTimeCountSink<Element, Observer: ObserverType>
...
+    func synchronized_on(_ event: Event<Element>) {
+        switch event {
+        case .next(let element):
+            _buffer.append(element)
+        case .error(let error):
+            _buffer = []
+            forwardOn(.error(error))
+            dispose()
+        case .completed:
+            if _buffer.count > 0 {
```
freak4pc, line 259:
```suggestion
            if !buffer.isEmpty {
```

Preferable over explicit count check

Tests/RxSwiftTests/Observable+BufferTests.swift
```diff
@@ -142,4 +142,68 @@ extension ObservableBufferTest {
...
+            .next(320, 5),
+            .next(350, 6),
+            .next(370, 7),
+            .next(875, 8),
+            .next(900, 9),
+            .next(999, 10),
+            .completed(1000)
+            ]).share()
+        
+        let debounced = hot.debounce(.seconds(30), scheduler: testScheduler)
```
freak4pc, line 163: What's the reason for the debouncing here? might be missing something

Tests/RxSwiftTests/Observable+BufferTests.swift
```diff
@@ -142,4 +142,68 @@ extension ObservableBufferTest {
...
+            .next(240, 3),
+            .next(255, 4),
+            .next(320, 5),
+            .next(350, 6),
+            .next(370, 7),
+            .next(875, 8),
+            .next(900, 9),
+            .next(999, 10),
+            .completed(1000)
+            ]).share()
```
freak4pc, line 161: don't think the share() is needed for a test observable

freak4pc (changes requested):
Hey, thank you. Was finally able to review this and have a few notes :) 
Thanks for your time!

freak4pc:
Also two more things:

1. Please add a mirror of this operator to **Infallible.swift**.
2. Run `./scripts/package-spm.swift` locally to make sure the test suite is updated or the CI won't pass.

Thank you!

RxSwift/Observables/Buffer.swift
simpers: Ah, can you tell me more of what that could look like?

freak4pc: You won't like this, but the right way is probably passing an optional `debounceTime` or something similar named to the `Producer` and `Sink` and have that logic sit there, without relying on `debounce` at all. 

In general it isn't a good practice for custom operators inside RxSwift to count on other operators as we prefer atomic units, as much as possible.

RxSwift/Observables/Buffer.swift
simpers: `fileprivate` is necessary as `BufferTriggerSink` is accessing them through its `parent` property. So it is a private field for this file.

freak4pc: perfect. Just make sure that's true for all uses of `fileprivate`. Thanks:)

RxSwift/Observables/Buffer.swift
freak4pc: PS this definitely looks like something that'll need locking at the buffer could change while you made a copy of it ,etc.

RxSwift/Observables/Buffer.swift
simpers: I think we'd have noticed that by now if that had been an issue as we've used this in production for over a year now haha. Or more 🤔 So I'd say it's safe?

freak4pc: Might've been a bit opportunistic, but I'll go with your battle-tested experience here :) 
Worst case we can fix it later on.

RxSwift/Observables/Buffer.swift
simpers: Any suggestions about this one though?

RxSwift/Observables/Buffer.swift
simpers: I'll ponder on this one a bit just to make sure though while I wait for a response about the `shared` part.

Tests/RxSwiftTests/Observable+BufferTests.swift
simpers: I don't remember so I am not sure haha. I'll have to test whether that has any affect or why it may be there.

simpers: Oh, well I guess this was the quick and dirty "generic" trigger I came up with in the heat of the moment. It is essentially the same thing as `buffer(debounce:)` though so one could change this one for something else. Any ideas? Maybe something that counts? Makes sure each buffer is of a specific size, perhaps?

simpers:
> Also two more things:
> 
> 1. Please add a mirror of this operator to **Infallible.swift**.
> 2. Run `./scripts/package-spm.swift` locally to make sure the test suite is updated or the CI won't pass.
> 
> Thank you!

1. Could you give me some pointers about this one?
2. I ran the script and it updated `Sources/AllTestz/main.swift` ✅

RxSwift/Observables/Buffer.swift
simpers: Sorry, as I asked for a suggestion the page hadn't refreshed so I missed the above.

Tests/RxSwiftTests/Observable+BufferTests.swift
freak4pc: Manual signal sounds good, thanks!

freak4pc:
> Could you give me some pointers about this one?

Yes, in **Infallible+Operators.swift** there are mirrors of _most_ operators. Basically just make one that mirrors this operator for the `Infallible` unit (should be as easy as wrapping with the initializer but let me know if you have questions)

simpers:
> > Could you give me some pointers about this one?
> 
> Yes, in **Infallible+Operators.swift** there are mirrors of _most_ operators. Basically just make one that mirrors this operator for the `Infallible` unit (should be as easy as wrapping with the initializer but let me know if you have questions)

Right. I think I got it in regards what I've got to do but I do have two questions:

1. Regarding Infallible:
Am I correct when reasoning that for the `buffer(trigger:)` case I'd need the trigger to be a `Infallible<TriggerElement>`, as the Infallible is guaranteed to not itself emit errors, it should also be guaranteed to not receive any errors? 

2. Does this project support XCode 12? I'm struggling a lot with bugs in XCode atm so just wanted to know if that's something you know about

freak4pc:
1. Yes
2. It does, what bugs ?

simpers:
> 1. Yes
> 2. It does, what bugs ?

1. Great, done!
2. 

It insists on not opening the Workspace, and instead opens it like a Swift package (I assume), which it cannot compile and if I try to edit a file after attempting to compile it will tell me the file does not exist so I cannot save. At this point I cannot exit the project either, as that would require saving haha. So I basically have to force-quit and then make sure to not edit anything when I start it up again so that I don't get stuck in this loop.

If I open the project's workspace file it's fine, but XCode remembers the project as if I opened it as a Swift package meaning I have to remember to ignore it in the "recent projects" menu. Is this something you've experienced before?

freak4pc:
Yup, this happens since Xcode 11 IMO. It's Xcode's default, nothing related to this project.

freak4pc:
About SPM not building:
https://bugs.swift.org/browse/SR-12303

About "File doesn't exist":
image

simpers:
> Yup, this happens since Xcode 11 IMO. It's Xcode's default, nothing related to this project.

Okay, thanks for the help! :)

parse:
@freak4pc Has there been any further thoughts on implementation of the buffer operator from your end? http://reactivex.io/documentation/operators/buffer.html

PR #2245, opened 2020-10-17
Check Android as well as Linux as a separate OS

gonzalolarralde:
`os(Android)` does not imply `os(Linux)`, therefore without this explicit check code just won't run.

I've run the tests on both Android and Linux, they are all passing when applying this changes to `main`. There are either test-issues on `develop`, or my toolchain version doesn't match what's being used to verify them when changed, because of that I cannot confirm that tests remain passing on `develop` but my assumption is that the will once tests are runnable on that platform.

As part of this changeset I'm also removing some redundant ` && !os(Linux)` checks, but please let me know if it's preferred to keep those even tho they're wrapped with broader checks. In that case I can just append ` && !os(Android)` as I did with the other cases.


freak4pc:
Hey @gonzalolarralde - TBH this project has never supported Android out of the box. 

Could you explain your motivation towards this? I appreciate the PR, but I don't know who would actually maintain this support when you're not here :)

freak4pc:
PS - If we _will_ actually merge this, we need a basic test suite for Android that validates at least _some_ functionality works. 
Another option is just ensuring it builds correctly on Android by adding a CI workflow, but I'm not strong enough in Android to know how to bundle this Swift framework into an Android project TBH.

freak4pc:
Also `develop` wasn't align to the latest since we just released RC1 of RxSwift 6, so you'll have to rebase or reapply the changes unfortunately :(

gonzalolarralde:
Thanks for your prompt response and for being open to consider this PR!

Android support for Swift, fully based on community support, has been gaining some momentum through the year. While there are applications using it in production it can still be considered experimental in many aspects. I'm at the moment exploring libraries that provide full support on Linux and testing them on the Android platform.

Although Android relies on bionic instead of glibc for its C standard library, on swift this keeps get imported under the `Glibc` namespace, as the compatibility level that both libraries have allows it. libdispatch is fully capable on Android and Foundation mostly is, with some exceptions that to my understanding are not relevant to RxSwift (like the impossibility of getting backtraces, lack of multi-user, and shorter thread naming support on certain versions) so currently there are not any specifics needed to support RxSwift on Android that I'm aware of. As so demonstrates the test-suite execution results at least.

I've executed the same test suite used for Linux and all tests are passing on `main` (will recheck `develop` after the rebase, thanks for pointing that out!). I can't think on any Android-specific test but if you have any ideas of tests that are currently not targeted for Linux but that might be appropriate in a sandboxed environment (iOS-specific tests maybe?) I could try to enable those.

Regarding configuring a CI instance, I can definitely try to help on setting that up for building using one of the toolchains available. Executing the test suite is a more complicated as it requires a real device connected. I've tried in the past tho executing tests on a headless simulator and had partial success, this test executions can become flaky tho. 

Thanks for your time!

freak4pc:
That's super interesting. Could be quite seeing how this could unfold in sharing logic in frameworks, etc.
Unfortunately, I don't think it fits inside this main repo due to two main reasons:

1. Android is not an officially supported platform for Swift (as opposed to Swift for Windows, for example), so it's hard to say how long term and stable it is.

2. You did make the code changes but there is unfortunately nothing that guarantees you would be here to maintain it if anything goes wrong on Android, which is a platform I wouldn't want to receive GitHub Issues / support tickets for at the moment.

Maybe a separate fork would make more sense, and if things evolve well on the Swift on Android side, we can pul l it in at some point.

gonzalolarralde:
@freak4pc that makes sense, will close this. Thanks for your time and feedback!

PR #2251, opened 2020-10-26
Add namespace to Logging

marksands:
xcframeworks can exhibit a lingering bug in the generated `swiftinterface` file that fails to properly resolve a module name if there exists a type with the same name as a module. When important RxCocoa as an xcframework along with another dependency named `Logging`, the swift compiler gets hung up on resolving the correct `Logging` when referenced from the target app.

Related discussions:
https://forums.swift.org/t/frameworkname-is-not-a-member-type-of-frameworkname-errors-inside-swiftinterface/28962
https://bugs.swift.org/browse/SR-898

There's a few ways to resolve this:

1. Do nothing, let other developers attempt to avoid using a type or module named `Logging` or rename their code.
2. Rename `Logging` to avoid any collision. This is what this pull request does, by prefixing `Logging` to `RXLogging`. No hard feelings if this is closed, but I figured this would make raising the issue and resolving it easier for the maintainers.
3. Make `Logging` into something more generic. Right now it is very specific in that it's used by RxCocoa for only logging `URLRequest`s. Perhaps it should do more? Or maybe it should be renamed to something specific about network requests?

Since RxSwift 6.0 is about to land, I was hoping this would be an opportunity to rename this type since this is the time for breaking changes.

Happy to dialog about this issue here.


freak4pc:
Thanks @marksands - that's a crazy little bug given how much Swift guidelines ask us to avoid name spacing. 
Does it conflict with something _inside_ your own app called `Logging` ?

marksands:
> Thanks @marksands - that's a crazy little bug given how much Swift guidelines ask us to avoid name spacing.
> Does it conflict with something _inside_ your own app called `Logging` ?

In our case, we have a module named Logging, and any properties or functions inside are ignored, since it's trying to resolve to `RxCocoa.Logging` 😬

freak4pc:
Gotya. Ok, here's what I have in mind. 

Remove **Logging.swift** entirely. In **URLSession+Rx.swift**, add inside the reactive extension:

```
extension Reactive where Base: URLSession {
    public static var shouldLogRequest: (URLRequest) -> Bool = { _ in
        #if DEBUG
            return true
        #else
            return false
        #endif
    }
}
```

This should allow simply doing `URLSession.rx.shouldLogRequest = { request in ??? }` as needed while not needing any new namespace

marksands:
@freak4pc Updated PR. Side note: it seems that `Static stored properties not supported in generic types`, so I had to create an extension where `Base == URLSession`, instead of using the existing extension `where Base: URLSession`

freak4pc:
Yup, using an `==` extension is the right thing to do here :) Thanks!

RxCocoa/Foundation/URLSession+Rx.swift
```diff
@@ -114,7 +114,7 @@ extension Reactive where Base: URLSession {
             // smart compiler should be able to optimize this out
             let d: Date?
 
-            if Logging.URLRequests(request) {
+			if URLSession.rx.shouldLogRequest(request) {
```
freak4pc, line 117: indentation is broken herre

RxCocoa/Foundation/URLSession+Rx.swift
```diff
@@ -228,3 +228,13 @@ extension Reactive where Base: URLSession {
...
+extension Reactive where Base == URLSession {
+/// Log URL requests to standard output in curl format.
+   public static var shouldLogRequest: (URLRequest) -> Bool = { _ in
+	   #if DEBUG
+		   return true
+	   #else
+		   return false
+	   #endif
+   }
+}
```
freak4pc, line 240: Indentation seems wrong here
```suggestion
    /// A closure to determine if cURL-formatted request information shroud be loged
    /// for a specific `URLRequest`
    public static var shouldLogRequest: (URLRequest) -> Bool = { _ in
        #if DEBUG
            return true
        #else
            return false
        #endif
    }
}
```

Documentation/GettingStarted.md
```diff
@@ -1089,20 +1089,11 @@ URLSession.shared.rx.response(myURLRequest)
...
-    public static var URLRequests: LogURLRequest =  { _ in
-    #if DEBUG
-        return true
-    #else
-        return false
-    #endif
-    }
+URLSession.rx.shouldLogRequest = { _ in
+	return true
 }
```
freak4pc, line 1097: Maybe a more practical example here? 
```suggestion
URLSession.rx.shouldLogRequest = { request in
  // Only log requests to reactivex.org     
  request.url?.host == "reactivex.org" || request.url?.host == "www.reactivex.org"
}
```

Documentation/GettingStarted.md
```diff
@@ -1089,20 +1089,11 @@ URLSession.shared.rx.response(myURLRequest)
 ```
 ### Logging HTTP traffic
 
-In debug mode RxCocoa will log all HTTP request to console by default. In case you want to change that behavior, please set `Logging.URLRequests` filter.
+In debug mode RxCocoa will log all HTTP request to console by default. In case you want to change that behavior, please set the `URLSession.rx.shouldLogRequest` filter.
```
freak4pc, line 1092: As long as you're here:
```suggestion
RxCocoa will log all HTTP request info to the console by default when run in debug mode. You may overwrite the `URLSession.rx.shouldLogRequest` closure to define which requests should and shouldn't be logged. 
```

freak4pc (commented):
few tiny notes but LGTM afterwards

freak4pc (approved):
:shipit: pending CI

Thank you @marksands 🙌

PR #2257, opened 2020-11-12
Bump arity from 8 to 9 in zip and combineLatest

sebastianvarela:
I made this suggestion because in some case we need to combine or zip 9 observables and see that the change is easy to do with the templates but is hard to implement in an extension.

Thanks in advice!


freak4pc:
Hey @sebastianvarela - the current arity feels enough in the scope of RxSwift core itself. It is quite rare to need additional arity. In those cases, it's preferable the app itself would host the extra code as needed. 

Thanks for the suggestions:)

PR #2265, opened 2020-12-10
use present() instead of show() with UIAlertController

yodagamaheshan:
In current iOS versions `show()` does not work as expected with UIAlertController. So that newcomers may confuse with the sample projects. I change it to the correct one.


freak4pc:
Thanks!

freak4pc (approved):

PR #2269, opened 2021-01-01
Fix drive() parameter documents

yanamura:
Fixed documents like [this commit](https://github.com/ReactiveX/RxSwift/commit/dc3aebb680c64d607c92dc1f4ac0e463adb8ec90)


freak4pc (approved):
Thank you for catching this! 👀 💯

PR #2280, opened 2021-01-28
RxTest: Allow interlocked sleep for all scheduled events

serges147:
This allows to emulate computer sleep (suspend) during which all timers are "frozen".


freak4pc:
Thanks for the contribution! This is a bit of a big change so I'll have to think about it. Thanks for the idea :) 
Could you fix the CI issues, though?

serges147:
Strange errors in CI... I'm not sure that's mine. And, of cause, I run all tests locally. @freak4pc Could you please take a look? Thnx

serges147:
Oh, I see... I need to run `./scripts/package-spm.swift`

serges147:
@freak4pc Hi! Have you had a chance to think about it? Thanks!

freak4pc:
My intuition is that this is too much of a niche use case to add it. Can you explain what’s you’re use case? What are you using this for?

serges147:
The use case is following:
- you have a daily schedule for an action, let's say, every 24 h.
- if you just do usual `.schedule(due: .seconds(86400), period: .seconds(86400)) { ... }` you are in troble if machine spend most of it time in sleep. It may wake up for 30 min, and then sleep the rest of 23h and 30mins. As a result what is supposed to to run daily will in reality triggered in 48 days (24h / 30 min = 48).
- to prevent this you should schedule with much more frequent period (like maybe every 5 or 10 mins) and then just track does the target next daily date has been reached or not. in such case machine can sleep even days, and then as soon it will wakeup it will run the "missed" target date action in couple minutes.

And now, to be able test that such "smart" 24h scheduling was indeed made correctly, I need that VirtualTimeScheduler supports emulation of "delay" due to "machine sleep" (when all timer counters are "frozen" which essentially means that everything have been scheduled before should be prolonged by interval of such "deep" sleep.

freak4pc:
Thanks for the explanation. It does seem quite of a niche use-case, I'm not entirely positive it needs to go into the core library. 

Is there any more common use case you can think of for this aside for 24 hours interval, for example?

serges147:
It might be "a niche use-case" especially if most of develpers don't realize how scheduler behaves during sleeps - it was quite a surprise for me at first. Also, I belive that a lot of products have something like daily schedules - so they all might be affected.

Can we do something different, like instead expose something so that who needs such behavior could add it...

RxSwift/Schedulers/VirtualTimeScheduler.swift
```diff
@@ -195,15 +195,29 @@ open class VirtualTimeScheduler<Converter: VirtualTimeConverterType>
...
     }
 
-    /// Advances the scheduler's clock by the specified relative time.
-    public func sleep(_ virtualInterval: VirtualTimeInterval) {
+    /// Advances the scheduler's clock by the specified relative time interval.
+    ///
+    /// When `deep` is `true` it also advances time of all already scheduled item by the interval.
+    /// This allows to emulate computer sleep (suspend) during which all timers are "frozen".
+    ///
+    public func sleep(_ virtualInterval: VirtualTimeInterval, deep: Bool = false) {
```
freak4pc, line 203: I thought of this some more and I think I can "live with it" since you presented a solid use-case.
I don't like the naming `deep` though - I think something that indicates it pushes all events simultaneously/synchronized/interlocked makes more sense.

Either:

```suggestion
    public func sleep(_ virtualInterval: VirtualTimeInterval, synchronized: Bool = false) {
```

O:

```suggestion
    public func sleep(_ virtualInterval: VirtualTimeInterval, interlocked: Bool = false) {
```

WDYT ?

serges147: 👍 I'l go with `interlocked`.

freak4pc:
Made a tiny fix to revert making time mutable

PR #2281, opened 2021-01-29
fix "'RxCocoaRuntime.h' file not found" compile error

faimin:
fix `'RxCocoaRuntime.h' file not found` error when not open the `use_frameworks!` configuration  in `podfile`.

generally occurs in the scene of mixed `Swift` and `Objective-C` project with `multiple_pod_projects`.

Add this configuration to enable it to be compiled as a static library in mixed project.


freak4pc (approved):
Nice catch! Thank you :)

PR #2284, opened 2021-02-03
remove link with libswiftXCTest.dylib deprecated and now removed on i…

srabiller:
…OS 14.5Beta, instead use its replacement, libXCTestSwiftSupport.dylib using ENABLE_TESTING_SEARCH_PATHS = YES.
Source: https://developer.apple.com/documentation/xcode-release-notes/xcode-12_5-beta-release-notes

XCTest’s legacy Swift overlay library (libswiftXCTest.dylib) has been removed. Don’t use this library; instead use its replacement, libXCTestSwiftSupport.dylib. Any frameworks that link XCTest and encounter build issues due to this removed legacy library need to replace their manual framework or library search paths with the build setting ENABLE_TESTING_SEARCH_PATHS=YES, which automatically configures a target with the search paths needed to locate XCTest libraries. (70365050)


freak4pc (approved):
thanks!

Issue #2270, opened 2021-01-03
ReactiveCompatible is a class protocol in RxSwift 6


groue:
**Short description of the issue**:

`ReactiveCompatible` is a class-only protocol in RxSwift 6, which prevents declaring handy `.rx` extensions on structs. This affects the RxGRDB repository.

**Expected outcome**:

`ReactiveCompatible` can be declared on struct extension, OR there exists another way to declare handy `.rx` extensions on structs.

**What actually happens**:

Compiler complains, and I don't know how to manage this breakage.

**Self contained code example that reproduces the issue**:

```swift
// No longer compiles
protocol P { ... }
struct S: P { ... }
extension S: ReactiveCompatible { } // Non-class type 'S' cannot conform to class protocol 'ReactiveCompatible'
extension Reactive where Base: P { /* handy extensions */ }

let s = S(...)
s.rx.yeah()
```

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

RxSwift 6.0.0

**Platform/Environment**

  - [x] iOS
  - [x] macOS
  - [x] tvOS
  - [x] watchOS
  - [x] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
Version 12.3 (12C33)
```

freak4pc:
Hey @groue - Thanks for reporting :) 
This is a breaking change and a known downside of the new `@dynamicMemberLookup` approach we took for RxSwift 6. 

Luckily, you can make a wrapper to resolve this issue:

```swift
struct MyObject { } 

extension MyObject {
    struct Rx { 
        private let base: MyObject
        init(_ base: MyObject) { self.base = base }
    }
}

public extension MyObject.Rx { 
    /// Reactive extensions
}
```

Everything else can remain as-is. 

Does this solution accommodate that specific use-case?

groue:
Thanks for your help and explanation 👍 !

I have too many structs to extend, so I'll basically have to restore the old `Reactive` struct-compatible wrapper. Now I'm just struggling to find a distinct name so that it does not conflict with RxSwift's ;-)

'll close this issue. Maybe, if several users report similar problems, this breaking change will turn into a regression that deserves to be addressed.

iDevid:
[...]

PR #2285, opened 2021-02-03
Removed ReactiveCompatible & Reactive conformance to objects

iDevid:
As explained in the issue #2270, ReactiveCompatible & Reactive protocols are from RxSwift 6 only class protocols due to the @dynamicMemberLookup approach.

With this pull request, I removed the limit of the class protocols without removing the @dynamicMemberLookup approach.

### DIFF
--- a/RxSwift/Reactive.swift
+++ b/RxSwift/Reactive.swift
@@ -24,7 +24,7 @@
  */
 
 @dynamicMemberLookup
-public struct Reactive<Base: AnyObject> {
+public struct Reactive<Base: Any> {
     /// Base object to extend.
     public let base: Base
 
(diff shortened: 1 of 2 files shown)


freak4pc:
Very simple and elegant solution! I'll run a few tests but if it passes I'm happy to merge it. 
Thank you!

iDevid:
Seems that this will break Object base mutation.

I tried to fix this with this solution:

    /// Base object to extend.
    private var _base: Base
   
    /// Base object to extend.
    public var base: Base {
        get { _base }
        // this enables using Reactive to "mutate" base type
        // swiftlint:disable:next unused_setter_value
        set { }
    }


But the I got a strange error during tests:

Schermata 2021-02-03 alle 12 25 12

Any hint?

iDevid:
Hi checked test errors, and I discovered that during RxSwift 6.0 development you changed the ReactiveTests paradigm.
I reverted that to precedent version because now ReactiveCompatible could be also a Struct, so the compiler doesn't know what's behind the rx wrapper.

freak4pc:
@iDevid looks good. I'll add a few more tests to improve the coverage of cases here, thank you!

PR #2286, opened 2021-02-04
Make object of withUnretained optional

lammertw:
Seems to me there is no reason that the object passed in to `withUnretained` cannot be optional. By making it optional it allows rewriting the following:

```
class MyView: UIView {

    let subview: UIView? = UIView()

    let observable = UIButton().rx.tap
    let disposeBag = DisposeBag()

    func setup() {
        observable
            .withUnretained(self)
            .subscribe(onNext: { owner, _ in
                guard let subview = owner.subview else { return }
                print("My subview: \(subview)")
            })
            .disposed(by: disposeBag)
    }

}
```

to this:

```
class MyView: UIView {

    let subview: UIView? = UIView()

    let observable = UIButton().rx.tap
    let disposeBag = DisposeBag()

    func setup() {
        observable
            .withUnretained(self.subview)
            .subscribe(onNext: { subview, _ in
                print("My subview: \(subview)")
            })
            .disposed(by: disposeBag)
    }

}
```

### DIFF
--- a/RxCocoa/Traits/SharedSequence/SharedSequence+Operators.swift
+++ b/RxCocoa/Traits/SharedSequence/SharedSequence+Operators.swift
@@ -455,7 +455,7 @@ extension SharedSequenceConvertibleType {
      - returns: An observable sequence that contains the result of `resultSelector` being called with an unretained reference on `obj` and the values of the original sequence.
      */
     public func withUnretained<Object: AnyObject, Out>(
-        _ obj: Object,
+        _ obj: Object?,
         resultSelector: @escaping (Object, Element) -> Out
     ) -> SharedSequence<SharingStrategy, Out> {
         SharedSequence(self.asObservable().withUnretained(obj, resultSelector: resultSelector))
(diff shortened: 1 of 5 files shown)


freak4pc:
Hey @lammertw. This isn't really what this operator about - you're talking about unwrapping, not retaining. 
Maybe using `compactMap { [weak subview] _ in subview }` would be a better alternative in this case. 

Also, this is a breaking change so even if it made sense to pull it in, it would have to wait for the next major release of RxSwift.

Thanks!

lammertw:
@freak4pc thanks for the feedback.

To me it seems that there aren't real downsides of making it optional. Also what exactly is the breaking change? Non optionals still work, so it should be completely backwards compatible.

freak4pc:
It's a breaking change as you're changing the interface (e.g. the method signature), which breaks binary stability. 

Regardless, the fact something "doesn't hurt" doesn't make it correct. You could make `map` also work on Optionals, but we have a separate `compactMap` to do this.

In the same way, I don't see this specific implementation as correct for this use case, as the value is Optional, and might not even exist.

What you're describing isn't "getting a retained copy of an object", but "unwrapping a wrapped value safely", which doesn't have the same semantic meaning and feels like needlessly overloading this operator.

lammertw:
Yes I believe you’re right. And in this sample I guess self isn’t captured at all anyway.

PR #2289, opened 2021-02-08
fix typo : subscription to subscriptions

sosuiiii:


freak4pc:
Sounds like something in your branch isn't aligned to the main branch

freak4pc:
@sosuiiii ?

freak4pc:
Sorry just realized this isn't actually a typo. This is creating a single subscription, not many.

Issue #2204, opened 2020-07-14
.rx.observeWeakly fails for KVO-compliant key path OperationQueue.operationCount on macOS 10.15+


DivineDominion:
**Short description of the issue**:

When you use KVO to subscribe to `OperationQueue.operationCount` changes, this will fail on macOS 10.15:

    backgroundQueue.rx
            .observeWeakly(Int.self, "operationCount")
            .subscribe(onNext: {
                print("reached the subscription:", $0) // never happens
            })

It's probably related to the deprecation of the `operationCount` property. The Objective-C header points to `progress.completedUnitCount` which was introduces in 10.15.

Manual KVO and KVC work fine, though. The `class_getProperty` check fails even though there's a KVO-compliant path:

https://github.com/ReactiveX/RxSwift/blob/6b2a406b928cc7970874dcaed0ab18e7265e41ef/RxCocoa/Foundation/NSObject%2BRx.swift#L466-L468

**Expected outcome**:

A `.next` event would go through.

**What actually happens**:

The change event doesn't reach the Observer and an unhandled error is logged instead:

> 2020-07-14 17:51:55.523075+0200 RxTestApp[64237:1304657] <RxTestApp.AppDelegate: 0x6000002408c0>: An -observeValueForKeyPath:ofObject:change:context: message was received but not handled.

**Self contained code example that reproduces the issue**:

See attached Xcode project:  
**[Archive.zip](https://github.com/ReactiveX/RxSwift/files/4919849/Archive.zip)**

```swift
import Cocoa
import RxSwift
import RxCocoa

@NSApplicationMain
class AppDelegate: NSObject, NSApplicationDelegate {
    @IBOutlet var window: NSWindow!

    let disposeBag = DisposeBag()
    let backgroundQueue = OperationQueue()

    func applicationDidFinishLaunching(_ aNotification: Notification) {
        backgroundQueue.addObserver(self, forKeyPath: "operationCount", options: .new, context: nil)
        backgroundQueue.rx
            .observeWeakly(Int.self, "operationCount")
            .subscribe(onNext: {
                print("reached the subscription:", $0) // never happens
            })
            .disposed(by: disposeBag)
        backgroundQueue.addOperation { print("operation 1") }
    }
}
```

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

Version 5.1.1

**Platform/Environment**

  - [ ] iOS
  - [x] macOS
  - [ ] tvOS
  - [ ] watchOS
[...]

PR #2294, opened 2021-03-01
Attempt to KVO key path without declared property

DivineDominion:
## Problem 

`observeWeakly` throws an error when you try to use it for `OperationQueue.operationCount` because it's not a formally declared property (anymore?). The key path "operationCount" is perfectly key-value observable as of the time of me writing this, though. 

This PR is an attempt to make KVO-based observing work for these kinds of properties, too.

## Approach 

The current approach is to look if the key path corresponds to a declared `@property` on the NSObject target. This is an okay heuristic, but not 100% accurate, because you can declare a formal property and then override the synthesized setter and never call the KVO will change/did change methods.

At the moment, this fails for e.g. `OperationQueue.operationCount`. That's a deprecated property, but it's still KVO'able. We discussed the issue in #2204 and how and why it fails. It's not a declared formal property, so the existing heuristic cannot work and an error is raised.

I added another fallback branch to the observable check: Checking for the selector via `responds(to: Selector(propertyName))` to determine if the object might respond to the key path at all. This could produce false positives, but I'm not sure and would be happy for ideas to make this break.

## Caveats 

- Please note that `responds(to: Selector(propertyName))` might indicate the property is **key-value codable (KVC),** but not key-value observable. In that regard, it's not worse than the current approach in principle, although the likelihood of a formal property to not be KVO'able is higher than "anything with a selector of this name". 
- It's a fallback when the first test fails, but it might break down the line in other ways. For example, if the path is not KVO'able but the selector check passes, you can write your Rx code but never receive an event and wouldn't know why. -- I'm not sure if this change in behavior is worse than the current approach of throwing an error if the key path is not a formal property.

## Alternatives considered 

- I looked at other options, i.e. since `class_getProperty` fails when the path is not a declared property, would `object_getClass(target)?.keyPathsForValuesAffectingValue(forKey: propertyName)` return something? Apparently, no, not in the case of `OperationQueue.operationCount`. It might work for custom objects in the user's app, though, where this static method is used to declare KVO conformance. Could still be a worthwhile addition in another PR if anyone wants to look into this.

- You can verify that a key path is Key-Value _Codable_ (albeit not observable) much better with `value(forKey:)` and `setValue(_:forKey:)` instead of `responds(to:)` because they operate on stringified key paths. Problem is that there's no sane way to use this for checking for availability of the property: accessing a property via key path that doesn't exist raises a runtime exception. That's arguably worse than throwing.

### DIFF
--- a/RxCocoa/Foundation/NSObject+Rx.swift
+++ b/RxCocoa/Foundation/NSObject+Rx.swift
@@ -488,14 +488,19 @@ private extension KeyValueObservingOptions {
         let propertyName = keyPathSections[0]
         let remainingPaths = Array(keyPathSections[1..<keyPathSections.count])
 
-        let property = class_getProperty(object_getClass(target), propertyName)
-        if property == nil {
+        // should dealloc hook be in place if week property, or just create strong reference because it doesn't matter
+        let propertyIsWeak: Bool
+        if let property = class_getProperty(object_getClass(target), propertyName) {
+            // Key path corresponds to a @property declaration that, unless setter is overridden, are KVO'able
+            let propertyAttributes = property_getAttributes(property)
+            propertyIsWeak = isWeakProperty(propertyAttributes.map(String.init) ?? "")
+        } else if target.responds(to: Selector(propertyName)) {
+            // Key path exist and might be KVC'able, but unsure if it's also observable for changes
+            propertyIsWeak = false
+        } else {
+            // Not found at all
             return Observable.error(RxCocoaError.invalidPropertyName(object: target, propertyName: propertyName))
         }
-        let propertyAttributes = property_getAttributes(property!)
-
-        // should dealloc hook be in place if week property, or just create strong reference because it doesn't matter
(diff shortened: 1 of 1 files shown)


freak4pc:
Thanks for the thorough explanation! I just want to make sure I understand the core issue - 
You're trying to KVO-observe `operationCount` - which is deprecated in itself. Did you notice this for any other properties or these changes are only to support this case? 

Also, does this work with the "new" keypath/closure-based `observe(_:options:)` ?

DivineDominion:
I dont' use KVO that much, so this was a rabbit hole when the app went crashing -- I don't know whereever else this might be relevant. I'm fine with closing this and keeping a record for posterity in case someone else stumbles upon similar problems.

The new key-path based methods didn't help; I believe I left details in the issue this was based on.

freak4pc:
I think it would make sense to close this for now, but if this pops up again for several other users, we can rethink If this is worth fixing in some way. Thanks for the thorough investigation!

PR #2300, opened 2021-03-19
Additional UIAccessbilityCustomActions bindings for triggering events.

devandanger:


devandanger:
Still trying to make sense some warnings I'm getting on the ```scripts/all-tests.sh```.  I'm not sure what going on there but I get the following output.

```
JIT session error: Symbols not found: .....
```

devandanger:
Well was fairly certain this was good, but looks like a runaway CI machine.  Hopefully it eventually times out.

freak4pc:
Hey @devandanger - 
I'm not too familiar with this API. Is it hugely popular and frequent for Cocoa developers? 
I'm wondering if it might be a bit too niche of a use case to include in the core of RxSwift.

devandanger:
If you have more complicated UI controls these customized accessibility actions are used to bridge that complexity to say voice over users.

Because of this I thought it would be a good fit in RxCocoa.

I attempted to implement outside of RxCocoa, but there was a lot of convenience methods I was missing.

RE: Niche, I don't think so. I mean making complex controls more accessible, might be niche but shouldn't be IMO.

freak4pc:
I think it's a bit niche, but probably also small enough that I'm OK with pulling it in. 
Would you mind opening up a new PR with `main` as the base branch? We removed the `develop` branch which is why the PR was automatically closed.

PR #2302, opened 2021-03-26
Improve withLatestFrom documentation

pimnijman:
The inline documentation for `withLatestFrom` could use some attention.
Especially `withLatestFrom(_:)` (which doesn't accept a result selector function) seemed incorrect and got me confused.
I hope this PR can prevent some confusion in the future.

### DIFF
--- a/RxSwift/Observables/WithLatestFrom.swift
+++ b/RxSwift/Observables/WithLatestFrom.swift
@@ -9,13 +9,13 @@
 extension ObservableType {
 
     /**
-     Merges two observable sequences into one observable sequence by combining each element from self with the latest element from the second source, if any.
+     Merges two observable sequences into one observable sequence by combining each element from self with the latest element from the second source.
 
      - seealso: [combineLatest operator on reactivex.io](http://reactivex.io/documentation/operators/combinelatest.html)
 
      - parameter second: Second observable source.
-     - parameter resultSelector: Function to invoke for each element from the self combined with the latest element from the second source, if any.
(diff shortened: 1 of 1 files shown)


RxSwift/Observables/WithLatestFrom.swift
```diff
@@ -9,13 +9,13 @@
 extension ObservableType {
 
     /**
-     Merges two observable sequences into one observable sequence by combining each element from self with the latest element from the second source, if any.
```
freak4pc, line 12: can you please revert the `, if any` portions throughout?
Aside for that, looks good. Thank you !

freak4pc (changes requested):

RxSwift/Observables/WithLatestFrom.swift
pimnijman: @freak4pc  Could you please explain what the `if any` portion is supposed to communicate?

`Merges two observable sequences into one observable sequence by combining each element from self with the latest element from the second source, if any.`

I interpret this to mean that the latest element from the second source is optional.
In other words: If the second source has no latest element, the result observable would still fire an event.
However this is not true. If the second source has no latest element, the result observable will not fire an event.

freak4pc: If you want to be more verbose you should say `[...] from the second source. if the second source has no latest element, no combined value will be emitted by this operator`.

freak4pc:
Closed due to inactivity

PR #2297, opened 2021-03-12
Fix build failure on Xcode 12.5 beta 3

vincentisambart:
When I tried building our app with Xcode 12.5 beta 3, it failed building RxCocoa.
Here are fixes for RxSwift and AllTests to build (and run) successfully on Xcode 12.5 (I also made sure I did not break Xcode 12.4).

PR #2305, opened 2021-04-08
fix: adjust method signatures for Xcode 12.5 compatibility

thedavidharris:
Xcode gives errors now around these function declarations starting in Xcode 12.5, tested in b3 and this properly adds the declarations as well as gets rid of what seems to be a now-unneeded ternary if. 

This is needed for the new Xcode versions, as well as anything depending on those.

### DIFF
--- a/RxCocoa/Runtime/_RXObjCRuntime.m
+++ b/RxCocoa/Runtime/_RXObjCRuntime.m
@@ -1022,7 +1022,7 @@ replacementImplementationGenerator:(IMP (^)(IMP originalImplementation))replacem
 
 #if TRACE_RESOURCES
 
-NSInteger RX_number_of_dynamic_subclasses() {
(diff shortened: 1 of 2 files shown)


freak4pc:
I just love Xcode's definition of "non breaking changes" 🤦 

Thank you!

freak4pc (approved):

vincentisambart:
That was already fixed in #2297 🤔
[CONTRIBUTING.md](https://github.com/ReactiveX/RxSwift/blob/main/CONTRIBUTING.md) says that PR should target the `develop` branch, but is that still the case?

freak4pc:
We don't work on develop anymore, I should fix that. Sorry for the confusion, didn't realize it's a dupe 🤭

vincentisambart:
No problem, fixing CONTRIBUTING.md would be a great idea 😄

PR #2308, opened 2021-04-22
fix: restore Xcode 12.4 functionality due to resultBuilder rename

thedavidharris:
Annoyingly there's no way to fix this warning, but `@resultBuilder` doesn't exist for Swift versions prior to 5.4.


freak4pc (approved):
Hey, thanks! 
Just FYI we do not guarantee main to be stable - it's a playground for development, so I don't necessarily recommend linking directly to it (perhaps linking to a stable commit makes sense if you have that use case)

thedavidharris:
@freak4pc agreed! Just something I noticed while prepping some of our codebases for 12.5 things

PR #2309, opened 2021-04-24
fix:Pass as a parenthesized argument to remove confusion with trailing closure

aomathwift:
Remove the warning `trailing closure in this context is confusable with the body of the statement; pass as a parenthesized argument to silence this warning` that is issued when `package-spm` scripts is executed.


freak4pc (approved):
Thanks !

Jarenbeaver:
https://github.com/ReactiveX/RxSwift/commit/7219024cd02502e9faa6528a655709257f28a105

PR #2320, opened 2021-05-10
The develop branch should not be used anymore

vincentisambart:
According to https://github.com/ReactiveX/RxSwift/pull/2305#issuecomment-825380723, the development does not happen on the `develop` branch anymore so I updated CONTRIBUTING.md accordingly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ Please take the time to carefully read the following guide. These rules help mak
 
 ### General rules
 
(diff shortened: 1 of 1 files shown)


freak4pc (approved):
Thank you! 
In kinda wondering if it's worth just removing that paragraph all together since it's the default anyways. WDYT?

vincentisambart:
I think removing the part mentioning a branch is fine, but in that case you should probably delete the existing `develop` branch (or rename it to some name making it clear it's not used anymore).

freak4pc (approved):

PR #2321, opened 2021-05-10
fix: missing UIApplication+Rx.swift reference

ElonPark:
## issue
- UIApplication+Rx.swift, UIApplication+RxTests.swift is missing reference in Rx.xcodeproj
- If you install Rxcocoa using Carthage, you will get this error.

스크린샷 2021-05-10 오후 7 34 04


freak4pc (approved):
Thanks!

freak4pc (approved):
Thanks! Sorry for the delay looking into this.

PR #2326, opened 2021-05-31
fix Adding a Period to the End of a PrimitiveSequence's summary

gwonii:
It keeps bothering my eyes...


freak4pc (commented):
Sorry, but we usually don't have periods at line comments. Appreciate the thought, though!

gwonii:
Thanks for your reply.
I'll pay more attention in RxSwift!!

PR #2327, opened 2021-06-08
fix playground deprecated code

konabe:
this commits are to fix warnings by Xcode in playgrounds.


freak4pc (approved):
Good fixes, thanks!

PR #2332, opened 2021-06-26
fix GettingStarted.md deprecated code

wogus3602:
I found deprecated while studying.

This commits are to fix warnings by Xcode

I will also study by referring to the documentation. I'll let you know when I find it again.


freak4pc (approved):
Thanks a lot!

PR #2336, opened 2021-07-07
using OptionSet instead of  enum for DisposeState

Gnoah89:
in SingleAssignmentDisposable  class,  DisposeState.disposableSet and DisposeState.disposed can coexist，using optionSet  should be better

### DIFF
--- a/RxSwift/Disposables/SingleAssignmentDisposable.swift
+++ b/RxSwift/Disposables/SingleAssignmentDisposable.swift
@@ -13,9 +13,11 @@ If an underlying disposable resource has already been set, future attempts to se
 */
 public final class SingleAssignmentDisposable : DisposeBase, Cancelable {
 
-    private enum DisposeState: Int32 {
-        case disposed = 1
-        case disposableSet = 2
+    private struct DisposeState: OptionSet {
+        let rawValue: Int32
+        
+        static let disposed = DisposeState(rawValue: 1 << 0)
+        static let disposableSet = DisposeState(rawValue: 1 << 1)
     }
 
     // state


freak4pc:
Thanks foe the contribution ! 
I'm wondering what makes this better ? We are still representing the values as Uint32 and the callsite hasn't changed...

I like OptionSet but doesn't seem necessarily different for this private callsite. What do you think?

Gnoah89:
@freak4pc  Thanks for your replay,  

In my understanding, enum only one value can be represented at the same time，the usual usage is to use Switch-case match one of the values

but in code `private let state = AtomicInt(0)`  , state seems like a collection and use the bitset, it is more like OptionSet 

```
public func setDisposable(_ disposable: Disposable) {
        self.disposable = disposable

        let previousState = fetchOr(self.state, DisposeState.disposableSet.rawValue)

        if (previousState & DisposeState.disposableSet.rawValue) != 0 {  // like bitset operation 
            rxFatalError("oldState.disposable != nil")
        }

        if (previousState & DisposeState.disposed.rawValue) != 0 {  // like bitset operation 
            disposable.dispose()
            self.disposable = nil
        }
 }

```

from Apple Document:  https://developer.apple.com/documentation/swift/optionset

> You use the OptionSet protocol to represent bitset types, where individual bits represent members of a set

I agree with your opinion `We are still representing the values as Uint32 and the callsite hasn't changed`, but It's more like a grammatical ambiguity。

freak4pc (approved):
Sure, I don't super mind, but I think as an internal dependency the "grammatical clarity" of it doesn't make a prominent difference.

Regardless this is a welcome change and I appreciate your help!

Issue #2316, opened 2021-05-03
Changing system time affects Throttle/Delay etc.


leonylcatwork:
*If you still want to report issue, please delete above statements before submitting an issue.*

**Short description of the issue**:

If we set throttle to a control, say, a button, we could prevent button being tapped multiple times when it is processing task:
```
button.rx.tap.throttle(.milliseconds(500), latest: false, scheduler: MainScheduler.instance)
```

This works fine normally, except when you change system time backwards (by, say, one day) after first tap when `lastSentTime` was recorded.
When ThrottleSink is calculating reduceWithSpanBetween() on the second tap, its remainder time will be a great positive number making it not firing event, because `laterDate` is now in fact earlier than `earlierDate` (since we changed system time to an earlier date). I assume `delay` operator would be similar since it calls reduceWithSpanBetween() in a similar fashion.
 
line 43 Date+Dispatch.swift

```
internal func reduceWithSpanBetween(earlierDate: Date, laterDate: Date) -> DispatchTimeInterval {
    return self.map { value, factor in
        let interval = laterDate.timeIntervalSince(earlierDate)
        let remainder = Double(value) - interval * factor
        guard remainder > 0 else { return 0 }
        return Int(remainder.rounded(.toNearestOrAwayFromZero))
    }
}
```

**Expected outcome**:

I suppose when using throttle/debounce/delay operators we focus on relative time changing rather than absolute time changing.
What if we take absolute value of `interval` when calculating `remainder`? This way we only care about the magnitude and not the direction of the difference of two Dates. I know this is not perfect (a range of value in the opposite direction still prevent it from firing), but it maybe the simplest  way to deal with above problem.

BTW, is there better way to prevent a control from firing multiple times if the user tap too fast?

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

6.0.0

**Platform/Environment**

iOS

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

easy, 100% repro

**Xcode version**: 

```
12.5
```

**Installation method**:
CocoaPods

**I have multiple versions of Xcode installed**:
no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
I have a significant code base

freak4pc:
I don't find it to be a reasonable use case to support (user tapping a button, going over to change their time, coming back to tap again). 

Is this a real use case or more theoretical ? 

[...]

PR #2341, opened 2021-07-17
fix: Changing system time affects Throttle/Delay. (#2316)

zyn381285190:
As title. And I'd like to share my use case. My app is used on the iPad in the meeting room to initiate and operate online meetings. The app will basically remain running and will not exit. Sometimes users will return to the background to modify the system time zone during the opening of the app (in order to keep the time consistent with overseas customers). If the time is adjusted forward, the button event using the throttle operator will become invalid.

### DIFF
--- a/RxSwift/Date+Dispatch.swift
+++ b/RxSwift/Date+Dispatch.swift
@@ -32,6 +32,17 @@ extension DispatchTimeInterval {
         }
     }
     
+    func toNanoseconds() -> Int {
+        switch self {
+        case .nanoseconds(let value): return value
+        case .microseconds(let value): return value * 1_000
+        case .milliseconds(let value): return value * 1_000_000
+        case .seconds(let value): return value * 1_000_000_000
+        case .never: fatalError()
+        @unknown default: fatalError()
+        }
+    }
+    
     var isNow: Bool {
         switch self {
         case .nanoseconds(let value), .microseconds(let value), .milliseconds(let value), .seconds(let value): return value == 0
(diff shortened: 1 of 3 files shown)


zyn381285190:
> Can you add some tests to confirm your fix fixes the issue you had, if possible?

I have tested it manually, because this case needs to modify the system time, it seems that there is no way to build an automated test case in the App

zyn381285190:
> Can you add some tests to confirm your fix fixes the issue you had, if possible?

I found a better way to fix the problem, which is using mach_continuous_time to get the CPU launch time. I have already finished coding and manual testing, but it cannot pass the auto test cases because the test cases are currently creating virtual time spans to test. The test cases require using Date to compare time span, but my implementation uses CPU time. I changed the cases to use DispatchQueue.asyncAfter like following demo, is that OK? If yes, I'll modify all the test cases.
`    func asyncAfter(timeMS: Int, block: @escaping (() -> Void)) {
        DispatchQueue.main.asyncAfter(deadline: .now() + Double(timeMS) / 1000 , execute: block)
    }
    
    func test_ThrottleTimeSpan_NotLatest_Completed() {
        let subject: PublishSubject<Int> = PublishSubject()
        var results: [Int] = []
        _ = subject
            .throttle(.milliseconds(20), latest: false, scheduler: MainScheduler.instance)
            .subscribe(onNext: { (value) in
            results.append(value)
        })
        
        asyncAfter(timeMS: 15) { subject.onNext(1) }
        asyncAfter(timeMS: 21) { subject.onNext(2) }
        asyncAfter(timeMS: 25) { subject.onNext(3) }
        asyncAfter(timeMS: 31) { subject.onNext(4) }
        asyncAfter(timeMS: 35) { subject.onNext(5) }
        asyncAfter(timeMS: 41) { subject.onNext(6) }
        asyncAfter(timeMS: 45) { subject.onNext(7) }
        asyncAfter(timeMS: 46) { subject.onCompleted() }
        
        asyncAfter(timeMS: 100) {
            let correct = [1, 6]
            XCTAssertEqual(results, correct)
        }
    }`

freak4pc:
Hey @zyn381285190,
I think the whole point of our TestScheduler is to use virtual time and not "real" time but perhaps in this specific case, we can rethink. Can you share the implementation you're using?

zyn381285190:
> Hey @zyn381285190,
> I think the whole point of our TestScheduler is to use virtual time and not "real" time but perhaps in this specific case, we can rethink. Can you share the implementation you're using?

I have push my implementation, please have a check!

freak4pc:
Hey, I thought about this a lot over the past couple of weeks.

I have to say I'm feeling very uncomfortable changing the entire timing mechanism for schedulers to use something like this.

This might have a more expected outcome when going to background and returning but there's no telling how much breakage it would cause existing consumers, and with such a drastic change it would affect every consumer of RxSwift in existence which is just too much risk for relatively little value. 

I know you've put work into this but I think I'll have to skip this change for the sake of our existing consumers. How you can understand, and thank you for your help!

Issue #2347, opened 2021-08-17
Main actor-isolated class 'RxNavigationControllerDelegateProxy' has different actor isolation


nalexn:
**Short description of the issue**:

I'm on v6.2.0 of RxSwift/RxCocoa, the code base compiles ok in Xcode 12.5.1

However, in Xcode 13.0 beta 5 there is the following compilation error in RxCocoa:

`Main actor-isolated class 'RxNavigationControllerDelegateProxy' has different actor isolation from nonisolated superclass 'DelegateProxy'`

**Expected outcome**:

No compilation error

**What actually happens**:

A compilation error

**Self contained code example that reproduces the issue**:

Not applicable

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

  _v6.2.0_

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  13.0 beta 5
```

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [x] yes (which ones) 13.0 beta 5 and 12.5.1
  - [ ] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [ ] I have a small code base
  - [x] I have a significant code base

markst:
Also applies to:
```swift
Main actor-isolated class 'RxScrollViewDelegateProxy' has different actor isolation from nonisolated superclass 'DelegateProxy'
```

PR #2348, opened 2021-08-19
Opt-out from RxNavigationControllerDelegateProxy being a @MainActor

marcelofabri:
Apparently, moving the `UINavigationControllerDelegate` conformance to an extension makes it so `RxNavigationControllerDelegateProxy` is not recognized as `@MainActor`, avoiding compilation errors when using RxCocoa.

Fixes https://github.com/ReactiveX/RxSwift/issues/2347

### DIFF
--- a/RxCocoa/iOS/Proxies/RxCollectionViewDataSourcePrefetchingProxy.swift
+++ b/RxCocoa/iOS/Proxies/RxCollectionViewDataSourcePrefetchingProxy.swift
@@ -31,8 +31,7 @@ private final class CollectionViewPrefetchDataSourceNotSet
 @available(iOS 10.0, tvOS 10.0, *)
 open class RxCollectionViewDataSourcePrefetchingProxy
     : DelegateProxy<UICollectionView, UICollectionViewDataSourcePrefetching>
-    , DelegateProxyType
-    , UICollectionViewDataSourcePrefetching {
+    , DelegateProxyType {
 
     /// Typed parent object.
     public weak private(set) var collectionView: UICollectionView?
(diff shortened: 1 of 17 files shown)


freak4pc:
I'm wondering if this is a patch or an actual fix. _Should this_ be a @MainActor eventually?

marcelofabri:
> I'm wondering if this is a patch or an actual fix. _Should this_ be a @mainactor eventually?

I think this would require `DelegateProxy` to be a `@MainActor` as well. Besides that, we probably want to think about actors adoption as a whole, since it'll bring downstream consequences for projects using RxSwift. I'm afraid I'm not super into the loop about the new Swift 5.5 concurrency to help here - I just want to be able to use RxSwift with Xcode 13 😅

freak4pc:
Hey @marcelofabri, overall no issue for me with merging this but can you provide a dummy project showing this is actually a problem? I just made a dummy project on the latest Xcode 13 beta and it's building without this error. Is it related to some experimental flag related to concurrency?

marcelofabri:
> Hey @marcelofabri, overall no issue for me with merging this but can you provide a dummy project showing this is actually a problem? I just made a dummy project on the latest Xcode 13 beta and it's building without this error. Is it related to some experimental flag related to concurrency? 

I don't think it's related to any experimental flags - I definitely didn't enable any. I'll try to create a sample project until Monday.

marcelofabri:
I was able to repro it by opening `Rx.xcworkspace` from `main` with Xcode 13.0 beta 5 and compiling the `RxExample-iOS` scheme

freak4pc:
Hey @marcelofabri - Thanks! 
Can you take care of other delegate proxies in this PR so we have a full solution? Seems like some others were listed in #2347 

Appreciate your help:)

marcelofabri:
> Hey @marcelofabri - Thanks! 
> 
> Can you take care of other delegate proxies in this PR so we have a full solution? Seems like some others were listed in #2347 
> 
> 
> 
> Appreciate your help:) 

I won't be able to get to it until next Tuesday, so feel free to push to this PR 😅

markst:
thanks @marcelofabri !

freak4pc (approved):
This looks great! Thank you so much 🔥

nalexn:
Thanks for addressing the issue! Could this be released asap, please? September 14 is just around the corner (with a likely iOS 15 release), but I'm still unable to compile the project in Xcode 13 to test on iOS 15

freak4pc:
@nalexn I tend to wait for the GM before releasing a version since we don't know what surprises Apple will have. For example, right now there is a somewhat critical issue with swiftinterface(s) in Combine that wasn't fixed in b5, so I'm assuming there will surely be a b6. 

Pointing to `main` sounds like a good idea for the time being.

tylerjames:
It looks like this issue is already fixed in the GM? I wasn't able to compile with Beta 5 but just tried the GM and it works fine

vburojevic:
Yes, it seems to work on Xcode 13 RC for me as well.

PR #2349, opened 2021-08-21
Check that aSelector is not nil before passing it on...

danielt1263:
Both `voidDelegateMethodsContain(_ :)` and `hasObservers(selector:)` require that `aSelector` is not nil. The `responds(to:)` method assumes it isn't nil but doesn't check. Some older objective-c libraries (eg, FSCalendar) check to see if the delegate responds to a `nil` selector.

This change adds code to return false if the selector is nil.

### DIFF
--- a/RxCocoa/Common/DelegateProxy.swift
+++ b/RxCocoa/Common/DelegateProxy.swift
@@ -225,9 +225,10 @@
         }
 
         override open func responds(to aSelector: Selector!) -> Bool {
-            return aSelector != nil && (super.responds(to: aSelector)
+            guard let aSelector = aSelector else { return false }
+            return super.responds(to: aSelector)
                 || (self._forwardToDelegate?.responds(to: aSelector) ?? false)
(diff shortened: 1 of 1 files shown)


danielt1263:
For more information, check out this StackOverflow question: https://stackoverflow.com/questions/68872523/error-occurred-when-i-using-rxswifts-delegateproxy-with-fscalendar/68873186#68873186

faimin:
`guard` maybe a better way

```swift
guard aSelector != nil else {
    return false
}

return super.responds(to: aSelector)
                || (self._forwardToDelegate?.responds(to: aSelector) ?? false)
                || (self.voidDelegateMethodsContain(aSelector) && self.hasObservers(selector: aSelector))
```

freak4pc:
Hey @danielt1263, sorry for the delay in replying :) 
I agree with @faimin, can you switch to an early return with `guard` for that case? Reads much better in my opinion. 

Thanks!

danielt1263:
Done.

freak4pc (approved):
Thanks you!

PR #2350, opened 2021-08-21
support setting UIButton attribute for muti-states

tanpengsccd:
Now we Can set  attribute more convenient.
```swift
...
    _ = Observable.just("normal").subscribe(button.rx.title(for:[ .normal,.disable]))
    //or
    _ = Observable.just("normal").subscribe(button.rx.title(for: .normal,.disable))
    //or
    _ = Observable.just("normal").subscribe(button.rx.title(for: UIControl.State.allCases))
...
```

### DIFF
--- a/RxCocoa/iOS/UIButton+Rx.swift
+++ b/RxCocoa/iOS/UIButton+Rx.swift
@@ -42,25 +42,52 @@ extension Reactive where Base: UIButton {
 import RxSwift
 import UIKit
 
+extension UIControl.State: CaseIterable{
+    public static var allCases: [UIControl.State] {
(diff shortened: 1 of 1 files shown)


freak4pc (commented):
Hey there. I appreciate this change and the work you've done but:

1. This is breaking the public, which we can't do for both ABI stability and generally semantic versioning

2. I'm not sure this is a broad enough use case (it was never asked until today) to include in the framework

freak4pc:
Closing, I'm not convinced this is a fits in RxCocoa core itself. Perhaps RxSwiftExt is a better fit :) 
Thank you again for your time and suggestion.

PR #2356, opened 2021-09-02
Correct misspelled words

dungntm58:
I found there were some misspelled words. I hope it will be helpful


freak4pc (approved):
Great catch, thanks a lot!

PR #2357, opened 2021-09-03
Update build phases about SwiftLint

kimxwan0319:
The Swift Lint part of the build phases, the code that continues to be repeated is made into a shell script file and modularized.


freak4pc (approved):
Small but nice change, thank you!

PR #2362, opened 2021-09-14
Fix parameter names in method Documentation Comments

jolyot:
This PR is to fix parameter names in method Documentation Comments.

The following is the revised content:

- Fix difference between actual parameter names and those in comments
- Replace argument labels with actual parameter names
- Add missing parameter names.


freak4pc (approved):
That's a great clean up! Thanks a lot :)

PR #2367, opened 2021-09-22
Support Concurrency in RxSwift (async/await)

freak4pc:
This PR introduces support for using async/await on every reactive stream available in RxSwift - `Observable`, `Infallible`, `Driver`, `Signal`, `Single`, `Maybe` and `Completable`. 

A simple use case for a stream that may emit multiple values could look as follows:

```swift
do {
  for try await value in observable.values {
      print("Observable emitted value: \(value)")
  }
} catch {
    print("Observable threw error: \(error)")
}
```

For `Infallible`, `Driver` and `Signal` you will not need to try/catch since they are guaranteed to never emit a failure.

For variations that emit a single value (i.e. Primitive Sequences), you can simply do:

```swift
do {
    let value = try await single.value
} catch {
    print("Single made a booboo: \(error)")
}
```


RxCocoa/Traits/SharedSequence/SharedSequence+Concurrency.swift
```diff
@@ -0,0 +1,38 @@
...
+    /// asynchronously via Swift's concurrency features (`async/await`)
+    ///
+    /// A sample usage would look like so:
+    ///
+    /// ```swift
+    /// for await value in driver.values {
+    ///     // Handle emitted values
+    /// }
+    /// ```
+    @MainActor var values: AsyncStream<Element> {
```
yakushevichsv, line 25: Why values should be accessed on main thread ?

freak4pc: `Driver` and `Signal` emit on the main thread:
https://github.com/ReactiveX/RxSwift/blob/main/RxCocoa/Traits/Driver/Driver.swift#L15

RxSwift/Observable+Concurrency.swift
```diff
@@ -0,0 +1,39 @@
...
+    var values: AsyncThrowingStream<Element, Error> {
+        AsyncThrowingStream<Element, Error> { continuation in
+            _ = asObservable().subscribe(
+                onNext: { value in continuation.yield(value) },
+                onError: { error in continuation.finish(throwing: error) },
+                onCompleted: { continuation.finish() },
+                onDisposed: { continuation.onTermination?(.cancelled) }
+            )
+        }
+    }
```
inamiy, line 37: Returned `Disposable` is not used, so cancellation is probably not working?

inamiy: Maybe this code might be useful.
https://twitter.com/inamiy/status/1439431498457759745

freak4pc: Generally speaking `Disposable`s are deallocated when leaving the closest scope so this will only be an issue when there's explicit cancellation of the task or its parent. Anyways, I don't think `withExtendedLifetime` will be needed :)

Great thoughts anyways, I'll add that. Thank you 🙏

inamiy: Yes, above twitter link is about Combine, not RxSwift 🚜
But the cancellation approach will still be similar where `onTermination` handler registration becomes necessary.

In RxSwift, it will require manual `disposable.dispose()` call rather than Combine's auto-cancellation on deinit.

For example:

```swift
continuation.onTermination = { @Sendable termination in
    switch termination {
    case .cancelled:
        disposable.dispose()
    case .finished:
        disposable.dispose() // probably won't need, but just in case...
    }
}
```

I need to elaborate more on `@Sendable`, but I guess this attribute will probably be needed too.

freak4pc: Yup, makes sense. I'll take a look, thank you!

RxSwift/Traits/PrimitiveSequence/PrimitiveSequence+Concurrency.swift
```diff
@@ -0,0 +1,102 @@
...
+    /// do {
+    ///     let value = try await single.value
+    /// } catch {
+    ///     // Handle error
+    /// }
+    /// ```
+    var value: Element {
+        get async throws {
+            try await withCheckedThrowingContinuation { continuation in
+                _ = self.subscribe(
```
mikchmie, line 29: Shouldn't subscriptions be disposed when tasks gets cancelled? So e.g. this property would look like something like this:
```
var value: Element {
    get async throws {
        var disposable: Disposable?

        return try await withTaskCancellationHandler(
            operation: {
                try await withCheckedThrowingContinuation { continuation in
                    disposable = self.subscribe(
                        onSuccess: { continuation.resume(returning: $0) },
                        onFailure: { continuation.resume(throwing: $0) }
                    )
                }
            },
            onCancel: { [disposable] in
                disposable?.dispose()
            }
        )
    }
}
```
Here's the best piece of documentation that I've found for this topic:
https://github.com/apple/swift-evolution/blob/main/proposals/0304-structured-concurrency.md#cancellation-handlers

inamiy: +1. Boxing might be needed though.
cf. https://gist.github.com/inamiy/0f11034f3dd658ad699a9ac5d2714b2d

mikchmie: @inamiy Huh, I thought I've found an easy workaround for the compiler error I was facing, but you're right - `disposable` will always be nil in `onCancel` and boxing should do the trick.

danielt1263: Use a single assignment disposable.

RxCocoa/Traits/SharedSequence/SharedSequence+Concurrency.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+//  Created by Shai Mishali on 22/09/2021.
+//  Copyright © 2021 Krunoslav Zaher. All rights reserved.
+//
+
+#if swift(>=5.5) && canImport(_Concurrency) && !os(Linux)
+import Foundation
+
+// MARK: - Shared Sequence
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
```
mikchmie, line 13: Could these new APIs be made available on older OSes too, as Xcode 13.2 brought Concurrency support for them? ([release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-13_2-release-notes#Swift))
`@available(macOS 10.15, iOS 13, tvOS 13, watchOS 6, *)`

That would also require a change in the `#if` statement:
`#if compiler(>=5.5.2) && canImport(_Concurrency)`

freak4pc: Yup:) it's the next commit on the list

freak4pc: See https://github.com/ReactiveX/RxSwift/pull/2367/commits/e498cbeb1239d4cba87680a4ef55912a06db36ce

freak4pc:
This looks good enough to merge. I'll cut 6.5 over the course of the next few days, if you have any thoughts or changes you'd like to see as part of this release, do let me know :)

RxSwift/Observable+Concurrency.swift
```diff
@@ -0,0 +1,67 @@
...
+                do {
+                    for try await value in self {
+                        observer.onNext(value)
+                    }
+                } catch {
+                    observer.onError(error)
+                }
+            }
+
+            observer.onCompleted()
```
mikchmie, line 62: Shouldn't `onCompleted` be called inside the `Task` and only in the `do` block, after all values get emitted?

freak4pc: Oy, it should be inside the the Task, but it doesn't matter if it's in the do or not (the catch would short-circuit it). Anyways, thanks for the catch :)

Tests/RxCocoaTests/SharedSequence+ConcurrencyTests.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+
+#if swift(>=5.5.2) && canImport(_Concurrency) && !os(Linux)
+import Dispatch
+import RxSwift
+import RxCocoa
+import XCTest
+import RxTest
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
```
mikchmie, line 16: I guess platform availibility should be updated in tests as well:
`@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)`

Jarenbeaver:
https://github.com/ReactiveX/RxSwift/commit/c03353aee4b8ba811dcb28468be90ef99cf7ef46

Tests/RxCocoaTests/SharedSequence+ConcurrencyTests.swift
freak4pc: No, I'm using `.values` there which is only available on iOS 15.

RxSwift/Observable+Concurrency.swift
mikchmie: > the catch would short-circuit it
It wouldn't as there's no `return` statement in `catch` block, so once all statement's in `catch` were executed, any code that's after the `do-catch` would be executed as well.

Tests/RxCocoaTests/SharedSequence+ConcurrencyTests.swift
mikchmie: Are you sure? As far as I can tell you've made it available in iOS 13 in c03353a

RxSwift/Observable+Concurrency.swift
```diff
@@ -0,0 +1,67 @@
...
+//  RxSwift
+//
+//  Created by Shai Mishali on 22/09/2021.
+//  Copyright © 2021 Krunoslav Zaher. All rights reserved.
+//
+
+#if swift(>=5.5) && canImport(_Concurrency)
+import Foundation
+
+@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
```
mikchmie, line 12: Is there anything stopping us from making it available in older systems? (same question applies to the `AsyncSequence` extension below)
`@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)`

freak4pc: I'll check a bit later, thanks !

Tests/RxCocoaTests/SharedSequence+ConcurrencyTests.swift
freak4pc: I'll check a bit later, thanks :)

RxSwift/Observable+Concurrency.swift
freak4pc: Should be good now, wasn't entirely focused this noon :) Thanks for the great review!

jpsim:
@freak4pc I see you bumped versions to 6.5 in several places (e.g. d632db1972ee292cfdaa0283f2bf51455292d1d8) but there's still no corresponding [git tag](https://github.com/ReactiveX/RxSwift/tags) or [GitHub Release](https://github.com/ReactiveX/RxSwift/releases). Sharing in case you forgot to do those manual steps. Thanks for getting this in!

freak4pc:
> I'll cut 6.5 over the course of the next few days

@jpsim ;-) 
Coming today or tomorrow. Thanks for your patience !

jpsim:
No rush.

PR #2370, opened 2021-10-05
chore: add dynamic support to libraries

barcister:
Add support for dynamic library options

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -50,6 +50,11 @@ let package = Package(
       .library(name: "RxRelay", targets: ["RxRelay"]),
       .library(name: "RxBlocking", targets: ["RxBlocking"]),
       .library(name: "RxTest", targets: ["RxTest"]),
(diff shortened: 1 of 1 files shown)


freak4pc:
Doesn't SPM pick this by itself? Seems strange to me, having this duplication. Is there any other large framework that does it this way?

gomfucius:
@freak4pc It does seem strange, however, the provider of the package must define this and not the user of the package. This is an example of a SP that does this: https://github.com/apollographql/apollo-ios/blob/main/Package.swift
I think I may need this commit to merge so please take a look again.

freak4pc:
> @freak4pc It does seem strange, however, the provider of the package must define this and not the user of the package. This is an example of a SP that does this: https://github.com/apollographql/apollo-ios/blob/main/Package.swift
> I think I may need this commit to merge so please take a look again.

Thanks, that's actually a good reference.

freak4pc (approved):

PR #2373, opened 2021-10-12
Replace .error with .failure in Single call

philosopherdog:


philosopherdog:
I'll consolidate all the changes in a single PR

PR #2382, opened 2021-12-09
iOS 15 navigation bar appearance fixes

yodagamaheshan:
Set the appearance settings for the navigation bar when the edge of scrollable content aligns with the edge of the navigation bar. Otherwise, it's black in iOS 15 for every type of navigation bar.


freak4pc (approved):
Thanks!

PR #2384, opened 2021-12-30
fix: type method from class to static in final class

yeonduing:
There are two type methods, class and static, and class type method is used to allow subclasses to override the superclass’s implementation of that method.
MainScheduler is a final class, so the object can't be subclassed.
I suggest using static rather than class.

### DIFF
--- a/RxSwift/Schedulers/MainScheduler.swift
+++ b/RxSwift/Schedulers/MainScheduler.swift
@@ -41,14 +41,14 @@ public final class MainScheduler : SerialDispatchQueueScheduler {
     public static let asyncInstance = SerialDispatchQueueScheduler(serialQueue: DispatchQueue.main)
 
     /// In case this method is called on a background thread it will throw an exception.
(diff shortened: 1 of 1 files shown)


Jarenbeaver (approved):

Jarenbeaver:
Approve

Jarenbeaver:
> Approve

Jarenbeaver:
Assignengin

Jarenbeaver:
Apro

Jarenbeaver:
> > Approve
> 
>

Jarenbeaver:
def fibonacci(n)
  return n if ( 0..1 ).include? n
  (fibonacci(n - 1) + fibonacci(n - 2)) #recursive calls
end

Jarenbeaver:
https://github.com/ReactiveX/RxSwift/pull/2384#issuecomment-1006992738

freak4pc (approved):
Hey @yeonduing - 
To my knowledge - given the class is `final`, the method can't be overridden even if it's `class` and not `static`.
It's still a good change, thank you!

PR #2407, opened 2022-03-18
add recorded operator 

O-O-wl:
added operators to make testing hot observables easier

ASIS
```swift
func testHotObservable() {
    // given
    let hotObservable = PublishSubject<Int>()
    
    // when
    hotObservable.onNext(1)
    
    // then
    let event = try! hotObservable.toBlocking(timeout: 1).first() // ❌ Sequence timeout.
    XCTAssertEqual(event, 1)
  }
```

TOBE
```swift
func testHotObservable() {
    // given
    let hotObservable = PublishSubject<Int>()
    let recorded = hotObservable.recorded()
    
    // when
    hotObservable.onNext(1)
    
    // then
    let event = try! recorded.toBlocking(timeout: 1).first() 
    XCTAssertEqual(event, 1) // ✅ 
  }  
```

### DIFF
--- a/RxBlocking/ObservableType+Record.swift
+++ b/RxBlocking/ObservableType+Record.swift
@@ -1,0 +1,21 @@
+//
+//  ObservableType+Record.swift
+//  
+//
+//  Created by Booung on 2022/03/19.
+//
+
+import RxSwift
+import Foundation
+
+extension ObservableType {
+    /// Converts an `Observable` that records last `N` events
+    ///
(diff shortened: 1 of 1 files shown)


freak4pc:
Hey, thank you - I appreciate the suggestion for the addition.
Looking at it - it seems like a relatively simple helper that might fit better in specific code bases that want this behavior. I don't feel it fits in the "core" of RxSwift. Perhaps something like RxTestExt or RxSwiftExt itself could be a good place for it.

Thank you, again!

PR #2411, opened 2022-03-30
Allow converting async block to Traits

pookjw:
This PR allows converting `async` block to RxSwift Traits - `Infailable`, `Single`, `Maybe`, `Completable`, `Signal`, `Driver`.

```swift
asSingle {
    await UIImage(named: "background")?.byPreparingThumbnail(ofSize: CGSize(width: 100, height: 100))
}
.compactMap { $0 }
.subscribe(on: MainScheduler.instance)
.asObservable()
.bind(to: imageView.rx.image)
```

### DIFF
--- a/RxSwift/Traits/Infallible/Infallible+Concurrency.swift
+++ b/RxSwift/Traits/Infallible/Infallible+Concurrency.swift
@@ -34,4 +34,19 @@ public extension InfallibleType {
         }
     }
 }
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+public func asInfailable<Element>(_ fn: @escaping () async -> Element) -> Infallible<Element> {
+    return .create { observer in
+        let task = Task {
+            let element = await fn()
+            observer(.next(element))
+            observer(.completed)
+        }
+        
+        return Disposables.create {
+            task.cancel()
+        }
+    }
+}
 #endif
--- a/RxSwift/Traits/Infallible/Infallible+Concurrency.swift
+++ b/RxSwift/Traits/Infallible/Infallible+Concurrency.swift
@@ -35,11 +35,17 @@ public extension InfallibleType {
     }
 }
 
+/**
+ Allows converting asynchronous block to `Infailable` trait.
+ 
+ - parameter block: An asynchronous block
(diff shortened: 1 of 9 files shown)


RxSwift/Traits/Infallible/Infallible+Concurrency.swift
```diff
@@ -27,11 +27,41 @@ public extension InfallibleType {
...
+            let task: Task<Void, Swift.Error>
+            
+            if detached {
+                task = Task.detached(priority: priority, operation: operation)
+            } else {
+                task = Task(priority: priority, operation: operation)
+            }
+            
+            return Disposables.create {
+                task.cancel()
```
rtharston, line 62: ~Good call to cancel the tasks, but the `operation` never checks if the task was cancelled. This will only do anything if the code passed in as `block`~

~Would it make sense to add `if Task.isCancelled { return }` after `let element = await block()` and not send anything to `observer`, as if the observable never fired anything?
Or would it make more sense to not send `.next(element)` but to still send `.completed`?~

I left that note to show my train of thought, but now I am thinking it _is_ best to leave this as is and not check `Task.isCancelled` since it is up to the user passing in Block to decide what to do when their Task is cancelled. They still may want to return a partial result perhaps. That should be up to them.

I was thinking there should also be a way to indicate they want to return nothing when their Task gets cancelled, but I think the catch block (in the other three, not Infallible) is probably sufficient for that. If they end user cares to know their Single/Maybe/Completable threw an error because the Task was cancelled and not some other error they can check the error type.

From the [documentation](https://developer.apple.com/documentation/swift/task):
"Tasks include a shared mechanism for indicating cancellation, but not a shared implementation for how to handle cancellation. Depending on the work you’re doing in the task, the correct way to stop that work varies. Likewise, it’s the responsibility of the code running as part of the task to check for cancellation whenever stopping is appropriate. In a long-task that includes multiple pieces, you might need to check for cancellation at several points, and handle cancellation differently at each point. If you only need to throw an error to stop the work, call the Task.checkCancellation() function to check for cancellation. **Other responses to cancellation include returning the work completed so far, returning an empty result, or returning nil.**"

rtharston (commented):
Some comments about Task

PR #2412, opened 2022-03-31
Fix: Class Unavailable on Main.storyboard

JacksonJang:
I found error that message is "Navigation Bar Appearances before iOS 13.0" when running RxExample-iOS project
Main.storyboard need to update for than iOS 13.0!

so I changed storyboard and PR

Have a good day!


freak4pc (approved):
Thank you!

PR #2414, opened 2022-04-12
Rename Cancelable -> Cancellable

kyungpyoda:
Why don't you change `Cancelable` to `Cancellable`?
I know `Cancelable` is not wrong spelling.
But in [RxJava](https://reactivex.io/RxJava/javadoc/index.html?io/reactivex/functions/Cancellable.html), [Combine](https://developer.apple.com/documentation/combine/cancellable) and some other frameworks, it is named `Cancellable`.
I couldn't find any frameworks using `Cancelable`.

I've changed tons of files related about `Cancelable`. If you find out some omissions, plz let me know.


freak4pc:
> This is a hard one... UIKit and SwiftUI both use the American spelling for all words across the board (Color vs Colour for example.) I find it rather odd that Combine doesn't use the American spelling...

Might have been a "hard one" if we just started planning RxSwift - but given its public interface, it's unfortunately not up for discussion to rename it. 

I agree with you that it's very strange (Apple's decision)

freak4pc:
I'm closing this one, the reasoning has been described in my initial post. 
Again, sorry for all the hard work you put - I suggest opening the issue with the suggestion _before_ making large code changes next time 🙏

danielt1263:
That said, I guess you could always `typealias Cancellable = Cancelable` somewhere in your code?

PR #2422, opened 2022-05-06
Change SingleEvent to Result<Element, Swift.Error>

Kareem-Gamal-Badr:
Change SingleEvent to Result<Element, Swift.Error> to match the changes in RxSwift v6


freak4pc (approved):
Thank you!

PR #2425, opened 2022-05-21
Relax constraint for convenient distinctUntilChanged operator on InfallibleType

asiliuk:
Please correct me if this is wrong to do. But at first sight looks like a typo, because all other `distinctUntilChanged()` are available on types that only conform to `Equatable`


freak4pc (approved):
That does seem like a mistake, indeed. Thank you!

Issue #2426, opened 2022-06-02
Value leaks its continuation


kjaklinovic:
**Short description of the issue**:

Example code will produce a console output: `SWIFT TASK CONTINUATION MISUSE: value leaked its continuation!`

**Expected outcome**:

  I would expect no console output indicating issues with continuation. 

**What actually happens**:

  If you look at the example you will notice that the error triggers the dispose block in `doSomething1` and the task is canceled. If you look at the code for the `value` computed property you can notice that `onCancel` will call `dispose` function which will dispose the subscription that never got success or error event. Because of that `resume` was never called for the `continuation` which results in the console output described above. 
  
  ```swift
    var value: Element {
        get async throws {
            let disposable = SingleAssignmentDisposable()
            return try await withTaskCancellationHandler(
                operation: {
                    try await withCheckedThrowingContinuation { continuation in
                        disposable.setDisposable(
                            self.subscribe(
                                onSuccess: { continuation.resume(returning: $0) },
                                onFailure: { continuation.resume(throwing: $0) }
                            )
                        )
                    }
                },
                onCancel: { [disposable] in
                    disposable.dispose()
                }
            )
        }
    }
  ```

**Self contained code example that reproduces the issue**:

```swift
_ = Observable<Int>.interval(
    .seconds(1),
    scheduler: MainScheduler.instance
)
    .enumerated()
    .do { if $0.index > 0 { throw SomeError.error } }
    .map { _ in () }
    .flatMap(doSomething1)
    .subscribe()

func doSomething1() -> Single<Void> {
    Single.create { single in
        let task = Task {
            do { single(.success(try await doSomething2())) }
            catch { single(.failure(error)) }
        }

        return Disposables.create { task.cancel() }
    }
}

func doSomething2() async throws {
    try await doSomething3().value
}

func doSomething3() -> Single<Void> {
    Single
        .just(())
        .delay(.seconds(5), scheduler: MainScheduler.instance)
}
```

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  13.4
```

**Installation method**:
  - [x] CocoaPods
[...]

PR #2427, opened 2022-06-02
Fix for value leaks its continuation

kjaklinovic:
This PR contains a fix for Swift task continuation misuse where value leaks its continuation. #2426

### DIFF
--- a/Tests/RxSwiftTests/PrimitiveSequence+ConcurrencyTests.swift
+++ b/Tests/RxSwiftTests/PrimitiveSequence+ConcurrencyTests.swift
@@ -50,6 +50,7 @@ extension PrimitiveSequenceConcurrencyTests {
                 try await single.value
                 XCTFail("Should not proceed beyond try")
             } catch {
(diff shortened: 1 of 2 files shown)


Tests/RxSwiftTests/PrimitiveSequence+ConcurrencyTests.swift
```diff
@@ -41,6 +41,37 @@ extension PrimitiveSequenceConcurrencyTests {
             XCTAssertTrue(true)
         }
     }
+
+    func testSingleThrowsCancellationErrorWhenParentTaskIsCancelledButNoEventIsProduced() async throws {
```
freak4pc, line 45: Lets shorten to something like this across the board

```suggestion
    func testSingleThrowsCancelleationWithoutEvents() async throws {
```

freak4pc (changes requested):
Great fix, thank you! Just a few tiny comments.

freak4pc (approved):
Looks great, thanks a lot!

PR #2439, opened 2022-08-11
Change the target of some operator extensions from Observable to ObservableConvertibleType

### LINKED ISSUE #2438: `Infallible` is missing a number of useful extensions

The new `Infallible` is useful when you know your observable won't have errors, but in some cases you have to resort to using `Observable` to do something `Infallible` should be able to do, or to simply remove some boilerplate.

**Expected outcome**:

Use on of the following methods on an Infallible:

* combineLatest without providing resultSelector
* combineLatest on a Collection/Sequence
* distinctUntilChanged on Equatable type (the Observable extension is where Element: Equatable, but for Infallible it is where Element: Comparable)
* .observe(on:)

**What actually happens**:

Compiler errors

```swift
// Error: "Missing argument for parameter 'resultSelector' in call"
Infallible.combineLatest(infallibleOne, infallibleTwo)

// Error: "Missing argument for parameter #2 in call"
// because the compiler thinks you are trying to use combineLatest(_:_:resultSelector)
Infallible.combineLatest(arrayOfBools.map { Infallible.just($0) }) 

// Error: "Referencing instance method 'distinctUntilChanged()' on 'InfallibleType' requires that 'ConversationAppActivity' conform to 'Comparable'"
someInfallible.distinctUntilChanged()

// Error: "Value of type 'Infallible<ConversationAppActivity>' has no member 'observe'"
someInfallible.observe(on: MainScheduler.asyncInstance)

```

6.5.0

**Platform/Environment**

  - [ ] iOS
  - [x] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
13.4.1
```

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

  - [x] yes (13.4.1, 14 beta)
  - [ ] no

  - [ ] just starting

TTOzzi:
Resolves #2438 

The issue is simply resolved when `Infallible` adopts `Observable Type`.
But protocol `ObservableType` that provides method `subscribe` with `onError` cannot be adopted because `Infallible` must be guaranteed not to fail.
Therefore, the useful extensions that `ObservableType` provides by default are not available.

I have confirmed that most of the `ObservableType` extensions work by calling `asObservable` internally.
This logic can work sufficiently as an `ObservableConvertibleType`, not necessarily an extension of the `ObservableType`.
I've converted the related operators mentioned in the issue first, and if this seems to be a good way, I'll try to convert other similar parts.

### DIFF
--- a/RxSwift/Observables/CombineLatest+Collection.swift
+++ b/RxSwift/Observables/CombineLatest+Collection.swift
@@ -6,7 +6,7 @@
 //  Copyright © 2015 Krunoslav Zaher. All rights reserved.
 //
 
-extension ObservableType {
+extension ObservableConvertibleType {
     /**
      Merges the specified observable sequences into one observable sequence by using the selector function whenever any of the observable sequences produces an element.
 
@@ -16,7 +16,7 @@ extension ObservableType {
      - returns: An observable sequence containing the result of combining elements of the sources using the specified result selector function.
      */
     public static func combineLatest<Collection: Swift.Collection>(_ collection: Collection, resultSelector: @escaping ([Collection.Element.Element]) throws -> Element) -> Observable<Element>
-        where Collection.Element: ObservableType {
+        where Collection.Element: ObservableConvertibleType {
         CombineLatestCollectionType(sources: collection, resultSelector: resultSelector)
     }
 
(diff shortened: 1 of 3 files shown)


TTOzzi:
There is a problem that the return value of the common implementation cannot be limited to `Infallible` 🤔

PR #2440, opened 2022-08-11
Modify the closure argument of subscribe to InfallibleEvent

TTOzzi:
`Infallible` ensures that it does not fail, but `Event` in the `subscribe` method have an error case.
This is not in line with `Infallible`'s implementation intentions.

Modified the closure argument of `subscribe` to `InfallibleEvent` and modified the associated logics as well.


TTOzzi:
I don't think the failed test cases are related to this changes 🤔

freak4pc (approved):
Nice change! Thanks :)

PR #2443, opened 2022-08-31
Reactive wrapper for `setTitleColor(_:for:)`

knottx:
Reactive wrapper for `setTitleColor(_:for:)`


freak4pc:
Hey @knottx,
We usually don't accept relatively simple Reactive Wrappers since most codebases that need them can just add their own.

Thanks,

PR #2444, opened 2022-09-07
Add array bind

danielt1263:
I noticed there is a bind that takes a var arg, but not one that take an array. This fixes the oversight.


RxCocoa/Common/Observable+Bind.swift
```diff
@@ -17,11 +17,22 @@ extension ObservableType {
...
     }
 
+	/**
+	 Creates new subscription and sends elements to observer(s).
+	 In this form, it's equivalent to the `subscribe` method, but it better conveys intent, and enables
+	 writing more consistent binding code.
+	 - parameter to: Observers to receives events.
+	 - returns: Disposable object that can be used to unsubscribe the observers.
+	 */
+	public func bind<Observer: ObserverType>(to observers: [Observer]) -> Disposable where Observer.Element == Element {
```
freak4pc, line 30: Thank you! 
The indentation here seems misaligned with the rest of the project, can you fix it?

Issue #2428, opened 2022-06-14
Application crash sending `text` selector to RxTableViewDelegateProxy or RxCollectionViewDelegateProxy


SlashDevSlashGnoll:
**Short description of the issue**:

We're seeing a crash among our userbase that is happening with alarming frequency however we have been unable to reproduce it or identify the root cause.   A slack [discussion was started here](https://rxswift.slack.com/archives/C052N01RE/p1654005762316679) among 2 of us that are seeing this problem but we wanted to post here with the hopes that someone else may have an idea about something to look at.

This is the stack trace/crash:

```Fatal Exception: NSInvalidArgumentException
-[RxCocoa.RxTableViewDelegateProxy text]: unrecognized selector sent to instance 0x28378a0d0

Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x99288 __exceptionPreprocess
1  libobjc.A.dylib                0x16744 objc_exception_throw
2  CoreFoundation                 0x176fc4 +[NSObject(NSObject) _copyDescription]
3  CoreFoundation                 0x2de98 ___forwarding___
4  CoreFoundation                 0x2cf70 _CF_forwarding_prep_0
5  ContextKitExtraction           0x59ec +[CKContextContentProviderUIScene _bestVisibleStringForView:usingExecutor:]
6  ContextKitExtraction           0x4d7c +[CKContextContentProviderUIScene _donateContentsOfWindow:usingExecutor:withOptions:]
7  ContextKitExtraction           0x48fc __78+[CKContextContentProviderUIScene extractFromScene:usingExecutor:withOptions:]_block_invoke
8  ContextKitExtraction           0x54ac __64-[CKContextExecutor addWorkItemToQueue:withWorkItem:andContext:]_block_invoke
9  libdispatch.dylib              0x1e6c _dispatch_call_block_and_release
10 libdispatch.dylib              0x3a30 _dispatch_client_callout
11 libdispatch.dylib              0x11fa4 _dispatch_main_queue_drain
12 libdispatch.dylib              0x11b98 _dispatch_main_queue_callback_4CF
13 CoreFoundation                 0x51800 __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
14 CoreFoundation                 0xb704 __CFRunLoopRun
15 CoreFoundation                 0x1ebc8 CFRunLoopRunSpecific
16 GraphicsServices               0x1374 GSEventRunModal
[...]

PR #2445, opened 2022-09-09
Table/CollectionViewDelegateProxy crash workaround

SlashDevSlashGnoll:
This PR implements the workaround for the crash discussed at https://github.com/ReactiveX/RxSwift/issues/2428 as requested by @freak4pc 

I put the extensions in a separate file to make later removal simple as this isn't expected to stay permanently.   I didn't add any tests as we do not know the actual source of the problem/bug at this time however existing tests passed without issue.

### DIFF
--- a/RxCocoa/iOS/Proxies/RxDelegateProxyCrashFix.swift
+++ b/RxCocoa/iOS/Proxies/RxDelegateProxyCrashFix.swift
@@ -1,0 +1,30 @@
+//
+//  RxDelegateProxyCrashFix.swift
+//  RxCocoa
+//
+//  Created by SlashDevSlashGnoll (SlashDevSlashGnoll@users.noreply.github.com) on 9/9/22.
+//  Copyright © 2022 Krunoslav Zaher. All rights reserved.
+//
+
+import Foundation
+
+
+// This extension exists solely to get around a crash found on iOS 15.4+ where a `text`
+// method invocation is being sent to the RxCollectionViewDelegateProxy which doesn't implement it.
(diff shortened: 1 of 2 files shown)


RxCocoa/iOS/Proxies/RxDelegateProxyCrashFix.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+
+import Foundation
+
+
+// This extension exists solely to get around a crash found on iOS 15.4+ where a `text`
+// method invocation is being sent to the RxCollectionViewDelegateProxy which doesn't implement it.
+// This is tracked at this bug: https://github.com/ReactiveX/RxSwift/issues/2428 This can be
+// removed if/when the actual source of the problem is found
+@objc extension RxCollectionViewDelegateProxy {
```
freak4pc, line 16: did you bump into this only for Table / Collection or delegate proxies in general? 
If it's broader you could theoretically add this to DelegateProxy itself as an extension.
Thoughts?

SlashDevSlashGnoll: The reports we discussed only listed crashes for TableView proxy and CollectionView proxy so it seemed specific to those as far as I've seen.

freak4pc:
Test failures are unrelated.

cruisediary:
Nice 💯

DemonionFF:
I'm receiving same issue, but for 'RxScrollViewDelegateProxy text'.

Added extension to it in a similar way to what is shown for collection/tableViews, but for ScrollView. 

Crash appeared again. 

Do you think adding 'text' properties to collection/table proxies may resolve it?

PR #2474, opened 2023-01-01
Use MainScheduler.instance in MessageDispatcher

minhaaan:
Use `MainScheduler.instance` in MessageDispatcher


freak4pc (commented):
What's the reason for this change?

PR #2475, opened 2023-01-06
Explicitly declare `APPLICATION_EXTENSION_API_ONLY` for CocoaPods

lickel:
This explicitly declares that RxSwift only uses extension safe APIs (and/or APIs in an extension-safe way).
The code was already extension safe; this just forward declares it and enforces it at compile time for CocoaPods.

If I can enforce this here, then I can enforce it in [Workflow](https://github.com/square/Workflow-swift) which is dependent on this library. Without this explicit conformance, it is not straightforward to write CI validation that these libraries remain extension safe.

There is zero harm in explicit conformance. Non-extension applications can call into extension-safe libraries without any restrictions.

There was an equivalent [PR](https://github.com/ReactiveCocoa/ReactiveSwift/commit/618b8026275c285b940aab1495ca4c3255925e73) for ReactiveSwift).

If possible, once this lands, would you be willing to publish a new release as well?
That way I would be unblocked in making/landing the change to Workflow.
And, thanks!


lickel:
Thanks!

Issue #2397, opened 2022-02-07
RxSwift 6.5.0 XCFramework breaks console po command


pmanuelli:
**Short description of the issue**:

With RxSwift and RxCocoa 6.5.0 integrated using the precompiled XCFrameworks, the object debug command (`po`) on the Xcode console stops working.

When trying to print an object with the `po` command I get an error. This happens with any object that I try to debug on the console.

This happens with the XCFrameworks integrated manually on the project or using CocoaPods.

**Expected outcome**:

The object is printed on the console normally.

**What actually happens**:

An error is printed on the console:

```
po myObject

error: virtual filesystem overlay file '/Users/shaimishali/Library/Developer/Xcode/DerivedData/Rx-dxolnknlpjuxspdqvtqlinrlckhg/Build/Intermediates.noindex/ArchiveIntermediates/RxCocoa/IntermediateBuildFilesPath/Rx.build/Release-iphonesimulator/RxCocoa.build/all-product-headers.yaml' not found
error: virtual filesystem overlay file '/Users/shaimishali/Library/Developer/Xcode/DerivedData/Rx-dxolnknlpjuxspdqvtqlinrlckhg/Build/Intermediates.noindex/ArchiveIntermediates/RxCocoa/IntermediateBuildFilesPath/Rx.build/Release-iphonesimulator/RxCocoa.build/all-product-headers.yaml' not found

error: couldn't IRGen expression. Please check the above error messages for possible root causes.
```

The error complains about RxCocoa. But if I remove it and leave only RxSwift it now complains about RxSwift:

```
po myObject

error: virtual filesystem overlay file '/Users/shaimishali/Library/Developer/Xcode/DerivedData/Rx-dxolnknlpjuxspdqvtqlinrlckhg/Build/Intermediates.noindex/ArchiveIntermediates/RxSwift/IntermediateBuildFilesPath/Rx.build/Release-iphonesimulator/RxSwift.build/all-product-headers.yaml' not found
error: virtual filesystem overlay file '/Users/shaimishali/Library/Developer/Xcode/DerivedData/Rx-dxolnknlpjuxspdqvtqlinrlckhg/Build/Intermediates.noindex/ArchiveIntermediates/RxSwift/IntermediateBuildFilesPath/Rx.build/Release-iphonesimulator/RxSwift.build/all-product-headers.yaml' not found

error: couldn't IRGen expression. Please check the above error messages for possible root causes.
```

[...]

PR #2487, opened 2023-02-15
fix debugging when using xcframework

Cyberbeni:
solves #2397

source: https://github.com/facebook/facebook-ios-sdk/issues/1628


freak4pc:
Very cool, thanks!

PR #2488, opened 2023-02-18
Fix missing typealiased import warnings

keith:
With Swift 5.8 a warning is produced unless you import all the modules that contain types you use typealises of. This adds the missing imports to fix the warnings.

Screenshot 2023-02-17 at 19 49 32


keith:
these went away in beta 2, so we can wait and see if they come back or close

keith:
dropping since it never came back in the beta period

PR #2489, opened 2023-02-18
Fix deployment target warnings in Xcode 14

keith:
Some platforms' deployment targets were lower than Xcode 14 supports, causing warnings. In this case I bumped to the minimum supported version.


keith:
Maybe we shouldn't merge this depending on older Xcode version support

freak4pc:
Thanks @keith, we can possibly push it to the next release:
> Please note, starting April 2023, all iOS and iPadOS apps submitted to the App Store must be built with Xcode 14.1 and the iOS 16.1 SDK.

DivineDominion:
Ran into this by accident and noticed it's already April, so might be good for a point-release, soon? :)

DivineDominion:
@freak4pc Thanks for merging -- and so quickly! I was looking at a day of work of changing repositories across dependencies otherwise :)

PR #2490, opened 2023-02-18
Fix SwiftLint build phase warning

keith:
Xcode warns since this build phase is always run, but it was marked as only running when the defined files changed.


freak4pc (approved):
Thanks!

PR #2506, opened 2023-03-22
Add fatal synchronization error hook

gabriellanata:
This change introduces a setting to trigger a fatal error when a synchronization error is detected. The `FATAL_SYNCHRONIZATION` flag has precedence over this setting and it does not change the current default behavior.

Reasoning: We're trying to work through many synchronization errors in our app, but this process is going to take a while. In the meantime, we want to prevent new errors from being added to places that don't have them already. Our approach is to enable this setting for tests and provide a way to disable it for specific tests that currently have synchronization issues. One of the things we need for this to be possible is a way to turn this feature on or off easily.

### DIFF
--- a/RxSwift/Rx.swift
+++ b/RxSwift/Rx.swift
@@ -81,7 +81,11 @@ func decrementChecked(_ i: inout Int) throws -> Int {
             #if FATAL_SYNCHRONIZATION
                 rxFatalError(message)
             #else
-                print(message)
+                if Hooks.fatalSynchronizationError {
+                    rxFatalError(message)
+                } else {
+                    print(message)
+                }
             #endif
         }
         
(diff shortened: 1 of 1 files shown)


freak4pc:
> This change introduces a setting to trigger a fatal error when a synchronization error is detected. The `FATAL_SYNCHRONIZATION ` flag has precedence over this setting and it does not change the current default behavior.
> 
> Reasoning: We're trying to work through many synchronization errors in our app, but this process is going to take a while. In the meantime, we want to prevent new errors from being added to places that don't have them already. Our approach is to enable this setting for tests and provide a way to disable it for specific tests that currently have synchronization issues. One of the things we need for this to be possible is a way to turn this feature on or off easily.

I'm sorry for not replying so far, I guess you have worked around this in a different way but I'm wondering what's the reaosn to not add `FATAL_SYNCHRONIZATION` to your test suite directly, instead of an additional flag from the RxSwift side?

freak4pc:
No response, closing

PR #2508, opened 2023-04-11
Provide `Infallible` versions of `combineLatest` without `resultSelector` requirement

hallee:
`Observable` has helper extensions for `combineLatest` that implement a basic default `resultSelector` which makes `combineLatest` much more ergonomic to work with. 

This PR implements `Infallible` versions of these helpers.

### DIFF
--- a/RxSwift/Traits/Infallible/Infallible+CombineLatest+arity.swift
+++ b/RxSwift/Traits/Infallible/Infallible+CombineLatest+arity.swift
@@ -29,6 +29,24 @@ extension Infallible {
     }
 }
 
+extension InfallibleType {
+    /**
+    Merges the specified observable sequences into one observable sequence of tuples whenever any of the observable sequences produces an element.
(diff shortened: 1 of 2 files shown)


RxSwift/Traits/Infallible/Infallible+CombineLatest+arity.swift
```diff
@@ -29,6 +29,24 @@ extension Infallible {
     }
 }
 
+extension InfallibleType where Element == Any {
```
freak4pc, line 32: Is this constraint actually needed? Can't you just extend all InfallibleTypes ?

hallee: @freak4pc Probably not, I just copied this from the `Observable` version, which has the constraint. Removing.

freak4pc (approved):

PR #2512, opened 2023-04-26
replace operator name : Rx.playground Content

doulos76:
### As is

- elementAt(3)

### To be

- element(at: 3)


freak4pc (approved):
Thanks !

PR #2520, opened 2023-05-11
Fix Infallible CombineLatest arity helpers

hallee:
From https://github.com/ReactiveX/RxSwift/pull/2508#discussion_r1181269601, I was wrong; the `where Element == Any` constraint is required to help the type system automatically infer the element type of the tuple. Without that constraint, combineLatest has this issue: 

Screenshot 2023-05-11 at 4 48 13 PM

Added back the constraint and also added tests to verify.


freak4pc (approved):
Very weird, thanks for covering with a test!

freak4pc (approved):

PR #2531, opened 2023-06-21
visionOS added to ISSUE_TEMPLATE.md

ilkerulusoy:
VisionOS SDK 1.0 has been released on June 21.  VisionOS has been added to the issue template.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -42,6 +42,7 @@
   - [ ] tvOS
   - [ ] watchOS
   - [ ] playgrounds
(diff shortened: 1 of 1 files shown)


freak4pc:
Hey, it's just a beta SDK. We won't support it until an official release and when we understand how / if it related to RxCocoa.

Thanks!

ilkerulusoy:
@freak4pc  it is issue template. i think we should open issue for visionOS.

freak4pc:
Nothing personal, we just don't support beta software. At this point it would be a waste of time debugging things that are constantly changing. We can investigate related bugs closer to the RC.

PR #2533, opened 2023-07-13
Factory closure lifetime differs between .create and .deferred, subscriptions to the latter keep its factory closure alive

nikolaykasyanov:
⚠️ it's an issue report in the form of a pull request, to have a self-contained runnable examples.

Each subscription to `Observable.deferred` captures its factory closure, unlike `Observable.create`.

**Expected outcome**:

Intuitively I would expect `Observable.deferred` to not keep its factory closure alive longer than required in subscriptions, but that's probably not the best argument.

Alternatively, I'd expect this behavior documented because I recognize that changing it at this point might be a very breaking change.

**What actually happens**:

`.deferred` is implemented in such a way each subscription to it retains its factory closure: https://github.com/ReactiveX/RxSwift/blob/2ff073fc1b9c600ac9b54734aff3348110437c19/RxSwift/Observables/Deferred.swift#L71

See the test cases in this PR. `testDeferredFactoryClosureLifetime` fails while `testObservableFactoryClosureLifetime` passes.

6.6.0

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  Version 14.3.1 (14E300c)
```

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

  - [x] yes (15.0.0-Beta.4)
  - [ ] no

  - [ ] just starting

### DIFF
--- a/Tests/RxSwiftTests/Observable+Tests.swift
+++ b/Tests/RxSwiftTests/Observable+Tests.swift
@@ -228,6 +228,78 @@ extension ObservableTest {
     }
 }
 
+// MARK: - Deferred
+extension ObservableTest {
+    func testDeferredFactoryClosureLifetime() {
+        class Foo {
+            let expectation: XCTestExpectation
+
+            init(expectation: XCTestExpectation) {
+                self.expectation = expectation
+            }
+
+            func bar() -> Observable<Void> {
+                Observable<Void>
+                    .deferred {
+                        self.expectation.fulfill()
+                        return .never()
+                    }
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


danielt1263:
I made a PR with the fix. #2534

nikolaykasyanov:
Closing in favour of #2534.

PR #2534, opened 2023-07-13
Fix lifetime issue with Deferred.

danielt1263:
This is in response to PR #2533. It includes the test that was added in that PR along with the fix.

### DIFF
--- a/RxSwift/Observables/Deferred.swift
+++ b/RxSwift/Observables/Deferred.swift
@@ -22,18 +22,16 @@ extension ObservableType {
 }
 
 final private class DeferredSink<Source: ObservableType, Observer: ObserverType>: Sink<Observer>, ObserverType where Source.Element == Observer.Element {
-    typealias Element = Observer.Element 
-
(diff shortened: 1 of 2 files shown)


freak4pc:
Thanks both @danielt1263 & @nikolaykasyanov 
The fix looks good but I'm quite worried about changing the behavior (even if it's wrong) with a patch / minor release. Not sure it's a huge issue but let me think about it a bit for a day or two :) 

If you have large code-bases it might be worth running it to confirm it works as expected at those scales.

danielt1263:
I have some largish code bases, but I almost never use deferred in my RxSwift code. (I use it all the time in Combine, but not Rx). So it wouldn't be a good test.

nikolaykasyanov:
I did it a bit of smoke testing and manual review (fairly big code base, around 20 instances of `.deferred` only though, if we don't count RxSwift itself), the change looks good from our side.

freak4pc (approved):

PR #2536, opened 2023-07-18
Fix Infallible's flatMap family of operators that can return an Infallible that fails silently

nikolaykasyanov:
I believe it's a big hole in the `Infallible` contract because this will a) compile b) return an `Infallible` that will fail silently, without calling any subscription closures:

```
_= someInfallible.flatMap { Observable.error(anError) }
    subscribe(onCompleted: { print("it's not gonna call me") }) 
```

There's two ways to fix this:
1. Have overloads for both variants, i.e.
  1. a flatMap that expects its closure to only return `Infallible` streams, and that itself returns `Infallible`.
  1. a flatMap that expects its closure to return any `ObservableConvertibleType` and itself returns `Observable`.
1. Only have the first of the above.

While 1. is better for backwards compatibility, it complicates overload resolution, although it seems to be working in the test code I've written.

As soon as the proposed change is agreed on, I'll add tests for the first and latest variants.

### DIFF
--- a/RxSwift/Traits/Infallible/Infallible+Operators.swift
+++ b/RxSwift/Traits/Infallible/Infallible+Operators.swift
@@ -263,7 +263,7 @@ extension InfallibleType {
 // MARK: - FlatMap
 extension InfallibleType {
     /**
-     Projects each element of an observable sequence to an observable sequence and merges the resulting observable sequences into one observable sequence.
+     Projects each element of an infallible sequence to an observable sequence and merges the resulting observable sequences into one observable sequence.
 
      - seealso: [flatMap operator on reactivex.io](http://reactivex.io/documentation/operators/flatmap.html)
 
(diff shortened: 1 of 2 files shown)


RxSwift/Traits/Infallible/Infallible+Operators.swift
```diff
@@ -288,12 +301,29 @@ extension InfallibleType {
...
-        -> Infallible<Source.Element> {
+        -> Observable<Source.Element> {
+        asObservable().flatMapLatest(selector)
+    }
+
+    /**
+     Projects each element of an infallible sequence into a new sequence of infallible sequences and then
+     transforms the infallible sequence of infallible sequences into an infallible sequence producing values only from the most recent infallible sequence.
+
+     It is a combination of `map` + `switchLatest` operator
```
nikolaykasyanov, line 312: Given that `switchLatest` doesn't exist for `Infallible,` should this line be kept or not?

RxSwift/Traits/Infallible/Infallible+Operators.swift
hoc081098, line 1: Good catch. It should be merged and released 🙏

freak4pc:
I think the fix could make sense indeed, but unfortunately it's a breaking API change so we can't easily merge it right now. It will have to wait for a later major version, but I'm keeping it open to not forget. Thanks!

PR #2537, opened 2023-07-21
Add primary associated types to protocols

gabriellanata:


freak4pc (commented):
Hey Gabriel,
This isn’t something we’ll be able to accept before RxSwift 7 since it’s a breaking API change.

gabriellanata:
:( Ok i understand, thanks for replying quickly. I'll close the PR

PR #2544, opened 2023-09-20
Add primary associated type Element to ObservableConvertibleType

Schemetrical:
This allows us to pass this around while ensuring matching types, such as when writing generic methods around ObservableConvertibleTypes

(just want to see if this builds without anything else changing in the repo, so ignore if needed)


Schemetrical:
nvm I guess the swift version is too low for CI

PR #2549, opened 2023-10-17
Tidied up some code using optional map

BenMaer:
Simply shortens/tidies up a few places of code, each of which were 3 lines, but could be written with just one using Optional's map function.

This feels like too small of change to mention in the changelog. If you disagree, let me know and I can make a mention.

I was also thinking of further condensing this code to the existing extension on `Disposables` in `AnonymousDisposable.swift`:
```
extension Disposables {
.....
    /// Constructs a new disposable with the given action used for disposal. Provides an optional `dispose` param for convenience
    ///
    /// - parameter dispose: Optional disposal action which would be run upon calling `dispose`.
    public static func create(with dispose: (() -> Void)?) -> Disposable {
        return dispose.map({ create(with: $0) }) ?? create()
    }

}
```

That way, the three places I cleaned up wouldn't each be repeating the same logic. Let me know if that's something you'd want.

### DIFF
--- a/RxSwift/Traits/PrimitiveSequence/Completable.swift
+++ b/RxSwift/Traits/PrimitiveSequence/Completable.swift
@@ -123,12 +123,7 @@ extension PrimitiveSequenceType where Trait == CompletableTrait, Element == Swif
                 let callStack = [String]()
         #endif
 
-        let disposable: Disposable
-        if let onDisposed = onDisposed {
-            disposable = Disposables.create(with: onDisposed)
-        } else {
-            disposable = Disposables.create()
-        }
+        let disposable: Disposable = onDisposed.map({ Disposables.create(with: $0) }) ?? Disposables.create()
 
         let observer: CompletableObserver = { event in
             switch event {
--- a/RxSwift/Traits/PrimitiveSequence/Completable.swift
+++ b/RxSwift/Traits/PrimitiveSequence/Completable.swift
@@ -123,7 +123,7 @@ extension PrimitiveSequenceType where Trait == CompletableTrait, Element == Swif
                 let callStack = [String]()
         #endif
 
-        let disposable: Disposable = onDisposed.map({ Disposables.create(with: $0) }) ?? Disposables.create()
+        let disposable: Disposable = onDisposed.map(Disposables.create(with:)) ?? Disposables.create()
 
         let observer: CompletableObserver = { event in
             switch event {
--- a/RxSwift/Traits/PrimitiveSequence/Maybe.swift
+++ b/RxSwift/Traits/PrimitiveSequence/Maybe.swift
@@ -138,7 +138,7 @@ extension PrimitiveSequenceType where Trait == MaybeTrait {
         #else
             let callStack = [String]()
         #endif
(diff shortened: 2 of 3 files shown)


RxSwift/Traits/PrimitiveSequence/Completable.swift
```diff
@@ -123,12 +123,7 @@ extension PrimitiveSequenceType where Trait == CompletableTrait, Element == Swif
                 let callStack = [String]()
         #endif
 
-        let disposable: Disposable
-        if let onDisposed = onDisposed {
-            disposable = Disposables.create(with: onDisposed)
-        } else {
-            disposable = Disposables.create()
-        }
+        let disposable: Disposable = onDisposed.map({ Disposables.create(with: $0) }) ?? Disposables.create()
```
freak4pc, line 126:
```suggestion
        let disposable: Disposable = onDisposed.map(Disposables.create(with:)) ?? Disposables.create()
```

RxSwift/Traits/PrimitiveSequence/Maybe.swift
```diff
@@ -138,12 +138,7 @@ extension PrimitiveSequenceType where Trait == MaybeTrait {
         #else
             let callStack = [String]()
         #endif
-        let disposable: Disposable
-        if let onDisposed = onDisposed {
-            disposable = Disposables.create(with: onDisposed)
-        } else {
-            disposable = Disposables.create()
-        }
+        let disposable: Disposable = onDisposed.map({ Disposables.create(with: $0) }) ?? Disposables.create()
```
freak4pc, line 141: ditto

RxSwift/Traits/PrimitiveSequence/Single.swift
```diff
@@ -135,12 +135,7 @@ extension PrimitiveSequenceType where Trait == SingleTrait {
             let callStack = [String]()
         #endif
 
-        let disposable: Disposable
-        if let onDisposed = onDisposed {
-            disposable = Disposables.create(with: onDisposed)
-        } else {
-            disposable = Disposables.create()
-        }
+        let disposable: Disposable = onDisposed.map({ Disposables.create(with: $0) }) ?? Disposables.create()
```
freak4pc, line 138: ditto :)

freak4pc (commented):
Nice fix! I added a clean-up suggestion.

RxSwift/Traits/PrimitiveSequence/Completable.swift
BenMaer: Clever suggestion! I committed your suggestion, and will commit the others as well.

RxSwift/Traits/PrimitiveSequence/Maybe.swift
BenMaer: Updated

RxSwift/Traits/PrimitiveSequence/Single.swift
BenMaer: Updated

BenMaer:
Hey @freak4pc , I made the updates you suggested. Clever call with that!

Let me know if there's anything else I can do.

freak4pc (approved):
Thanks!

PR #2555, opened 2023-11-08
Resolves issue 2554 (Missing imports warn 'this is an error in Swift 6')

drewster99:
## Summary
This commit resolves issue 2554, where Release builds of RxSwift (and other targets) emit many warnings that various references to Foundation types, such as `Foundation.Date` and `Dispatch.DispatchTimeInterval`, cannot be used because their definition was never imported, and that this warning will be a hard error in Swift 6.

## Implementation detail
These are resolved by simply adding "import Foundation" to the files generating the warnings.

## Other thoughts
If those files weren't importing Foundation on purpose, for some reason, the individual types needed could be imported instead.

### DIFF
--- a/RxSwift/Observables/Buffer.swift
+++ b/RxSwift/Observables/Buffer.swift
@@ -6,6 +6,8 @@
 //  Copyright © 2015 Krunoslav Zaher. All rights reserved.
 //
 
+import Foundation
+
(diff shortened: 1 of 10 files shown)


freak4pc:
Nice, thank you!

Issue #2548, opened 2023-10-16
Cannot call value of non-function type '[Observable<Event>.Type]'


dd9913762113:
Xcode 15.0
Mac Book Pro M1

Feedbacks.swift
```
private func bindingsStrongify<Event, O, WeakOwner>(_ owner: WeakOwner, _ bindings: @escaping (WeakOwner, O) -> (Bindings<Event>))
    -> (O) -> (Bindings<Event>) where WeakOwner: AnyObject {
    return { [weak owner] state -> Bindings<Event> in
        guard let strongOwner = owner else {
            return Bindings(subscriptions: [], events: [Observable<Event>]())
        }
        return bindings(strongOwner, state)
    }
}
```
- Moya/RxSwift (15.0.0):
    - Moya/Core
    - RxSwift (~> 6.0)
  
<img width="781" alt="截屏2023-10-16 下午5 20 00" src="https://github.com/ReactiveX/RxSwift/assets/85228967/ce0404d3-bd3d-4ec6-86ff-5d3ff22cde18">

danielt1263:
The compiler is confusing the Swift.Observable type with the RxSwift.Observable type. Use: `return Bindings(subscriptions: [], events: [RxSwift.Observable<Event>]())` instead.

freak4pc:
Feedbacks.swift is part of the example project and not the library itself so I don't think it's very relevant. Worth fixing but not a blocker for consumers. Closing for now, thanks

PR #2556, opened 2023-11-09
Fixes issue #2548, Cannot call value of non-function type RxExample-iOS and RxExample-macOS fail build

drewster99:
## Summary

Building RxExample-iOS or RxExample-macOS from Xcode both fail with the warning, `Cannot call value of non-function type '[Observable<Event>.Type]'`.

This is detailed in issue #2548 (https://github.com/ReactiveX/RxSwift/issues/2548).  While the issue is already marked as Closed and details of the needed fix mentioned in that issue, the fix itself was not actually implemented.

This resolves the issue for both the RxExample-iOS and RxExample-macOS targets.

## Implementation Detail
Replaced `Observable<Event>` with `RxSwift.Observable<Event>`, alleviating the compiler's confusion with Apple's `Swift.Observable` type.


CoderKingXY (commented):
it works

freak4pc:
Hey, thanks for the fix but on the latest version of Xcode (15.3) it seems to build without it, so I'm not merging it for now. If the issue persists please do let me know :)

PR #2560, opened 2023-11-17
Remove unused/unneeded method.

danielt1263:
The unsubscribe(_:) method is not used internally and isn't exposed publicly.


danielt1263:
Why would removing an unused internal method fail all the checks? The tests pass on my end...

freak4pc (approved):

freak4pc:
> Why would removing an unused internal method fail all the checks? The tests pass on my end...

The CI isn't working. Could use an upgrade to Xcode 15 when time allows.

PR #2568, opened 2023-12-14
Fix a typo in comments in Recorded+Event.swift

nicholaskim94:
A small typo in the commentss on the paramters


freak4pc (approved):
Nice catch, thanks!

PR #2160, opened 2020-03-31
Add asDriver() method to PublishRelay

yhkaplan:

PR #2575, opened 2024-01-22
Adds `asDriver()` operator to `PublishRelay`

Gongcu:
Currently, only `BehaviorRelay` has `asDriver()`, so I added the same operator to `PublishRelay`.

[BehaviorRelay+Driver.swift](https://github.com/ReactiveX/RxSwift/blob/main/RxCocoa/Traits/Driver/BehaviorRelay%2BDriver.swift)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


freak4pc:
https://github.com/ReactiveX/RxSwift/pull/2160
https://github.com/ReactiveX/RxSwift/pull/1769

PR #2577, opened 2024-01-23
Fix the crash issue with objc_sync_enter lock when using Firebase.

pandaleecn:
- Related crashes
https://github.com/firebase/firebase-ios-sdk/issues/5998
https://github.com/firebase/firebase-ios-sdk/issues/4790

- Theoretical reference
https://github.com/apple/darwin-libplatform/blob/main/src/os/lock.c#L494
https://developer.apple.com/documentation/foundation/posixerror/2293610-eownerdead
https://straypixels.net/swift-dictionary-locking/

### DIFF
--- a/RxCocoa/Foundation/NSObject+Rx.swift
+++ b/RxCocoa/Foundation/NSObject+Rx.swift
@@ -536,15 +536,31 @@ private extension KeyValueObservingOptions {
 // MARK: Constants
 
 private let deallocSelector = NSSelectorFromString("dealloc")
+private var lockKeyReactive: UInt8 = 0
 
 // MARK: AnyObject + Reactive
 
 extension Reactive where Base: AnyObject {
+
+    private var lock: NSRecursiveLock {
+		get {
+			if let l = objc_getAssociatedObject(self.base, &lockKeyReactive) as? NSRecursiveLock {
+				return l
+			}
+			let l = NSRecursiveLock()
+			objc_setAssociatedObject(self.base,  &lockKeyReactive, l, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+			return l
+		}
+	}
+
     func synchronized<T>( _ action: () -> T) -> T {
         objc_sync_enter(self.base)
         let result = action()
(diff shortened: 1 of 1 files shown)


RxCocoa/Foundation/NSObject+Rx.swift
```diff
@@ -536,15 +536,31 @@ private extension KeyValueObservingOptions {
...
+	}
+
     func synchronized<T>( _ action: () -> T) -> T {
         objc_sync_enter(self.base)
         let result = action()
         objc_sync_exit(self.base)
         return result
+        lock.lock()
+		defer { lock.unlock() }
+        return action()
```
freak4pc, line 563: Seems you left both the old and new. action() will run twice.
Can you further explain what's the issue ? Add some more context to your fix.

RxCocoa/Foundation/NSObject+Rx.swift
```diff
@@ -536,15 +536,27 @@ private extension KeyValueObservingOptions {
...
 private let deallocSelector = NSSelectorFromString("dealloc")
+private var lockKeyReactive: UInt8 = 0
 
 // MARK: AnyObject + Reactive
 
 extension Reactive where Base: AnyObject {
+
+    private var lock: NSRecursiveLock {
+		get {
+			if let l = objc_getAssociatedObject(self.base, &lockKeyReactive) as? NSRecursiveLock {
```
freak4pc, line 547: I have to say I'm not feeling comfortable creating an associated value like this for each `Reactive` object, especially given the fact we conform many types to it by default. 

Can you share where did you see references that swapping to a lock is the right solution? I saw the issues you linked but not any solutions.

freak4pc: Should we close this?

Naveen-C-Ramachandrappa (approved):
Thanks for addressing this crash. Looks good.

freak4pc:
This PR looks quite dangerous IMO and I am not seeing a reasonable explanation it's needed. With no response from the original author, I'm closing this for now.

PR #2587, opened 2024-03-27
Link XCTest weakly in RxTest to enable static linking

mkj-is:
When static linking is enabled (usually [a CocoaPods plugin](https://github.com/microsoft/cocoapods-pod-linkage)) the build might fail due to RxTest. We have encountered this on tvOS target using app extension.

Screenshot 2024-03-27 at 11 28 19

By enabling weak linking and correct linker flags for tests the pod will compile successfully. Inpired by [solution in Nimble framework](https://github.com/Quick/Nimble/blob/256e242d294b691103305588cbab176ed617a39e/Nimble.podspec#L26).


freak4pc (approved):
Nice workaround! Thanks :)

PR #2596, opened 2024-05-07
Reactive GitHub Client with RxSwift

tospery:
Added an example of an application implemented in RxSwift


freak4pc:
I apprecaite the contribution but we keep that section mainly for official reference material / official materials in general and less about demo projects, but regardless appreciate your work!

Issue #2098, opened 2019-10-31
A bug using `.seconds` in `throttle`.


danielt1263:
**Short description of the issue**:

I found a `throttle` bug.

```
_ = Observable<Void>.merge(Observable.just(()), Observable.just(()).delay(.milliseconds(500), scheduler: MainScheduler.instance))
	.throttle(.seconds(1), latest: false, scheduler: MainScheduler.instance)
	.debug("Doesn't Work")
	.subscribe()


_ = Observable<Void>.merge(Observable.just(()), Observable.just(()).delay(.milliseconds(500), scheduler: MainScheduler.instance))
	.throttle(.milliseconds(1000), latest: false, scheduler: MainScheduler.instance)
	.debug("works")
	.subscribe()
```

**Expected outcome**:

Each of the two observable chains above should emit a single next event.

**What actually happens**:

The "Doesn't Work" chain emits two events.

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

5.0.0

**Platform/Environment**

  - [X] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [X] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
11.1
```
**Installation method**:
  - [X] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [ ] yes (which ones)
  - [X] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [ ] I have a small code base
  - [X] I have a significant code base

kzaher:
Hi @danielt1263,

your code doesn't prove that we have a bug. It seems to me that you have defined a race in your code between `throttle` and `delay`.

There is no guarantee regarding timing in these cases. We are relying on the underlying dispatch queue behavior mostly.

If you think that something can be improved, you are welcome to make a PR so we can discuss your proposal, but if you presented this code to someone without "works" and "doesn't work" nobody could tell you what is expected from that code IMHO.

freak4pc:
[...]

PR #2600, opened 2024-05-14
Fix `.throttle` with `.seconds()` rounding issue.

VAndrJ:
When using `.seconds(1)`, 2 elements are emitted per second.
MRE with problem reproduction:

<details>

<summary>MRE</summary>

```swift

import SwiftUI
import RxSwift
import RxCocoa

struct ContentView: View {
    private let valuesObs = PublishRelay<Int>()
    private let bag = DisposeBag()

    init() {
        valuesObs
            .throttle(.milliseconds(1000), latest: true, scheduler: MainScheduler.instance)
            .subscribe(onNext: {
                print(".milliseconds(1000) throttle date: \(formatter.string(from: Date())), value: \($0)")
            })
            .disposed(by: bag)
        valuesObs
            .throttle(.seconds(1), latest: true, scheduler: MainScheduler.instance)
            .subscribe(onNext: {
                print(".seconds(1) throttle date: \(formatter.string(from: Date())), value: \($0)")
            })
            .disposed(by: bag)
    }

    var body: some View {
        Image(systemName: "globe")
            .task {
                for i in 0...999 {
                    valuesObs.accept(i)
                    try? await Task.sleep(for: .milliseconds(100))
                }
            }
    }
}

let formatter: DateFormatter = {
    let formatter = DateFormatter()
    formatter.dateFormat = "mm:ss.SSS"

    return formatter
}()

```

</details>

Result of MRE:

```

.milliseconds(1000) throttle date: 24:50.981, value: 0
.seconds(1) throttle date: 24:50.981, value: 0
.seconds(1) throttle date: 24:51.508, value: 5
.milliseconds(1000) throttle date: 24:51.983, value: 9
.seconds(1) throttle date: 24:52.037, value: 10
.seconds(1) throttle date: 24:52.566, value: 15
.milliseconds(1000) throttle date: 24:52.984, value: 19
.seconds(1) throttle date: 24:53.084, value: 20
.seconds(1) throttle date: 24:53.604, value: 25
...

```

As we can see, due to rounding, 2 elements are emitted per second, but only one should.

### DIFF
--- a/RxSwift/Date+Dispatch.swift
+++ b/RxSwift/Date+Dispatch.swift
@@ -45,7 +45,7 @@ extension DispatchTimeInterval {
             let interval = laterDate.timeIntervalSince(earlierDate)
             let remainder = Double(value) - interval * factor
             guard remainder > 0 else { return 0 }
-            return Int(remainder.rounded(.toNearestOrAwayFromZero))
+            return Int(remainder.rounded(.up))
         }
     }
 }
--- a/Tests/RxSwiftTests/Observable+ThrottleTests.swift
+++ b/Tests/RxSwiftTests/Observable+ThrottleTests.swift
@@ -393,7 +393,33 @@ extension ObservableThrottleTest {
         XCTAssertEqual(2, end.timeIntervalSince(start), accuracy: 0.5)
         XCTAssertEqual(a, [0, 1])
     }
-    
+
+    func test_ThrottleTimeSpan_WithRealScheduler_secondsRounding() {
+        let scheduler = ConcurrentDispatchQueueScheduler(qos: .default)
+
+        let throttlingSubject = PublishSubject<Int>()
+        for i in 1...5 {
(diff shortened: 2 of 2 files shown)


danielt1263:
I like this change. It's something that has always bugged me. It should probably wait for the major version bump.

freak4pc:
Seems like a good fix! 
I'm wondering if we can somehow pull it into the current release by doing something like `thorttle(..., rounding: FloatingPointRoundingRule = . toNearestOrAwayFromZero)` and being able to override it to `up`. 

(Or think of a different "behavior" argument). This will allow to be backward compatible during RxSwift 6.x and then make this the default behavior in RxSwift 7.x.

Thoughts?

danielt1263:
I know this has been an issue since at least 5.0.0 (see for eg: https://github.com/ReactiveX/RxSwift/issues/2098) And I've always used `milliseconds` in production code because of it. So I know none of my apps would be affected by the fix.

I like the idea of adding the parameter so we can put this fix in sooner...

Issue #2601, opened 2024-05-15
Add support for visionOS in Podspec


Hari51215:
Need to add the deployment_target for visionOS in Podspec

Hari51215:
Not Available on main branch

PR #2603, opened 2024-05-18
add the deployment_target for visionOS in Podspec

sanghun0724:
fix this issue https://github.com/ReactiveX/RxSwift/issues/2601#issue-2297554328 
-  add the deployment_target for visionOS in Podspec

- add extra if sentence ->  ` if s.respond_to?(:visionos)` 
because of this issue [pod install not working with lower version of Cocoapods due to visionOS](https://github.com/software-mansion/react-native-svg/pull/2240)


freak4pc:
Looks great, thanks!

Issue #2609, opened 2024-07-10
Why is VirtualTimeScheduler checking to ensure it's on the main thread?


danielt1263:
**RxTest's TestScheduler causes a crash when used in Swift Testing**:

  When running a test inside the Swift Testing system, the TestScheduler uses VirtualTimeScheduler which, at several points, checks to ensure it's running on the main thread. However since Swift Testing runs tests on background threads, the test harness will crash on the assertion.

**Expected outcome**:

  I expected that VirtualTimeScheduler would be thread agnostic. At worst, it would ensure that all its methods were run on the same thread as the start method was run on, but that's about it.

**What actually happens**:

  The VirtualTimeScheduler asserts because it's not being run on the main thread.

**Self contained code example that reproduces the issue**:

```swift
struct ExampleTests {
    @Test func testExample() async throws {
        let scheduler = TestScheduler(initialClock: 0)
        scheduler.start()
    }
}
```

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

  6.7.1

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  Version 16.0 beta 2 (16A5171r)
```

**Installation method**:
  - [x] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [x] yes (Version 16.0 beta 2 (16A5171r), 15.4)
  - [ ] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [ ] I have a small code base
  - [x] I have a significant code base

PR #2610, opened 2024-07-10
Allow VirtualTimeScheduler to run on any thread 

danielt1263:
…as long as critical methods are all called on the same thread.

This is designed to fix issue #2609

### DIFF
--- a/RxSwift/Schedulers/VirtualTimeScheduler.swift
+++ b/RxSwift/Schedulers/VirtualTimeScheduler.swift
@@ -24,6 +24,8 @@ open class VirtualTimeScheduler<Converter: VirtualTimeConverterType>
 
     private var nextId = 0
 
+    private var thread: Thread!
+
     /// - returns: Current time.
     public var now: RxTime {
         self.converter.convertFromVirtualTime(self.clock)
@@ -106,7 +108,10 @@ open class VirtualTimeScheduler<Converter: VirtualTimeConverterType>
      - returns: The disposable object used to cancel the scheduled action (best effort).
      */
     public func scheduleAbsoluteVirtual<StateType>(_ state: StateType, time: VirtualTime, action: @escaping (StateType) -> Disposable) -> Disposable {
-        MainScheduler.ensureExecutingOnScheduler()
+        guard thread == nil || Thread.current == thread else {
+            rxFatalError("Executing on the wrong thread. Please ensure all work on the same thread.")
+        }
+        thread = Thread.current
 
         let compositeDisposable = CompositeDisposable()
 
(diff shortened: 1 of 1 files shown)


danielt1263:
As far as I can tell, this is 100% backwards compatible and doesn't require any API change from users of the library. It may not solve every future use-case that Swift Testing might throw at it, but it solves the single most obvious one.

I see no reason that this can't be accepted into the next minor release/update.

RxSwift/Schedulers/VirtualTimeScheduler.swift
```diff
@@ -106,7 +108,10 @@ open class VirtualTimeScheduler<Converter: VirtualTimeConverterType>
      - returns: The disposable object used to cancel the scheduled action (best effort).
      */
     public func scheduleAbsoluteVirtual<StateType>(_ state: StateType, time: VirtualTime, action: @escaping (StateType) -> Disposable) -> Disposable {
-        MainScheduler.ensureExecutingOnScheduler()
+        guard thread == nil || Thread.current == thread else {
+            rxFatalError("Executing on the wrong thread. Please ensure all work on the same thread.")
+        }
+        thread = Thread.current
```
nikolaykasyanov, line 114: Is it safe to read and write `thread` without any locking?

danielt1263: Isn't that what we always do on the main thread? Read/write without locking? Maybe I don't understand the question...

nikolaykasyanov: Indeed, it should be completely fine as long scheduler's methods are actually called on one thread, but what if there would be two close calls from different threads, it's not impossible that both will see `thread == nil` and proceed, for example.

danielt1263: Right! Good point. I will update to account for that.

nikolaykasyanov: @danielt1263 sounds good!

bennnjamin:
What's the hold up in getting this merged? Would love to start being able to migrate our RxSwift tests to Swift Testing

freak4pc:
We'll need someone with a meaningful RxSwift-based test suite to test this to make sure it doesn't break existing tests.

nikolaykasyanov:
@freak4pc I'll try it out and report back.

freak4pc:
> @freak4pc I tried this PR on our relatively big test suite (around 1800 tests, most of which use `TestScheduler`) and didn't find any issues ✅ . To be sure, I also ran the suite repeatedly 10 times ✅ and with thread sanitizer ✅ .

That's awesome, thanks Nikolay!

PR #2611, opened 2024-07-13
introduce reduce with inout accumulator parameter

kjaklinovic:
The `scan` function has a variant in which it takes an input parameter as the first accumulator argument. I find that function more convenient since it requires a bit less code for the accumulator. I noticed that the same is not true for `reduce`, so I decided to introduce this behavior. I pretty much followed the same code style as `scan` has. I did the same for tests too, for every ~second existing test for `reduce` I introduced the new variant of the operator.

### DIFF
--- a/RxSwift/Observables/Reduce.swift
+++ b/RxSwift/Observables/Reduce.swift
@@ -8,6 +8,40 @@
 
 
 extension ObservableType {
+
+    /**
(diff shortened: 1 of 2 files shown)


kjaklinovic:
Hi @freak4pc , are there any further actions I need do perform for this one ? I can't check the logs of the failing checks.

PR #2624, opened 2024-08-27
Add Sendable conformance to AtomicInt

marcelofabri:
This fixes the following warning that shows up in Xcode 16 beta 6:

> Class 'AtomicInt' must restate inherited '@unchecked Sendable' conformance


freak4pc:
Thanks!

PR #2627, opened 2024-09-06
Added some documentations to AsyncLock file

herlandro:


Naveen-C-Ramachandrappa (approved):

ankushkushwaha (approved):

ankushkushwaha (approved):

ankushkushwaha:
@herlandro thanks for this addition. There are some failed workflow, I dont know what is the cause of failing these workflows. Could you please try to fix these?

freak4pc:
The main issue is that the CI is still using an older version of Xcode, I'll try to get to tackling it soon and bump to Xcode 16.

herlandro:
@freak4pc Did you see about the Xcode update to version 16?

.github/workflows/tests.yml
```diff
@@ -19,20 +19,48 @@ jobs:
...
+          echo 'export SWIFTENV_ROOT="$HOME/.swiftenv"' >> ~/.bashrc
+          echo 'export PATH="$SWIFTENV_ROOT/bin:$PATH"' >> ~/.bashrc
+          echo 'eval "$(swiftenv init -)"' >> ~/.bashrc
+          export SWIFTENV_ROOT="$HOME/.swiftenv"
+          export PATH="$SWIFTENV_ROOT/bin:$PATH"
+          eval "$(swiftenv init -)"
+          swiftenv install 5.5
+          swiftenv global 5.5
+          swift --version
+
```
freak4pc, line 63: I'm unsure how this change relates to documentation updates. Would you mind removing it from this PR and have a separate one with an explanation on why it's needed?

freak4pc:
I've rebased your branch - 
The previous CI failure was on docs generation which isn't really an issue because it's not a required check. Yet, I fixed it.
Please remove the updates to the GH workflows and leave only the doc updates.

Thanks 🙏

.github/workflows/tests.yml
herlandro: Removed. Done!

freak4pc (approved):
Ok, so - tl;dr
I personally don't think it's usually worth the effort to document so deeply the internal layers of RxSwift, since it's not public interface that consumers are using - 

But since you've already done this, it definitely doesn't hurt. Thank you for the work!

PR #2635, opened 2024-10-16
Update SwiftConcurrency.md

piotrprzeliorz:
This PR adds a note about the `async throws` overload availability. Hopefully, it’ll save everyone from trying to use it on version 6.5.0 😅.

### DIFF
--- a/Documentation/SwiftConcurrency.md
+++ b/Documentation/SwiftConcurrency.md
@@ -66,7 +66,7 @@ stream.asObservable()
 
 ### Wrapping an `async` result as a `Single`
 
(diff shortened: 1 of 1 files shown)


freak4pc (commented):
Uhm it's an interesting question - the docs are attached to specific versions, do you think it's worthwhile noting this here ?

piotrprzeliorz:
I think it’s worth adding because, at the beginning, we have a note stating that ‘Starting with RxSwift 6.5, you can await on your Observables.’ This might lead someone to assume that the new overload is also available from 6.5 version, like the rest of the async/await features (at least that was the case for me).

Issue #2563, opened 2023-12-04
Crash: value tried to resume its continuation more than once


ifabijanovic:
**Short description of the issue**:

  Seeing a new crash after updating to RxSwift 6.6.0 

**Expected outcome**:

  Does not crash

**What actually happens**:

  After updating to RxSwift 6.6.0 started seeing crashes with this message:

```
`_Concurrency/CheckedContinuation.swift:167: Fatal error: SWIFT TASK CONTINUATION MISUSE: value tried to resume its continuation more than once, returning <some_value>!`
```

  We downgraded to 6.5.0 and the crash went away.

**Self contained code example that reproduces the issue**:

  I do not have a code sample, the crash stack trace just shows RxSwift code, it is unclear where in the App code is it originating from.

**RxSwift/RxCocoa/RxBlocking/RxTest version/commit**

  Crashes in RxSwift 6.6.0, does not in 6.5.0

**Platform/Environment**

  - [x] iOS
  - [ ] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

**How easy is to reproduce? (chances of successful reproduce after running the self contained code)**

  - [ ] easy, 100% repro
  - [X] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  15.0.0
```

:warning: Fields below are optional for general issues or in case those questions aren't related to your issue, but filling them out will increase the chances of getting your issue resolved. :warning:

**Installation method**:
  - [ ] CocoaPods
  - [ ] Carthage
  - [ ] Git submodules
  - [X] SPM

**I have multiple versions of Xcode installed**:
  (so we can know if this is a potential cause of your issue)
  - [ ] yes (which ones)
  - [x] no

**Level of RxSwift knowledge**:
  (this is so we can understand your level of knowledge
    and formulate the response in an appropriate manner)
  - [ ] just starting
  - [ ] I have a small code base
  - [x] I have a significant code base

ifabijanovic:
```
Crashed: rxswift.queue.DispatchQoS(qosClass: Dispatch.DispatchQoS.QoSClass.default, relativePriority: 0)
0  libswiftCore.dylib             0x38870 _assertionFailure(_:_:file:line:flags:) + 264
1  libswift_Concurrency.dylib     0x59a0 CheckedContinuation.resume(returning:) + 504
[...]

PR #2641, opened 2024-11-04
Fix race conditions in PrimitiveSequence+Concurrency

0xpablo:
This caused continuations to be resumed more than once, crashing clients due to a violation of the Swift Concurrency contract.

I added tests that make it easy to reproduce the issue under the previous implementation

Screenshot 2024-11-04 at 09 38 15

This should fix https://github.com/ReactiveX/RxSwift/issues/2563

### DIFF
--- a/RxSwift/Traits/PrimitiveSequence/PrimitiveSequence+Concurrency.swift
+++ b/RxSwift/Traits/PrimitiveSequence/PrimitiveSequence+Concurrency.swift
@@ -62,19 +62,29 @@ public extension PrimitiveSequenceType where Trait == SingleTrait {
                 operation: {
                     try await withCheckedThrowingContinuation { continuation in
                         var didResume = false
+                        let lock = RecursiveLock()
                         disposable.setDisposable(
                             self.subscribe(
-                                onSuccess: {
-                                    didResume = true
-                                    continuation.resume(returning: $0)
+                                onSuccess: { value in
+                                    lock.withLock {
+                                        guard !didResume else { return }
+                                        didResume = true
+                                        continuation.resume(returning: value)
+                                    }
                                 },
-                                onFailure: {
-                                    didResume = true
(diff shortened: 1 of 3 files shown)


RxSwift/Traits/PrimitiveSequence/PrimitiveSequence+Concurrency.swift
```diff
@@ -111,27 +121,37 @@ public extension PrimitiveSequenceType where Trait == MaybeTrait {
             return try await withTaskCancellationHandler(
                 operation: {
                     try await withCheckedThrowingContinuation { continuation in
-                        var didEmit = false
                         var didResume = false
+                        let lock = RecursiveLock()
```
nikolaykasyanov, line 125: Take this with a grain of salt, as far as I can tell `RecursiveLock` is the to-go tool in this codebase.
- does it have to be recursive?
- can it be replaced with atomic compare-and-swap?

0xpablo: Hey, that's a good point. I used `RecursiveTool` precisely because of that, I saw that was the tool being used to synchronize code in the rest of the codebase. I initially thought about that but given that even `AtomicInt` inside RxSwift is backed by a `NSLock`, I didn't want to introduce a different way to synchronize this.
If a maintainer wants me to switch to CAS for this I'm happy to do that.

hoc081098: I think changing didResume to AtomicInt and using fetchOr might be better 🙏

0xpablo: I don't think so, `AtomicInt` is backed by an `NSLock` so the only thing that will add is making this code less clear.

freak4pc (approved):

PR #2642, opened 2024-11-10
Update cURL command for easier copy-pasting

EliaCereda:
Print a backslash at the end of each line in the cURL command printed by URLSession requests, so that the command can be directly copy-pasted in the Terminal


danielt1263:
Great addition! However, it doesn't handle the optional body correctly. Also, you might consider switching to using `#"..."#` syntax to make things a bit less confusing.

EliaCereda:
It’s not clear to me what you mean exactly. Multi-line bodies work out of the box for me, since the body is inside quotes and those newlines don’t break the command.

Let me know if there is a specific case that is broken for you

freak4pc:
The fix looks good. If you don't mind I would switch to use `#` delimiters so you can do less escaping with slashes and juse use `\n`. Maintaining / debugging this will be painful later on :) 

Let me know your thoughts 💪

EliaCereda:
Yes, I understand your point, it's a good point. I don't have my computer on hand to test this right now, if you want just go ahead and make the change yourself ;)

PR #2647, opened 2024-12-18
Don't treat \0 in description as literal null

matt-phylum:
I think in the Gem description you meant to write `\0` as human-readable text instead of a literal null byte. Heredoc strings interpret escapes unless you put single quotes like this. You can see it happening in the JSON version here where it says `\u0000` instead: https://github.com/CocoaPods/Specs/blob/98a5cfac82ecccd406eeff0d549c8eccadf1b699/Specs/2/e/c/RxSwift/6.8.0/RxSwift.podspec.json#L5

https://docs.ruby-lang.org/en/master/syntax/literals_rdoc.html#label-Here+Document+Literals

### DIFF
--- a/RxSwift.podspec
+++ b/RxSwift.podspec
@@ -2,7 +2,7 @@ Pod::Spec.new do |s|
   s.name             = "RxSwift"
   s.version          = "6.8.0"
   s.summary          = "RxSwift is a Swift implementation of Reactive Extensions"
-  s.description      = <<-DESC
+  s.description      = <<-'DESC'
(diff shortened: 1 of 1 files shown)


freak4pc:
Hey, thanks, I'm not entirely understanding this fix. I've never seen a podspec that has the single-quote in the multiline literal like this. Could you elaborate?

matt-phylum:
I hadn't seen it before either. The linked documentation on heredocs says that interpolation and escape sequences can be disabled by using `<<-'EXPECTED'`. Alternatively, it'd probably work to write `\\0` instead of `\0`, but then the description in the podspec file would look different from the rendered description.

PR #2648, opened 2024-12-19
feature: create map, compactMap, filter, flatMap ObservableType with with object

DevVenusK:
To facilitate chaining methods of `Observable`, a method with a `with` parameter has been overloaded.

When using `withUnretained`, it requires two different approaches.
1. Keep returning owner.
```swift
Observable.just<1>
  .withUnretained(self)
  .map { (owner, value) in
        return (owner, value * 2)
  }
  .map { (owner, value) in
        return (owner, value + 1)
  }
```
2. Continue using the withUnretained method.
```swift   
Observable.just<1>
  .withUnretained(self)
  .map { (owner, value) in
         return (value * 2)
  }
  .withUnretained(self)
  .map { (owner, value) in
        return (value + 1)
  }
```
However, to eliminate such inconvenience, the method has been overloaded with a `with` parameter.
```swift
Observable.just<1>
  .map(with: self) { $1 * 2 }
  .map(with: self) { $1 + 1 }
```

### DIFF
--- a/RxSwift/Observables/CompactMap.swift
+++ b/RxSwift/Observables/CompactMap.swift
@@ -19,6 +19,14 @@ extension ObservableType {
         -> Observable<Result> {
         CompactMap(source: self.asObservable(), transform: transform)
     }
+    
+    public func compactMap<Object: AnyObject, Result>(
+        with object: Object,
+        _ transform: @escaping ((Object, Element)) throws -> Result?
+    ) -> Observable<Result> {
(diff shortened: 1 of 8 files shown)


freak4pc:
Hey @DevVenusK, I appreciate the suggestion, but this doesn't feel general engouh to me to include in the core RxSwift package. You can possibly add it to RxSwiftExt. 

Thanks!

Issue #2525, opened 2023-05-29
Locks during Event Forwarding in RxSwift


firecore:
**Short description of the issue**:

We have identified a commonly encountered problem in RxSwift where event forwarding is performed under a lock. It leads to deadlocks and unexpected locks on threads, causing issues in code execution. We created a sample that demonstrates two specific problems with`testDeadlock` and `testTemporaryLock` methods within the `RxTester` class.

**Expected outcome**: 

We assume that event forwarding should be performed outside of the locked sections. However, if there are cases that require forwarding under a lock to guarantee that only one event is forwarded simultaneously, it should use a separate lock.

**What actually happens**: 

The issue of event forwarding under a lock has widespread implications throughout the RxSwift codebase.

**Problem 1: testDeadlock** 

This example illustrates the deadlock caused by a specific combination of Rx operators. The major root causes:

* The `share(replay: 1)` operator performs event forwarding under its lock for a non-first subscriber if there is an event to replay.
* The `combineLatest` operator uses the `SynchronizedOnType` protocol and performs event forwarding under a lock.

**Problem 2: testTemporaryLock** 

The testTemporaryLock example showcases how a lightweight event generation (e.g., on the main thread) can unexpectedly wait on a lock due to the heavyweight processing of events in a background thread. It happens because some operators (e.g., ‘debounce’) may receive and forward events in different threads but do that under the same lock.

Presumably, similar problems might reproduce with other operators. For example, all classes using ‘SynchronizedOnType’ protocol automatically follow the pattern; some classes not using it might have similar logic (like ShareReplay1WhileConnectedConnection). At the same time, many pieces of RxSwift code perform event forwarding after locked sections, and it looks right.

**Self contained code example that reproduces the issue**:

```swift
import Foundation
import RxSwift

class RxTester
{
    private let bag = DisposeBag()
    
    // Some observable that starts with an event and doesn't complete instantly.
    private func makeSomeObservable<T>(firstValue: T) -> Observable<T>
    {
        Observable<T>.create({ observer in
            observer.onNext(firstValue)
            return Disposables.create()
        })
    }
    
    /*
     This test demonstrates a deadlock, which is possible because:
     * share(replay: 1) performs event forwarding for non-first subscribers under its lock (see ShareReplay1WhileConnectedConnection.synchronized_subscribe(_:))
     * combineLatest, using SynchronizedOnType protocol, performs event forwarding under its lock
     
     Here's the scheme of the locked state of the threads:
[...]

PR #2654, opened 2025-02-23
Forward not under lock

### LINKED ISSUE #2653: `take(1)` + `share(replay: 1)` + `zip` = deadlock
Hey, guys. I have uncovered that there is a situation where you can easily cause RxSwift do deadlock.

I have prepared a repo with 100% reproduction rate of this feature.

It is caused by the fact that:
- `take(1)` emits `.completed`
- `zip` starts unsubscribing from `source` when it sees `.completed`
- `share(replay: 1)` synchronizes the unsubscribing under it's own lock
- BUT also `share(replay: 1)` replays the stored element under this very same lock
- which can also be forwarded into `.completed` event because `take(1)`
- which will be received by `zip`, which will try to unsubscribe from source
- which is the same `share(replay: 1)`
- ???
- PROFIT

In this deadlock situation, there are always 2 threads that get stuck:
- one has acquired the lock of `zip`, and tries to lock on `share(replay: 1)` to unsub
- the other has acquired the lock of `share(replay: 1)` to replay an event, and tries to lock on `zip`

If one thing happens earlier than the other in a concurrent environment, then they will deadlock.

Here's a link to my repository that reproes this issue, just build, Cmd+R, it's raw Xcodeproj with SPM:

https://github.com/isaac-weisberg/RxSwiftDeadlock

Separate thanks I would like to say to @firecore and specifically this issue: #2525 - one of the problems described in this issue is definitely related!

isaac-weisberg:
This pull request aims to fix #2653.

The original issue stems from the fact that:
- `ShareReplay1WhileConnect`, `ReplaySubject`, `BehaviorSubject` and `PublishSubject` emit their **initial events upon subscription** while holding onto their internal locks that protect their mutable state
- Their results can be forwarded to any operators
- These same operators might take decision to dispose of the upstream observables
- And these operators might do this while retaining their own locks
- And these operators might perform the handling of upstream events under the same locks

The last 3 points perfectly describe the behavior of a `zip` operator, which:
- will try to dispose of the source, while retaining its own lock
- will receive an initial event upon subscription from `ShareReplay1WhileConnect` or any `Subject`, with these sources retaining their locks
- will try to acquire its own lock to handle these events

In a multithreaded environment, these conditions can race, and this can produce a deadlock. This PR positions "emission upon subscription under lock" as the core issue and tries to fix specifically this behavior.

Special considerations:
- The current circumstances of the immediate emissions happening under lock might be both deliberate or accidental
- And with these changes, the sequentiality of these emissions will not be protected by the state lock anymore.
- Maybe I should add a separate lock to sequentialize specifically the emission of the initial elements? (as prophesied by @firecore in #2525) 
- I should say that didn't try hard enough to come up with a case in which the sequentialization of the emissions is absolutely necessary across threads.
- But I did ensure that at least the read out of the current state, and preparation of the values for the initial emission is synchronized.

Additionally:

@freak4pc I would like to say sorry for [this comment](https://github.com/ReactiveX/RxSwift/issues/2525#issuecomment-2669707983). I really wasn't planning to add anything valuable to the conversation, and my behavior was really counter productive here. This wasn't a troll comment, it was just - I was amused by the fact that RxSwift can break, I found it funny.

@danielt1263 Also I would like to say thank you for your time and the fact that you responded to me.

I would also like to say sorry to you both for [this comment](https://github.com/ReactiveX/RxSwift/issues/2653#issuecomment-2676433864) - the vagueness of my answer actually permitted an interpretation where I am not only posting nothing of value, but also deliberately try and insult you or the framework. This wasn't my intention, and I'm sorry that I came off this way.

### DIFF
--- a/RxSwift/Observables/ShareReplayScope.swift
+++ b/RxSwift/Observables/ShareReplayScope.swift
@@ -169,7 +169,7 @@ private final class ShareReplay1WhileConnectedConnection<Element>
     private let lock: RecursiveLock
     private var disposed: Bool = false
     fileprivate var observers = Observers()
-    private var element: Element?
+    fileprivate var element: Element?
 
     init(parent: Parent, lock: RecursiveLock) {
         self.parent = parent
@@ -203,20 +203,6 @@ private final class ShareReplay1WhileConnectedConnection<Element>
 
     final func connect() {
         self.subscription.setDisposable(self.parent.source.subscribe(self))
-    }
-
-    final func synchronized_subscribe<Observer: ObserverType>(_ observer: Observer) -> Disposable where Observer.Element == Element {
-        self.lock.performLocked {
-            let disposeKey = self.observers.insert(observer.on)
-
-            return SubscriptionDisposable(owner: self, key: disposeKey)
-        }
-    }
-        
-    func replayStoredElementIfNeeded<Observer: ObserverType>(_ observer: Observer) where Observer.Element == Element {
-        if let element = self.element {
-            observer.on(.next(element))
-        }
     }
 
     final private func synchronized_dispose() {
@@ -276,16 +262,20 @@ final private class ShareReplay1WhileConnected<Element>
         let connection = self.synchronized_subscribe(observer)
         let count = connection.observers.count
 
-        let disposable = connection.synchronized_subscribe(observer)
+        let disposeKey = connection.observers.insert(observer.on)
+
+        let initialValueToReplay = connection.element
         self.lock.unlock()
         
-        connection.replayStoredElementIfNeeded(observer)
+        if let initialValueToReplay {
+            observer.on(.next(initialValueToReplay))
+        }
         
         if count == 0 {
             connection.connect()
         }
 
-        return disposable
+        return SubscriptionDisposable(owner: connection, key: disposeKey)
     }
 
     @inline(__always)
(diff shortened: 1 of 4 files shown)


freak4pc:
I'll review this later, but wanted to say I highly appreciate your PR, and the apology 🙌
And even more appreciate the fact you submitted a fix to the issue that bothered you. Thank you!

freak4pc:
Could you add a unit tests that validates the fix against the issue you found?

isaac-weisberg:
I have prepared tests to test the specific issue, although, now that I think of it, they could be implemented better:
- Tecnically, the original problem really only offended `share(replay: 1, scope: .whileConnected)`. However, I wrote the same tests for other combinations of parameters for the `share` operator.
- At the same time, I deleted the implementation for `share(replay: 0, ...)` because it doesn't make sense semantically with my reproduction method of using `zip + take(1)` - they proper way to test these would be to: a) take `multicast` of `PublishSubject`; b) `subscribe` from multiple threads; c) `connect` once; d) count all the elements

But the main take away I guess is that it checks for the original issue, despite it intrinsically being impossible to happen.
So what's up, what are gonna be the next steps?

isaac-weisberg:
@freak4pc @danielt1263 Can you guys please take a look?

nikolaykasyanov:
It'd be great to get this merged if the trade-offs are acceptable.

PR #2660, opened 2025-03-31
feat: Enable Swift Concurrency for Linux

CrownedPhoenix:
Trivially removes the `!os(Linux)` guard from various concurrency related features.


CrownedPhoenix:
Salient bit of information about this: I tested this in a Linux Docker container with Swift 6.0.3.

I'm not sure what history there is related to Swift >=5.6 and <6 related to Linux so there may be a better way to gate these sections?

PR #2662, opened 2025-04-16
fix(#2553) : AsyncSequence.asObservable() runs on background thread

### LINKED ISSUE #2553: AsyncSequence.asObservable() effect is not as expected
The following code looks the same. cloneProgress returns an AsyncThrowStream, but using asObservable causes it to work on the main thread (the thread from which this code is executed).

```swift
// This will cause the code in the Task to run in the current thread (in my case, the main thread).
return AsyncSwiftGit.Repository.cloneProgress(from: remoteURL, to: localURL, credentials: credentials).asObservable()

// This works as expected, but it just pulls the asObservable code out
return .create { observer in
    let task = Task {
        do {
            for try await value in AsyncSwiftGit.Repository.cloneProgress(from: remoteURL, to: localURL, credentials: credentials) {
                observer.onNext(value)
            }
            observer.onCompleted()
        } catch {
            observer.onError(error)
        }
    }
    return Disposables.create {
        task.cancel()
    }
}
```

  6.6.0

**Platform/Environment**

  - [ ] iOS
  - [x] macOS
  - [ ] tvOS
  - [ ] watchOS
  - [ ] playgrounds

  - [x] easy, 100% repro
  - [ ] sometimes, 10%-100%
  - [ ] hard, 2% - 10%
  - [ ] extremely hard, %0 - 2%

**Xcode version**:

```
  15.0.0
```

**Installation method**:
  - [ ] CocoaPods
  - [ ] Carthage
  - [x] SwiftPM
  - [ ] Git submodules

  - [ ] yes (which ones)
  - [x] no

  - [ ] just starting
  - [x] I have a small code base
  - [ ] I have a significant code base

suojae:
Fixes #2553

### Problem Description
The current implementation of `AsyncSequence.asObservable()` can potentially execute on the main thread, which may cause UI blocking issues. This becomes especially problematic when the AsyncSequence performs time-consuming operations.

### Changes
Modified `AsyncSequence.asObservable()` to use `Task.detached`, ensuring the sequence always runs on a background thread. This change guarantees that AsyncSequence operations won't block the main thread.

### Testing
- Successfully passed all existing tests (1793 tests, 0 failures)
- Added and passed test cases that verify execution on background threads
- Included additional test cases covering edge scenarios (with sleep operations, error handling)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 * Provides `Infallible` versions of `combineLatest` without `resultSelector` requirement.
 * Provides `Infallible` versions of `CombineLatest+Collection` helpers.
 * Explicitly declare `APPLICATION_EXTENSION_API_ONLY` for CocoaPods 
+* Ensure `AsyncSequence.asObservable()` runs on background thread using `Task.detached`. 
 
 ## 6.5.0
 
--- a/RxSwift/Observable+Concurrency.swift
+++ b/RxSwift/Observable+Concurrency.swift
@@ -60,7 +60,7 @@ public extension AsyncSequence {
     /// - returns: An `Observable` of the async sequence's type
     func asObservable() -> Observable<Element> {
         Observable.create { observer in
-            let task = Task {
+            let task = Task.detached {
                 do {
                     for try await value in self {
                         observer.onNext(value)
--- a/Tests/RxSwiftTests/Observable+ConcurrencyTests.swift
+++ b/Tests/RxSwiftTests/Observable+ConcurrencyTests.swift
@@ -90,5 +90,111 @@ extension ObservableConcurrencyTests {
         task.cancel()
     }
 
+    func testAsyncSequenceToObservableRunsOnBackgroundThread() async throws {
+        
+        let asyncSequence = AsyncStream<Int> { continuation in
+            for i in 1...5 {
+                continuation.yield(i)
+            }
+            continuation.finish()
+        }
+        
+        let expectation = XCTestExpectation(description: "Observable completes")
+        
+        DispatchQueue.main.async {
+            let observable = asyncSequence.asObservable()
+            
+            var threadIsNotMain = false
+            var values = [Int]()
+            
+            _ = observable.subscribe(
+                onNext: { value in
+                    values.append(value)
+                    threadIsNotMain = !Thread.isMainThread
+                },
+                onCompleted: {
+                    XCTAssertEqual(values, [1, 2, 3, 4, 5])
+                    XCTAssertTrue(threadIsNotMain, "AsyncSequence.asObservable should not run on main thread")
+                    expectation.fulfill()
+                }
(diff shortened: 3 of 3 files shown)


RxSwift/Observable+Concurrency.swift
```diff
@@ -60,7 +60,7 @@ public extension AsyncSequence {
     /// - returns: An `Observable` of the async sequence's type
     func asObservable() -> Observable<Element> {
```
hoc081098, line 61: What do you think about adding a new param to decide whether to use Task or Task.detached?

suojae: I think adding a parameter is a good idea to provide flexibility. I'm considering implementing it like this:

```swift
func asObservable(detached: Bool = true) -> Observable<Element> {
    // Use Task.detached or Task based on the parameter
}
```

With true as the default value, it would still solve the original issue by running on a background thread, while giving users the option to use regular Task when they want to maintain the current context's priority.

What do you think about this approach and the default value being true?

hoc081098: It sounds good. I’ve also seen a similar implementation in another library, which takes the same approach

https://github.com/ReactiveX/RxSwift/blob/5dd1907d64f0d36f158f61a466bab75067224893/RxSwift/Traits/PrimitiveSequence/PrimitiveSequence%2BConcurrency.swift#L25

About the default value of `detached`, I think `false` should be used to avoid introducing a breaking change.

suojae: I've implemented the parameter as suggested, with false as the default value to maintain backward compatibility. 

```swift
func asObservable(detached: Bool = false) -> Observable<Element> {
    // Implementation uses Task.detached when detached=true, otherwise uses Task
}
```

All tests are passing. Thanks! 

commit log: [17e8102](https://github.com/ReactiveX/RxSwift/pull/2662/commits/17e81023ef45ff370b9622b140a5ac5913c26761)

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 * Provides `Infallible` versions of `combineLatest` without `resultSelector` requirement.
 * Provides `Infallible` versions of `CombineLatest+Collection` helpers.
 * Explicitly declare `APPLICATION_EXTENSION_API_ONLY` for CocoaPods 
+* Ensure `AsyncSequence.asObservable()` runs on background thread using `Task.detached`.
```
hoc081098, line 14:
```suggestion
* Add a new parameter `detached: Bool = false` to the `AsyncSequence.asObservable` function to allow iterating over the `AsyncSequence` on a background thread. 
```

RxSwift/Observable+Concurrency.swift
hoc081098: Nice 🙏

CHANGELOG.md
suojae: thanks!

PR #2663, opened 2025-04-26
Android support in Package.swift and CI testing

marcprux:
This PR sets up conditional exclusion of the `RxCocoa` target when building for Android. It is similar to how Linux is excluded, except that since we are cross-compiling, we can't check the OS of the host toolchain and instead check the "TARGET_OS_ANDROID" environment variable (which is set by both the [swift-android-action](https://github.com/marketplace/actions/swift-android-action) and `skip android build`, and can be manually set if you are compiling straight against the Swift SDK.

Note that this isn't sufficient to get the whole project building — you'll also need to change `#if os(Linux)` to `#if os(Linux) || os(Android)`, or else perhaps change them all to `#if !canImport(Darwin)` (which is my preferred way to distinguish Apple/non-Apple platforms).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,14 +25,14 @@ jobs:
         run: CI=1 ./scripts/all-tests.sh "${{ matrix.environment }}"
   
   linux:
-    name: "Test (Android)"
+    name: "Build (Android)"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - name: "Build Swift Package on Android"
         uses: skiptools/swift-android-action@v2
         with:
-          skip-tests: true
+          run-tests: false
 
   # We're having some issues with the Linux tests, so we're disabling them for now.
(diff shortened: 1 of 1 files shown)


marcprux:
Note that there's another (potentially better) way of doing this:

```swift
    static func rxCocoa() -> [Target] {
        if !targetsDarwin() {
            return [.rxTarget(name: "RxCocoa", dependencies: ["RxSwift", "RxRelay"])]
        } else {
            // change RxCocoaRuntime target to be conditional on Darwin platforms
            //return [.rxTarget(name: "RxCocoa", dependencies: ["RxSwift", "RxRelay", "RxCocoaRuntime"])]
            return [.rxTarget(name: "RxCocoa", dependencies: ["RxSwift", "RxRelay", .target(name: "RxCocoaRuntime", condition: .when(platforms: [.iOS, .macOS, .tvOS, .watchOS]))])]
        }
    }
```

PR #2676, opened 2025-11-24
Implement refCount with timeout

nikolaykasyanov:
This PR adds an overload of `ConnectableObservableType.refCount` that accepts a timeout that the source connection will be kept alive after the last subscription to the refcounted observable has been disposed.

RxJava counterpart: https://reactivex.io/RxJava/javadoc/io/reactivex/observables/ConnectableObservable.html#refCount-long-java.util.concurrent.TimeUnit-io.reactivex.Scheduler-

### DIFF
--- a/RxSwift/Observables/Multicast.swift
+++ b/RxSwift/Observables/Multicast.swift
@@ -304,20 +304,21 @@ final private class RefCountSink<ConnectableSource: ConnectableObservableType, O
                     return
                 }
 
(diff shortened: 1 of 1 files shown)


nikolaykasyanov:
@freak4pc thanks for the update! I doesn't seem like this operator is implementable without the access to the RxSwift internals though, at least not it won't be as simple, so I'm not sure about the suggested way forward.

PR #2690, opened 2026-02-02
macCatalyst compile fix

nighthawk:
6.10.0 and 6.10.1 broke support for macCatalyst, which doesn't compile anymore due to the missing RxCocoaRuntime dependency. This is the simple fix for that.


nighthawk:
Sorry, used the wrong base branch. Will apply to branch based on `main` instead.

PR #2707, opened 2026-06-02
Never call `dispose` and `subscribe` under lock in `Connection` and `ConnectionObservableAdapter`

### LINKED ISSUE #2698: Deadlock in `Connection` and `ConnectableObservableAdapter`
It is rare that we get an opportunity to look at an actual stack trace of an RxSwift system being in a deadlocked state, it's a very flaky behavior, and [my last time I was able to fix a deadlock in RxSwift](https://github.com/ReactiveX/RxSwift/issues/2653) was thanks for the issue being so easy to observe in development environment. 

But recently we had a stroke of luck where one of our crash reports has allowed us to catch very clearly that there are indeed deadlocks all over our codebase because of how RxSwift works with its locks and event forwarding.

- We have recently added a method of periodically checking the main thread for a deadlock, and crashing the app early.
- In this crash, we found our main thread deadlocked while trying to subscribe to an observable, which down the line calls to `ReplayBufferBase.subscribe<T> (ReplaySubject.swift:143)`
- It tries to acquire a lock there, but the lock is busy.
- And at the same time, we have threads 51 and 52 deadlocked against one another with, undoubtedly the same, ReplayBufferBase being an unlucky victim.

[trimmed crashlog.txt](https://github.com/user-attachments/files/26349155/trimmed.crashlog.txt)

In this particular instance, it is clear that the culprit is something along the lines of Connection disposing source subscription under its own lock, while this same subscription on another thread has acquired Connection's lock to 
a) connect a subscription in ConnectableObservableAdapter 
b) OR forward an event through the `Connection.on`
I'm a bit hazy on that detail here in this particular instance of the deadlock.

We are going to locally fix this on our side by moving `dispose` call on `subscription` in `Connection.dispose` outside the lock. However, to fix this in the RxSwift code base, we would need to provide a minimal reproducible example and write tests for it...

The overarching problem is that the library is written in such a way where every operator forwards events while retaining its own lock, and usage of `NSRecursiveLock` pretty much everywhere has worked out as a great enabler, making it that back and forward propagations of events don't cause any problems. 

For example, `.take(1)` causes a stop event, and thus, a dispose on the source subscription that is forwarding onto `.take(1)`. So the original source of the event gets its `dispose` called while having its own `forwardOn` in the stack, while holding its own lock. And this only works because the lock is recursive, otherwise, there would be a situation of a double lock on same thread. 

While recursive locks have saved this system within the boundaries of one thread, I would like to report that this paradigm can cause deadlocks in multithreaded environment.

isaac-weisberg:
Fixes [2698](https://github.com/ReactiveX/RxSwift/issues/2698)

### DIFF
--- a/RxSwift/Observables/Multicast.swift
+++ b/RxSwift/Observables/Multicast.swift
@@ -176,20 +176,24 @@ private final class Connection<Subject: SubjectType>: ObserverType, Disposable {
     }
 
     func dispose() {
-        lock.lock(); defer { lock.unlock() }
-        fetchOr(disposed, 1)
-        guard let parent else {
-            return
-        }
+        let subscriptionToDispose: Disposable? = lock.withLock {
+            fetchOr(disposed, 1)
+            guard let parent else {
+                return nil
+            }
 
-        if parent.connection === self {
-            parent.connection = nil
-            parent.subject = nil
-        }
-        self.parent = nil
+            if parent.connection === self {
+                parent.connection = nil
+                parent.subject = nil
+            }
+            self.parent = nil
 
-        subscription?.dispose()
-        subscription = nil
+            let subscriptionToDispose = subscription
+            subscription = nil
+            return subscriptionToDispose
+        }
+        
+        subscriptionToDispose?.dispose()
     }
 }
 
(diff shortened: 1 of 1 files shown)


freak4pc:

PR #2709, opened 2026-06-02
`RefCount` call `connect` without holding the lock

### LINKED ISSUE #2708: RefCount calls connect under its lock, and this cause the deadlock
Details here:

https://github.com/ReactiveX/RxSwift/issues/2698#issuecomment-4200082234

isaac-weisberg:
Closes #2708


freak4pc:

PR #2712, opened 2026-06-03
Fix the deadlock in MergeSink by not calling out from under state lock

### LINKED ISSUE #2711: MergeSink can deadlock
This deadlock has been detected by us monitoring the main thread aliveness, and crashing the app if we detect the main thread to be locked.

[MergeSink deadlock stack trace.txt](https://github.com/user-attachments/files/28549649/MergeSink.deadlock.txt)

isaac-weisberg:
However, the events are serialized by a separate lock for forwarding.

Fixes #2711


freak4pc:
Merged via #2724 with your commit's authorship preserved (plus the forwardLock teardown correction we discussed there) — closing as superseded. Thanks again for the diagnosis and the fix!

PR #2714, opened 2026-06-03
Make ReplayBufferBase propagate events without holding onto its lock

### LINKED ISSUE #2713: ReplayBufferBase can create deadlocks
We have tons of reports where replaying buffer from under lock in ReplayBufferBase and disposing under operator locks that goes into ReplayBufferBase, causes deadlocks

[ReplayBufferBase subscribe replay deadlock.txt](https://github.com/user-attachments/files/28551266/ReplayBufferBase.subscribe.replay.deadlock.txt)

[ReplayBufferBase unsubscribe dispose deadlock.txt](https://github.com/user-attachments/files/28551267/ReplayBufferBase.unsubscribe.dispose.deadlock.txt)

isaac-weisberg:
Otherwise, it can create deadlock situations

Fixes #2713

### DIFF
--- a/RxSwift/Subjects/ReplaySubject.swift
+++ b/RxSwift/Subjects/ReplaySubject.swift
@@ -103,7 +103,7 @@ private class ReplayBufferBase<Element>:
         rxAbstractMethod()
     }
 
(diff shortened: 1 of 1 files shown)


freak4pc:

PR #2716, opened 2026-06-25
Fix missing CFBundleShortVersionString

### LINKED ISSUE #2687: Missing CFBundleShortVersionString causing app upload failures on 6.10.0
We use Carthage to build RxSwift and recently upgraded from 6.9.1 to 6.10.0.

After trying to upload our updated apps to App Store Connect, we were hit with this validation error from Transporter:

> Validation failed (409)
> The bundle 'Payload/[app-name]/Frameworks/RxCocoa.framework' is missing plist key. The Info.plist file is missing the required key: CFBundleShortVersionString. 
> Please find more information about CFBundleShortVersionString at https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring (ID: 602a4e7e-f85f-4001-8c49-ca9cf7790b40)

After inspecting the Info.plist contents in the .ipa we uploaded, this key is missing from all of the Rx frameworks. 

Within the Carthage checkouts folder used to generate the .ipa, `CFBundleShortVersionString` is present in the Rx Info.plists, but they're all set to `$(RX_VERSION)` instead of a version number.

Info.plist from the .ipa:

Image

Info.plist from Carthage:

Image

GautamKumarOffical:
Fixes #2687

`Version.xcconfig` defines `RX_VERSION` and the Info.plists reference `$(RX_VERSION)`, but the xcconfig was never actually linked to any build configuration in the Xcode project. So the variable never resolves and `CFBundleShortVersionString` ends up empty when building with Carthage.

Changes:
- Added `Version.xcconfig` as a `PBXFileReference` and linked it in the project
- Set it as `baseConfigurationReference` for all 3 project-level build configs (Debug, Release, Release-Tests)
- Bumped `RX_VERSION` from 6.10.1 to 6.10.2 to match the latest tag


freak4pc:
Thanks for this @GautamKumarOffical, and sorry it sat so long — the diagnosis was spot on.

Closing as a duplicate: #2696 landed the same fix and was opened first (March). Both were verified to produce the correct `CFBundleShortVersionString`; I went with #2696 because it inserts the `PBXFileReference` in Xcode's own sort position.

Really appreciate you taking the time on it.

PR #2721, opened 2026-08-13
Replay ReplaySubject's buffer without holding its lock

### LINKED ISSUE #2713: ReplayBufferBase can create deadlocks

freak4pc:
Lands @isaac-weisberg's fix from #2714 (commit authorship preserved) together with a regression test.

Opened here rather than on #2714 because that PR comes from an organization-owned fork, where GitHub does not offer *Allow edits from maintainers*, and its CI run is too old for GitHub to re-run — so there was no way to add the test or get a fresh CI matrix on the original PR. Full credit to @isaac-weisberg for the diagnosis and the fix.

Fixes #2713.

### The bug

`ReplayBufferBase.synchronized_subscribe` replays the buffer to a new observer **while holding the subject's lock**. Any lock a downstream observer takes during that replay is therefore acquired underneath the subject's lock, so a thread that holds that other lock and then reaches into the subject deadlocks against it.

### The fix

The buffer is snapshotted under the lock and replayed outside it. A per-subscription wrapper observer queues events that arrive during the replay and drains them afterwards, so ordering is preserved and no event can slip between registration and replay.

### The test

`test2713ReplaySubjectDoesntReplayWhileHoldingItsLock` reproduces the inversion directly: one thread holds an unrelated lock and then subscribes to the subject, while another subscribes and has the replayed value delivered into an observer that needs that same lock.

It uses `lock(before:)` so a regression fails the assertion instead of hanging the suite.

| | result |
|---|---|
| on `main`, without the fix | **fails** after ~22s of lock timeouts |
| with the fix | **passes** in 0.002s |

Verified locally on the full `AllTests-macOS` suite (green ×2).

### DIFF
--- a/RxSwift/Subjects/ReplaySubject.swift
+++ b/RxSwift/Subjects/ReplaySubject.swift
@@ -103,7 +103,7 @@ private class ReplayBufferBase<Element>:
         rxAbstractMethod()
     }
 
-    func replayBuffer<Observer: ObserverType>(_: Observer) where Observer.Element == Element {
+    func replayBuffer() -> ReplayBufferSnapshot<Element> {
         rxAbstractMethod()
     }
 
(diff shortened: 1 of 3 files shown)


freak4pc:
Folded into #2723, which lands all five of @isaac-weisberg's deadlock fixes together with their regression tests. Consolidating because the fixes are not independent — #2707 + #2709 without #2714 deadlock deterministically, so landing them atomically removes the ordering hazard.

PR #2722, opened 2026-08-13
Never dispose or subscribe under Connection's lock

### LINKED ISSUE #2698: Deadlock in `Connection` and `ConnectableObservableAdapter`

freak4pc:
Lands @isaac-weisberg's fix from #2707 (commit authorship preserved) together with a regression test and two small cleanups.

Opened here rather than on #2707 because that PR comes from an organization-owned fork, where GitHub does not offer *Allow edits from maintainers*, and its CI run is too old to re-run. Full credit to @isaac-weisberg for the diagnosis and the fix.

Fixes #2698.

### The bug

`Connection.dispose()` disposes its source subscription while holding the adapter's lock, and `ConnectableObservableAdapter.connect()` subscribes to the source while holding that same lock. Any lock the upstream chain takes during teardown is therefore acquired underneath the adapter's lock, so a thread holding that other lock and reaching into the connectable deadlocks against it.

### The fix

Both callouts move outside the critical section. The subscription is handed out of the locked region and disposed after unlocking; `connect()` subscribes after releasing the lock and assigns into the `SingleAssignmentDisposable`, which already handles being disposed before assignment.

### The test

`test2698ConnectionDoesntDisposeItsSubscriptionUnderItsLock` reproduces the inversion: one thread holds an unrelated lock and then subscribes to the connectable, while another disposes the connection so its upstream teardown needs that same lock.

The second thread subscribes rather than connecting, so it needs the adapter's lock without re-subscribing upstream — that keeps the teardown running exactly once, on the other thread. `lock(before:)` makes a regression fail the assertion instead of hanging the suite.

| | result |
|---|---|
| on `main`, without the fix | **fails** after ~22s of lock timeouts |
| with the fix | **passes** in 0.002s |

### Cleanups

- `lock.withLock` → `performLocked`, matching every other lock site in the codebase.
- Stripped the trailing whitespace the original commit introduced.

Verified locally: `AnomaliesTest` + `ObservableMulticastTest` green, and the full `AllTests-macOS` suite green ×2 on the underlying fix.

### DIFF
--- a/RxSwift/Observables/Multicast.swift
+++ b/RxSwift/Observables/Multicast.swift
@@ -176,7 +176,7 @@ private final class Connection<Subject: SubjectType>: ObserverType, Disposable {
     }
 
     func dispose() {
-        let subscriptionToDispose: Disposable? = lock.withLock {
+        let subscriptionToDispose: Disposable? = lock.performLocked {
             fetchOr(disposed, 1)
             guard let parent else {
                 return nil
@@ -192,7 +192,7 @@ private final class Connection<Subject: SubjectType>: ObserverType, Disposable {
             subscription = nil
             return subscriptionToDispose
         }
-        
+
         subscriptionToDispose?.dispose()
     }
 }
@@ -233,7 +233,7 @@ private final class ConnectableObservableAdapter<Subject: SubjectType>:
             let subscription = self.source.subscribe(connectionToReturn)
             singleAssignmentDisposableToSubscribe.setDisposable(subscription)
         }
-        
+
         return connectionToReturn
     }
 
--- a/RxSwift/Observables/Multicast.swift
+++ b/RxSwift/Observables/Multicast.swift
@@ -176,20 +176,24 @@ private final class Connection<Subject: SubjectType>: ObserverType, Disposable {
     }
 
     func dispose() {
-        lock.lock(); defer { lock.unlock() }
-        fetchOr(disposed, 1)
-        guard let parent else {
-            return
-        }
+        let subscriptionToDispose: Disposable? = lock.withLock {
+            fetchOr(disposed, 1)
+            guard let parent else {
+                return nil
+            }
 
-        if parent.connection === self {
-            parent.connection = nil
-            parent.subject = nil
-        }
-        self.parent = nil
+            if parent.connection === self {
+                parent.connection = nil
+                parent.subject = nil
+            }
+            self.parent = nil
 
-        subscription?.dispose()
-        subscription = nil
+            let subscriptionToDispose = subscription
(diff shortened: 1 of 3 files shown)


freak4pc:
Folded into #2723, which lands all five of @isaac-weisberg's deadlock fixes together with their regression tests. Consolidating because the fixes are not independent — #2707 + #2709 without #2714 deadlock deterministically, so landing them atomically removes the ordering hazard.

PR #2724, opened 2026-08-14
Never call out to user code while holding Merge's locks

### LINKED ISSUE #2711: MergeSink can deadlock

freak4pc:
Stacked on #2723. Lands @isaac-weisberg's Merge deadlock fix — the fifth and largest piece of the lock-inversion campaign — plus one correction found in review. His fix commit keeps his authorship.

Supersedes #2712. Fixes #2711.

### What it does

Near-total rewrite of `MergeLimitedSink` and `MergeSink`: state transitions are decided under `lock` and acted on outside it, forwarding is serialized by a new `forwardLock`, and a `terminating` flag suppresses post-terminal work. Inner-subscription teardown (`group.remove`, `sourceSubscription.dispose`) moves outside all locks — that teardown-under-lock is the #2711 deadlock.

### Correction from review: teardown moved out of `forwardLock`

The original had `synchronizedForwardOnAndDispose` call `dispose()` — which tears down upstream — while holding `forwardLock`. Review found that leaves a narrow but real residual cycle: a thread can pass the `!terminating` check while holding some foreign lock `L` and park on `forwardLock`, while the terminating thread holds `forwardLock` and its teardown needs `L`. Same shape as the bug being fixed, error-terminal variant, much smaller window.

`dispose()` could not simply be hoisted out, because running it inside `forwardLock` was load-bearing: it sets the sink's disposed flag before the lock is released, and that flag is the only thing making `forwardOn` drop a straggler `.next` from the parked thread — hoisting it naively would forward an element *after* a terminal event. The fix splits the two concerns: a small `Sink.markDisposed()` (just the flag) runs inside `forwardLock`, and the actual teardown runs after it. `dispose()` is idempotent, so calling it afterwards is safe.

**The corrected window has no regression test.** It lives between the `!terminating` check and the `forwardLock` acquire — a few instructions, with no seam a test can widen from outside the library (any externally sequenced emitter bails at the check instead of parking). The #2711 shape itself is covered by `test2711MergeDoesntDisposeInnerSourcesUnderItsLock`, which fails on `main` with ~22s of lock timeouts and passes here in ~0.003s.

### Verification

- `test2711MergeDoesntDisposeInnerSourcesUnderItsLock`: fails on `main`, passes here; 100 stress iterations green.
- Full `ObservableMergeTest` class (merge / flatMap / concat semantics): 70/70.
- ThreadSanitizer over the Anomalies, Merge, Zip, Multicast, Concat, and ReplaySubject test classes: 233/233, zero data races.
- This tree is byte-identical to the combined branch that passed the full `AllTests-macOS` suite twice.

### Review gate

@isaac-weisberg — the teardown correction changes your fix, so I'd appreciate your eyes on that commit before this merges. The base PR (#2723) carries the other four fixes and can land independently of this one.

### DIFF
--- a/RxSwift/Observables/Merge.swift
+++ b/RxSwift/Observables/Merge.swift
@@ -136,9 +136,7 @@ public extension ObservableType {
 }
 
 private final class MergeLimitedSinkIter<SourceElement, SourceSequence: ObservableConvertibleType, Observer: ObserverType>:
-    ObserverType,
-    LockOwnerType,
-    SynchronizedOnType where SourceSequence.Element == Observer.Element
+    ObserverType where SourceSequence.Element == Observer.Element
 {
     typealias Element = Observer.Element
     typealias DisposeKey = CompositeDisposable.DisposeKey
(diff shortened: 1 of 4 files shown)


isaac-weisberg:
It's all good

PR #2727, opened 2026-08-18
Avoid primitive sequence cancellation deadlock

victorpimentel:
`Single.value`, `Maybe.value`, and `Completable.value` currently resume their continuations while holding the bridge lock. If completion races task cancellation, that can invert the bridge and Swift task locks and deadlock.

This is showing up frequently as an app hang in Goodnotes production.

This uses RxSwift's existing `AtomicInt` latch to select one callback, then resumes after the latch has released its lock. The race tests now wait for all 10,000 bridged tasks, so a stuck task cannot pass unnoticed. They are also faster: 1.87 seconds combined instead of 4.07 seconds, averaged across five isolated runs on my machine.

On RxSwift 6.10.2, the updated tests timed out in 5 of 30 race runs on an iOS 26 simulator. With this change, the full test class passed 180 of 180 runs.

This follows up on [#2641](https://github.com/ReactiveX/RxSwift/pull/2641), which added the bridge lock to prevent continuations from being resumed more than once.

### DIFF
--- a/Tests/RxSwiftTests/PrimitiveSequence+ConcurrencyTests.swift
+++ b/Tests/RxSwiftTests/PrimitiveSequence+ConcurrencyTests.swift
@@ -88,13 +88,14 @@ extension PrimitiveSequenceConcurrencyTests {
         )
     }
 
-    /// A previous implementation of the `Single` to swift concurrency bridge had a bug where it would sometimes call the continuation twice.
+    /// A previous implementation of the `Single` to swift concurrency bridge could call the continuation twice or deadlock when completion raced cancellation.
     /// The current number of iterations is a sweet spot to not make the tests too slow while still catching the bug in most runs.
     /// If you are debugging this issue you might want to increase the iterations and/or run this test repeatedly.
-    func testSingleContinuationIsNotResumedTwice() {
+    func testSingleCancellationRaceCompletesExactlyOnce() {
         let expectation = XCTestExpectation()
         let iterations = 10000
-        for i in 0 ..< iterations {
+        expectation.expectedFulfillmentCount = iterations
+        for _ in 0 ..< iterations {
             DispatchQueue.global(qos: .userInitiated).async {
                 let single = Single<Int>.create { observer in
                     DispatchQueue.global(qos: .userInitiated).asyncAfter(deadline: .now() + 0.005) {
@@ -104,6 +105,7 @@ extension PrimitiveSequenceConcurrencyTests {
                 }
 
                 let task = Task {
+                    defer { expectation.fulfill() }
                     _ = try await single.value
                 }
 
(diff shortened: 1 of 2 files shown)


victorpimentel:
Hi @nikolaykasyanov and @freak4pc. Since you were involved in #2641, I thought you might have useful context on this follow-up. We are seeing this lock-ordering race frequently as an app hang in Goodnotes production. Any thoughts would be appreciated. Thanks!

Tests/RxSwiftTests/PrimitiveSequence+ConcurrencyTests.swift
```diff
@@ -88,13 +88,14 @@ extension PrimitiveSequenceConcurrencyTests {
         )
     }
 
-    /// A previous implementation of the `Single` to swift concurrency bridge had a bug where it would sometimes call the continuation twice.
+    /// A previous implementation of the `Single` to swift concurrency bridge could call the continuation twice or deadlock when completion raced cancellation.
     /// The current number of iterations is a sweet spot to not make the tests too slow while still catching the bug in most runs.
```
freak4pc, line 92: this line says the iteration count catches the bug in most runs, but per the description the deadlock only reproduced 5/30 on 6.10.2. thats a pretty weak guard for something thats hanging in prod. worth bumping iterations or tightening the emit/cancel window so it fails reliably against the old impl? at minimum the sentence should reflect the new odds. same copy in the maybe + completable docs.

freak4pc (commented):
thanks for the contribution! 

not related to this PR, but same shape as what it fixes: values on ObservableConvertibleType still guards its terminal with a plain var isFinished, written from onError/onCompleted and read from onDisposed, no sync at all. finish() is idempotent so it's not a crash like the double-resume was, but it's a real race. probably wants the same AtomicInt latch. opening a follow-up.

Tests/RxSwiftTests/PrimitiveSequence+ConcurrencyTests.swift
victorpimentel: Yeah, I think 10k iterations were a compromise to avoid having very long test times. The changes in the tests right now do not guarantee failure in those cases but are much more stronger since now we wait for all blocks to finish exactly once.

However now the tests run a bit faster than before since we avoid sleeping too much, we can bump the number of iterations if you think it makes sense. What would be your accepted running time in an “average” machine?

freak4pc: uhm i missed the fact it's already at 10k, let's leave it like this. youll probably need over 100,000 runs to make it a meaningful card which probably isn't worth it.

freak4pc (approved):
if you want to take a look at the second place i suggested that'd be good, but not mandatory since a bit out of scope :) PR looks good to me otherwise.

victorpimentel:
I'll be able to take a look at it next week. Still, when do you think this change may get to a release? Thanks!

victorpimentel:
@freak4pc Do we need something else for this to be merged? Thank you!

PR #2728, opened 2026-08-26
Make Observable.values terminal handling atomic

victorpimentel:
While following up on [#2727](https://github.com/ReactiveX/RxSwift/pull/2727), I found the same terminal/disposal race in `ObservableConvertibleType.values`. `onError` and `onCompleted` write a captured `isFinished` flag while `onDisposed` reads it, and those callbacks can run on different threads. `AsyncThrowingStream.finish` is idempotent, so this does not cause the double-resume crash from #2727, but the flag is still a data race and cancellation can win over the source terminal event.

This replaces the flag with the existing `AtomicInt` latch. Each terminal callback claims the latch with `fetchOr`, and only the winner calls `finish`. The test exercises 10,000 completion/cancellation races and waits for every task to finish.

Tests: `xcodebuild -workspace Rx.xcworkspace -scheme AllTests-macOS -destination platform=macOS -only-testing:AllTests-macOS/ObservableConcurrencyTests test` (the scheme enables Thread Sanitizer and Undefined Behavior Sanitizer).


victorpimentel:
@freak4pc this is the fix related to the `Observable.values` property. The added tests do fail pretty much every time with the previous code, while the new code always passes the tests.

victorpimentel:
BTW I did audit other types and didn't find a similar issue so this should be all.

